Issue #111, opened 2022-07-27
Project Issue: Plus Intro Offers


emilylaguna:
## Description 
Provide an introductory offer of 1 month free when a new plus subscriber signs up for an annual plan.

### IAP Changes
- [x] Refresh the IAP products on launch: https://github.com/Automattic/pocket-casts-ios/pull/112
- [x] Add a server check to see if someone is eligible for the intro offer: https://github.com/Automattic/pocket-casts-ios/pull/171
    - [x] Add a method to check if someone is eligible for an offer: https://github.com/Automattic/pocket-casts-ios/pull/113
- [x] Add a way to see if a product has an intro offer: https://github.com/Automattic/pocket-casts-ios/pull/112
    - [x] Return information about the offer, ie: how long (1 week, 1 month, etc): https://github.com/Automattic/pocket-casts-ios/pull/112

#### Reuse `UpgradeRequiredViewController`
  - [x] Add a check when a product has an offer: https://github.com/Automattic/pocket-casts-ios/pull/113
      - [x] If there is more than 1 offer (ie: monthly and yearly have offers) promote the yearly offer
  - [x] Update the UI to reflect the offer: https://github.com/Automattic/pocket-casts-ios/pull/113
      - [x] Update the primary header with new copy: "Try Pocket Casts Plus free for TIME period"
          - Replace `TIME` with the trial duration: "3"
          - Replace `period` with the trial duration period: "days"
      - [x] Add a secondary label with information that the trial can be cancelled at anytime, and there's no payment required
          - "No Payment Now – Cancel Anytime"
      - [x] Update the next button label to "Start Free Trial"

#### Update `SelectAccountTypeViewController` 
- [x] Do a test of removing the monthly price from the view: #119
- [x] If not, update the price to reflect the available offers: #119

#### Update `SelectPaymentFreqViewController` 
- [x] Add a label to display the current offers for yearly and monthly: https://github.com/Automattic/pocket-casts-ios/pull/120

#### Update `ConfirmPaymentViewController`
- [x] Add a label to display the currently selected offer and the terms
    - "Free for TIME period then PRICE/frequency after trial": https://github.com/Automattic/pocket-casts-ios/pull/121
- [x] Update the next button label to "Subscribe with Free Trial": https://github.com/Automattic/pocket-casts-ios/pull/121

#### Update `PlusDetailsViewController`
- [x] Update the view to also display the offers available: https://github.com/Automattic/pocket-casts-ios/pull/124
[...]

PR #113, opened 2022-07-27
Free Trial: Reuse the UpgradeRequiredViewController to display free trial information

emilylaguna:
Project: #111
Depends on: #112 

## Screenshots
| Before | After Without Trial | After With Trial |
|:---:|:---:|:---:|
|Screen Shot 2022-07-27 at 4 18 33 PM|Screen Shot 2022-07-27 at 4 20 11 PM|Screen Shot 2022-07-28 at 11 30 17 AM|

## Description
- Updates the `UpgradeRequiredViewController` to also provide information about the free trial if it's available
- I cleaned up the xib design a bit:
   - Fixed the plus image size since it wasn't set correctly
   - Added a stack view for the title/trial detail label to allow for more consistent spacing between the views
   - Updated the spacing between views to be more consistent
- I took a bold approach in hiding the nav bar to provide a cleaner look to the view 
   - Side effect is the top left X button is no longer there, but the no thanks button does the same function
- I made the title label a bit larger to stand out more
- I added a new method to return "any" free trial duration
   - I didn't want to be stuck requesting a trial for a specific product on the views, this allows us to support trials for any product
   - I wanted to have a priority of display (~yearly then monthly~ monthly then yearly) so I updated the `productIdentifiers` to use the enum values, so I can loop through them in that order. This required some changes to the places where the string values were used. 

## To test

1. Follow the steps from #112 to setup the storekit config, and configure it with a free trial for either product or both to test the any order
2. Launch the app and make sure you don't have plus active
3. Tap on the podcast tab
4. Tap on the new folder button
5. ✅ Verify you see the free trial UI
6. ✅ Verify the title label says the correct duration
7. ✅ Verify the trial subheading label shows with "No payment required - cancel anytime"
8. ✅ Verify the blue button now reads "Start Free Trial"
9. Disable the trials in the storekit config
10. Open the upsell view again
11. ✅ Verify the UI does not reference any trials and the subheading is hidden
12. ✅ Verify the spacing looks good

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE_NOTES.md` if necessary.


leandroalonso (approved):

leandroalonso (approved):
@emilylaguna re-reviewed :shipit:

PR #271, opened 2022-09-22
Fix crash when selecting title of featured episode

### LINKED ISSUE #58: Crash: when selecting a title of a featured episode
This was initially reported to us in the [Beta Slack](https://pocketcastsbeta.slack.com/archives/C0GKZ8J4A/p1655487254424499). 

We have been unable to replicate it; however, the user can consistently replicate the issue and has sent the crash logs to Firebase. Upon closer look, it seems like the user is on the iOS 15.6 Beta, so we may need to double-check if the issue is iOS-related or something we need to fix. 

Screen Recording:

https://user-images.githubusercontent.com/16253818/174629866-95653d86-f5a6-498b-bf6a-762506228b97.mp4

```
App Version: 7.20
Device: iPhone14,5
OS: 15.6

```

#5306175-zen

jgcaruso:
Fixes #58 

This PR fixes a data issue that I haven't been able to reproduce in practice, but I have a hunch is occurring because it seems like the only hole is in the "Featured Episode" tap handler.

The issue stems from the fact that if the podcast has already been downloaded, the existing entry is being used instead of pulling new data. If the episode that is featured is a newer episode than what was downloaded the "featured episode" will be missing from the list and the view will crash trying to load the episode. Reasons why the episode might be missing could be the sync hasn't happened yet, or is happening but is slow due to slow network connection or the device being offline.

This PR adds a new optional parameter to the `loadPodcast` method to pass the episode that we are interested in, and if it doesn't exist, the podcast data will be refreshed. If it can't be refreshed or if the episode is still missing after it is refreshed, the episode view will not be opened.

These changes should prevent the crashes in the Discover tab, but has the weird user behaviour of tapping on an episode and having nothing happen. We could add some feedback here, however this is also currently the behaviour if you don't have internet access and you try to tap the episode when you don't already have the podcast itself downloaded.

There is a similar crash reported coming from the ListeningHistory view, but I suspect the cause of that one might be different even though the same child view is causing the crash.

## To test

1. Switch the Build configuration for the `pocketcasts` scheme to `Staging`, this will ensure you have a Featured Episode available to tap in the Discover view since the staging data has a few loaded
2. debug and in the simulator visit the Discover view, tap the Featured Episode title (**not** the "Play episode" button) and the episode should load as expected (ensuring everything still works correctly)
    

3. Simulate the episode is missing from the app db:
   - add the following code at line ~~144~~  ~~148~~ 146 of `DiscoverEpisodeViewModel.swift` : `DataManager.sharedManager.delete(episodeUuid: episodeUuid) // testing line to exercise code block`
   - debug again, tap the Featured Episode title and the episode should load as expected (simulates downloading the episode when its missing)
   - turn off wifi, tap the Featured Episode title and nothing should happen (because the episode can't be downloaded) - but importantly the app shouldn't crash either
4. Simulate the episode is missing from app db AND is completely missing from the podcast (maybe it was deleted by the creator):
   - ~~replace line 191 with `let episode:Episode? = nil`~~
   - ~~replace line 192 with `completion?(false)`~~
   - replace line 169 with `promise(.failure(.episodeNotFound))`
   - debug again, tap the Featured Episode title and nothing should happen (because we're simulating the episode not existing anymore) - but importantly the app doesn't crash either

## Checklist

Tests were considered but the function I'd need to write tests for is `static func loadPodcast` which means I'd have to inject the mocked data manager into that function, unless there is a way to override the static `DataManager.sharedManager` during testing? I'm happy to write the tests if there is a clean way to test this function, or if we're fine with the messy way of adding the DataManager as a param to 2 static functions 😄


emilylaguna (changes requested):
Hey @jgcaruso! 

Nice work on this tricky issue. I gave this a test and I did not get the crash 🥳 . 

I do think however we should have some handling to show an error or alert to the user if this happens because I found myself repeatedly tapping the episode because I thought I did something wrong. 

I also left some other comments about some code cleanup. Let me know what you think.

podcasts/DiscoverEpisodeViewModel.swift
```diff
@@ -124,7 +125,7 @@ class DiscoverEpisodeViewModel: ObservableObject {
             AnalyticsHelper.podcastEpisodeTapped(fromList: listId, podcastUuid: podcastUuid, episodeUuid: episodeUuid)
         }
 
-        DiscoverEpisodeViewModel.loadPodcast(podcastUuid)
+        DiscoverEpisodeViewModel.loadPodcast(podcastUuid, ensureEpisodeUuid: episode.uuid)
             .receive(on: RunLoop.main)
             .sink { [weak self] podcast in
                 guard let podcast = podcast else { return }
```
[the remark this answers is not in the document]
jgcaruso: Good idea! If you don't have an internet connection AND don't have the episode it already behaves like that (no error or feedback on tap), but I think we can fix it in all cases here.

pocketcasts:
## Danger has errored
[!] Invalid `Dangerfile` file: Error running SwiftFormat:
Error: [0mRunning SwiftFormat...
(lint mode - no files will be changed.)
/usr/local/var/buildkite-agent/builds/MV-MKE-X64-009/automattic/pocket-casts-ios/podcasts/DiscoverEpisodeViewModel.swift:155:1: warning: (redundantReturn) Remove unneeded return keyword.
/usr/local/var/buildkite-agent/builds/MV-MKE-X64-009/automattic/pocket-casts-ios/podcasts/DiscoverEpisodeViewModel.swift:156:1: warning: (redundantReturn) Remove unneeded return keyword.
SwiftFormat completed in 0.1s.
Source input did not pass lint check.
1/3 files require formatting, 1 file skipped.[0m. Updating the Danger gem might fix the issue. Your Danger version: 8.6.1, latest Danger version: 9.0.0

```
 #  from Dangerfile:23
 #  -------------------------------------------
 #  # rubocop:enable Layout/LineLength
 >  swiftformat.check_format(fail_on_error: true)
 #  
 #  -------------------------------------------
```

</p>

podcasts/DiscoverEpisodeViewModel.swift
```diff
@@ -135,15 +136,39 @@ class DiscoverEpisodeViewModel: ObservableObject {
...
+                        if exists {
+                            promise(.success(existingPodcast))
+                        }
+                        else {
+                            promise(.failure(.episodeNotFound))
+                        }
+                    }
+                    return
+                }
+
```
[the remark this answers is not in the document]
jgcaruso: After looking at this again, I also realized `episodeUuid` didn't need to be optional, so I did a little refactor that reduced the amount of code that needed to be duplicated... but also moved it all into a separate internal-to-the-function function.

If that's not a best practice (it felt a bit like a gross php/js-like thing to do here) please let me know if there is a more swift-y way to handle it.

podcasts/DiscoverEpisodeViewModel.swift
```diff
@@ -154,4 +179,26 @@ class DiscoverEpisodeViewModel: ObservableObject {
...
+        if episode == nil {
+            ServerPodcastManager.shared.updatePodcastIfRequired(podcast: podcast) { _ in
+                let episode = DataManager.sharedManager.findEpisode(uuid: episodeUuid)
+
+                if episode == nil {
+                    completion?(false)
+                }
+                else {
+                    completion?(true)
+                }
```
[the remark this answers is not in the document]
jgcaruso: I'll try to keep this pattern in mind (when the var is not used) for the future.

podcasts/DiscoverEpisodeViewModel.swift
```diff
@@ -121,51 +121,45 @@ class DiscoverEpisodeViewModel: ObservableObject {
             AnalyticsHelper.podcastEpisodeTapped(fromList: listId, podcastUuid: podcastUuid, episodeUuid: episodeUuid)
         }
 
-        DiscoverEpisodeViewModel.loadPodcast(podcastUuid, ensureEpisodeUuid: episode.uuid)
+        DiscoverEpisodeViewModel.loadPodcast(podcastUuid, episodeUuid: episodeUuid)
             .receive(on: RunLoop.main)
             .sink { [weak self] podcast in
-                guard let podcast = podcast else { return }
+                guard let podcast = podcast else {
+                    DispatchQueue.main.async {
```
emilylaguna: Since this is set to be received on the main thread: `.receive(on: RunLoop.main)` we don't need to dispatch this call on main.

podcasts/DiscoverEpisodeViewModel.swift
```diff
@@ -121,51 +121,45 @@ class DiscoverEpisodeViewModel: ObservableObject {
...
+                        SJUIUtils.showAlert(title: L10n.error, message: L10n.discoverFeaturedEpisodeErrorNotFound, from: self?.delegate)
```
emilylaguna: Since this is just a view model, we shouldn't include any UI items in here. 

Instead we can add a new delegate method called `func failedToLoadEpisode(_ episodeUuid: String)` and replace this with `self?.delegate?.failedToLoadEpisode(episodeUuid)`

podcasts/DiscoverEpisodeViewModel.swift
```diff
@@ -121,51 +121,45 @@ class DiscoverEpisodeViewModel: ObservableObject {
...
+                }
                 self?.delegate?.show(discoverEpisode: episode, podcast: podcast)
             }
             .store(in: &cancellables)
     }
 
     // MARK: Static helpers
 
-    static func loadPodcast(_ podcastUUID: String, ensureEpisodeUuid: String? = nil) -> AnyPublisher<Podcast?, Never> {
+    static func loadPodcast(_ podcastUUID: String, episodeUuid: String) -> AnyPublisher<Podcast?, Never> {
```
emilylaguna, line 143: There is a few different places where this is called, but we're only handling the failure by showing the alert in 1 place, this means nothing happens when:

- I tap on the play button and the episode can't be loaded
- I am on the Episode list view (Featuring Keanu Reeves list) and tap on an episode and the episode can't be loaded

We should also update the other areas this is called to show relevant alerts.

podcasts/DiscoverEpisodeViewModel.swift
```diff
@@ -121,51 +121,45 @@ class DiscoverEpisodeViewModel: ObservableObject {
...
+                DiscoverEpisodeViewModel.ensureEpisodeExists(podcast: podcast, episodeUuid: episodeUuid) { exists in
+                    guard exists else {
+                        promise(.failure(.episodeNotFound))
+                        return
                     }
-                    return
+
+                    promise(.success(podcast))
                 }
+            }
```
emilylaguna: Hmm.. Looking at this a bit more now, it feels like all we're doing is wrapping a handler around `ensureEpisodeExists`. What do you think about:

1. Deleting this nested func
2. Replacing the completion argument in `ensureEpisodeExists`  with `promise: @escaping (Result<Podcast?, ClientError>) -> Void`
3. Passing the promise directly into the call
4. Replacing the completion calls with promise ones

This will simplify the code a bit more. I made an example branch with this here: https://github.com/Automattic/pocket-casts-ios/compare/task/58-missing-episode-from-discover-view...emily/271-idea?expand=1

podcasts/DiscoverSummaryProtocol.swift
```diff
@@ -6,7 +6,7 @@ protocol DiscoverSummaryProtocol: AnyObject {
     func populateFrom(item: DiscoverItem)
 }
 
-protocol DiscoverDelegate: AnyObject {
+protocol DiscoverDelegate: UIViewController {
```
emilylaguna: In general there isn't much of a need to specify a delegate protocol to be a specific type and it limits the way you can use it in the future.

emilylaguna (changes requested):
Hey @jgcaruso! Nice work as always! While testing I noticed that when I tapped or tried to play some of the failed episodes I wasn't getting an alert, I left a comment about that below.

I also left some comments on how we can improve the code some more. Please let me know what you think.

I also made an example branch with some of the requested changes here:  https://github.com/Automattic/pocket-casts-ios/compare/task/58-missing-episode-from-discover-view...emily/271-idea?expand=1

podcasts/DiscoverEpisodeViewModel.swift
jgcaruso: Oooh right because of `.receive(on: RunLoop.main)`

podcasts/DiscoverEpisodeViewModel.swift
jgcaruso: > Passing the promise directly into the call

I tried this initially but couldn't quite figure it out, which is why I ended up with the extra function with the completion call. Thanks for pointing this out though, I'm glad it is possible.

I really appreciate the example branch! It has made a lot of these solutions very clear.

podcasts/DiscoverSummaryProtocol.swift
jgcaruso: I think I did this because passing the delegate into `SJUIUtils.showAlert` was complaining about the type. But that should have been a clue I was doing it wrong.

The new delegate methods you suggested fix this up.

podcasts/DiscoverEpisodeViewModel.swift
```diff
@@ -90,15 +90,20 @@ class DiscoverEpisodeViewModel: ObservableObject {
         AnalyticsHelper.listImpression(listId: listId)
     }
 
-    public func didSelectPlayEpisode() {
+    public func didSelectPlayEpisode(completion: ((Bool) -> Void)? = nil) {
```
emilylaguna: I hit an issue on the main discover screen, when tapping the play button and it failing the play button isn't reset. 
I think we can just pass the button directly into this function and reset its state on failure as a simple solution:

```suggestion
    public func didSelectPlayEpisode(from button: PlayPauseLabeledButton) {
```

https://user-images.githubusercontent.com/793774/197885952-6f26fe3c-7e71-47c0-8984-b9c2bf416056.mp4

emilylaguna (changes requested):
Nice!! Once the play button issue is fixed we can get this merged in!

Awesome work!

emilylaguna (approved):
Nice work! :shipit:

PR #282, opened 2022-09-27
Add and configure SwiftLint

mokagio:
This PR adds SwiftLint, but to do so, it first updates SwiftFormat and centralizes the Swift version definition in the Xcode project. These three steps could have been dedicated PRs but they ended up into such a small diff that it made sense to leave them in a single place.

Our [standard SwiftLint configuration](https://github.com/Automattic/swiftlint-config) picked up a lot of issues, the majority of them were trailing whitespaces. I separated fixing them in #283 .

**⚠️ This PR doesn't address the question of whether having _both_ SwiftLint and SwiftFormat is a good idea! ⚠️**

I was interested in implementing side-by-side support to get an idea of the work involved. From my experience here, I'd say running both tools is redundant and will only lead to issues when their default configurations diverge (see the many force pushes to this and the autocorrect application branch).

I think we should stick with SwiftLint only, because of the remote configuration with overrides support, but keen to hear what other folks think. Removing SwiftFormat would be straightforward and can be done in a dedicated PR later on.

## To test

Run `make swiftformat` and verify nothing changes. That is, the changes haven't broken or affected its setup. _CI already does this for us._

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary. – n.a.
- [x] I have considered adding unit tests for my changes. – n.a.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics. – n.a.


AliSoftware (commented):
Code changes look good to me, but:
 - I agree that it's debatable to keep SwiftFormat in addition to SwiftLint, as their rules could conflict, which is going to make the life of developers harder, not easier. Especially if both are run automatically in parallel
 - I'm not completely against keeping `swiftformat` around configured in the project, but if we do so:
    - I'd configure it in a CI step — so it scans the whole codebase not just the files changed — instead of Dangerfile?
    - Or more likely, I'd remove it as a CI check and from the Dangerfile, and instead only keep it in the `Makefile`, to be run manually by developers only on demand; keeping `swiftlint` as the main source of check and truth for the code styling rules we want to enforce.
    - We could also consider adding an entry in the `Makefile` to make running `swiftlint autocorrect` easier on demand, slowly pushing devs towards using `swiftlint autocorrect` instead of `swiftformat` when they want to auto-fix code styling issues on demand, in the hope that we could then remove `swiftformat` and only lean on `swiftlint` in the future — and thus only have a single source of truth for the code styling rules instead of conflicting ones.

leandroalonso (approved):
> I think we should stick with SwiftLint only, because of the remote configuration with overrides support, but keen to hear what other folks think. Removing SwiftFormat would be straightforward and can be done in a dedicated PR later.

Ha! @emilylaguna and I was talking about that a few days ago. We prefer to have only Swiftlint, so we're 100% in removing Swiftformat.

leandroalonso:
@mokagio I'm going to merge this PR before the next code freeze to be able to the other PR with the auto-corrections. Otherwise, we can possibly run into the issue of the release branch and `trunk` having some nasty conflicts.

PR #348, opened 2022-10-04
Configure Danger to run SwiftLint

mokagio:
~~**Update**: Still a draft. For some reason, the plugin is linting files from the Pods. I haven't figured out how to prevent this.~~ Solved via https://github.com/Automattic/pocket-casts-ios/pull/349#issuecomment-1267815239.

We already have Danger setup here, and as you can see making it run SwiftLint was straightforward.

However, I do wonder if this is worth doing at all. It's nice to have Danger comment on the PR, but, as of #349, CI is already running SwiftLint as part of the build process. I wonder if it would be better to augment the `make lint` so that it can fail the build or what not if in CI.

The conversation about the extent to which to run SwiftLint and other tools in CI and whether to include them in Danger is one that span across all our mobile projects. My suggestion here  is  to adopt the setup from this PR for the time being.

## To test

See #357, which shows Danger leaving a comment:

image

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary. – N.A.
- [x] I have considered adding unit tests for my changes. – N.A.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics. – N.A.

PR #349, opened 2022-10-04
Run SwiftLint as part of the PreBuildActions target

mokagio:
This is neat because it runs SwiftLint _before_ building the app, delivering fast feedback and, if/when there'll be `--strict` mode, even faster build failures.

I decided to split this PR from the one that runs SwiftLint via Danger in CI, #348, so that changes in one won't block the other.

The setup I propose is to run SwiftLint on every build but without the `--autocorrect` (`make format`) mode. Running the tool to also reformat the entire source code would _flush the Xcode undo history_. A possible semi-workaround is to run it with the unit tests, which reduces the reformat occurrences but does not address the undo history loss.

Keen to hear what @Automattic/pocket-casts-ios think about this and keen to iterate.

## To test

You can test this by checking out this branch and building the app:

image

Notice that those warnings will go away when #348 lands in `trunk`. In there, I disabled them via the config file `disabled_rule` option. 

## Checklist

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,2 +1,2 @@
-parent_config: https://raw.githubusercontent.com/Automattic/swiftlint-config/16560cc1db19447a7cd9ddf268b5e8290f153a6c/.swiftlint.yml
+parent_config: https://raw.githubusercontent.com/Automattic/swiftlint-config/0f8ab6388bd8d15a04391825ab125f80cfb90704/.swiftlint.yml
 remote_timeout: 10.0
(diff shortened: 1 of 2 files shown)


leandroalonso (approved):
Tested and it works as described!

> Keen to hear what https://github.com/orgs/Automattic/teams/pocket-casts-ios think about this and keen to iterate.

I'd rather prefer to run `make format` manually even if we didn't have the undo issue. Having it is just another reason to not do it. :P

leandroalonso:
@mokagio the unit tests are failing on CI. 🤔

mokagio:
@leandroalonso thanks for the ping @leandroalonso 

Interesting... The failures are because of SwiftLint violations that don't appear when testing locally via Xcode:

image


_Looking into it..._

mokagio:
The issue turned out to be that CI stores the CocoaPods cache in the project folder resulting in additional files being linted.

This configuration change addresses the problem: https://github.com/Automattic/swiftlint-config/commit/16560cc1db19447a7cd9ddf268b5e8290f153a6c. Here's the CI build I used to verify it: https://buildkite.com/automattic/pocket-casts-ios/builds/1424#0183a5d2-0787-452a-bc37-23ba9ceac22f.

Enabling auto-merge, as everything should work now.

mokagio:
_I knew when I wrote "everything should work now" I would have regretted it. Still failing 🤔_

PR #359, opened 2022-10-05
Migrate Crash reporting from Crashlytics to Sentry

### LINKED ISSUE #341: Migrate Crash reporting from Crashlytics to Sentry
### Overview
On Android, we recently made the switch from Crashlytics to Sentry. In an effort to keep the tooling consistent within the product and within A8C the iOS team decided to scope out the work to make the same move for iOS.

In my experience, having two crash logging libraries active at the same time can be a recipe for trouble, so I’m recommending that we do a clean cut from Crashlytics to Sentry.

Like Crashlytics, Sentry has [limited symbolication support and no crash handling](https://docs.sentry.io/platforms/apple/guides/watchos/) for WatchOS. So this project won’t do anything for our watch app.

### Tasks
- [x] Prepare a Sentry project for the Pocket Casts iOS app
- ~[ ] Create a Debug Developer-focused screen with a force crash button for testing.~
- [x] Remove Crashlytics
- [x] Integrate Sentry from https://github.com/Automattic/Automattic-Tracks-iOS using the [CrashLogging](https://github.com/Automattic/Automattic-Tracks-iOS/blob/f10e463b59358cc8391746247e511cdcad6b9e92/Sources/Remote%20Logging/Crash%20Logging/CrashLogging.swift) wrapper.
    - This should consider Opt-Out
    - Add necessary credential items 
- [x] Update the release scripts to upload the dsyms to sentry
    - We should be able to use the [upload_symbols_to_sentry](http://docs.fastlane.tools/actions/upload_symbols_to_sentry/#upload_symbols_to_sentry) action and replace our make command ([1](https://github.com/Automattic/pocket-casts-ios/blob/7786534abfbaf48c07d6a7c3d7c1be37deac7a2d/fastlane/Fastfile#L148), [2](https://github.com/Automattic/pocket-casts-ios/blob/7786534abfbaf48c07d6a7c3d7c1be37deac7a2d/fastlane/Fastfile#L415))

chipsnyder:
Fixes #341 

Removes Crashlytics in favor of Sentry.

## To test
You'll need to update your APICredential file to proceed. 

1.) Switch your scheme to "Release"
2.) Add a `fatalError` exception to the app 
3.) Build and run.
4.) Stop the run.
5.) Start the app again by selecting the icon on the home screen.
6.) Execute the code that has the `fatalError` or perform a [crash](https://github.com/Automattic/pocket-casts-ios/issues?q=is%3Aopen+is%3Aissue+label%3Acrash) repro steps.
7.) Check Sentry for the crash and expect Crashlytics to not have the crash.

**Opt-out**
1.) Set privacy to Opt-out.
2.) Perform the steps above
3.) Expect Sentry to not contain the crash. 

## Checklist


.configure
```diff
@@ -1,7 +1,7 @@
 {
   "project_name": "pocketcasts-ios",
-  "branch": "trunk",
-  "pinned_hash": "54932eb7c27f4bbdfa69467360e3b07e601b117b",
+  "branch": "pc/sentry",
```
chipsnyder: TODO: This needs to be updated to `trunk` before merging.

chipsnyder:
This is ready for review although we may want to time the merge with https://github.com/Automattic/pocket-casts-ios/pull/362 but that's not a hard requirement since dsyms can be uploaded manually.

podcasts/Analytics/Adapters/Tracks/CrashLoggingDataProvider.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+import AutomatticRemoteLogging
+
+class CrashLoggingDataProvider: AutomatticRemoteLogging.CrashLoggingDataProvider {
+    let sentryDSN = ApiCredentials.sentryDSN
+    var userHasOptedOut: Bool {
+        Settings.analyticsOptOut()
```
leandroalonso: I don't think we want the Analytics opt-out to work for the crash analytics. we should have two different toggles: one for Analytics and another for crashes (check WPiOS for an example).

As an opt-out already doesn't exist for Firebase, I'd recommend removing opt-out in this PR and tackling in another one.

podcasts/Analytics/Adapters/Tracks/TracksAdapter.swift
```diff
@@ -44,7 +48,7 @@ class TracksAdapter: AnalyticsAdapter {
         self.userDefaults = userDefaults
         self.subscriptionData = subscriptionData
         self.notificationCenter = notificationCenter
-
+        self.crashLogging = try? CrashLogging(dataProvider: CrashLoggingDataProvider()).start()
```
leandroalonso: Similar to the comment above. Adding this call here couples the opt-out from Analytics to the crashes.

leandroalonso (changes requested):
Works as described! I left some comments about the opt-out. :)

chipsnyder:
> I left some comments about the opt-out.

That works for me. I wasn't sure how deep we needed to go with the privacy settings. I'll decouple it and remove opt-out like you suggest. We can easily add it back in follow-up PRs.

podcasts/Analytics/Adapters/Crash Logging/CrashLoggingAdapter.swift
```diff
@@ -0,0 +1,12 @@
...
+import AutomatticRemoteLogging
+
+class CrashLoggingAdapter: AnalyticsAdapter {
+    let crashLogging: CrashLogging?
+
+    init() {
+        self.crashLogging = try? CrashLogging(dataProvider: CrashLoggingDataProvider()).start()
+    }
+
+    func track(name: String, properties: [AnyHashable: Any]?) { }
```
chipsnyder, line 11: Left as a no-op, for now, to match what Firebase was doing. If we'd like to add manual error logging levels we can do that using [these APIs](https://github.com/Automattic/Automattic-Tracks-iOS/blob/f21fcea46c2063840cccd19106fb70ab0ab6c680/Sources/Remote%20Logging/Crash%20Logging/CrashLogging.swift#L139-L229). We would probably want to identify the appropriate levels for the messages we're sending though.

chipsnyder:
@leandroalonso those changes have been applied now

leandroalonso (approved):

PR #360, opened 2022-10-06
Tracks: Fix some events being fired on app launch/install

emilylaguna:
This fixes:
- Some settings changed events being triggered on app install
- Some payment events being triggered when there is a renewal
- On reinstall the email was not being removed from the keychain which resulted in the "weird state" logs

### Settings changed
1. Enable the `tracksLoggingEnabled` feature flag
2. Delete the app
3. Launch it
4. ✅ Verify you do not see the `Sync account is in a weird state, logging user out` log in console
5. ✅ Verify you do not see any settings changed events
6. Go to the General Settings and change the Multi Select gesture enabled and verify you see the event in console
7. Go to the Appearance settings and change the "Use iOS Light/Dark Mode" and verify you see the event in console
8. Go to the Auto Archive settings and change the Starred/Inactive after and verify you see the event in console
9. Go to the Download settings and change the Up Next and Only on Wifi option
10. Go to the Storate and Data use and toggle the warn before using mobile data option 
11. Go to the settings for a Podcast and toggle the auto download enabled option

### Payment
1. Enable the `tracksLoggingEnabled` feature flag
2. Launch the app
3. Sign into an account without plus
4. Purchase the Monthly version of plus using your sandbox account
12. ✅ Verify after purchasing you see the `purchase_successful` event in console
13. Go back to the main app
14. Wait 5 minutes
15. ✅ Verify you do not see any purchase_successful events being fired again

## Checklist

### DIFF
--- a/podcasts/AppDelegate+Defaults.swift
+++ b/podcasts/AppDelegate+Defaults.swift
@@ -16,7 +16,7 @@ extension AppDelegate {
 
             Settings.setShouldDeleteWhenPlayed(true)
             Settings.setHomeFolderSortOrder(order: .dateAddedNewestToOldest)
-            Settings.setMobileDataAllowed(true, userInitiated: false)
+            Settings.setMobileDataAllowed(true)
             setWhatsNewAcknowledgeToLatest()
         }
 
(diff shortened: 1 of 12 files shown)


podcasts/AppDelegate+Defaults.swift
```diff
@@ -16,15 +16,15 @@ extension AppDelegate {
 
             Settings.setShouldDeleteWhenPlayed(true)
             Settings.setHomeFolderSortOrder(order: .dateAddedNewestToOldest)
-            Settings.setMobileDataAllowed(true)
+            Settings.setMobileDataAllowed(true, userInitiated: false)
```
leandroalonso: What do you think about doing the inverse?

The default value of `userInitiated` is `false`. If it's user-initiated this needs to be explicit. I think it might prevent calling these functions and triggering the tracks by mistake.

leandroalonso (approved):

podcasts/AppDelegate+Defaults.swift
emilylaguna: Done in: 2b800b1409112bdcb4376242a861aba0b6eb8394

PR #362, opened 2022-10-06
Add Sentry Dsyms upload Steps to CI Release Cycle

### LINKED ISSUE #341: Migrate Crash reporting from Crashlytics to Sentry

chipsnyder:
Fixes #341 

| Depends on https://github.com/Automattic/pocket-casts-ios/pull/359 |
|:---:|

Adds the DSYM upload steps to the release cycle. 

## To test

1. Set the environment variable `SENTRY_AUTH_TOKEN` with an appropriate auth token
2. Set up a test lane like:
```
lane :upload_test do |options|
    sentry_check_cli_installed
    configure_code_signing

    gym(
      scheme: 'pocketcasts',
      include_bitcode: false,
      include_symbols: true,
      clean: true,
      export_options: { method: 'app-store' }
    )

    symbols_upload
end
```
3. run your test lane `bundle exec fastlane upload_test`
4. Verify the Dsyms made it to sentry. 

## Checklist


chipsnyder:
@leandroalonso @mokagio I'm not sure what the best way to test this would be since it creates a build that is uploaded to App Store Connect and TestFlight. Also, for a proper test, this should probably be run on the CI. 

How have you guys been testing your release script changes? 

-----
The build also seems to be erroring, possibly because of the Gemfile changes. I'll look at that closer soon.

leandroalonso:
> How have you guys been testing your release script changes?

It depends. ™️ 

In this case what I believe you can do is:

`source source ~/.mobile-secrets/CI/secrets/pocket-casts-ios/env`

And then:

`bundle exec fastlane symbols_upload`

For this, you'll need to have the dSYMs for the current build, of course.

> Also, for a proper test, this should probably be run on the CI.

We should be able to run all the lanes in our machines IMHO. I'm not sure though exactly how `lane_context` works (@mokagio knows for sure). It looks like Woo references the dSYM directly: https://github.com/woocommerce/woocommerce-ios/blob/7942619547e49c0675b46bdc0bda1b2f8e4fd3b1/fastlane/Fastfile#L461

Not sure what's the difference between the two approaches.

Gemfile
```diff
@@ -10,6 +10,7 @@ gem 'danger-rubocop', '~> 0.10'
 gem 'danger-swiftlint'
 gem 'fastlane', '~> 2.210.0'
 gem 'fastlane-plugin-wpmreleasetoolkit', '~> 5.4'
+gem 'fastlane-plugin-sentry'
```
mokagio: It would be great to specify a version rule for this, even if we don't have any explicit requirement.

I'd suggest basing it off the requiring the current version: 

```ruby
gem 'fastlane-plugin-sentry', '~> 1.14'
```

I didn't leave this as a `suggestion` comment because applying this change will also require running `bundle install` to update the `Gemfile.lock` accordingly.

fastlane/Fastfile
```diff
@@ -133,6 +135,7 @@ platform :ios do
   lane :beta do
```
mokagio: I could be wrong, but I don't think this `beta` is unused. 

If that's the case, we could omit updating it, which would simplify the rest of the code by removing the need to extract the call to the Sentry action to upload dSYMs into its own lane.

We can then open another PR to remove the unused lane.

Gemfile
chipsnyder: We may want to consider making a similar change in [WordPress-iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/db0d9cc52057647aceea48aee0bf062a56769d18/Gemfile#L12).

fastlane/Fastfile
```diff
@@ -931,6 +934,16 @@ inal copy, and try again.")
...
 
+  desc 'Uploads DSYM Symbols'
+  lane :symbols_upload do
+    sentry_upload_dsym(
+      auth_token: get_required_env('SENTRY_AUTH_TOKEN'),
+      org_slug: SENTRY_ORG_SLUG,
+      project_slug: SENTRY_PROJECT_SLUG,
+      dsym_path: lane_context[SharedValues::DSYM_OUTPUT_PATH]
+    )
+  end
```
mokagio: Nitpick.

By defining this functionality in a `lane` we make it possible to call it individually with `bundle exec fastlane symbols_upload`. However, doing so would fail because there wouldn't be any `lane_context[SharedValues::DSYM_OUTPUT_PATH]` in memory.

I can thing of three alternatives:

- Make this a `private_lane` so that it cannot be called individually; or
- Make this a method and require the path `def upload_symbols(dsym_path:) ... end`. This will make it impossible to call in isolation, and also require the parameter to be passed to it if we'll ever call it somewhere else; or
- Inline this at the call site, since it's only used there

My favorite would be inlining because we don't need the extra abstraction right now.

mokagio: I didn't see @leandroalonso's [comment](https://github.com/Automattic/pocket-casts-ios/pull/362#issuecomment-1270590071) when I posted the one above,

> I'm not sure though exactly how `lane_context` works (@mokagio knows for sure).	

[`lane_context` is a Fastlane mechanism to allow actions to communicate with each other.](http://docs.fastlane.tools/advanced/lanes/#lane-context) It's a place where to dump information _at runtime_ for other components of one's pipeline to read.

mokagio:
@leandroalonso @chipsnyder 

> How have you guys been testing your release script changes?

Testing release scripts is a tricky business because they are inherently deterministic and destructive (as in they alter the App Store Connect and/or GitHub state in ways that sometimes cannot be reversed. The only way to have a genuine test it to make a release 🙃 

Having said that, there's ways to test as much of the flow as possible and @chipsnyder's suggestion to create a test lane from the [PR description](https://github.com/Automattic/pocket-casts-ios/pull/362#issue-1400167626) is spot on.

Something else I often do is add an early return to a lane I'd like to test. In this case, we could:

1. Move the `symbols_upload` call in `build_and_upload_app_store_connect ` between `gym` and `testflight`; then
2. Add `UI.user_error! 'Testing – Early exit'` after it

---

> We should be able to run all the lanes in our machines IMHO. – @leandroalonso 

Yes. Totally. In fact, we _have to_, in case Buildkite is down when we need to ship an urgent release.

mokagio (approved):

leandroalonso:
@chipsnyder do you plan on changing `lane_context[SharedValues::DSYM_OUTPUT_PATH]` for the path itself? I think it would be better in case we want to run that locally. :)

chipsnyder:
> do you plan on changing lane_context[SharedValues::DSYM_OUTPUT_PATH] for the path itself?

I was actually just looking into this. I was leaning toward making it a `private_lane`, but you're right it might still have value to run on its own. I'll make it configurable to allow a `dsyms_path`

chipsnyder:
@leandroalonso @mokagio https://github.com/Automattic/pocket-casts-ios/pull/362/commits/1e56dc1f2c85230eb255a07c2b71e46ebab5f806 adds support to now run the `symbols_upload` lane via

`bundle exec fastlane symbols_upload dsym_path:{YOUR_PATH}`

If that option isn't available, it falls back to the `lane_context`

fastlane/Fastfile
```diff
@@ -931,6 +934,17 @@ inal copy, and try again.")
     File.write("#{PROJECT_ROOT_FOLDER}/podcasts/acknowledgements.html", styled_html)
   end
 
+  desc 'Uploads DSYM Symbols'
+  lane :symbols_upload do |options|
+    symbols_path = options[:dsym_path].nil? ? lane_context[SharedValues::DSYM_OUTPUT_PATH] : options[:dsym_path]
```
leandroalonso: You can use this syntax:

```suggestion
    symbols_path = options[:dsym_path] || lane_context[SharedValues::DSYM_OUTPUT_PATH]
```

fastlane/Fastfile
```diff
@@ -931,6 +934,17 @@ inal copy, and try again.")
...
   end
 
+  desc 'Uploads DSYM Symbols'
+  lane :symbols_upload do |options|
+    symbols_path = options[:dsym_path] || lane_context[SharedValues::DSYM_OUTPUT_PATH]
+    sentry_upload_dsym(
+      auth_token: get_required_env('SENTRY_AUTH_TOKEN'),
+      org_slug: SENTRY_ORG_SLUG,
+      project_slug: SENTRY_PROJECT_SLUG,
+      dsym_path:  symbols_path
```
pocketcasts: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-ios/blob/2420d79ae3fd6efa3591c9aae07c5e1ada0e5953/fastlane/Fastfile#L944"></span>Align the keys of a hash literal if they span more than one line.</td>
    </tr>
  </tbody>
</table>

PR #373, opened 2022-10-10
Explicitly prevent Xcode from bumping the build number

mokagio:
Without any apparent change on our side, we saw a couple of builds with unexpected, and unexpectedly incrementing, build numbers on TestFlight:

image

This behavior can be replicated locally by running `bundle exec fastlane build_and_upload_to_app_store_connect` but not when archiving via Xcode.

I pin pointed it to the `manageAppVersionAndBuildNumber` option being `true` by default in the `ExportOptions.plist` that Fastlane / `xcodebuild` generates for `gym` to convert an archive into an IPA.

Explicitly setting the option to `false` solves the issue locally. The next step is to merge this and to see if App Store Connect allows us to upload a binary with version `7.24.0.4` after one with version `9`. It might not, but at least we address the source of the build number being different from what we set.


mokagio:
I'm going to admin-merge this so that we can move on with the deployment.

PR #387, opened 2022-10-12
Apple SSO: Setup

chipsnyder:
| 📘 Project: #381 |
|:---:|

Adds the baseline for the Sign in with Apple Flow by exposing the Sign In with Apple Button when the feature is enabled. 

We're using the standard Sign In with Apple button, so some design details differ from what's in Figma. 

## To test
- Enable the Feature Flag `signInWithApple`
- Navigate to the Log in screen
- Look for a continue with Apple Button
- Check that Login takes you to the Password Login flow
- Disabling the Feature should show the same options as prod. 

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

Issue #381, opened 2022-10-11
Project: SSO


chipsnyder:
This issue is to help track the work to be done to enable Social Sign In

- [x] Google Sign In #649 
- [x] Token refresh mechanism #650 
- [x] Sign In With Apple #652 
- [x] Apple Watch #656
- [x] Sonos #667
- [x] Events #679 
- [x] Hide password reset and email change #672

## Additional

- [x] The "You've been logged out" triggers the previous login flow #674 
- [x] Remove SIWA authentication revoked (we should not log out the user) - Done as part of #656
- [ ] Memory leak when the onboarding flow dismisses automatically (without user interaction)

emilylaguna:
Closing this as the project has been launched.

PR #388, opened 2022-10-12
Apple SSO: Happy Path

chipsnyder:
| 📘 Project: #381 | Depends on: https://github.com/Automattic/pocket-casts-ios/pull/387 |
|:---:|:---:|

Connects the Continue with Apple Button to the login flow for connecting an Apple ID to an existing Sync Account.

📓 This doesn't handle Errors yet and will need to handle [Token refreshes](https://github.com/Automattic/pocket-casts-ios/blob/33a541ed0af33691a6d1b462dc84a1a1956ad0c0/Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift#L53) still those will come in a follow-up PR. 

📓 You'll need a Sync account with the same email address as your Apple ID in the Staging environment to test this.

## To test
2. Switch your Scheme to `Staging`
3. Navigate to the Log in screen
4. Complete the Login flow and select to Share your email (hiding email will fail)

Expect to:
- Be logged in
- The log in screen to dismiss
- Your podcasts to sync to the device. 

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
@@ -200,6 +200,8 @@ struct Api_UserLoginResponse {
 
   var uuid: String = String()
 
+  var email: String = String()
+
   var unknownFields = SwiftProtobuf.UnknownStorage()
 
   init() {}
@@ -4956,6 +4958,7 @@ extension Api_UserLoginResponse: SwiftProtobuf.Message, SwiftProtobuf._MessageIm
   static let _protobuf_nameMap: SwiftProtobuf._NameMap = [
     1: .same(proto: "token"),
     2: .same(proto: "uuid"),
+    3: .same(proto: "email"),
   ]
 
   mutating func decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
@@ -4966,6 +4969,7 @@ extension Api_UserLoginResponse: SwiftProtobuf.Message, SwiftProtobuf._MessageIm
       switch fieldNumber {
       case 1: try { try decoder.decodeSingularStringField(value: &self.token) }()
       case 2: try { try decoder.decodeSingularStringField(value: &self.uuid) }()
+      case 3: try { try decoder.decodeSingularStringField(value: &self.email) }()
       default: break
       }
     }
@@ -4978,12 +4982,16 @@ extension Api_UserLoginResponse: SwiftProtobuf.Message, SwiftProtobuf._MessageIm
     if !self.uuid.isEmpty {
       try visitor.visitSingularStringField(value: self.uuid, fieldNumber: 2)
     }
+    if !self.email.isEmpty {
+      try visitor.visitSingularStringField(value: self.email, fieldNumber: 3)
+    }
     try unknownFields.traverse(visitor: &visitor)
   }
 
   static func ==(lhs: Api_UserLoginResponse, rhs: Api_UserLoginResponse) -> Bool {
     if lhs.token != rhs.token {return false}
     if lhs.uuid != rhs.uuid {return false}
+    if lhs.email != rhs.email {return false}
     if lhs.unknownFields != rhs.unknownFields {return false}
     return true
   }
--- a/Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
@@ -124,7 +124,27 @@ public extension ApiServerHandler {
         }
     }
 
-    private class func extractErrorResponse(data: Data?, error: Error? = nil) -> APIError? {
+    internal func obtainToken(request: URLRequest, completion: @escaping (Result<AuthenticationResponse, APIError>) -> Void) {
+        urlSession.dataTask(with: request) { data, response, error in
+            guard let responseData = data, error == nil, response?.extractStatusCode() == ServerConstants.HttpConstants.ok else {
+                let errorResponse = ApiServerHandler.extractErrorResponse(data: data, error: error)
+                FileLog.shared.addMessage("Unable to obtain token, status code: \(response?.extractStatusCode() ?? -1), server error: \(errorResponse?.rawValue ?? "none")")
+                completion(.failure(errorResponse ?? .UNKNOWN))
+                return
+            }
+
+            do {
+                let response = try Api_UserLoginResponse(serializedData: responseData)
+                completion(.success(AuthenticationResponse(from: response)))
+            } catch {
+                FileLog.shared.addMessage("Error occurred while trying to unpack token request \(error.localizedDescription)")
+                completion(.failure(.UNKNOWN))
+            }
+
+        }.resume()
+    }
+
+    internal class func extractErrorResponse(data: Data?, error: Error? = nil) -> APIError? {
         if let data = data {
             do {
                 let errorJson = try JSON(data: data)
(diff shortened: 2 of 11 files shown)


Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
```diff
@@ -124,7 +124,27 @@ public extension ApiServerHandler {
         }
     }
 
-    private class func extractErrorResponse(data: Data?, error: Error? = nil) -> APIError? {
+    internal func obtainToken(request: URLRequest, completion: @escaping (Result<AuthenticationResponse, APIError>) -> Void) {
```
chipsnyder: The [Obtain token flow](https://github.com/Automattic/pocket-casts-ios/pull/388/files#diff-032fe1f2cd3597f5dc28351d8a82ece28bea8ed8a091921206a45088eb1f5968L103) from the Login with password route can be re worked to use this method and consolidate repeat code, but I left that out of this change. I'll add that in its own PR for more focused testing.

Modules/Server/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
```diff
@@ -200,6 +200,8 @@ struct Api_UserLoginResponse {
 
   var uuid: String = String()
 
+  var email: String = String()
```
leandroalonso, line 203: I assume this is available only on Staging, right?

In that case, we probably wouldn't want to merge this change to `trunk`.

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
```diff
@@ -124,7 +124,27 @@ public extension ApiServerHandler {
...
+                FileLog.shared.addMessage("Unable to obtain token, status code: \(response?.extractStatusCode() ?? -1), server error: \(errorResponse?.rawValue ?? "none")")
+                completion(.failure(errorResponse ?? .UNKNOWN))
+                return
+            }
+
+            do {
+                let response = try Api_UserLoginResponse(serializedData: responseData)
+                completion(.success(AuthenticationResponse(from: response)))
+            } catch {
+                FileLog.shared.addMessage("Error occurred while trying to unpack token request \(error.localizedDescription)")
```
leandroalonso, line 140: Shall we add any mention of Apple SSO in the log message?

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
```diff
@@ -124,7 +124,27 @@ public extension ApiServerHandler {
...
+                completion(.failure(.UNKNOWN))
+            }
+
+        }.resume()
+    }
+
+    internal class func extractErrorResponse(data: Data?, error: Error? = nil) -> APIError? {
```
leandroalonso: I believe `internal` is the default modifier, you don't need to include it.

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+SocialAuth.swift
```diff
@@ -0,0 +1,20 @@
...
+import PocketCastsUtils
+
+public extension ApiServerHandler {
+    func validateLogin(identityToken: Data?, completion: @escaping (Result<AuthenticationResponse, APIError>) -> Void) {
+        let url = ServerHelper.asUrl(ServerConstants.Urls.api() + "user/login_apple")
+        guard var request = ServerHelper.createEmptyProtoRequest(url: url),
+              let identityToken = identityToken,
+              let token = String(data: identityToken, encoding: .utf8)
+        else {
+            FileLog.shared.addMessage("Unable to create protobuffer request to obtain token")
```
leandroalonso: Same comment as before (about adding Apple SSO mention).

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler.swift
```diff
@@ -5,6 +5,11 @@ import SwiftProtobuf
 
 public class ApiServerHandler {
     public static let shared = ApiServerHandler()
+    internal let urlSession: URLSession
```
leandroalonso: I wonder if we change this to `private`?

leandroalonso (commented):
@chipsnyder works as described!

I left a few comments though, let me know wdyt.

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
chipsnyder: I would like to [consolidate](https://github.com/Automattic/pocket-casts-ios/pull/388/files/784a9d8e24e038190c824909660d2c06b00fe842#r993796549) the code to prevent duplication because this is copied and pasted from the login flow. Ony the completion block is changed slightly. We could include a log here by passing in a source but perhaps a log at the start of the Apple SSO attempts would provide the context.

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
chipsnyder: You're right It's not necessary, but I like to be explicit in cases like this. Ideally, I would make it `fileprivate` or `private` but that would force me to pull the extensions into the same file. 

I'm happy to drop it though.

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler.swift
chipsnyder: Same idea mentioned [here](https://github.com/Automattic/pocket-casts-ios/pull/388/files/784a9d8e24e038190c824909660d2c06b00fe842#r994747790).

Happy to drop the identifier

Modules/Server/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
chipsnyder: This field is only available on staging, but I think the decision to merge or not might be based on the concern you're seeing. 
I see two possible concerns:

1) Will this impact login if the value isn't available?
- No, the email will default to an empty string and be ignored.

2) What happens if we regenerate the protobuff from `trunk`?
- The developer would get an error, and we can regenerate from Staging, which will include all of prod plus what's in development. I can see this as a reason not to merge, but I think its a small one.

leandroalonso: If everything works as it is I think we're good. :)

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
leandroalonso: I'd just vote for consistency. :)

The other methods don't have `internal` so I think this can lead someone else to think that these methods have a different accessor type. Thoughts?

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler.swift
leandroalonso: Oh, I haven't realized that this is not `private` because it's used in an extension. That seems a wrong reason to have this `internal`. 😬

But refactoring it is out of the scope of this task.

leandroalonso (approved):
Left a few more comments, not blockers. Up to you to change it or not.

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
chipsnyder: Your [comment](https://github.com/Automattic/pocket-casts-ios/pull/388#discussion_r994801832) here probably captures my true feelings on it. 

These methods are different than the others in the sense that I want them more restrictive than internal, so I want to call them out. But I'll drop the accessor, and the variable added [here](https://github.com/Automattic/pocket-casts-ios/pull/388#discussion_r994801832) we can add that variable back when we start adding tests.

Issue #376, opened 2022-10-10
Project: End of Year


leandroalonso:
## Description

Show end of year stats.

### Initial work

- [x] Add modal structure #377
- [x] Add stories structure #383
- [x] Add stories data source #389
- [x] Add profile card structure and badge #393

### Presenting and sharing Stories

- [x] Generate a story with dummy data #397
- [x] Generate an image based on the story shown #404
- [x] Share the generated image through a share sheet #404

### Stories data

- [x] Show total listening time in 2022 #411
- [x] Show number of different categories they’ve listened to #413
- [x] Show top 5 categories and how many podcasts in each of these they’ve listened to #413
- [x] Show total number of podcasts and episodes they’ve listened to. #414
- [x] Show number 1 podcast for 2022 #415
- [x] Show top 5 podcasts of the year #415
- [x] Show longest episode they listened to in 2022 #416
- [x] Show the first and last story (with the option to replay) #420
- [x] Refactor the data source to remove the queries to another place #428
- [x] Check eligibility and if the user has the whole 2022 Listening History, if not, grab their 2022 listening history from the server. #433

### Tracks

- [x] Add tracks #476

### Design

- [x] Finish the modal design and profile card design #441
- [x] Intro story #448 
- [x] Listening time story #452
- [x] Listened categories story #453 
- [x] Top categories story #468 
- [x] Listened podcasts story #454 
- [x] Top 1 podcast story #455 
- [x] Top 5 podcasts story #462 
- [x] Longest episode story #463 
- [x] Epilogue story #467 
- [x] Download colors before presenting podcasts and optimistically download podcasts covers #471
- [ ] Refactor views to reduce code duplication

### Misc

- [x] Implement the flow for logged out users and add a remote feature flag to toggle this behavior #474
- [x] Sharing: text and links #478
- [x] Test on iOS 14/15/16 and iPadOS 14/15/16 #495 

## Listening History Sync

- [x] Sync Listening History #483

## Nice to have

- [ ] Instead of using bottom sheet on iPad, present it as "a card"
- [ ] Add default colors to stories for when the colors don't load

leandroalonso:
### Comments from reviews:

In #452

- [x] Make height/position dynamic
- [x] Display a placeholder while loading podcasts
- [x] Rename `randomPodcasts`

In #454

- [x] When there aren't many podcasts available it looks weird

All addressed in: https://github.com/Automattic/pocket-casts-ios/pull/489

PR #389, opened 2022-10-13
End of Year: add stories Data Source and gestures

leandroalonso:
| 📘 Project: #376 |
|:---:|

This PR adds a data source protocol to populate the Stories view with stories and also adds a few gestures.

https://user-images.githubusercontent.com/7040243/195642253-8da709b6-d2da-43ef-8bdb-743f727fbe45.mp4

## About the data source

I tried to be as generic as possible with how we display stories to allow different cases to be used without much effort.

In order to show stories, we need to pass a data source that conforms to `StoriesDataSource`. Each story is simply a `View`.

This makes it possible for a story to be a custom view, an image, or a video. It's also generic enough to be used for other purposes in the future.

Take a look at the `TestStoriesDataSource` for an example.

The overall API is still in its early days, so we'll change it as the feature gets developed.

## To test

2. Run the app
4. ✅ Check that each story is shown for 5s and that the story indicator (the rectangles at the top) are reflecting the state
5. ✅ Tap and hold, check that the current story is "paused". The rectangle stop progressing and the story doesn't change
6. ✅ While seeing the yellow story, tap the left portion of the screen, you should go back to the purple one
7. ✅ While seeing the purple story, tap the right portion of the screen, you should go to the next story (yellow)
8. ✅ Swipe down and make sure the view is dismissed ¶

¶ Swipe down can be improved with better effects (like with the view following the gesture) but this will be done in another task

## Checklist


podcasts/End of Year/StoriesModel.swift
```diff
@@ -0,0 +1,56 @@
...
+
+class StoriesModel: ObservableObject {
+    @Published var progress: Double
+
+    @Published var currentStory: Int = 0
+
+    private let dataSource: StoriesDataSource
+    private let publisher: Timer.TimerPublisher
+    private var cancellable: Cancellable?
+    private var interval: TimeInterval = 5.seconds
```
emilylaguna, line 12: I wonder if in the future we should move this into the datasource to allow different story views to have different durations?

podcasts/End of Year/StoriesModel.swift
```diff
@@ -0,0 +1,56 @@
...
+    func story(index: Int) -> AnyView {
+        dataSource.storyView(for: index)
+    }
+
+    func next() {
+        progress = Double(Int(progress) + 1)
+    }
+
+    func previous() {
+        progress = Double(Int(progress) - 1)
```
emilylaguna, line 50: If I tap the previous action while on the first view the progress gets set to -1 which causes progress to be delayed by 5 seconds before it starts ticking up again.

https://user-images.githubusercontent.com/793774/195682972-c83ff059-76a0-49b7-8b83-0e1652ea9ecb.mov

podcasts/End of Year/StoriesModel.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    func start() {
+        cancellable = publisher.autoconnect().sink(receiveValue: { _ in
+            var newProgress = self.progress + (0.01 / self.interval)
+
+            let currentStory = Int(newProgress)
+
+            if currentStory >= self.numberOfStories {
+                newProgress = 0
+                self.currentStory = 0
```
emilylaguna: What do you think about offloading this behavior to the datasource to allow it to decide how it wants to handle when we reach the end of the stories?

podcasts/End of Year/Views/StoriesView.swift
```diff
@@ -63,36 +71,41 @@ struct StoriesView: View {
...
+                .onChanged { _ in
+                    model.pause()
+                }
+                .onEnded { value in
+                    let velocity = CGSize(
+                        width: value.predictedEndLocation.x - value.location.x,
+                        height: value.predictedEndLocation.y - value.location.y
+                    )
+
+                    // If a quick swipe down is performed, dismiss the view
```
emilylaguna, line 101: Are we also planning on adding a swipe next/back gesture?

emilylaguna (changes requested):
Nice, this is shaping up! I found a bug with tapping back while on the first story (see video below). 

I also left a few other comments below. Let me know what you think.

podcasts/End of Year/StoriesModel.swift
leandroalonso: Yes, that's the idea! I'll do it in another PR. :)

podcasts/End of Year/StoriesModel.swift
leandroalonso: Good catch! Fixed in e44ffee77f93bb6d011212d0a7cd20f623607c65

podcasts/End of Year/StoriesModel.swift
leandroalonso: This is also another idea I have for subsequent tasks. :)

podcasts/End of Year/Views/StoriesView.swift
leandroalonso: Hmm, is swipe next/back gesture common in stories inside the same "context"?

I know it's a thing when you want to "advance" (to the next user, for example), but not for the next story.

emilylaguna: Good point. I don't think it's needed anymore.

emilylaguna (approved):
Nice work!

PR #392, opened 2022-10-13
Apple SSO: Corrupt Token Failure

chipsnyder:
| 📘 Project: #381 | Depends on: https://github.com/Automattic/pocket-casts-ios/pull/388 |
|:---:|:---:|

This handles the path for when a user's Auth token has been revoked, and the app cannot log them in again. Previously the app navigated the user to the Password login flow. Now, it will take the user to the start of the login flow so they can select their social login provider again. 

## To test

📓  The simulator doesn't always return the email address so please test with a physical device to ensure every thing works as expected 

2. Switch your Scheme to `Staging`
3. Login with SSO 
4. Navigate to the new developer menu
    - Profile > Settings > Developer
    - 🗒️ This Menu will only appear in Debug and Staging
5. Select "Corrupt Sync Login Token" 
6. Trigger an Authenticated request like Refresh
7. Expect the Token to be restored.
8. Corrupt your token again
9. Remove Apple Sign-In from the app.
    - Apple Settings App > Your Name > Password & Security > Apps Using Apple ID > Pocket Casts > Stop Using Apple ID
10. Return to the App 
11. Expect to see the logged out message with the prompt to sign in

12. Tapping Sign-In should take you to step 1 of the authentication flow

## Checklist

### DIFF
--- a/Modules/Server/Package.swift
+++ b/Modules/Server/Package.swift
@@ -30,7 +30,10 @@ let package = Package(
                 "SwiftyJSON"
             ],
             path: "Sources",
-            linkerSettings: [.linkedFramework("CFNetwork", .when(platforms: [.iOS]))]
+            linkerSettings: [
+                .linkedFramework("CFNetwork", .when(platforms: [.iOS])),
+                .linkedFramework("AuthenticationServices", .when(platforms: [.iOS, .watchOS]))
+            ]
         ),
         .testTarget(
             name: "PocketCastsServerTests",
--- a/Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
@@ -51,18 +51,9 @@ class TokenHelper {
     }
 
     class func acquireToken() -> String? {
-        guard let email = ServerSettings.syncingEmail(), let password = KeychainHelper.string(for: ServerConstants.Values.syncingPasswordKey) else {
-            // if the user doesn't have an email and password, they aren't going to be able to acquire a sync token
-            if ServerSettings.syncingEmail() == nil {
-                FileLog.shared.addMessage("Acquire Token was called, however the user has no email address")
-            } else {
-                FileLog.shared.addMessage("Acquire Token was called, and the user has an email, but no password")
-            }
-
-            FileLog.shared.addMessage("Sync account is in a weird state, logging user out")
-            SyncManager.signout()
-
-            return nil
+        guard let email = ServerSettings.syncingEmail(), let password = ServerSettings.syncingPassword() else {
+            // if the user doesn't have an email and password, then we'll check if they're using SSO
+            return acquireIdentityToken()
         }
 
         let url = ServerHelper.asUrl(ServerConstants.Urls.api() + "user/login")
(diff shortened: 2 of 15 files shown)


Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
```diff
@@ -105,4 +109,43 @@ class TokenHelper {
...
+        ApiServerHandler.shared.refreshIdentityToken { result in
+            switch result {
+            case .success(let token):
+                refreshedToken = token
+            case .failure:
+                refreshedToken = nil
+            }
+            semaphore.signal()
+        }
+        semaphore.wait()
```
chipsnyder, line 127: This follows the same Sync flow that the Password option uses. I would like to iterate this to track where the synchronous requirement originates from, but that would extend the scope of this change.

Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
```diff
@@ -90,8 +112,6 @@ class TokenHelper {
 
             if httpResponse.statusCode == ServerConstants.HttpConstants.ok {
                 let token = try Api_UserLoginResponse(serializedData: validData).token
-                KeychainHelper.save(string: token, key: ServerConstants.Values.syncingV2TokenKey, accessibility: kSecAttrAccessibleAfterFirstUnlock)
-
```
chipsnyder, line 94: This is moved to the common acquire token

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+SocialAuth.swift
```diff
@@ -1,20 +1,62 @@
...
+        guard
+            let identityToken = ServerSettings.appleAuthIdentityToken,
+            let request = tokenRequest(identityToken: identityToken, cachePolicy: .reloadIgnoringLocalCacheData, timeoutInterval: 30.seconds)
+        else {
+            FileLog.shared.addMessage("Unable to locate Apple SSO token in Keychain")
+            throw APIError.UNKNOWN
+        }
+
+        if try await hasValidSSOToken() {
+            let response = try await obtainToken(request: request)
```
chipsnyder, line 46: I'm using `async/await` to try and contain the complexity of nested async tasks

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncManager.swift
```diff
@@ -61,5 +61,6 @@ public class SyncManager {
         KeychainHelper.removeKey(ServerConstants.Values.syncingEmailKey)
         KeychainHelper.removeKey(ServerConstants.Values.syncingPasswordKey)
         KeychainHelper.removeKey(ServerConstants.Values.syncingV2TokenKey)
+        KeychainHelper.removeKey(ServerConstants.Values.appleAuthIdentityTokenKey)
```
chipsnyder, line 64: @leandroalonso I just noticed this fails to remove the `appleAuthUserIDKey` so I'll push that change after your review

Modules/Server/Package.swift
```diff
@@ -30,7 +30,10 @@ let package = Package(
                 "SwiftyJSON"
             ],
             path: "Sources",
-            linkerSettings: [.linkedFramework("CFNetwork", .when(platforms: [.iOS]))]
+            linkerSettings: [
+                .linkedFramework("CFNetwork", .when(platforms: [.iOS])),
+                .linkedFramework("AuthenticationServices", .when(platforms: [.iOS, .watchOS]))
```
leandroalonso, line 35: Just wondering: do we really need to include on watchOS?

leandroalonso (approved):
Works as described. In the next PRs might be good to state that testing on a real device is better than on the simulator.

I additionally tested with the flag off and logging in with the good old username/password and tested the token refresh and it's working normal.

Modules/Server/Package.swift
chipsnyder: We could remove the WatchOS requirement and add predef macros to remove the relevant code from the watch. 

However, the watch does have its own session management that might need to use these methods. I'm looking into that flow now in a separate PR.

PR #394, opened 2022-10-13
Apple SSO: Error Display

chipsnyder:
| 📘 Project: #381 | 
|:---:|

Adds an error label to the initial login screen to inform the user that an error has occurred and allows them to choose their recovery path. 

## To test

2. Set your scheme to Debug that will force an error when trying to use the API
3. Login with SSO
4. Expect to see an Error label above the login options. 
5. Select "Log In"
6. Go back
7. Expect the error label to have been cleaned up. 

## Checklist


podcasts/ProfileIntroViewController.swift
```diff
@@ -199,8 +207,14 @@ extension ProfileIntroViewController: ASAuthorizationControllerDelegate {
     func showError(_ error: Error) {
         FileLog.shared.addMessage("Failed to connect SSO account: \(error.localizedDescription)")
         let error = (error as? APIError) ?? .UNKNOWN
-        Analytics.track(.userSignInFailed, properties: ["source": AuthenticationSource.ssoApple.rawValue, "error_code": error.rawValue])
+        var message = L10n.accountSsoFailed
+        if error != .UNKNOWN, !error.localizedDescription.isEmpty {
+            message = error.localizedDescription
```
leandroalonso: I wonder if we should show specific errors on the interface. The regular user probably wouldn't understand it, so a generic "Sign in with Apple failed. Please try again." is better IMHO.

leandroalonso (approved):

podcasts/ProfileIntroViewController.swift
chipsnyder: Sounds good. I was looking to maintain consistency with the other screens, but I'm happy to keep it fully generic.

PR #397, opened 2022-10-14
End of Year: create a dummy story

leandroalonso:
| 📘 Project: #376 |
|:---:|

Creates a "dummy" story based on randomly generated podcasts. This PR also adds the foundation to allow the Data Source to work async — something we'll need when requesting data from the API.

Same as the previous tasks, the designs here are not final and are mainly placeholders.

https://user-images.githubusercontent.com/7040243/195937498-983ee03f-3ff6-4841-98ef-7452193f7af5.mp4

### Story generation with thumbs and dynamic color

2. Run the app
3. When the prompt appears tap "View my 2022"
4. ✅ A list of podcasts should appear ¶
5. Dismiss the view by tapping X or swiping down
6. Tap "View my 2022" again
7. ✅ Check the displayed podcasts
8. You can check with multiple iPhone devices/simulator and podcasts to see the result

¶ If you don't see any podcasts, tap some podcasts on Discover or login.

### Async loading

Change `isReady` on `EndOfYearStoriesDataSource.swift` to:

```swift
    func isReady() async -> Bool {
        let delay = UInt64(5 * 1_000_000_000)
        try? await Task<Never, Never>.sleep(nanoseconds: delay)
        return true
    }
```

1. Run the app
3. When the prompt appears tap "View my 2022"
4. ✅ A loading indicator should appear
4. ✅ After 5 seconds the stories should appear

## Checklist


podcasts/End of Year/Stories/DummyStory.swift
```diff
@@ -0,0 +1,63 @@
...
+
+    var tintColor: Color {
+        .white
+    }
+
+    var body: some View {
+        ZStack {
+            backgroundColor.ignoresSafeArea()
+
+            VStack {
```
leandroalonso, line 21: A lot of constants here can be extracted to their own constants.

However, given that this is an early and dummy version I wonder if that's really needed.

emilylaguna (approved):
LGTM!

PR #399, opened 2022-10-17
Remove legacy localization automation

mokagio:
It was still part of the project because it had to be used to solve an issue during the previous, 7.23 and 7.24, release cycles.


leandroalonso:
@mokagio I'll keep this open for now just in case we need to revert back to it (hopefully we won't need to do that). :)

mokagio:
@leandroalonso @emilylaguna just checking how the 7.25 release cycle and 7.26 code freeze went and wether we can merge this 😄

leandroalonso (approved):

leandroalonso:
@mokagio Merging it!

PR #400, opened 2022-10-17
Tooling: for beta and code freeze, submit the PR using a "merge branch"

leandroalonso:
During the release of 7.24, we ran into an issue where the release branch conflicted with `trunk`.

The conflict was somewhat simple and I solved it using GitHub until I realized this led to a mistake: `trunk` had been merged into the release branch.

We want the release process to be able to avoid such mistakes, so I changed the release process to do this:

* For the final release and hotfix: the PR will be created from the release branch. Eg.: `release/7.25` or `release/7.25.1`
* For intermediate beta and code freeze: the PR will be created from a "merge release branch". Eg.: `merge/release-7.25-into-trunk`

In this case, if a conflict appears, we can easily fix it and not run into the issue we had last week.

## To test

Not much to test, just check the code.

Steps to test your PR.

## Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -299,6 +299,10 @@ platform :ios do
     "release/#{ios_get_app_version}"
   end
 
+  def merge_branch_name
+    "merge/release-#{build_version}-into-trunk"
+  end
+
   # Triggers a beta build on CI
   #
   # @option [String] branch_to_build The name of the branch we want the CI to build, e.g. `release/19.3`
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -956,6 +956,8 @@ inal copy, and try again.")
   def push_release_branches
     [merge_branch_name, release_branch_name].each do |branch_name|
       push_to_git_remote(local_branch: branch_name, remote_branch: branch_name, tags: false)
+    rescue StandardError
+      UI.message("#{branch_name} doesn't exist. Skip pushing it.")
     end
   end
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -460,6 +460,8 @@ platform :ios do
     lint_localizations
 
     ios_bump_version_beta
+    create_merge_release_branch
+    push_release_branches
 
     trigger_beta_build(branch_to_build: release_branch_name)
   end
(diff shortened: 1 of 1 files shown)


leandroalonso:
@emilylaguna @chipsnyder once you have a chance can you please review this PR? I'll only start the code freeze after it has been merged.

@mokagio please feel free to suggest any changes and I can do it in another PR. :)

chipsnyder (approved):
LGTM

fastlane/Fastfile
```diff
@@ -416,7 +421,7 @@ platform :ios do
       body: {
         title: "#{ios_get_app_version} Release: Merge changes from #{ios_get_build_version} into `trunk`",
         body: "Merge changes from #{ios_get_app_version} (#{ios_get_build_version}) to `trunk`.\n\n## To test\n\n- Ensure the build is 🟢\n- The code changes here were tested in their own PRs",
-        head: release_branch_name,
+        head: options[:beta_release] ? merge_branch_name : release_branch_name,
```
mokagio, line 424: We might always want to use a `merge/release-x.y-into-trunk` branch in this scenario, for code freeze and betas as well as the release finalization. There are chances of running into merge conflicts during the release finalization as well.

It's a straightforward change to make, though, so we might see how the next couple of releases go and then update?

leandroalonso: @mokagio I just thought that this approach:

1. Don't create issues for the "final merge" (hotfix and release finalization)
2. It removes the need to delete the branch after the process is finished manually

The original issue I'm trying to prevent is merging `trunk` into the release branch. However, if we just did the final build or the hotfix process, if we run into a conflict there's no problem in fixing those and merging the branch, because the process (in theory) is already finished. And we also remove the manual step of "Ensure the release branch is deleted."

Do you think that makes sense?

mokagio: I like that this would remove the need to "ensure the release branch is deleted."

I'm not sure about merging `trunk` into `release/x.y.z`, even if the release process is finished, because I'd always expect release to go into `trunk`, not vice versa.

However, I realize this might be overly pedantic from a process point of view, particularly given we don't keep track of those branches, but use tags instead. Pragmatically, as long as a release tag comes from a release branch that doesn't have `trunk` merged into it, then we should be good.

Keen to hear what other folks from @Automattic/apps-infrastructure think of this.

PR #403, opened 2022-10-18
Apple SSO: Watch token refresh

chipsnyder:
| 📘 Project: #381 |
|:---:|

| 📓 In my testing, Apple SSO didn't always return the email in the simulator, so testing with a physical device is recommended. |
|:---:|

Syncs the user login tokens to the watch to preserve session management. 

### Setup
2. Set your scheme to `Staging` 

### Testing
You'll want to perform these tests with Username and Password, and Apple SSO as the flow changes touched each path.

1. Login on the App
2. Open the watch app 
3. Trigger an account refresh if the app doesn't trigger one itself
    - The app can be slow to send messages on builds from Xcode so be patient. 
4. Expect to be logged in to the same account as the phone.
5. Log out and try the other path.  

## Checklist


Pocket Casts Watch App Extension/WatchSyncManager.swift
```diff
@@ -116,37 +121,40 @@ class WatchSyncManager {
...
 
-                // clear any previously stored tokens as we're signing in again and we might have one in Keychain already
-                SyncManager.clearTokensFromKeyChain()
-                ServerSettings.saveSyncingPassword(password)
-                ServerSettings.clearLastSyncTime()
-                ServerSettings.setSyncingEmail(email: username)
-                ServerSettings.userId = userId
+    private func handleLogin() {
+        FileLog.shared.addMessage("Login successful")
+        self.checkSubscriptionStatus()
```
leandroalonso, line 145: We don't call `DataManager.sharedManager.markAllPodcastsUnsynced()` anymore. Is that on purpose?

Pocket Casts Watch App Extension/WatchSyncManager.swift
```diff
@@ -164,12 +172,15 @@ class WatchSyncManager {
             let password = response[WatchConstants.Messages.LoginDetailsResponse.password] as? String ?? ""
             ServerSettings.setSyncingEmail(email: username)
             ServerSettings.saveSyncingPassword(password)
-            if SyncManager.isUserLoggedIn(), username.count == 0, password.count == 0 {
+
+            ServerSettings.appleAuthIdentityToken = response[WatchConstants.Messages.LoginDetailsResponse.appleAuthToken] as? String
+
+            if SyncManager.isUserLoggedIn(), username.isEmpty {
```
leandroalonso, line 184: There's any chance the username might not be empty but the password is?

leandroalonso (commented):
@chipsnyder it looks good from the tests I made.

However I just switched devices and I'd like to properly test it, I'll need to add my new device to the portal to be able to build to it.

As soon as I test on it and everything goes well I'll approve here. :)

Pocket Casts Watch App Extension/WatchSyncManager.swift
chipsnyder: That's part of one of the reused code paths and is handled as part of the `refreshLogin` flow [here](https://github.com/Automattic/pocket-casts-ios/pull/403/files/8e8d963b370d8f4f26d9290f9588b00371682d4b#diff-e1ee3e8d716e15b59667a3ed1ad5c97c7da85c0300ffe890847c97d48252289eR89)

Pocket Casts Watch App Extension/WatchSyncManager.swift
chipsnyder: Yeah, the username will be set on both Password and SSO flows. So if a user logs in with SSO, they'll have the username set but not a password. If there is no username, then we'll sign out. If we decide to try one of the other routes and fail to find the necessary Auth credentials, then it will throw an error and sign out as well, preserving the original flow.

leandroalonso (changes requested):
@chipsnyder I tested the following steps:

1. Clean install
2. Log in using Apple SSO
3. Watch reflects the change
4. Log out
5. Log in using email/password
6. 💥 The changes never reflected on the Apple Watch. It kept me showing that I was logged out.

leandroalonso:
To add more info:

* I tried with the flag disabled
* Logged in with email/password
* I briefly see my email
* The interface changes to "Not Signed In"

chipsnyder:
@leandroalonso 🤔 

I wonder if I need to bring [this downstream change](https://github.com/Automattic/pocket-casts-ios/pull/409/files#diff-e1ee3e8d716e15b59667a3ed1ad5c97c7da85c0300ffe890847c97d48252289eR102-R104) up to this branch. I'll take a look today.

chipsnyder:
@leandroalonso I believe [that](https://github.com/Automattic/pocket-casts-ios/pull/403#issuecomment-1285667577) was the problem. Can you test this again?

leandroalonso:
@chipsnyder I'll test again. However, I think we should expand this task's test scenario to ensure it doesn't affect any possible paths before merging. Here's my plan:

### With the flag disabled

1. Test a currently logged-in user, and make sure the app behaves correctly (maybe force some token refresh)
2. Login with email/password and make sure the app behaves correctly

### With the flag enabled

1. Test a currently logged-in user, and make sure the app behaves correctly (maybe force some token refresh)
2. Login with email/password and make sure the app behaves correctly
3. Test a currently logged-in user that used Apple SSO, and make sure the app behaves correctly (maybe force some token refresh)
4. Login with Apple SSO and make sure the app behaves correctly

leandroalonso (approved):

PR #404, opened 2022-10-18
End of Year: generate image and share it through Share Sheet

leandroalonso:
| 📘 Project: #376 |
|:---:|

Includes the logic for image generation based on the SwiftUI story View and passes the image to the Share Sheet.

## Additional details

The image generation needs to happen on the main thread — so I had to come up with a way to generate the image without affecting the presentation of the Share Sheet. If we generate the image right before the Share Sheet is presented the UI hands for a few milliseconds.

I postponed this process to right after the Share Sheet is presented. If we want to also share directly to Instagram/Facebook this process will need to apply to them too, but I'll leave that for later.

## To test

I recommend testing that on a real device because on Simulator the image generation is quite quick and almost doesn't affect the main thread.

2. Run the app
4. Tap "Share"
5. Tap Save Image
6. Go to Photos
7. ✅ Make sure the generated image is correct ¶
8. Go back to the app, and share the story again
9. Select Instagram or Facebook (story)
10. ✅ Check that the image appears correctly in those apps

¶ You'll notice that the generated image is slightly different from the presented Story. I'll leave those tweaks for when we'll work on the final designs.

## Checklist


chipsnyder (approved):
Code Looks good and the photo path works as expected.

> Select Instagram or Facebook (story)

I don't have Instagram or Facebook set up on my test device so I wasn't able to check these paths.

PR #406, opened 2022-10-18
Updates the method of detecting the iPhone 14 to reuse the UIDeviceIdentifier module

### LINKED ISSUE #327: iPhone 14 UI Issues
### Expected behavior

On a podcast's page, there should be no gap between the podcast title and the search bar. Here's an example of how it appears on an iPhone 11:

<details>
<summary>Screenshot</summary>

IMG_D3B67627A87E-1

</details>

There is also no issue on an iPhone SE 2022 and iPhone 13.

### Actual behavior

On an iPhone 14 Pro, there is a gap between the podcast title and the search bar, which results in the episode list being visible through the gap:

<details>
<summary>Screenshot</summary>

Attachment (1)

</details>

1. On an iPhone 14 Pro, open a podcast page
2. Notice that there is a gap between the podcast title and the search bar

- We've been able to replicate this.
- Reported on #5581011-zen: iOS 16.0.2, Pocket Casts 7.22

---

### Expected behavior

When either (1) swiping left to archive an episode, or (2) toggling Show/Hide Archived, there should be no white bar at the top of the podcast page. This is how it is on an iPhone SE 2022, for example.

### Actual behavior

On an iPhone 14 Pro, when either (1) swiping left to archive an episode, or (2) toggling Show/Hide Archived, a white bar appears briefly at the top of the podcast page.

<details>
<summary>Screen Recordings</summary>

https://user-images.githubusercontent.com/74066241/193181068-fed34d74-86ff-493b-8894-012fa7d2e235.mov

https://user-images.githubusercontent.com/74066241/193181105-c7611ceb-6020-4749-82de-8d969c798f89.MP4

</details>

1. On an iPhone 14 Pro, open a podcast page
2. Swipe left to archive an episode or toggle Show/Hide Archived
3. Notice how a white bar briefly appears at the top of the podcast page

- We've been able to replicate this.
- Reported on #5589565-zen: iOS 16.0.2, Pocket Casts 7.22

emilylaguna:
Fixes #327

There is a strange issue where the iPhone 14 Pro fix is applied to the release and debug versions but not the TestFlight version. So instead of relying on the status bar height we'll instead detect the model of the device starts with iPhone 14 Pro. 

This also reuses the UIDeviceIdentifier module that has been bundled in the app.

I also added some simulator model detection to allow this to be tested in the simulator.

### On iPhone 14 Pro and iPhone 14 Pro Max (devices with Dynamic Island)

1. Run the app
2. Open any podcast screen
3. Toggle the header
4. ✅ Check that the top doesn't have a white line blinking
5. Scroll down
6. ✅ Check that there isn't a space between the search bar and the nav bar

### Other devices
1. Repeat the steps above and make sure everything works as it was.

## Checklist

### DIFF
--- a/podcasts/PodcastViewController.swift
+++ b/podcasts/PodcastViewController.swift
@@ -3,6 +3,7 @@ import PocketCastsDataModel
 import PocketCastsServer
 import PocketCastsUtils
 import UIKit
+import UIDeviceIdentifier
 
 protocol PodcastActionsDelegate: AnyObject {
     func isSummaryExpanded() -> Bool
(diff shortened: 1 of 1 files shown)


podcasts/PodcastViewController.swift
```diff
@@ -214,12 +215,17 @@ class PodcastViewController: FakeNavViewController, PodcastActionsDelegate, Sync
     }
 
     private func updateTopConstraintForiPhone14Pro() {
-        guard let window = UIApplication.shared.windows.filter({ $0.isKeyWindow }).first else {
-            return
+        // Retrieve the name of the device
+        var deviceName = UIDeviceHardware.platformString()
+
+        #if targetEnvironment(simulator)
```
emilylaguna, line 221: I could have just done a #if #else #endif for this but I wanted to allow the simulator to fallback to the original device name if needed.

leandroalonso (approved):
Tested with "14 Pro" devices and other devices and it works as expected.

PR #409, opened 2022-10-18
Apple SSO: Login Consolidate

chipsnyder:
| 📘 Project: #381 | Depends on #408 |
|:---:|:---:|

| 📓 For Apple SSO testing: In my testing, Apple SSO didn't always return the email in the simulator, so testing with a physical device is recommended. |
|:---:|

Consolidates some code paths when using the Password Auth:
- Reduces duplicated code in the obtain Token flow to use the paths created for Apple Watch and Sonos
- Moves Sync Progress updates to a shared view.

## To test
1. Log into an account with Username and Password with the `signInWithApple` feature flag enabled and disabled to ensure prod is fine.
2. ✅ Validate you are logged in successfully and the Sync Progress indicator updates with sync status changes.
3. Enable `signInWithApple`  switch to `Staging` use Apple SSO to login 
4. ✅ Validate you are logged in successfully, and the Sync Progress indicator updates with sync status changes.

## Checklist

### DIFF
--- a/podcasts/AppDelegate+Analytics.swift
+++ b/podcasts/AppDelegate+Analytics.swift
@@ -39,23 +39,15 @@ extension AppDelegate {
     /// This should only need to be ran once.
     func retrieveUserIdIfNeeded() {
         guard
-            let username = ServerSettings.syncingEmail(),
-            let password = ServerSettings.syncingPassword(),
+            ServerSettings.syncingEmail() != nil,
             ServerSettings.userId == nil
         else {
             return
         }
 
         FileLog.shared.addMessage("Missing User ID - Retrieving from the server")
-
-        // Refresh the login, but only retrieve the userId
-        ApiServerHandler.shared.validateLogin(username: username, password: password) { success, userId, _ in
-            guard success, let userId else {
-                return
-            }
-
-            ServerSettings.userId = userId
-            NotificationCenter.default.post(name: .userLoginDidChange, object: nil)
+        Task {
+            try? await AuthenticationHelper.refreshLogin()
         }
     }
 }
--- a/podcasts/AuthenticationHelper.swift
+++ b/podcasts/AuthenticationHelper.swift
@@ -30,16 +30,19 @@ class AuthenticationHelper {
 
     // MARK: Password
 
-    static func validateLogin(username: String, password: String, scope: AuthenticationScope) async throws -> AuthenticationResponse {
+    @discardableResult
+    static func validateLogin(username: String, password: String, scope: AuthenticationScope = .mobile) async throws -> AuthenticationResponse {
         let response = try await ApiServerHandler.shared.validateLogin(username: username, password: password, scope: scope.rawValue)
-        handleSuccessfulSignIn(response, .password)
+        SyncManager.clearTokensFromKeyChain()
+        ServerSettings.setSyncingEmail(email: username)
         ServerSettings.saveSyncingPassword(password)
-
+        handleSuccessfulSignIn(response, .password)
         return response
     }
 
     // MARK: Apple SSO
 
+    @discardableResult
     static func validateLogin(appleIDCredential: ASAuthorizationAppleIDCredential) async throws -> AuthenticationResponse {
         guard let identityToken = appleIDCredential.identityToken,
               let token = String(data: identityToken, encoding: .utf8)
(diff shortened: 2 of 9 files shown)


podcasts/AppDelegate+Analytics.swift
```diff
@@ -39,23 +39,15 @@ extension AppDelegate {
...
         FileLog.shared.addMessage("Missing User ID - Retrieving from the server")
-
-        // Refresh the login, but only retrieve the userId
-        ApiServerHandler.shared.validateLogin(username: username, password: password) { success, userId, _ in
-            guard success, let userId else {
-                return
-            }
-
-            ServerSettings.userId = userId
-            NotificationCenter.default.post(name: .userLoginDidChange, object: nil)
```
chipsnyder, line 58: these items are handled in the completion of `refreshLogin`

podcasts/AuthenticationHelper.swift
```diff
@@ -30,16 +30,19 @@ class AuthenticationHelper {
 
     // MARK: Password
 
-    static func validateLogin(username: String, password: String, scope: AuthenticationScope) async throws -> AuthenticationResponse {
+    @discardableResult
+    static func validateLogin(username: String, password: String, scope: AuthenticationScope = .mobile) async throws -> AuthenticationResponse {
         let response = try await ApiServerHandler.shared.validateLogin(username: username, password: password, scope: scope.rawValue)
-        handleSuccessfulSignIn(response, .password)
+        SyncManager.clearTokensFromKeyChain()
+        ServerSettings.setSyncingEmail(email: username)
```
chipsnyder, line 37: Repeating saving the email here as a safety measure for Prod since the response won't have an email item yet.

podcasts/AuthenticationHelper.swift
```diff
@@ -93,16 +97,15 @@ class AuthenticationHelper {
     // MARK: Common
 
     private static func handleSuccessfulSignIn(_ response: AuthenticationResponse, _ source: AuthenticationSource) {
-        SyncManager.clearTokensFromKeyChain()
-
         ServerSettings.userId = response.uuid
         ServerSettings.syncingV2Token = response.token
+        if let email = response.email, !email.isEmpty {
+            ServerSettings.setSyncingEmail(email: email)
+        }
```
emilylaguna, line 104: From my testing if the syncing email is not set, then the rest of the sync will fail since this is used to determine if a user is logged in or not throughout the app: https://github.com/Automattic/pocket-casts-ios/blob/5e6181fa26d6e1b97dc2fd84b8fb3653f95fa048/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncManager.swift#L8

I wonder if we should we consider a nil email as a failed login?

podcasts/SyncSigninViewController.swift
```diff
@@ -271,37 +237,37 @@ class SyncSigninViewController: PCViewController, UITextFieldDelegate {
...
+        })
+    }
 
-                self.progressAlert = ShiftyLoadingAlert(title: L10n.syncAccountLogin)
-                self.progressAlert?.showAlert(self, hasProgress: false, completion: {
-                    // clear any previously stored tokens as we're signing in again and we might have one in Keychain already
-                    SyncManager.clearTokensFromKeyChain()
+    private func handleError(_ error: Error) {
+        let error = (error as? APIError) ?? APIError.UNKNOWN
+        Analytics.track(.userSignInFailed, properties: ["source": self.authSource, "error_code": error.rawValue])
```
emilylaguna: Nit: 
```suggestion
        Analytics.track(.userSignInFailed, properties: ["source": authSource, "error_code": error.rawValue])
```

emilylaguna (changes requested):
Nice work, I tested with and without the feature flag enabled and everything works as intended. 

I left a few comments below.  Let me know what you think.

podcasts/AuthenticationHelper.swift
chipsnyder: In general, I agree, and I think that should be the case when the endpoint changes for `user/login_apple` are merged to prod, and we can expect an email from all User Login Responses. For now, though, this acts as a safety net to not clear out the email address if it happens to be missing from the response. I'll create a clean-up task to address that for when the endpoint is available. 

**Edit:** Created Task https://github.com/Automattic/pocket-casts-ios/issues/412

chipsnyder: @emilylaguna Do you have any thoughts on the plan to revisit this when the endpoint merges to prod?

emilylaguna: Thanks for creating the task, I think we can merge this in and tackle it later.

emilylaguna (approved):

emilylaguna:
There are some changes coming to the login/token endpoints that will require more changes how we handle the token refresh/etc. So I am going to close this for now and revisit these improvements later.

PR #410, opened 2022-10-18
Apple SSO: Consolidate Token Refresh paths

chipsnyder:
| 📘 Project: #381 | Depends on #409 |
|:---:|:---:|

| 📓 For Apple SSO testing: In my testing, Apple SSO didn't always return the email in the simulator, so testing with a physical device is recommended. |
|:---:|

Consolidates the refresh password token flow to use the existing API methods.

The token refresh methods have a `30.seconds` timeout which seems very long, especially when the default is already `15.seconds`, so I didn't maintain the `30.seconds` period when porting to the existing methods. 

I did, however, modify the login paths (SSO and Username/Password) to use the `.reloadIgnoringLocalCacheData` policy, as that just seems wise for auth calls. 

## To test

This can be tested with the same process as https://github.com/Automattic/pocket-casts-ios/pull/392 but using Username and Password log in instead of SSO

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
@@ -17,7 +17,7 @@ class TokenHelper {
 
         if SyncManager.isUserLoggedIn() {
             let token: String
-            if let storedToken = KeychainHelper.string(for: ServerConstants.Values.syncingV2TokenKey) {
+            if let storedToken = ServerSettings.syncingV2Token {
                 token = storedToken
             } else if let newToken = TokenHelper.acquireToken() {
                 token = newToken
@@ -58,7 +58,6 @@ class TokenHelper {
             switch result {
             case .success(let token):
                 refreshedToken = token
-                ServerSettings.syncingV2Token = token
             case .failure:
                 refreshedToken = nil
             }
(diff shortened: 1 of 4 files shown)


leandroalonso:
@chipsnyder it looks like there are build errors on this PR. 🤔

chipsnyder:
😅 Got moving too fast yesterday and didn't push the last change to this branch.

Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
```diff
@@ -79,76 +78,50 @@ class TokenHelper {
...
-                return token
+            else if let token = await acquireIdentityToken() {
+                completion(.success(token))
             }
-
-            if httpResponse.statusCode == ServerConstants.HttpConstants.unauthorized {
-                FileLog.shared.addMessage("TokenHelper logging user out, invalid password")
-                SyncManager.signout()
+            else {
+                completion(.failure(.UNKNOWN))
```
emilylaguna, line 90: Nit: Since we're touching this code, how do you feeling about adding some better context to this failure by adding a new error type for "unable to acquire token"?

emilylaguna (approved):
Tested using the steps and everything worked as intended. I left a comment about improving a failure reason

Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
chipsnyder: That should be fine. The caveat here is that this change will touch a few spots that have default messaging for errors of type `.Unknown`. 

Many of those check for a valid Error message, but we may want to add a property like:

```
extension APIError {
    var isGenericError: Bool {
         switch self:
            case .Unkown, .TokenFailure:
                return true
            default:
               return false
    }
}
```

That would then apply to cases like [here](https://github.com/Automattic/pocket-casts-ios/blob/c9a937af6f4d017e337fad91d22915edea4aac57/podcasts/NewEmailViewController.swift#L177) and a few others

emilylaguna: I don't think we need to do that. We can just update https://github.com/Automattic/pocket-casts-ios/blob/7786534abfbaf48c07d6a7c3d7c1be37deac7a2d/podcasts/Server%2BStrings.swift#L25

With: `case .ACQUIRE_TOKEN_FAILED: return L10n.serverErrorUnknown`. This way the same default error message will be shown to the user, but internally we would log the more specific reason. 

I would probably also update the initial log message to display the raw error code that is returned: 
https://github.com/Automattic/pocket-casts-ios/blob/c9a937af6f4d017e337fad91d22915edea4aac57/podcasts/NewEmailViewController.swift#L176

chipsnyder: I'll branch off of this one to capture the changes. 

The places I was considering are:

https://github.com/Automattic/pocket-casts-ios/blob/c9a937af6f4d017e337fad91d22915edea4aac57/podcasts/NewEmailViewController.swift#L177-L182


In these cases, we don't use the enum default value but override that with a context-specific message.

emilylaguna (approved):

emilylaguna:
There are some changes coming to the login/token endpoints that will require more changes how we handle the token refresh/etc. So I am going to close this for now and revisit these improvements later.

PR #413, opened 2022-10-19
End of Year: show total categories listened and top categories

leandroalonso:
| 📘 Project: #376 |
|:---:|

Adds two new stories:

* Total categories listened
* Top 5 categories listened

| Total categories | Top 5 categories |
| ---------------- | ---------------- |
| Simulator Screen Shot - iPhone 14 - 2022-10-19 at 15 01 35 | Simulator Screen Shot - iPhone 14 - 2022-10-19 at 15 01 44 |

As you already know, design is gonna be tackled in another PR.

## To discuss

The list is being ordered by total time listened in each category. However, I wonder if the user might find it strange that maybe they have a top 1 category that has fewer podcasts than the top 2, for example.

We can either change the design to show the listening time, change the query (to order by the number of podcasts) or do nothing.

## To test

2. Run the app and login with an account that has a Listening History with at least a few items
4. Skip the first story
5. ✅ Check that the second story shows the total number of categories you listened to
6. ✅ Check that the third story shows the top 5 categories you listened to with the number of different podcasts for each one

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -26,4 +26,46 @@ class EndOfYearDataManager {
         return listeningTime
     }
 
+    /// Returns all the categories the user has listened to podcasts
+    ///
+    /// The returned array is ordered from the most listened to the least
+    func listenedCategories(dbQueue: FMDatabaseQueue) -> [ListenedCategory] {
+        var listenedCategories: [ListenedCategory] = []
+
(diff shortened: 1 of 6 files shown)


emilylaguna (approved):
> We can either change the design to show the listening time, change the query (to order by the number of podcasts) or do nothing.

I think showing the listening time makes more sense than the number of podcasts on this view. 

I know the design isn't finalized yet, but I want to bring up that the clipping looks a bit weird on the category name label.

leandroalonso:
> I think showing the listening time makes more sense than the number of podcasts on this view.

Shared here pdeCcb-1is-p2#comment-1311

> but I want to bring up that the clipping looks a bit weird on the category name label.

I'll def tackle that on the design task. :)

PR #414, opened 2022-10-19
End of Year: show number of listened podcasts and episodes

leandroalonso:
| 📘 Project: #376 |
|:---:|

Adds a story that shows the number of listened podcasts and episodes. Designs are not contemplated in this task.

## To test

4. Skip to the fourth story
5. ✅ Check that the number of played podcasts and episodes is shown

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -63,9 +63,46 @@ class EndOfYearDataManager {
         return listenedCategories
     }
 
+    /// Return the number of podcasts and episodes listened
+    ///
+    func listenedNumbers(dbQueue: FMDatabaseQueue) -> ListenedNumbers {
+        var listenedNumbers: ListenedNumbers = ListenedNumbers(numberOfPodcasts: 0, numberOfEpisodes: 0)
+
+        dbQueue.inDatabase { db in
+            do {
+                let query = """
+                            SELECT COUNT(\(DataManager.episodeTableName).id) as episodes,
+                                COUNT(DISTINCT \(DataManager.podcastTableName).uuid) as podcasts
+                            FROM \(DataManager.episodeTableName), \(DataManager.podcastTableName)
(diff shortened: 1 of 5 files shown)


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -63,9 +63,51 @@ class EndOfYearDataManager {
...
+
+        dbQueue.inDatabase { db in
+            do {
+                let query = """
+                            SELECT COUNT(\(DataManager.episodeTableName).id) as episodes,
+                                COUNT(DISTINCT \(DataManager.podcastTableName).uuid) as podcasts
+                            FROM \(DataManager.episodeTableName), \(DataManager.podcastTableName)
+                            WHERE `\(DataManager.podcastTableName)`.uuid = `\(DataManager.episodeTableName)`.podcastUuid and
+                                lastPlaybackInteractionDate IS NOT NULL AND
+                                lastPlaybackInteractionDate BETWEEN strftime('%s', date('now','start of year')) and strftime('%s', 'now')
```
emilylaguna, line 79: It seems like this bit of SQL is being used in every query. I wonder if can extract this out so it can be reused instead of C&P it?

emilylaguna (changes requested):
The stats look good, I left a comment about reusing the date range SQL fragment. 

Let me know what you think.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
leandroalonso: Would you mind if I change that on #416? Just to avoid a bunch of merges/rebases.

emilylaguna (approved):

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
emilylaguna: Works for me. :shipit:

PR #426, opened 2022-10-21
Modify accessibility label of ThemeableSelectionView based on its selection state

adincebic:
When user was creating a new account, plan selection was not indicated to VoiceOver users. When user taps "regular" or "plus" account, VoiceOver now  prefixes with word "selected" if that plan is selected. This is inspired by UIButton selection state.

## To test
1. Turn voiceover on
2. Go to profile
3. Tap create account
4. Navigate with voiceover

If selected plan is regular, voiceover should say "Selected, Regular". If regular is not selected, voiceover should just read the accessibility label "regular". Same of course applies for the plus plan.

### DIFF
--- a/podcasts/ThemeableSelectionView.swift
+++ b/podcasts/ThemeableSelectionView.swift
@@ -70,11 +70,6 @@ class ThemeableSelectionView: UIView {
     }
 
     private func updateAccessibilityLabel() {
-        if unselectedAccessibilityLabel == nil {
-            unselectedAccessibilityLabel = accessibilityLabel
-        }
-        let unselectedAccessibilityLabel = self.unselectedAccessibilityLabel ?? ""
-        let selectedAccessibilityLabel = "\(L10n.statusSelected), \(unselectedAccessibilityLabel)"
-        accessibilityLabel = isSelected ? selectedAccessibilityLabel : unselectedAccessibilityLabel
+        accessibilityValue = isSelected ? L10n.statusSelected : ""
     }
 }
--- a/podcasts/ThemeableSelectionView.swift
+++ b/podcasts/ThemeableSelectionView.swift
@@ -16,7 +16,7 @@ class ThemeableSelectionView: UIView {
     var isSelected: Bool = false {
         didSet {
             updateColor()
-            updateAccessibilityLabel()
+            updateAccessibilityTraits()
         }
     }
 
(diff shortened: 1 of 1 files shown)


podcasts/ThemeableSelectionView.swift
```diff
@@ -65,4 +68,13 @@ class ThemeableSelectionView: UIView {
...
         layer.borderColor = isSelected ? AppTheme.colorForStyle(selectedStyle, themeOverride: themeOverride).cgColor : AppTheme.colorForStyle(unselectedStyle, themeOverride: themeOverride).cgColor
     }
+
+    private func updateAccessibilityLabel() {
+        if unselectedAccessibilityLabel == nil {
+            unselectedAccessibilityLabel = accessibilityLabel
+        }
+        let unselectedAccessibilityLabel = self.unselectedAccessibilityLabel ?? ""
+        let selectedAccessibilityLabel = "\(L10n.statusSelected), \(unselectedAccessibilityLabel)"
+        accessibilityLabel = isSelected ? selectedAccessibilityLabel : unselectedAccessibilityLabel
```
emilylaguna: I think we can achieve the same result in a bit cleaner fashion by using the accessibilityValue and adding a new string for statusUnselected "Unselected".

You can view the docs for adding a new string here: https://github.com/Automattic/pocket-casts-ios/blob/trunk/docs/localization.md

```suggestion
        accessibilityValue = isSelected ? L10n.statusSelected : L10n.statusUnselected
```

emilylaguna (changes requested):
Awesome! Thank you so much for your contribution! 🙌 

I left a suggestion below about using the accessibilityValue instead of updating the accessibilityLabel. Let me know what you think.

podcasts/ThemeableSelectionView.swift
adincebic: @emilylaguna  thank you so much for suggesting to use accessibilityValue. I did not add new string for unselected as that is not needed because it is enough to indicate when something is selected. In fact, this is how Apple's controls work. Try setting isSelected to true on UIButton and word selected will be pronounced when button is selected. However, when button is not selected it wont pronounce unselected but will just read its accessibility label.

emilylaguna: Ah, I see. I think in that case we can stay consistent with how other buttons work and use the selected property on [UIAccessibilityTraits](https://developer.apple.com/documentation/uikit/uiaccessibilitytraits).

```
if isSelected {
    accessibilityTraits.insert(.selected)
} else {
    accessibilityTraits.remove(.selected)
}
```

wdyt?

adincebic: Wow, this is even better.

leandroalonso (approved):
Thanks for the changes and your contribution @adincebic!

PR #428, opened 2022-10-21
End of Year: refactor data source

leandroalonso:
| 📘 Project: #376 |
|:---:|

Refactor the data source to move data initialization to another entity and also add handling failure in the process:

Most of the changes are merely changes in the code though.

### Stories hasn't changed

4. ✅ Check that the stories are presented correctly (there are no changes)

### Forcing error

We might run into an error when loading stories. This is now handled, first, change `EndOfYearStoriesDataSource.isReady()` to just `return false`. Then:

1. Run the app
4. ✅ Check that you see a "Failed to load stories" error

## Checklist

### DIFF
--- a/podcasts/End of Year/EndOfYearStoriesBuilder.swift
+++ b/podcasts/End of Year/EndOfYearStoriesBuilder.swift
@@ -59,6 +59,11 @@ class EndOfYearStoriesBuilder {
                 stories.append(.topOnePodcast)
             }
 
+            // Top 5 podcasts
+            if topPodcasts.count > 1 {
+                stories.append(.topFivePodcasts)
+            }
+
             continuation.resume(returning: (stories, data))
         }
     }
--- a/podcasts/End of Year/EndOfYearStoriesBuilder.swift
+++ b/podcasts/End of Year/EndOfYearStoriesBuilder.swift
@@ -29,10 +29,19 @@ class EndOfYearStoriesBuilder {
     // Call this method to build the list of stories and the data provider
     func build() async -> ([EndOfYearStory], EndOfYearStoriesData) {
         await withCheckedContinuation { continuation in
+            // Listening time
             if let listeningTime = dataManager.listeningTime(),
                 listeningTime > 0 {
                 stories.append(.listeningTime)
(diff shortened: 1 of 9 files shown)


podcasts/End of Year/EndOfYearStoriesBuilder.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    private var stories: [EndOfYearStory] = []
+
+    private let data = EndOfYearStoriesData()
+
+    init(dataManager: DataManager = DataManager.sharedManager) {
+        self.dataManager = dataManager
+    }
+
+    // Call this method to build the list of stories and the data provider
```
emilylaguna: Nit. 

```suggestion
    /// Call this method to build the list of stories and the data provider
```

podcasts/End of Year/EndOfYearStoriesBuilder.swift
```diff
@@ -0,0 +1,92 @@
...
+                data.topPodcasts = topPodcasts
+                stories.append(.topOnePodcast)
+            }
+
+            // Top 5 podcasts
+            if topPodcasts.count > 1 {
+                stories.append(.topFivePodcasts)
+            }
+
+            if let longestEpisode = dataManager.longestEpisode(),
```
emilylaguna, line 68: Nit: Can you also add a comment above this section like you have on the others?

podcasts/End of Year/Stories/EndOfYearStoriesDataSource.swift
```diff
@@ -2,59 +2,57 @@ import SwiftUI
...
-            self.longestEpisode = DataManager.sharedManager.longestEpisode()
+        if !stories.isEmpty {
+            stories.insert(.intro, at: 0)
+            stories.append(.epilogue)
 
-            continuation.resume(returning: true)
+            return true
+        } else {
+            return false
         }
```
emilylaguna: Nit: We don't really need the else here. 

```suggestion
            return true
        }

        return false
```

emilylaguna (approved):
Tested and works well. Left some nit comments.

PR #433, opened 2022-10-21
End of Year: check eligibility and listening history

leandroalonso:
| 📘 Project: #376 |
|:---:|

* Check if a user is eligible to see their EoY stats (aka they listened to at least one episode for more than 30 minutes this year)
* Check user Listening History, if their history is not complete, call a block to sync it

## Checking Listening History

Right now, the app assumes that the user has the full listening history based on:

1. If their oldest item on Listening History is from a previous year
2. If the above is false check if their total number of episodes in Listening History is smaller or equal to 100 (the server limit)

Conversations about this is ongoing in pdeCcb-1mZ-p2#comment-1325

Note that we also still don't have the "listening history" sync endpoint, so there's no sync happening yet. However, everything is in place for doing so.

## To test

2. Do a clean install of the app and run it
3. ✅ Check that nothing happens, no modal is shown and there's no card under Profile
4. Login to account in which you have a Listening History with a few episodes
5. ✅ Check that as soon as you logged in the modal is shown and the card is now visible on Profile

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -64,7 +64,7 @@ class EndOfYearDataManager {
                 if resultSet.next() {
                     isFullListeningHistory = true
                 } else {
-                    isFullListeningHistory = numberOfItemsInListeningHistory(dbQueue: dbQueue) >= 100
+                    isFullListeningHistory = numberOfItemsInListeningHistory(db: db) <= 100
                 }
             } catch {
                 FileLog.shared.addMessage("EndOfYearDataManager.isFullListeningHistory error: \(error)")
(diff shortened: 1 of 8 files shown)


podcasts/End of Year/EndOfYearStoriesBuilder.swift
```diff
@@ -22,13 +22,22 @@ class EndOfYearStoriesBuilder {
...
+        self.sync = sync
     }
 
     // Call this method to build the list of stories and the data provider
     func build() async -> ([EndOfYearStory], EndOfYearStoriesData) {
         await withCheckedContinuation { continuation in
+            // Check if the user has the full listening history for this year
+            if !dataManager.isFullListeningHistory() {
+                // TODO: update with the correct endpoint to sync history (eoy-todo)
+                sync?()
```
leandroalonso, line 38: Here we'll be able to use something like `SyncHistoryTask` without much effort.

The Stories view is already prepared for it and it will show a loading indicator as long as this goes on.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -9,6 +9,71 @@ class EndOfYearDataManager {
...
+                FileLog.shared.addMessage("EndOfYearDataManager.listeningTime error: \(error)")
+            }
+        }
+
+        return isEligible
+    }
+
+    /// Check if the user has the full listening history or not.
+    ///
+    /// This is not 100% accurated. In order to determine if the user
```
emilylaguna:
```suggestion
    /// This is not 100% accurate. In order to determine if the user
```

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -9,6 +9,71 @@ class EndOfYearDataManager {
...
+    /// has the full history we check for their latest episode listened.
+    /// If this episode was interacted in 2021 or before, we assume they
+    /// have the full history.
+    /// If this is not true, we check for the total number of items of
+    /// this year. If the number is small or equal 100, we assume they
```
emilylaguna:
```suggestion
    /// this year. If the number is less than or equal 100, we assume they
```

emilylaguna (approved):
Nice work, tested on a fresh install and with an account with minimal listening history. Confirmed that after listening to some episodes the EOY then displayed.

PR #436, opened 2022-10-22
Make SwitchCell easier to use with assistive technologies

adincebic:
Current implementation of SwitchCell does not support easier navigation with VoiceOver or other assistive technologies due to the fact that all UI elements are exposed as accessibility children. When cells have a switch and it is the only action button, tapping on the whole cell using assistive technologies should toggle its value. More over, it should not be separate accessibility element.

## To test

1. Turn voiceover on
2. Go to app settings from the profile tab
3. Tap on Notifications
4. Use swipe left or right gestures with one finger and try toggling notifications on and off

## Current behavior:

Voiceover treats label of the cell and the switch as two elements requiring users to swipe twice before reaching the switch. This gets more problematic if there are a lot of switches like in General section of app settings.

## Behaviour after this fix:

Voiceover treats cell label and switch control as single element and double tapping on it toggles its value just like in Apple's apps.

### DIFF
--- a/podcasts/SwitchCell.swift
+++ b/podcasts/SwitchCell.swift
@@ -1,14 +1,12 @@
 import UIKit
 
 class SwitchCell: ThemeableCell {
-    @IBOutlet var cellSwitch: ThemeableSwitch!
+    let cellSwitch: ThemeableSwitch = {
+        let cellSwitch = ThemeableSwitch()
+        cellSwitch.isAccessibilityElement = false
+        return cellSwitch
+    }()
     @IBOutlet var cellLabel: UILabel!
-    @IBOutlet var cellSecondaryLabel: ThemeableLabel! {
-        didSet {
-            cellSecondaryLabel.style = .primaryText02
-        }
-    }
-
     @IBOutlet var cellImage: UIImageView!
     @IBOutlet var cellTextToImageConstraint: NSLayoutConstraint!
 
@@ -23,7 +21,7 @@ class SwitchCell: ThemeableCell {
 
     override func awakeFromNib() {
         super.awakeFromNib()
-
+        accessoryView = cellSwitch
         setNoImage()
     }
 
@@ -47,28 +45,7 @@ class SwitchCell: ThemeableCell {
     override func setSelected(_ selected: Bool, animated: Bool) {}
     override func setHighlighted(_ highlighted: Bool, animated: Bool) {}
 
-    override func layoutSubviews() {
-        super.layoutSubviews()
-        setupAccessibility()
-    }
-
     override func accessibilityActivate() -> Bool {
-        cellSwitch.setOn(!cellSwitch.isOn, animated: true)
-        cellSwitch.sendActions(for: .valueChanged)
-        matchAccessibilityValueWithSwitchControl()
-        return true
-    }
-
-    private func setupAccessibility() {
-        isAccessibilityElement = true
-        let labelText = cellLabel.text ?? ""
-        let secondaryLabelText = cellSecondaryLabel.text ?? ""
-        accessibilityLabel = "\(labelText), \(secondaryLabelText)"
-        accessibilityTraits = cellSwitch.accessibilityTraits
-        matchAccessibilityValueWithSwitchControl()
-    }
-
-    private func matchAccessibilityValueWithSwitchControl() {
-        accessibilityValue = cellSwitch.accessibilityValue
+        return isLocked
     }
 }
--- a/podcasts/SwitchCell.swift
+++ b/podcasts/SwitchCell.swift
@@ -46,4 +46,29 @@ class SwitchCell: ThemeableCell {
 
     override func setSelected(_ selected: Bool, animated: Bool) {}
     override func setHighlighted(_ highlighted: Bool, animated: Bool) {}
+
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        setupAccessibility()
+    }
+
+    override func accessibilityActivate() -> Bool {
+        cellSwitch.setOn(!cellSwitch.isOn, animated: true)
+        cellSwitch.sendActions(for: .valueChanged)
+        matchAccessibilityValueWithSwitchControl()
+        return true
+    }
+
+    private func setupAccessibility() {
+        isAccessibilityElement = true
+        let labelText = cellLabel.text ?? ""
+        let secondaryLabelText = cellSecondaryLabel.text ?? ""
+        accessibilityLabel = "\(labelText), \(secondaryLabelText)"
+        accessibilityTraits = cellSwitch.accessibilityTraits
+        matchAccessibilityValueWithSwitchControl()
+    }
+
+    private func matchAccessibilityValueWithSwitchControl() {
+        accessibilityValue = cellSwitch.accessibilityValue
+    }
 }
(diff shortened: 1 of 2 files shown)


podcasts/SwitchCell.swift
```diff
@@ -46,4 +46,29 @@ class SwitchCell: ThemeableCell {
 
     override func setSelected(_ selected: Bool, animated: Bool) {}
     override func setHighlighted(_ highlighted: Bool, animated: Bool) {}
+
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        setupAccessibility()
```
adincebic: I did try calling this method from "awakeFromNib" but it was too early as "cellSwitch" and "cellLabels" had values defined in xib file which is not desirable.

podcasts/SwitchCell.swift
```diff
@@ -46,4 +46,29 @@ class SwitchCell: ThemeableCell {
...
+
+    override func accessibilityActivate() -> Bool {
+        cellSwitch.setOn(!cellSwitch.isOn, animated: true)
+        cellSwitch.sendActions(for: .valueChanged)
+        matchAccessibilityValueWithSwitchControl()
+        return true
+    }
+
+    private func setupAccessibility() {
+        isAccessibilityElement = true
```
adincebic: I tried setting isAccessibilityElement to true, but it didn't work as voiceover for some reason tries to read accessibility label from its parent view (this cell).

podcasts/SwitchCell.swift
```diff
@@ -46,4 +46,29 @@ class SwitchCell: ThemeableCell {
...
+        let labelText = cellLabel.text ?? ""
+        let secondaryLabelText = cellSecondaryLabel.text ?? ""
+        accessibilityLabel = "\(labelText), \(secondaryLabelText)"
+        accessibilityTraits = cellSwitch.accessibilityTraits
```
adincebic: I tried manually setting .switch as accessibility trait, but such trait does not exist. Accessibility trait for switch is not exposed and this is the only way to match behaviour like in Apple's apps.

adincebic:
@emilylaguna  could you please add Accessibility label to this pr?

podcasts/SwitchCell.swift
fsalata: To behave like Apple apps you only have to add the switch to the `accessoryView` instead of `contentView` of the cell.
I can't test on a real device right now but you can see in the video that the cell is read entirely and preforming the action (double tap) activates/deactivates the switch. 
Removing the switch from interface builder, changing line 4 to `var cellSwitch = ThemeableSwitch(frame: .zero)` and adding it to cell's `accessoryView` ( `self.accessoryView = cellSwitch`) should fix the behavior.
Hope it helps.

https://user-images.githubusercontent.com/7861088/197657365-78441bb8-3385-4bb1-ac9d-7c565460ecad.mov

adincebic: > To behave like Apple apps you only have to add the switch to the `accessoryView` instead of `contentView` of the cell.
> 
> I can't test on a real device right now but you can see in the video that the cell is read entirely and preforming the action (double tap) activates/deactivates the switch. 
> 
> Removing the switch from interface builder, changing line 4 to `var cellSwitch = ThemeableSwitch(frame: .zero)` and adding it to cell's `accessoryView` ( `self.accessoryView = cellSwitch`) should fix the behavior.
> 
> Hope it helps.
> 
> 
> 
> https://user-images.githubusercontent.com/7861088/197657365-78441bb8-3385-4bb1-ac9d-7c565460ecad.mov
> 
> 
> 
> 

Thanks for the comment. I am aware of accessory view and accessibility behaviour when switch is its subview. I wanted to make as little changes as possible.

fsalata: Oh ok.
You can try to hide the switch from the voiceover and set a `accessibilityCustomActions` for the cell to change the switch value programmatically

podcasts/SwitchCell.swift
```diff
@@ -46,4 +46,29 @@ class SwitchCell: ThemeableCell {
...
     override func setSelected(_ selected: Bool, animated: Bool) {}
     override func setHighlighted(_ highlighted: Bool, animated: Bool) {}
+
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        setupAccessibility()
+    }
+
+    override func accessibilityActivate() -> Bool {
+        cellSwitch.setOn(!cellSwitch.isOn, animated: true)
```
emilylaguna: If the isLocked value is set to false then we should return false here to prevent the switch from being activated.

You can reproduce by doing the following:

1. Sign into an account without plus
2. Navigate to the Apple Watch or Files settings
3. Select a Plus only switch
4. Activate it

podcasts/SwitchCell.swift
```diff
@@ -46,4 +46,29 @@ class SwitchCell: ThemeableCell {
...
+    private func setupAccessibility() {
+        isAccessibilityElement = true
+        let labelText = cellLabel.text ?? ""
+        let secondaryLabelText = cellSecondaryLabel.text ?? ""
+        accessibilityLabel = "\(labelText), \(secondaryLabelText)"
+        accessibilityTraits = cellSwitch.accessibilityTraits
+        matchAccessibilityValueWithSwitchControl()
+    }
+
+    private func matchAccessibilityValueWithSwitchControl() {
```
emilylaguna: I think we can make this method name a bit more generic. 

```suggestion
    private func updateAccessibilityValue() {
```

podcasts/SwitchCell.swift
```diff
@@ -46,4 +46,29 @@ class SwitchCell: ThemeableCell {
...
+        cellSwitch.sendActions(for: .valueChanged)
+        matchAccessibilityValueWithSwitchControl()
+        return true
+    }
+
+    private func setupAccessibility() {
+        isAccessibilityElement = true
+        let labelText = cellLabel.text ?? ""
+        let secondaryLabelText = cellSecondaryLabel.text ?? ""
+        accessibilityLabel = "\(labelText), \(secondaryLabelText)"
```
emilylaguna: The secondaryLabel isn't actually used anywhere for the switch cell so we don't need to pass this value in here. In the future we should remove the label from the Xib and class. 
```suggestion
        accessibilityLabel = cellLabel.text
```

emilylaguna (changes requested):
Hey @adincebic, thank you so much for your contribution 🙌 👏 .

I gave this a test and everything is working with the exception of it bypassing the isLocked state (See the comment below) 

However I agree with @fsalata below that if we're updating the accessibility for this view we should also update it to reuse the system accessibility features by moving the switch into `accessoryView`.

What do you think about making that change?

adincebic:
> Hey @adincebic, thank you so much for your contribution 🙌 👏 .
> 
> 
> 
> I gave this a test and everything is working with the exception of it bypassing the isLocked state (See the comment below) 
> 
> 
> 
> However I agree with @fsalata below that if we're updating the accessibility for this view we should also update it to reuse the system accessibility features by moving the switch into `accessoryView`.
> 
> 
> 
> What do you think about making that change?

Sure, I'll get to it as soon as possible.

adincebic:
@emilylaguna  @fsalata  

* I moved cellSwitch to accessoryView
* Took isLocked into account
* Removed secondaryLabel from xib and Swift file

Please review.

adincebic:
Hey @fsalata  @emilylaguna  can I ask you to review this code once more? :)

podcasts/SwitchCell.swift
```diff
@@ -1,14 +1,12 @@
 import UIKit
 
 class SwitchCell: ThemeableCell {
-    @IBOutlet var cellSwitch: ThemeableSwitch!
+    let cellSwitch: ThemeableSwitch = {
+        let cellSwitch = ThemeableSwitch()
+        cellSwitch.isAccessibilityElement = false
```
fsalata, line 6: I believe you don't need to set it to false, when you add it to `accessoryView` it should work as expected

adincebic: Not realy, it gets exposed as accessibility element along side parent cell. I tested it manually and needed to set it to false.

emilylaguna (approved):
Nice work! This works nicely in my testing. 

Thank you so much for your contribution! I'll get this merged in.

PR #450, opened 2022-10-26
Add filter chip for subscription status - WIP

### LINKED ISSUE #438: Enhancement: Add filter criterion for subscription state
I'd like to show, for example, any Downloaded, In Progress episodes irrespective of subscription state.

Proposal: Create a new filter that allows for podcast subscription state to be defined ( _All_ | _Subscribed_ | _Not Subscribed_)

Since current behavior is that a filter automatically includes only episodes of subscribed podcasts, perhaps the above criterion can default to _Subscribed_ when presented in the UI so that existing filters remain backwards compatible, and are updated only if the user changes the value to include _All_ or _Not Subscribed_.

aidano:
Fixes #

This fixes https://github.com/Automattic/pocket-casts-ios/issues/438 by allowing the user to filter by podcast subscription status (_All_ | _Subscribed_ | _Not Subscribed_ )

Due to some naming challenges that this functionality introduces, while the code is fully functional in my testing, and should be very close to production-ready (pending any review feedback), the UX is not. 

The primary challenge is confusion around the naming and related functionality of the _All Your Podcasts_ filter chip versus this one (_Any Podcast_ / _Subscribed_ / _Not Subscribed_): it's too similar and will likely introduce user confusion.

Maybe someone on the Automattic team can chime in with thoughts or ideas.

## To test

* Subscribe to a podcast
* Download an episode from that podcast
* Download an episode from a podcast to which you're not subscribed
* Create a filter for 'Downloaded' Download status
* Filter by subscription status using the right-most filter chip

The below screenshots and video provide more context based on the above test scenario.

#### 1) Filtering by Subscribed
Filtering by Subscribed
Filtering by Subscribed

#### 2) Filtering by Not Subscribed
Filtering by Not Subscribed
Filtering by Not Subscribed

#### 3) Filtering by All
Filtering by All
Filtering by All

#### Video of a Playlist of Unplayed, Downloaded Podcasts, Filtering by Subscription Status
https://user-images.githubusercontent.com/621864/198107003-6e55af77-8d15-483b-ba17-db6e8a198f74.mp4

## Checklist

This is a WIP due to the UX nuances introduced and as such is not ready for the below checklist items.

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeFilterDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeFilterDataManager.swift
@@ -25,7 +25,9 @@ class EpisodeFilterDataManager {
         "wasDeleted",
         "filterDuration",
         "longerThan",
-        "shorterThan"
+        "shorterThan",
+        "filterSubscribed",
+        "filterNotSubscribed"
     ]
 
     func count(includeDeleted: Bool, dbQueue: FMDatabaseQueue) -> Int {
@@ -227,6 +229,8 @@ class EpisodeFilterDataManager {
         filter.filterDuration = rs.bool(forColumn: "filterDuration")
         filter.longerThan = rs.int(forColumn: "longerThan")
         filter.shorterThan = rs.int(forColumn: "shorterThan")
+        filter.filterSubscribed = rs.bool(forColumn: "filterSubscribed")
+        filter.filterNotSubscribed = rs.bool(forColumn: "filterNotSubscribed")
 
         return filter
     }
@@ -256,6 +260,8 @@ class EpisodeFilterDataManager {
         values.append(filter.filterDuration)
         values.append(filter.longerThan)
         values.append(filter.shorterThan)
+        values.append(filter.filterSubscribed)
+        values.append(filter.filterNotSubscribed)
 
         if includeUuidForWhere {
             values.append(filter.uuid)
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
@@ -166,6 +166,19 @@ class PodcastDataManager {
         return allPodcasts
     }
 
+    func allSubscribedPodcastUuids(dbQueue: FMDatabaseQueue) -> [String] {
+        var allSubscribed = [String]()
+        cachedPodcastsQueue.sync {
+            for podcast in cachedPodcasts {
+                if !podcast.isSubscribed() { continue }
+
+                allSubscribed.append(podcast.uuid)
+            }
+        }
+
+        return allSubscribed
+    }
+    
     func allUnsubscribedPodcastUuids(dbQueue: FMDatabaseQueue) -> [String] {
         var allUnsubscribed = [String]()
         cachedPodcastsQueue.sync {
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
@@ -640,6 +640,16 @@ class DatabaseHelper {
                 return
             }
         }
+        if schemaVersion < 41 {
+            do {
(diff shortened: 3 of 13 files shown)


leandroalonso:
Hey @aidano, thanks for your contribution.

Could you please add images/videos showing the changes you added to this PR? This helps a lot so other people can chime in with their thoughts without necessarily having to build the code.

aidano:
@leandroalonso  Sure thing. I've edited the original comment to contain some screenshots and a demo of the filter being used in a playlist.

geekygecko:
Thanks for taking the time with this contribution. We might have to consider some other issues before accepting this change:

- The idea behind filters was it only displayed subscribed podcasts. Allowing unsubscribed podcasts could raise new issues, such as how would remove an unsubscribed podcast from a filter if you are no longer interested in it.
- I don’t believe we sync all unsubscribed podcasts so two mobile devices with the same filter won’t look the same which might cause confusion and support issues.
- We would need to make sync server code changes to support this first.
- If you just open a podcast in discover without playing an episode on Android it adds it to your database as unsubscribed.

aidano:
Thanks for taking the time to outlay the challenges with this. While navigating the code more, I did gain a sense that there are some systemic challenges around making episodes from unsubscribed podcasts a 1st class citizen, so it's not a big surprise. 

That said, I would love it if there was a path to allowing these unsubscribed episodes to be treated largely the same in the app as subscribed episodes, so I hope there are others who feel motivated to contribute to a wider effort if that's needed.  

Some comments regarding the specific challenges you raised:

> The idea behind filters was it only displayed subscribed podcasts. Allowing unsubscribed podcasts could raise new issues, such as how would remove an unsubscribed podcast from a filter if you are no longer interested in it.

I may be misunderstanding the nomenclature and functionality but I thought filters were for podcast episodes rather than podcasts. But understood that there may be, for example, some special logic needed to deal with non-subscribed podcast episodes. Swiping left to archive an episode in the filter I implemented works just fine on iOS in my testing, FWIW.

> I don’t believe we sync all unsubscribed podcasts so two mobile devices with the same filter won’t look the same which might cause confusion and support issues.

Good point!

> We would need to make sync server code changes to support this first.

Understood (presumably for the new filter DB columns, and the above point).

> If you just open a podcast in discover without playing an episode on Android it adds it to your database as unsubscribed.

Would you mind elaborating on the issue with this behavior on Android?

geekygecko:
> Would you mind elaborating on the issue with this behavior on Android?

The way Android currently works is that to display the podcast page the podcast and its episodes are added to the database. The process is the same if you are subscribed, unsubscribed, or just browsing the podcast through the discover section. The podcast can be in any of these states, and the user can tap one of the actions on an episode, such as play, and the app will sync the change to their account. Having all the states share the same logic has made it simpler, but I'm open to other ideas on how it could work.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeFilterDataManager.swift
```diff
@@ -25,7 +25,9 @@ class EpisodeFilterDataManager {
         "wasDeleted",
         "filterDuration",
         "longerThan",
-        "shorterThan"
+        "shorterThan",
+        "filterSubscribed",
+        "filterNotSubscribed"
```
yaelirub, line 30: test

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeFilterDataManager.swift
```diff
@@ -25,7 +25,9 @@ class EpisodeFilterDataManager {
         "wasDeleted",
         "filterDuration",
         "longerThan",
-        "shorterThan"
+        "shorterThan",
```
yaelirub, line 28: Im reviewing

CookieyedCodes:
@geekygecko @leandroalonso while I get this is an old issue that wasn't brought forward I would still see the value in a similar option, would it make more sense to put three these three checkbox options under downloads pill,
- [ ] include loose downloads in filter
- [ ] include downloaded files in filter
- [ ] include cloud files in filter

As this would allow those loose files being included but am I missing something on either platform 🤔 (+ yes it wouldnt match platform to platform but as its under downloads the intent is clear)

I think theirs also a need for a folder filter pill by the way (sorry if I repeat myself) with the option to have the sort order to be based on the folder position basically enabling podcast grouping in filters which would be a massive enhancement for filters & plus in my opinion 😅

SergioEstevao:
Taking the recent work on playlists, this changes are no longe relevant. Thanks for your contribution @aidano

PR #452, opened 2022-10-27
End of Year: listening time story design

leandroalonso:
| 📘 Project: #376 |
|:---:|

Add the design for the "listening time" story.

| iPhone 8 | iPhone 14 | iPhone 14 Pro Max |
| -------- | ---------- | ------------------- |
| Simulator Screen Shot - iPhone 8 - 2022-10-27 at 09 09 13 | Simulator Screen Shot - iPhone 14 - 2022-10-27 at 09 09 47 | Simulator Screen Shot - iPhone 14 Pro Max - 2022-10-27 at 09 10 13 |

## Details

* This PR also adds the dynamic background based on pdeCcb-wX-p2
* In order to display some podcasts on this screen, I added logic to retrieve the top 10 podcasts, reverse them, and then display the last 3. This way we can display podcasts that the user also listened to (while saving the top 5 for later)
* Colors needed to be downloaded, I didn't added logic for that in this PR to avoid being too big

## To test

I recommend using these devices to test, given they provide different screen sizes:

* iPhone 8
* iPhone 14
* iPhone 14 Pro Max

4. Navigate to the second story
4. ✅ Check the design

## Checklist


podcasts/End of Year/Stories/ListeningTimeStory.swift
```diff
@@ -1,27 +1,111 @@
...
     var duration: TimeInterval = 5.seconds
 
     let listeningTime: Double
 
+    let podcasts: [Podcast]
+
     var body: some View {
-        ZStack {
-            Color.black
+        GeometryReader { geometry in
```
leandroalonso, line 13: I didn't extract constants to allow changes to be faster.

Also, a lot of components can be reusable between the stories but my goal was to finish the design of each one before cleaning up the code.

podcasts/End of Year/Stories/ListeningTimeStory.swift
```diff
@@ -1,27 +1,111 @@
...
+    }
+
+    @ViewBuilder
+    func podcastCover(_ index: Int) -> some View {
+        Group {
+            if let podcast = podcasts[safe: index] {
+                ImageView(ServerHelper.imageUrl(podcastUuid: podcast.uuid, size: 280))
+            } else {
+                Rectangle().opacity(0)
             }
```
emilylaguna, line 71: When I loaded this up with 3g throttling enabled in Charles and all I saw was a blank screen for a few seconds until the images loaded up. I wonder if we can display a placeholder so it doesn't appear so empty for users who are on slower connections?

A simple idea could be to just show a rectangle with the podcast color while it's loading in.

Screen Shot 2022-10-31 at 1 20 25 PM

podcasts/End of Year/Stories/ListeningTimeStory.swift
```diff
@@ -1,27 +1,111 @@
...
+                VStack {
+                    Spacer()
+
+                    HStack {
+                        ForEach([1, 0, 2], id: \.self) {
+                            podcastCover($0)
+                        }
+                    }
+                    .modifier(PodcastCoverPerspective())
+                    .position(x: geometry.frame(in: .local).midX, y: geometry.size.height - 230)
```
emilylaguna, line 48: Is there a way to have the view position be a bit more dynamic? On larger screens the large gap at the top looks out of place. Especially when there's less than 3 podcasts to be displayed:

| One item | 2 items |
|:---:|:---:|
|||

podcasts/End of Year/EndOfYearStoriesBuilder.swift
```diff
@@ -62,9 +62,10 @@ class EndOfYearStoriesBuilder {
             }
 
             // Top podcasts
-            let topPodcasts = dataManager.topPodcasts()
+            let topPodcasts = dataManager.topPodcasts(limit: 10)
             if !topPodcasts.isEmpty {
-                data.topPodcasts = topPodcasts
+                data.topPodcasts = Array(topPodcasts.prefix(5))
+                data.randomPodcasts = Array(topPodcasts.suffix(8)).map { $0.podcast }.reversed()
```
emilylaguna, line 68: Is the name `randomPodcasts` still accurate? It seems like this is just returning a subset of podcasts?

emilylaguna (changes requested):
Heya, I tested on a few different devices and left some comments below.
- When testing on larger devices such as the iPhone 14 Pro Max, the gap between the bottom of the top text and the top of the podcasts images seems very large. 
- When testing with a slower internet connection the view appeared very bare while the images were being loaded
- When testing with more limited data (only 1 or 2 "top" podcasts) the view design looked a bit odd on larger devices.

podcasts/End of Year/Stories/ListeningTimeStory.swift
leandroalonso: > A simple idea could be to just show a rectangle with the podcast color while it's loading in.

I think we can add a black translucent square. The colors are tricky because the podcast might have them or not, and if they don't we need to make a request to retrieve them.

podcasts/End of Year/Stories/ListeningTimeStory.swift
leandroalonso: Hmm, having 2 podcasts feel super weird. I think we can add the same as before, a translucent black color.

I wonder how many people will run into having less than 3 podcasts to display though. Do you mind if I tackle that in a separated task?

podcasts/End of Year/Stories/ListeningTimeStory.swift
emilylaguna: True. I think that makes sense.

podcasts/End of Year/EndOfYearStoriesBuilder.swift
emilylaguna: 👆 Since this is used in the subsequent PRs we can address this later.

emilylaguna (approved):

PR #457, opened 2022-10-27
Apple SSO: Fixes Watch app token refreshing issues

emilylaguna:
| 📘 Project: #381 |
|:---:|

This fixes a few issues:

1. When performing a token refresh the some of the required information (email, auth tokens) would be deleted and never restored when a successful refresh occurred.  To fix this I moved when the info is stored to after `handleSuccessfulSignIn` is called ensuring the information is set. If the endpoint doesn't return an email then we manually set our previously stored email again. 

2. If the user revoked the Sign in With Apple the app would be logged out, but the watch wouldn't. This is calling `setLoginDetailsUpdated` in `handleSSOTokenRevoked`

**3. Apple Watch and Apple SSO token refreshes**

If the stored `SJSyncV2Token` is invalid or needs to be refreshed, and the user has signed in with Apple, the code previously would try to do the following to refresh the token

1.  On Failure, it determined the user used Apple SSO
2. It calls `ApiServerHandler.shared.refreshIdentityToken` from `TokenHelper.acquireIdentityToken`
3. `refreshIdentityToken` calls `hasValidSSOToken`
4. Which calls `ssoCredentialState`
5. Which tries to retrieve the current Apple credential state (revoked, authorized, etc) using `ASAuthorizationAppleIDProvider`

The problem with this is that it assumes that:
1. The iOS and Watch apps are able to **share** the Apple SSO login information and check it
    -  The apps don't share this info, and in order for this to work we would need the user to sign in to Apple directly on their watch in the app
2. The Watch app has the correct entitlement for SIWA
   - It's currently not enabled, so it's not able to perform ANY SIWA code so the check will always fail

The change I made in this PR is to completely disable the SIWA checks on the Watch app and rely on the server to refresh, and then a connection with the app 

## To test

> **Note**
> Make sure you enabled the `signInWithApple` FeatureFlag, and switch to the Staging Build Configuration. 
> You can however test the email/password steps without the flag enabled. 

#### Main App
1. Launch the app
2. Sign into an existing account with email/password
3. Go to Profile Tab > Settings > Developer > Corrupt Sync Token
4. Perform a sync
5. ✅ The sync succeeds and you are not logged out

#### Watch App
1. Launch the app and the watch app
7. Sign into an existing account with Plus with email/password on the app
8. Sync the account with the watch
9. In Xcode open KeychainHelper.swift
10. Replace the first 2 methods with the contents of the `KeychainHelper Code` section at the bottom of this PR
    - This will log out any keychain changes making debugging easier
11. in Xcode open: `SourceInterfaceController`
12. Go to line 80 (`override func awake(withContext context: Any?)` method)
13. Add a new line with: `KeychainHelper.save(string: "invalid", key: "SJSyncV2Token", accessibility: kSecAttrAccessibleAfterFirstUnlock)`
    - 👆Upon relaunch of the Watch app this will force the login token to be invalid forcing the token refresh
14. Relaunch the Watch app
15. Once it's launched, tap the `Refresh Data` button
16. ✅ The refresh works correctly
17. ✅ Verify you see `🔐 Removing SJSyncingEmail`,  `🔐 Removing SJSyncingPwd`, and `🔐 Removing SJSyncV2Token`
18. ✅ Verify after all the removals you also see and `🔐 Saving:` log for each of the items above
    - If you were to reproduce the issue you would not see these

#### Main App
1. Launch the app
2. Sign using SIWA
3. Go to Profile Tab > Settings > Developer > Corrupt Sync Token
4. Perform a sync
5. ✅ The sync succeeds and you are not logged out

#### Watch App
1. Launch the app and the watch app
7. Sign into an existing account with Plus using SIWA on the app
8. Sync the account with the watch
9. In Xcode open KeychainHelper.swift
10. Replace the first 2 methods with the contents of the `KeychainHelper Code` section at the bottom of this PR
    - This will log out any keychain changes making debugging easier
11. in Xcode open: `SourceInterfaceController`
12. Go to line 80 (`override func awake(withContext context: Any?)` method)
13. Add a new line with: `KeychainHelper.save(string: "invalid", key: "SJSyncV2Token", accessibility: kSecAttrAccessibleAfterFirstUnlock)`
    - 👆Upon relaunch of the Watch app this will force the login token to be invalid forcing the token refresh
14. Relaunch the Watch app
15. Once it's launched, tap the `Refresh Data` button
16. ✅ The refresh works correctly
18. ✅ Verify you also see and `🔐 Saving: SJSyncV2Token` log with a large value

#### Revoking the login

1. Launch the app
2. Sign in with Apple
3. Sync the account to the watch
4. Go to Settings.app
5. Tap your name at the top of the view
6. Tap on Password & Security
7. Tap on Apps using Apple ID
8. Locate Pocket Casts
9. Tap Stop using Apple ID
10. Go back to the main app
11. ✅ Verify you are signed out
12. Open the Watch app
13. Tap Refresh Data
14. ✅ Verify you are signed out

----

#### KeychainHelper Code
```    @discardableResult
    public class func save(string: String?, key: String, accessibility: CFTypeRef) -> Bool {
        print("🔐 Saving: \(key): \(string ?? "<empty>")")
        return KeychainHelper.shared.save(value: string, key: key, accessibility: accessibility)
    }

    @discardableResult
    public class func removeKey(_ key: String) -> Bool {
        print("🔐 Removing: \(key)")
        // the accessibility flag is ignored on saving a nil value, so it's safe for this helper to put whatever in that field
        return KeychainHelper.shared.save(value: nil, key: key, accessibility: kSecAttrAccessibleAfterFirstUnlock)
    }
```

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+SocialAuth.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+SocialAuth.swift
@@ -4,12 +4,12 @@ import PocketCastsUtils
 
 /**
  * The Watch app and the iOS app don't share SIWA credentials, so if we try to do an "Credentials State" check from SIWA
- * from the watch, it will fail regardless. So, instead we will rely on the server to validate the identity token and fail
- * there if needed.
+ * from the watch, it will fail regardless. So, instead we will rely on the server to validate the identity token and fail there if needed. If not, then the watch will check the login state next time the user is connected to the main app on their phone.
  *
- * We will still do the credential state check in the main app, since it's better to fail early if possible.
+ * We will still do the credential state check in the main app since it's recommended by Apple.
  *
  */
+
 #if !os(watchOS)
 import AuthenticationServices
 
--- a/Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+SocialAuth.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+SocialAuth.swift
@@ -1,6 +1,16 @@
 import Foundation
 import PocketCastsDataModel
 import PocketCastsUtils
+
+/**
+ * The Watch app and the iOS app don't share SIWA credentials, so if we try to do an "Credentials State" check from SIWA
+ * from the watch, it will fail regardless. So, instead we will rely on the server to validate the identity token and fail
+ * there if needed.
+ *
+ * We will still do the credential state check in the main app, since it's better to fail early if possible.
+ *
+ */
+#if !os(watchOS)
 import AuthenticationServices
 
 public extension ASAuthorizationAppleIDProvider.CredentialState {
@@ -19,6 +29,7 @@ public extension ASAuthorizationAppleIDProvider.CredentialState {
         }
     }
 }
+#endif
 
 public extension ApiServerHandler {
     func validateLogin(identityToken: String?) async throws -> AuthenticationResponse {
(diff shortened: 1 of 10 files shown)


Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+SocialAuth.swift
```diff
@@ -66,14 +91,12 @@ public extension ApiServerHandler {
...
-        let url = ServerHelper.asUrl(ServerConstants.Urls.api() + "user/login_apple")
-        guard let identityToken = identityToken,
-              var request = ServerHelper.createEmptyProtoRequest(url: url, cachePolicy: cachePolicy, timeoutInterval: timeoutInterval)
-        else { return nil }
-
-        request.setValue("Bearer \(identityToken)", forHTTPHeaderField: ServerConstants.HttpHeaders.authorization)
-        return request
+extension ApiServerHandler {
+    func hasValidSSOToken() async throws -> Bool {
+        return true
```
emilylaguna, line 99: Intentionally always returning true on the watch app to "skip" the token check. I thought about modifying the place where this is called, but this seemed cleaner.

podcasts/WatchManager.swift
```diff
@@ -334,8 +334,9 @@ class WatchManager: NSObject, WCSessionDelegate {
         if let password = ServerSettings.syncingPassword() {
             response[WatchConstants.Messages.LoginDetailsResponse.password] = password
         }
-        else if let authToken = ServerSettings.appleAuthIdentityToken {
+        else if let authToken = ServerSettings.appleAuthIdentityToken, let appleUserId = ServerSettings.appleAuthUserID {
```
emilylaguna, line 337: The user ID wasn't being sent to the watch before, so this adds that.

podcasts/AuthenticationHelper.swift
```diff
@@ -75,31 +114,7 @@ class AuthenticationHelper {
...
-        if let email = response.email, !email.isEmpty {
-            ServerSettings.setSyncingEmail(email: response.email)
-        }
-
-        NotificationCenter.default.post(name: .userLoginDidChange, object: nil)
-        Analytics.track(.userSignedIn, properties: ["source": source.rawValue])
-
-        RefreshManager.shared.refreshPodcasts(forceEvenIfRefreshedRecently: true)
-        Settings.setPromotionFinishedAcknowledged(true)
         Settings.setLoginDetailsUpdated()
```
emilylaguna, line 119: This wasn't being called before in `handleSSOTokenRevoked` which meant when the main was app signed out due to the token being revoked the watch app wouldn't also sign the user out.

leandroalonso (approved):

PR #471, opened 2022-11-01
End of Year: preload stories

leandroalonso:
| 📘 Project: #376 |
|:---:|

To provide a better experience for the users, the app preloads the next 2 stories. The assets that are preloaded are:

1. Image covers
2. Podcast colors

Those are requested if they do not exist, and they are requests only once (images by using Kingfisher and colors by using the current `ColorManager`).

This also changes how the `StoriesView` is rendered to avoid Stories being re-created hundreds of time.

### Playing stories without skipping to next/previous

2. Do a fresh install of the app
3. Login to an account with a few items on Listening History
3. Use a tool like Network Link Conditioner to change the speed of your connection (to 3G, for example)
3. ✅ Leave the stories to play and check that most stuff should be ready when the stories appear

### Skipping

2. Do a fresh install of the app
3. Login to an account with a few items on Listening History
3. Use a tool like Network Link Conditioner to change the speed of your connection (to 3G, for example)
3. ✅ Quickly goes to the next story and you should be able to see that the background is opaque but once colors are downloaded it changes

### Number of Stories View

A bit of context: `StoriesModel` had the `progress` property as a published one. Each time that changed, `StoriesView` was invalidated and re-rendered again. In turn, this caused the current story to be invalidated and re-rendered again.

In terms of allocation, this wasn't a big problem because the views are mainly `structs`, but the CPU usage was going up due to the view being invalidated every 0.01s.

I moved the progress to another object that is used only by the progress indicators, now the `StoriesView` is only invalidated when the current story changes.

To test:

3. Run Instruments > Allocations and start recording the current Pocket Casts instance
4. Add "Story" to the "Detail Filter"
3. Check that only 1 to 2 total Story views area created

## Checklist


podcasts/End of Year/Stories/DynamicBackgroundView.swift
```diff
@@ -34,13 +30,64 @@ struct DynamicBackgroundView: View {
...
+    private let podcast: Podcast?
+
+    init(podcast: Podcast) {
+        backgroundColor = podcast.bgColor().color
+        foregroundColor = ColorManager.lightThemeTintForPodcast(podcast).color
+        self.podcast = podcast
+
+        if !ColorManager.podcastHasBackgroundColor(podcast) {
+            _ = ColorManager.backgroundColorForPodcast(podcast)
+            NotificationCenter.default.addObserver(self, selector: #selector(podcastColorsLoaded(_:)), name: Constants.Notifications.podcastColorsDownloaded, object: nil)
```
emilylaguna: Since the `podcastColorsDownloaded` notification posts using the Podcast UUID as the object, we can pass it in here to only receive the notification for this podcast, which also means we can reduce the boilerplate checking code needed in `podcastColorsLoaded()`.

```suggestion
            NotificationCenter.default.addObserver(self, selector: #selector(podcastColorsLoaded(_:)), name: Constants.Notifications.podcastColorsDownloaded, object: podcast.uuid)
```

podcasts/End of Year/Stories/DynamicBackgroundView.swift
```diff
@@ -34,13 +30,64 @@ struct DynamicBackgroundView: View {
...
+
+    private let podcast: Podcast?
+
+    init(podcast: Podcast) {
+        backgroundColor = podcast.bgColor().color
+        foregroundColor = ColorManager.lightThemeTintForPodcast(podcast).color
+        self.podcast = podcast
+
+        if !ColorManager.podcastHasBackgroundColor(podcast) {
+            _ = ColorManager.backgroundColorForPodcast(podcast)
```
emilylaguna: Nit: What do you think about adding `@discardableResult` to the `backgroundColorForPodcast` so we don't need to do `_ =`?

podcasts/End of Year/Stories/DynamicBackgroundView.swift
```diff
@@ -34,13 +30,64 @@ struct DynamicBackgroundView: View {
...
 }
 
+@MainActor
+class DynamicBackgroundProvider: ObservableObject {
+    @Published var backgroundColor: Color
+    @Published var foregroundColor: Color
+
+    private let podcast: Podcast?
+
+    init(podcast: Podcast) {
```
emilylaguna, line 47: Not a blocker: I was wondering if we could provide a static default color to use in case the podcast colors don't load in? This way we can still have some pretty looking backgrounds.

podcasts/End of Year/Views/StoryIndicator.swift
```diff
@@ -11,7 +13,7 @@ struct StoryIndicator: View {
                         .cornerRadius(Constants.storyIndicatorBorderRadius)
 
                     Rectangle()
-                        .frame(width: geometry.size.width * progress, height: nil, alignment: .leading)
+                        .frame(width: geometry.size.width * (min(max((CGFloat(model.progress) - CGFloat(index)), 0.0), 1.0)), height: nil, alignment: .leading)
```
emilylaguna: Nit: I found out recently we have a clamp function available: 


```suggestion
                        .frame(width: geometry.size.width * (model.progress - CGFloat(index)).clamped(to: 0.0 ..< 1.0), height: nil, alignment: .leading)
```

emilylaguna (approved):
Nice work, left some comments to clean up a few items and an idea for default colors. 

None of them are blockers though.

podcasts/End of Year/Stories/DynamicBackgroundView.swift
leandroalonso: Done in d829d221e54c88516e2caecfe5235137941c4650

podcasts/End of Year/Stories/DynamicBackgroundView.swift
leandroalonso: Instead of doing that I exposed `updateColorsIfRequired` in 1455376bfbb58a96651d95a575ec8aba62b49260

podcasts/End of Year/Stories/DynamicBackgroundView.swift
leandroalonso: I'll add that to the list of nice to have. :)

podcasts/End of Year/Views/StoryIndicator.swift
leandroalonso: Done in 9b7351efb7e6d615df4f0d303a0a2c1ecf388b9a

PR #473, opened 2022-11-01
Use the official, working way of setting user-agent in iOS16+

johnspurlock-skymethod:
> As we’ve documented in the past, [many different apps use a useragent called AppleCoreMedia to request podcasts](https://podnews.net/article/applecoremedia-user-agent), making it hard for podcasters, and podcast hosting companies, to know which app is really downloading them. iOS 16 will fix that: [AVURLAssetHTTPUserAgentKey has just appeared in Apple’s documentation](https://developer.apple.com/documentation/avfoundation/avurlassethttpuseragentkey?utm_source=podnews.net&utm_medium=web&utm_campaign=podnews.net:2022-06-07) for the next version of iOS, allowing all podcast apps to correctly state their user agent.
>
> — https://podnews.net/update/applecoremedia-gone-off

Apple doc: https://developer.apple.com/documentation/avfoundation/avurlassethttpuseragentkey

This will ensure Pocket Casts is properly credited for all requests it makes when folks are analyzing backend server logs for app attribution.

## To test

1. Use a tool like Charles proxy (or any other proxy tool)
2. Run Pocket Casts
3. Check the requests for the asset
4. ✅ Ensure they have "Pocket Casts" as the user-agent:

Screen Shot 2022-11-02 at 09 25 12

## Checklist

### DIFF
--- a/podcasts/PlaybackItem.swift
+++ b/podcasts/PlaybackItem.swift
@@ -15,7 +15,7 @@ class PlaybackItem: NSObject {
 
     func createPlayerItem() -> AVPlayerItem? {
         guard let url = EpisodeManager.urlForEpisode(episode) else { return nil }
-        var options = [String : Any]()
+        var options: [String: Any] = [:]
         if #available(iOS 16, *), #available(watchOSApplicationExtension 9.0, *) {
             // there is now an official, working way to set the user-agent for every request
             // https://developer.apple.com/documentation/avfoundation/avurlassethttpuseragentkey
--- a/podcasts/PlaybackItem.swift
+++ b/podcasts/PlaybackItem.swift
@@ -24,7 +24,6 @@ class PlaybackItem: NSObject {
             let customHeaders = [ServerConstants.HttpHeaders.userAgent: ServerConstants.Values.appUserAgent]
             options["AVURLAssetHTTPHeaderFieldsKey"] = customHeaders
         }
-        
         let asset = AVURLAsset(url: url, options: options)
 
         return AVPlayerItem(asset: asset)
--- a/podcasts/PlaybackItem.swift
+++ b/podcasts/PlaybackItem.swift
@@ -15,9 +15,17 @@ class PlaybackItem: NSObject {
 
     func createPlayerItem() -> AVPlayerItem? {
         guard let url = EpisodeManager.urlForEpisode(episode) else { return nil }
-
-        let customHeaders = [ServerConstants.HttpHeaders.userAgent: ServerConstants.Values.appUserAgent]
-        let asset = AVURLAsset(url: url, options: ["AVURLAssetHTTPHeaderFieldsKey": customHeaders])
+        var options = [String : Any]()
+        if #available(iOS 16, *) {
+            // there is now an official, working way to set the user-agent for every request
+            // https://developer.apple.com/documentation/avfoundation/avurlassethttpuseragentkey
+            options[AVURLAssetHTTPUserAgentKey] = ServerConstants.Values.appUserAgent
+        } else {
+            let customHeaders = [ServerConstants.HttpHeaders.userAgent: ServerConstants.Values.appUserAgent]
+            options["AVURLAssetHTTPHeaderFieldsKey"] = customHeaders
+        }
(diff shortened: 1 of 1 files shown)


podcasts/PlaybackItem.swift
```diff
@@ -15,9 +15,17 @@ class PlaybackItem: NSObject {
 
     func createPlayerItem() -> AVPlayerItem? {
         guard let url = EpisodeManager.urlForEpisode(episode) else { return nil }
-
-        let customHeaders = [ServerConstants.HttpHeaders.userAgent: ServerConstants.Values.appUserAgent]
-        let asset = AVURLAsset(url: url, options: ["AVURLAssetHTTPHeaderFieldsKey": customHeaders])
+        var options = [String : Any]()
```
leandroalonso: There's a warning here:

> Colon Spacing Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (colon)

```suggestion
        var options: [String: Any] = [:]
```

podcasts/PlaybackItem.swift
```diff
@@ -15,9 +15,17 @@ class PlaybackItem: NSObject {
...
+        if #available(iOS 16, *) {
+            // there is now an official, working way to set the user-agent for every request
+            // https://developer.apple.com/documentation/avfoundation/avurlassethttpuseragentkey
+            options[AVURLAssetHTTPUserAgentKey] = ServerConstants.Values.appUserAgent
+        } else {
+            let customHeaders = [ServerConstants.HttpHeaders.userAgent: ServerConstants.Values.appUserAgent]
+            options["AVURLAssetHTTPHeaderFieldsKey"] = customHeaders
+        }
+
```
leandroalonso: Also a warning:

> Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)

```suggestion

```

podcasts/PlaybackItem.swift
```diff
@@ -15,9 +15,17 @@ class PlaybackItem: NSObject {
 
     func createPlayerItem() -> AVPlayerItem? {
         guard let url = EpisodeManager.urlForEpisode(episode) else { return nil }
-
-        let customHeaders = [ServerConstants.HttpHeaders.userAgent: ServerConstants.Values.appUserAgent]
-        let asset = AVURLAsset(url: url, options: ["AVURLAssetHTTPHeaderFieldsKey": customHeaders])
+        var options = [String : Any]()
+        if #available(iOS 16, *) {
```
leandroalonso: The code doesn't compile because `AVURLAssetHTTPUserAgentKey` is only available for watch OS extension 9+, we need to do this check too.

```suggestion
        if #available(iOS 16, *), #available(watchOSApplicationExtension 9.0, *) {
```

leandroalonso (changes requested):
Hi @johnspurlock-skymethod, thanks for your contribution.

However, the code doesn't compile and there are two warnings.

On another note, I tested both your and current changes and `trunk` and I saw no difference in the requests to get the podcast assets. In both of them, the user-agent is "Pocket Casts".

We can use the new iOS 16 key but in practice I wasn't able to verify any difference — it seems our user-agent is already correct. Let me know if I'm missing something, and thanks again.

Ps.: I also took the liberty to edit your PR description with testing steps.

johnspurlock-skymethod:
Hi Leandro, thanks for taking a look!  Good point about compiling on the watch as well. Feel free to modify as you see fit according to your project's style conventions!

It's important to make this change: Apple added this official constant specifically so you don't have to resort to what you're doing now.  Right now you're using something that's undocumented, doesn't work in all cases, and technically part of the Apple private API, which runs afoul of the Developer Agreement and can get you blocked in review if they find out.

Looking forward to one of the best podcast players on iOS be one of the first apps to make this change!

podcasts/PlaybackItem.swift
leandroalonso: @johnspurlock-skymethod can you please commit this suggestion? Then we'd be able to merge this PR :)

leandroalonso (approved):

PR #483, opened 2022-11-04
End of Year: sync Listening History

leandroalonso:
| 📘 Project: #376 |
|:---:|

If we identify the user doesn't have the entire 2022 listening history, we sync this with the server adding the missing episodes.

This is a case that might happen for a very active user that switched devices during this year — given we only sync the latest 100 items.

**How do we identify if a user needs a sync?**

~We look at their listening history. If they have an episode from last year, we assume that their 2022 listening history is complete.~

~A much more bulletproof strategy would be to call the endpoint to check the number of items for the user, and sync if needed, **just once**. But I'm not sure what would be the impact of such a call on the backend (@Ferdev curious to hear your thoughts about it).~

We request the number of items the server has. If the returned number is smaller than what we have locally, we sync the necessary items.

## To test

You'll need an account with more than 100 items. Please ping me and I can provide one.

2. Do a clean run of the app in Staging mode (Schemes > pocketcasts > change "Build Configuration" to Staging)
3. Login with an account with more than 100 items
4. Check your listening history, you should see just ~40 items
5. Put a breakpoint on `EndOfYearStoriesBuilder.swift` line `39`
6. Check your EoY stories
7. ✅ The hitpoint should be hit, proceed with execution
8. ✅ You should see the spinner for a few seconds
9. ✅ The stories should appear
10. Dismiss the stories
11. Check your Listening History
12. ✅ You should have more than 100 items there now (the oldest episode is an archived episode called "39. Mother"
13. Check your stories again
14. ✅ The breakpoint should not be hit (2022 sync happens just once)

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -74,6 +74,31 @@ class EndOfYearDataManager {
         return isFullListeningHistory
     }
 
+    /// Returns the number of episodes we have for this year
+    func numberOfEpisodes(dbQueue: FMDatabaseQueue) -> Int {
+        var numberOfEpisodes: Int = 0
+
+        dbQueue.inDatabase { db in
+            do {
+                let query = """
+                            SELECT COUNT(*) as numberOfEpisodes from \(DataManager.episodeTableName)
+                            WHERE
+                            \(listenedEpisodesThisYear)
+                            """
+                let resultSet = try db.executeQuery(query, values: nil)
+                defer { resultSet.close() }
+
+                if resultSet.next() {
+                    numberOfEpisodes = Int(resultSet.int(forColumn: "numberOfEpisodes"))
+                }
+            } catch {
+                FileLog.shared.addMessage("EndOfYearDataManager.numberOfEpisodes error: \(error)")
+            }
+        }
+
+        return numberOfEpisodes
+    }
+
     /// Returns the approximate listening time for the current year
     func listeningTime(dbQueue: FMDatabaseQueue) -> Double? {
         var listeningTime: Double?
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -63,8 +63,6 @@ class EndOfYearDataManager {
 
                 if resultSet.next() {
                     isFullListeningHistory = true
-                } else {
-                    isFullListeningHistory = numberOfItemsInListeningHistory(db: db) <= 100
                 }
             } catch {
                 FileLog.shared.addMessage("EndOfYearDataManager.isFullListeningHistory error: \(error)")
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -11,7 +11,7 @@ class EndOfYearDataManager {
 
     /// If the user is eligible to see End of Year stats
     ///
-    /// All it's needed is a single episode listened for more than 30 minutes.
+    /// All it's needed is a single episode listened for more than 5 minutes.
     func isEligible(dbQueue: FMDatabaseQueue) -> Bool {
         var isEligible = false
 
@@ -20,7 +20,7 @@ class EndOfYearDataManager {
                 let query = """
                             SELECT playedUpTo from \(DataManager.episodeTableName)
                             WHERE
-                            playedUpTo > 1800 AND
+                            playedUpTo > 300 AND
                             \(listenedEpisodesThisYear)
                             LIMIT 1
                             """
--- a/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
@@ -6,7 +6,6 @@ import SwiftProtobuf
 class SyncYearListeningHistoryTask: ApiBaseTask {
     private var token: String?
 
-    private let podcastHelper = PodcastExistHelper()
     private let yearToSync: Int32
 
     var success: Bool = false
@@ -125,8 +124,8 @@ class SyncYearListeningHistoryTask: ApiBaseTask {
 
 /// Helper that checks for podcast existence
 /// It caches database requests
-class PodcastExistHelper {
-    static let shared = PodcastExistHelper()
+class PodcastExistsHelper {
+    static let shared = PodcastExistsHelper()
 
     var checkedUuidsThatExist: [String] = []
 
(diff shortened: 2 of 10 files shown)


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -20,7 +20,7 @@ class EndOfYearDataManager {
                 let query = """
                             SELECT playedUpTo from \(DataManager.episodeTableName)
                             WHERE
-                            playedUpTo > 1800 AND
+                            playedUpTo > 300 AND
```
leandroalonso, line 23: If a user had listened to multiple episodes with less than 30 minutes they were not eligible, I changed that to just 5 minutes.

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
```diff
@@ -0,0 +1,158 @@
...
+            if let episodes = ApiServerHandler.shared.retrieveEpisodeTaskSynchronouusly(podcastUuid: $0) {
+                DataManager.sharedManager.saveBulkEpisodeSyncInfo(episodes: DataConverter.convert(syncInfoEpisodes: episodes))
+            }
+        }
+    }
+}
+
+/// Helper that checks for podcast existence
+/// It caches database requests
+class PodcastExistHelper {
```
leandroalonso: I created this to avoid calling the database over and over to check if an episode exists.

Ferdev:
Hey @leandroalonso!
> A much more bulletproof strategy would be to call the endpoint to check the number of items for the user, and sync if needed, just once. But I'm not sure what would be the impact of such a call on the backend (@Ferdev curious to hear your thoughts about it)

I don't think it would be a problem if we just call it once. The database query is pretty simple and the whole request doesn't execute any crazy calculations, so we should be fine if we just run it one time :+1: .

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
```diff
@@ -0,0 +1,156 @@
...
+        self.token = token
+
+        performRequest(token: token, shouldSync: false)
+    }
+
+    private func performRequest(token: String, shouldSync: Bool) {
+        var dataToSync = Api_YearHistoryRequest()
+        dataToSync.deviceTime = TimeFormatter.currentUTCTimeInMillis()
+        dataToSync.version = apiVersion
+        dataToSync.year = 2022
```
emilylaguna: What do you think about allowing this to be injected?

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
```diff
@@ -0,0 +1,156 @@
...
+                }
+
+                dispatchGroup.leave()
+            }
+        }
+
+        dispatchGroup.wait()
+
+        // Sync episode status for the retrieved podcasts' episodes
+        let uniqueUuidsToUpdate = Array(Set(podcastsToUpdate))
```
emilylaguna: Instead of converting the array -> set -> array. What do you think about just changing podcastsToUpdate to a Set?

`var podcastsToUpdate: Set<String> = []`

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
```diff
@@ -0,0 +1,156 @@
...
+        if exists {
+            checkedUuidsThatExist.append(uuid)
+        }
+
+        return exists
+    }
+
+    func markAsExistent(uuid: String) {
+        checkedUuidsThatExist.append(uuid)
+    }
```
emilylaguna: This doesn't seem to be used, should we remove it?

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
```diff
@@ -0,0 +1,156 @@
...
+            if let episodes = ApiServerHandler.shared.retrieveEpisodeTaskSynchronouusly(podcastUuid: $0) {
+                DataManager.sharedManager.saveBulkEpisodeSyncInfo(episodes: DataConverter.convert(syncInfoEpisodes: episodes))
+            }
+        }
+    }
+}
+
+/// Helper that checks for podcast existence
+/// It caches database requests
+class PodcastExistHelper {
```
emilylaguna: Nit: 
```suggestion
class PodcastExistsHelper {
```

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
```diff
@@ -0,0 +1,156 @@
...
+        dataToSync.deviceTime = TimeFormatter.currentUTCTimeInMillis()
+        dataToSync.version = apiVersion
+        dataToSync.year = 2022
+        dataToSync.count = !shouldSync
+
+        let url = ServerConstants.Urls.api() + "history/year"
+        do {
+            let data = try dataToSync.serializedData()
+            let (response, httpStatus) = postToServer(url: url, token: token, data: data)
+            if let response = response, httpStatus == ServerConstants.HttpConstants.ok {
```
emilylaguna: Nit: 
```suggestion
            if let response, httpStatus == ServerConstants.HttpConstants.ok {
```

emilylaguna (changes requested):
Heya! This worked well in my testing. I left some comments, let me know what you think.

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
leandroalonso: Done in 9001df4ea623989bf8817411c56c2ddb6e85e0a4

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
leandroalonso: Done in fecb93b21f5195525be729b208bb70c9aa94148b

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
leandroalonso: Done in 194ddece6dbffacc694d6c9be8a12dde2f04598d

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
leandroalonso: Done in 21f03cb5f02e7962c5cb7a00d451b1685f0f7996

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
leandroalonso: Done in fdd751b6a44ff5613e559ba8b30e585eb37ce88c

emilylaguna (approved):

PR #484, opened 2022-11-04
Resume downloads on next launch after force quit

### LINKED ISSUE #472: Episode Downloads Do Not Resume When Reopening After Force Quitting App
### Expected behavior
If a download is in progress when the app is force quit, the download should resume when the app is reopened.

### Actual behavior
The episodes will appear as if they are downloading — with the download spinner displayed. But the app isn't actually downloading the episode file.

1. Force quit the app during an episode download.
2. Reopen the app.

jgcaruso:
Fixes #472 

This PR adds a check for episodes in the `downloading` status that aren't actually downloading, and re-adds them to the queue. This is performed when other "resume download" logic is run in `checkForPendingAndAutoDownloads`

## To test

1. pull branch and build
2. start downloading an episode
3. once the download has started, force quit the app
4. reopen the app and browse to the episode again (or check the Profile -> Downloads section)
5. The episode should have auto resumed download
6. Ensure download only on wifi is enabled ( Profile -> Gear -> auto download -> Only on WiFi )
7. Force `isConnectedToWifi()` to return `false` and rebuild (or if there is some simulator setting I couldn't find that simulates a cell connection, do that)
8. Repeat the test plan, when you get to step 5 the episode SHOULD NOT resume download and you should see a wifi icon in the episode cell
9. Remove the forced cell connection behaviour and rebuild
10. When the app opens, return to the Profile -> Downloads list.
11. The episode should resume downloading.

NOTE: currently, the download resumes from 0% since there doesn't appear to be any download progress state saved. Also, it might not be desirable to resume from the existing file's position since this download could resume several days later which might cause issues with playback due to auto ad insertion.

NOTE 2: I tested in an iOS 14.1 simulator prior to applying this branch, and downloads didn't resume there either, so it doesn't appear that they ever resumed if no code changes have been made around this recently... or at least, its not the result of a fundamental change to iOS.

## Checklist


podcasts/PodcastManager.swift
```diff
@@ -91,6 +91,14 @@ class PodcastManager: NSObject {
             }
         }
 
+        // check if any existing episodes were downloading, but aren't currently (caused by app force quit while episode was downloading)
+        let stuckDownloadingEpisodes = DataManager.sharedManager.findEpisodesWhere(customWhere: "episodeStatus == ?", arguments: [DownloadStatus.downloading.rawValue])
```
leandroalonso, line 95: @jgcaruso if someone configured the app to auto-download only on Wi-fi, I believe this will bypass this setting.

The way it works now is that if someone closes the app downloads won't be resumed, so it's up to the user to add them again — whether they are on Wi-fi or cellular network. With this change, downloads will resume as soon as the user opens the app — it doesn't matter if on Wi-Fi or not. If they have 10 episodes to download, this will download from their cellular data if they're not on Wi-fi.

While your change works as expected, I wonder whether not resuming pending downloads was an intentional decision.

~I think we can check how that works on Android. :)~

It seems that Android downloads the episode even if the user closes the app, so it's a key difference between OSes. 

@emilylaguna any thoughts?

jgcaruso: > With this change, downloads will resume as soon as the user opens the app — it doesn't matter if on Wi-Fi or not.

Hmm good point. I suppose an extra check to the Podcast's setting for "only download on wifi" could be done before re-starting the download, and fixing the status to `waitingForWifi` if the device is currently on cellular, so it gets picked up next time the device is on wifi.

leandroalonso: @jgcaruso that looks like a good solution!

leandroalonso (approved):
Sorry for the delay on finish the review on this one @jgcaruso, tested and it works just fine.

You might want to update the `CHANGELOG.md` before merging though.

PR #494, opened 2022-11-08
Onboarding: Adds some initial SwiftUI extensions and views

emilylaguna:
| 📘 Project: #500 |
|:---:|

This adds some base views that I made while implementing the updated onboarding views:

- `ScrollViewIfNeeded`: Checks the size of its content and will dynamically wrap itself in a scrollview if the content will overflow outside the original bounds
- `ContentSizeReader`: Reads the size of the wrapped content and returns the contentSize
- `HorizontalScrollView`: A horizontally scrolling view that adjusts its height to the its content size
- `Action`:  A helper view that allows you to easily run code without needing to bind to .onAppear

This also adds a FeatureFlag, the ability to make a Color from Hex, dynamic font scaling for SwiftUI, and a Button modifier.

These will be used in the upcoming view updates and changes. 

## To test

Each of the classes has a demo preview in their respective files:

1. In Xcode go to Pocket Casts/Utilities/SwiftUI
2. Click on each of the items
3. Show the Canvas if needed
4. Verify each of them work as intended

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Extensions/UI/Color+Hex.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Extensions/UI/Color+Hex.swift
@@ -1,0 +1,28 @@
+import SwiftUI
+
+extension Color {
+    public init(hex: String) {
+        let hex = hex.trimmingCharacters(in: CharacterSet.alphanumerics.inverted)
+        var int: UInt64 = 0
+        Scanner(string: hex).scanHexInt64(&int)
+        let a, r, g, b: UInt64
+        switch hex.count {
+        case 3: // RGB (12-bit)
+            (a, r, g, b) = (255, (int >> 8) * 17, (int >> 4 & 0xF) * 17, (int & 0xF) * 17)
+        case 6: // RGB (24-bit)
+            (a, r, g, b) = (255, int >> 16, int >> 8 & 0xFF, int & 0xFF)
+        case 8: // ARGB (32-bit)
+            (a, r, g, b) = (int >> 24, int >> 16 & 0xFF, int >> 8 & 0xFF, int & 0xFF)
+        default:
+            (a, r, g, b) = (1, 1, 1, 0)
+        }
+
+        self.init(
+            .sRGB,
+            red: Double(r) / 255,
+            green: Double(g) / 255,
+            blue: Double(b) / 255,
+            opacity: Double(a) / 255
+        )
+    }
+}
--- a/podcasts/SwiftUI/ActionView.swift
+++ b/podcasts/SwiftUI/ActionView.swift
@@ -1,0 +1,56 @@
+import SwiftUI
+
+
+/// This is a helper view that allows you to run code inline of SwiftUI without needing to move logic into .onAppear
+/// It works by creating a completely empty view, waiting for .onAppear, and then trigering the action. Once the action
+/// has been triggered, the emptyview is replaced with a Group to prevent any possible layout issues
+/// Example Usage:
+///
+/// ```
+///  ZStack{
+///     Action { print("debugging is cool 😎") }
+///     VStack {
+///         ... some views ...
+///     }
+///
+///     ... some views ...
+///
+///     Action {
+///         viewModel.loadContent()
+///     }
+/// }
+/// ```
+struct Action: View {
+    init(_ action: @escaping () -> Void) {
+        self.action = action
+    }
+
+    private let action: () -> Void
+    @State private var didPerform: Bool? = nil
+
+    var body: some View {
+        // If we performed the action already, remove the "caller" view from the stack
+        // and just return a group
+        if didPerform != nil {
+            Group { }
+        } else {
+            // If the action hasn't been performed yet, we'll create an empty view and listen for the onAppear
+            NoView().onAppear() {
+                action()
+                didPerform = true
+            }
+        }
+    }
+
+    /// This clears a "view" that has no frame, and appears very far off screen.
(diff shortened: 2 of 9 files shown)


Modules/Utils/Sources/PocketCastsUtils/Extensions/UI/Color+Hex.swift
```diff
@@ -0,0 +1,28 @@
+import SwiftUI
+
+extension Color {
+    public init(hex: String) {
```
emilylaguna, line 4: I know we can do: `UIColor(hex: "#ffe0ff").color` but it is much nicer and cleaner to just do `Color(hex: "#ffe0ff")`.

podcasts/UIFont+FontStyle.swift
```diff
@@ -39,8 +40,114 @@ extension UIFont {
...
+        let scaledSize = min(maxPointSize, metrics.scaledValue(for: size, compatibleWith: sizeCategory.traitCollection))
+
+        // Return the new calculated font
+        return content.font(.system(size: scaledSize, weight: weight))
+    }
+}
+
+// MARK: - Enums to map SwiftUI -> UIKit font types
+
+private extension ContentSizeCategory {
```
emilylaguna, line 87: This is A LOT of code for what it is, but unfortunately I don't have a better solution to map the values.

podcasts/SwiftUI/ActionView.swift
```diff
@@ -0,0 +1,78 @@
...
+///     }
+///
+///     ... some views ...
+///
+///     Action {
+///         viewModel.loadContent()
+///     }
+/// }
+/// ```
+struct Action: View {
```
leandroalonso, line 23: Isn't that similar to [.task](https://developer.apple.com/documentation/swiftui/view/task(priority:_:))? (which is iOS 15+ only)

podcasts/SwiftUI/ContentSizeReader.swift
```diff
@@ -0,0 +1,55 @@
+import SwiftUI
+
+/// Calculates the size of the containing view and returns it
+/// This allows you to get easily get the size of the view without needing all the boilerplate for `GeometryReader`
```
leandroalonso:
```suggestion
/// This allows you to easily get the size of the view without needing all the boilerplate for `GeometryReader`
```

podcasts/SwiftUI/ContentSizeReader.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @State private var size: CGSize?
+
+    var body: some View {
+        // If we've calculated the size, then just return it and the view
+        if let size {
+            content(size)
+        } else {
+            content(nil).background(
```
leandroalonso: I don't think using `GeometryReader` is too problematic in terms of boilerplate code by doing `reader.size.width` or `reader.size.height`.

I also don't see a problem using this wrapper, I just want to point one key difference: this wrapper forces the view to render twice (because of the `@State size`).

I'm not sure if there's any cost in terms of CPU — I guess not because SwiftUI is smart enough to only re-render views if something changed — but if someone uses `ContentSizeReader` together with a `.onAppear` this will get called twice, which I think is an unintended behavior.

podcasts/SwiftUI/ScrollViewIfNeeded.swift
```diff
@@ -0,0 +1,83 @@
...
+                    // Calculate the content size again, but this time of the "true" size
+                    ContentSizeReader { contentSize in
+                        Action {
+                            guard let contentSize else { return }
+
+                            willOverflow = contentSize.height > geometry.size.height
+                        }
+                    }
+                )
+            }
```
leandroalonso: Since `ContentSizeReader` already uses `.background` I think we can just do:

```swift
            GeometryReader { geometry in
                // Calculate the content size again, but this time of the "true" size
                ContentSizeReader { contentSize in
                    content()

                    Action {
                        guard let contentSize else { return }

                        willOverflow = contentSize.height > geometry.size.height
                    }
                }
            }
```

leandroalonso (approved):
@emilylaguna left a few comments, none of them blockers.

podcasts/SwiftUI/ContentSizeReader.swift
emilylaguna: > I don't think using GeometryReader is too problematic in terms of boilerplate code by doing reader.size.width or reader.size.height.

The big difference between GeometryReader and this is that since the ContentSizeReader doesn't calculate the value until after the view has appeared the size will be represent the actual size of the content. You can see this if you replace ContentSizeReader with the a GeometryReader in `HorizontalScrollView`

> I'm not sure if there's any cost in terms of CPU — I guess not because SwiftUI is smart enough to only re-render views if something changed — but if someone uses ContentSizeReader together with a .onAppear this will get called twice, which I think is an unintended behavior.

I fixed this by switching how I pass the contentSize around: 2a58a3ac7bb3267fe7331c445ff7b4e90e81c887 now it uses a binding and makes the code that uses it cleaner.

podcasts/SwiftUI/ScrollViewIfNeeded.swift
emilylaguna: I swapped this to use a second GeometryReader since using the ContentSizeReader for this wasn't necessary: dc6a038d0e6c8b2a25b43e3290058592fa8f4761

podcasts/SwiftUI/ActionView.swift
emilylaguna: :( Yeah it is. If we go iOS 15+ we can remove the Action and use .task

PR #495, opened 2022-11-08
End of Year: small fixes

leandroalonso:
| 📘 Project: #376 |
|:---:|

While testing the feature on iOS 14/15/16 and different scenarios I found some small issues and they're all fixed in this PR.

I explain every issue in the "To test" section.

## To test

Make sure you have `tracksLoggingEnabled` and `endOfYear` enabled in `FeatureFlag.swift`.

### Modal being cut at the bottom on iOS 15

On iOS 15 in a device without the Home button, the modal was transparent in the bottom part. This was fixed by using a `ThemeableView` in `MDCSwiftUIWrapper`.

1. Run the app on iOS 15
2. Login so the bottom sheet appears
3. ✅ Check that the bottom part is not transparent

### Profile card

1. Run the app
2. Go to profile
3. Toggle light/dark mode
4. ✅ Check that the background of the cell that the card is embedded changes accordingly

### Stories failed and gestures

Go to `EndOfYearStoriesBuilder` and add this to line `36`:

```swift
continuation.resume(returning: ([], data))
return
```

1. Run the app
2. Check your stories
3. ✅ `end_of_year_stories_failed_to_load` should be tracked
4. Tap the left and right portions of the screen
5. ✅ Nothing should happen
6. Swipe to dismiss
7. ✅ Stories should dismiss

### 16:9 proportion and sharing to Facebook

The image height was slightly changed to be in the 16:9 proportion.

1. Run the app
2. Check your stories
3. Share any story as a Facebook story
4. ✅ Check that there aren't any left or right space left when sharing on FB

### Share Sheet on iOS 15

Specifically, on iOS 15, we have an issue where the sheet is dismissed when the "Saving Image" option is used. The problem is that this also dismisses the stories, and we don't want that. A fix was applied for iOS 15 but is also used for iOS 14 and 16 given that the result there is the same.

1. Run the app on iOS 15 (device or simulator)
2. Tap Share
3. Tap "Save Image"
4. ✅ The sheet should dismiss but the stories should still display
5. Tap Share again and dismiss
6. ✅ Stories should play again
7. You can also try sharing to other places, the behavior should be the same as before

## Checklist

### DIFF
--- a/podcasts/End of Year/Stories/MDCSwiftUIWrapper.swift
+++ b/podcasts/End of Year/Stories/MDCSwiftUIWrapper.swift
@@ -13,6 +13,8 @@ class MDCSwiftUIWrapper<ContentView: View>: UIViewController {
     init(rootView content: ContentView) {
         super.init(nibName: nil, bundle: nil)
 
+        view = ThemeableView()
+
         view.addSubview(stackView)
         stackView.translatesAutoresizingMaskIntoConstraints = false
         NSLayoutConstraint.activate([
--- a/podcasts/End of Year/Stories/MDCSwiftUIWrapper.swift
+++ b/podcasts/End of Year/Stories/MDCSwiftUIWrapper.swift
@@ -13,8 +13,6 @@ class MDCSwiftUIWrapper<ContentView: View>: UIViewController {
     init(rootView content: ContentView) {
         super.init(nibName: nil, bundle: nil)
 
-        view = ThemeableView()
-
         view.addSubview(stackView)
         stackView.translatesAutoresizingMaskIntoConstraints = false
         NSLayoutConstraint.activate([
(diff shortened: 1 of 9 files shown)


podcasts/End of Year/Stories/MDCSwiftUIWrapper.swift
```diff
@@ -13,6 +13,8 @@ class MDCSwiftUIWrapper<ContentView: View>: UIViewController {
     init(rootView content: ContentView) {
         super.init(nibName: nil, bundle: nil)
 
+        view = ThemeableView()
```
emilylaguna: View overriding should be done by overriding `loadView()`

emilylaguna (changes requested):
Nice, everything worked as intended. I left 1 comment about overloading the view.

podcasts/End of Year/Stories/MDCSwiftUIWrapper.swift
leandroalonso: Done in 9d96fc93964af1710b6c670e71abb1345c2c67b4

emilylaguna (approved):

PR #497, opened 2022-11-09
Onboarding: Adds the styling for the updated plus landing view

emilylaguna:
| 📘 Project: #500 |
|:---:|

This adds the styling for the plus landing view, which includes:
- Image assets for the onboarding flow
- Reusable plus button styles
    - The stroke style can be used as a selectable button that shows a checkmark
- Includes a new common component, a dynamicProperty named ScalingValue (Maybe need a better name)
   - This allows you to define a simple var and have it update dynamically based on the view's frame without needing to do any other calculations.
   - I found this really useful when laying out "absolute" positioned that also needed to appear consistently across various view sizes without needing to do (geometry.size.width * percent) over and over
       - You can see this in use in the `PlusBackgroundGradientView`
        - Also see the Example preview in the ScalingValue 
- This also shows real use of a few components that were added:
    - ScrollViewIfNeeded
    - HorizontalScrollView
Currently the text isn't localized which I'm planning on doing in a future PR, iPad support is also not done either.

## Screenshots
| iPhone SE | iPhone 14 Pro | iPhone 14 Pro Max|
|:---:|:---:|:---:|
||||

## To test
1. With the feature flag off
2. Go to the Profile tab
3. Tap the Create Folder button
4. Verify the existing behavior works
5. Enable the `onboardingUpdates` FeatureFlag
6. Launch on a larger device
7. Go to the Podcasts view
8. Tap the Folder button
9. ✅ Compare [against the designs](https://www.figma.com/file/PviUP0aiZctOprDuuYRwpb/Onboarding-%26-Upgrade-v2?node-id=406%3A24621) to make sure the view matches and looks good
10. Relaunch on a smaller device like the iPhone SE, and verify the view now scrolls

## Checklist


podcasts/Onboarding/Plus/PlusButtonStyles.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    func makeBody(configuration: Configuration) -> some View {
+        configuration.label
+            .updateFont()
+            .frame(maxWidth: .infinity)
+            .padding()
+
+            // Overlay the gradient, or just set the color if not selected
+            .foregroundColor(isSelected ? nil : Color.unselectedColor)
+            .gradientOverlay(isSelected ? Color.plusGradient : nil)
```
emilylaguna, line 36: I couldn't just easily pass the value into one of the functions since they are 2 different types.

podcasts/Onboarding/Plus/PlusLandingView.swift
```diff
@@ -0,0 +1,243 @@
...
+                            dismissAction?()
+                        }.buttonStyle(PlusGradientStrokeButton())
+                    }
+                }.padding(ViewConfig.padding.view)
+            }
+        }.enableValueScaling()
+    }
+
+    // Static list of the feature models to display
+    private let features = [
```
emilylaguna, line 49: I may change how this works in the future if I need to reuse the models, but for now they're just hardcoded and static.

podcasts/SwiftUI/ScalingValue.swift
```diff
@@ -0,0 +1,156 @@
...
+
+/// Allows for easy scaling of the given value based on the size of the view its contained in (usually the full height)
+/// This is useful when you need to position and/or size a view consistently across device sizes
+///
+///
+/// Basic usage is: @ScalingValue(with: .width) var scaledSize = 0.5
+/// You can also specify a custom arithmetic operator (+, -, /, *) *to use in the calculations, the default is *
+///
+/// See the ExampleView in the PreviewProvider below for usage
+@propertyWrapper public struct ScalingValue<Value>: DynamicProperty where Value: BinaryFloatingPoint {
```
emilylaguna: I found myself using a lot of GeometryReader's and manual calculations when needing to position absolute views which in my opinion cluttered up the code a bit so I came up with a simpler way to have these done automagically for me. 

This also doesn't add any additional onAppears or Action hacks, so all the values are set at once and the view is only rendered once.

podcasts/Onboarding/Plus/PlusLandingView.swift
```diff
@@ -0,0 +1,243 @@
+import SwiftUI
+import PocketCastsUtils
+
+struct PlusLandingView: View {
+    // TODO: Remove this
```
leandroalonso, line 5: This is for a future task?

podcasts/SwiftUI/ScalingValue.swift
leandroalonso: I wonder if `ProportionalValue` might be a better name? Because it's always proportional to the screen size, right? Although it is quite similar to `ScalingValue` so not sure.

podcasts/SwiftUI/ScalingValue.swift
```diff
@@ -0,0 +1,156 @@
...
+    // The environment variable we base the calculations off of
+    @Environment(\.scalingValueFrame) var viewFrame
+
+    // A helper type to pass in a math operator: +, -, *, /, etc
+    public typealias ArithmeticOperator = (Double, Double)->Double
```
leandroalonso:
```suggestion
    public typealias ArithmeticOperator = (Double, Double) -> Double
```

leandroalonso (approved):
Tested on iPhone 14 (iOS 16), iPhone 8 (iOS 14), and iPad, and all look good.

podcasts/Onboarding/Plus/PlusLandingView.swift
emilylaguna: Yeah, I am going to add the functionality to the view and replace this with something better in a future PR.

podcasts/SwiftUI/ScalingValue.swift
emilylaguna: Done: fc296093e93e0b5f08daa5f3ffa1ef66d201428d

podcasts/SwiftUI/ScalingValue.swift
emilylaguna: Ah! That is the word I was looking for but got stuck with "Scaling". Thank you. Updated here: fc296093e93e0b5f08daa5f3ffa1ef66d201428d

Issue #500, opened 2022-11-09
Project: Onboarding Improvements


emilylaguna:
## Description
Update and improve the UI of the app onboarding process

### Setup
- [x] Add some common SwiftUI components: https://github.com/Automattic/pocket-casts-ios/pull/494

### Implement the updated designs
- [x] Login Landing View: https://github.com/Automattic/pocket-casts-ios/pull/503
    - [x] Add functionality: https://github.com/Automattic/pocket-casts-ios/pull/532
    - [x] Update the header for no sso
- [x] Plus Landing View: https://github.com/Automattic/pocket-casts-ios/pull/497
    - [x] Add functionality: https://github.com/Automattic/pocket-casts-ios/pull/506
- [x] Purchase view: https://github.com/Automattic/pocket-casts-ios/pull/506
    - [x] Add functionality: https://github.com/Automattic/pocket-casts-ios/pull/507
- [x] Post login (Welcome) view: https://github.com/Automattic/pocket-casts-ios/pull/508
    - [x] Add the newsletter: https://github.com/Automattic/pocket-casts-ios/pull/520
- [x] Plus Prompt View: https://github.com/Automattic/pocket-casts-ios/pull/518
    - [x] Add functionality: https://github.com/Automattic/pocket-casts-ios/pull/518

### Add the new onboarding flow
- [x] Update the existing flows to push to the new views: https://github.com/Automattic/pocket-casts-ios/pull/532
- [x] Present the onboarding process when launching the app for the first time: https://github.com/Automattic/pocket-casts-ios/pull/532
- [x] Migrate the existing IAP code into the new design: https://github.com/Automattic/pocket-casts-ios/pull/507

### Improve importing
- [x] Implement the view design: https://github.com/Automattic/pocket-casts-ios/pull/526
    - [x] Landing view: https://github.com/Automattic/pocket-casts-ios/pull/526
    - [x] Steps detail view: https://github.com/Automattic/pocket-casts-ios/pull/526
- [x] Create a list of steps to import from other apps: https://github.com/Automattic/pocket-casts-ios/pull/526
- [x] Add app detection to make the list only present relevant options: https://github.com/Automattic/pocket-casts-ios/pull/526
- [x] Add a way to import from Apple Podcasts: https://github.com/Automattic/pocket-casts-ios/pull/526

### Final touches
- [x] Tracks: https://github.com/Automattic/pocket-casts-ios/pull/542
- [x] iPad testing
[...]

PR #503, opened 2022-11-11
Onboarding Updates: Adds the updated Login landing view design

emilylaguna:
| 📘 Project: #500 | 
|:---:|

I apologize for the size of this PR, a lot of it is image assets that were added. 

This adds the updated login landing design with some basic navigation to the sign up / create account view:
- I also added a `EventDelegateHostingController` to allow a coordinator to react to the view controller events
    - This was needed because the the LoginCoordinator to interact with both the SwiftUI view and the hosting controller
    - I am using it to set the navigation bar style, because when I tried doing this through SwiftUI the changes didn't take effect
- I added a MotionManager class to easily interact with the device motion, I'm using this for a subtle parallax effect
- For fun I added an tap handler that makes the podcast image bounce
- I added an EnvironmentValue for `appTheme` that is setup in the `setupDefaultEnvironment`. This was done because iOS 14 ButtonStyle's doesn't get the EnvironmentObject passed to it, but it does take the custom environment value.
- I moved some assets from the single sign on feature branch into this one

## Screenshots

| iPhone SE 14.1 | iPhone 8  - 15.5 | iPhone 14 Pro Max - 16.1 |
|:---:|:---:|:---:|
||||

## To test

1. Launch the app with the onboardingUpdates Feature Flag enabled
2. Go to the Profile screen
3. Tap the Create Folder button
4. ✅ Verify the view looks good on multiple screen sizes, and themes
5. ✅ Verify the view [matches the designs](https://www.figma.com/file/PviUP0aiZctOprDuuYRwpb/Onboarding-%26-Upgrade-v2?node-id=305%3A21965)
6. Tap the Sign Up button
7. ✅ Verify you are brought to the new account view
8. Tap back
9. Tap the Log In button
10. ✅ Verify you are brought to the login view
11. Tap the Not Now button
12. ✅ Verify the view dismisses 

## Checklist


podcasts/AppTheme.swift
```diff
@@ -788,6 +820,13 @@ class AppTheme {
         ThemeColor.filter03()
     }
 
+    // MARK: - Getting Colors from ThemeStyles
+
+    /// Returns a SwiftUI color for the theme style
+    static func color(for style: ThemeStyle, theme: Theme? = nil) -> Color {
```
emilylaguna, line 826: I added this to make getting the color a bit easier.

podcasts/End of Year/Stories/Views/PodcastCover.swift
```diff
@@ -46,6 +46,35 @@ struct PodcastCover: View {
     }
 }
 
+/// Applies the podcast cover style to a static image
+struct PodcastCoverImage: View {
```
emilylaguna, line 50: I originally modified the PodcastCover struct, but this seemed like a cleaner way to accomplish the same thing.

podcasts/NewEmailViewController.swift
```diff
@@ -115,6 +114,12 @@ class NewEmailViewController: UIViewController, UITextFieldDelegate {
         Analytics.track(.createAccountShown)
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        emailField.becomeFirstResponder()
```
emilylaguna, line 120: This was causing glitches when I was pushing to it. So I moved it here.

podcasts/Onboarding/Login/LoginLandingView.swift
```diff
@@ -0,0 +1,290 @@
...
+    @EnvironmentObject var theme: Theme
+    @ObservedObject var coordinator: LoginCoordinator
+
+    var body: some View {
+        ProportionalValueFrameCalculator {
+            LoginLandingContent(coordinator: coordinator)
+        }
+    }
+
+    func configure(controller: UIViewController) {
```
emilylaguna: I know this is super hacky and I'm open to other ideas on how to accomplish this.

podcasts/Onboarding/LoginCoordinator.swift
```diff
@@ -0,0 +1,105 @@
+import Foundation
+
+class LoginCoordinator: ObservableObject {
```
emilylaguna: I originally tried to push to new views within SwiftUI itself and using a NavigationView but it was extremely buggy when pushing to UIViewController's so I moved to the coordinator.

podcasts/Onboarding/Login/LoginLandingView.swift
leandroalonso: What do you think about moving that to a VC?

This way we decouple this logic from the view.

podcasts/Onboarding/Login/LoginLandingView.swift
```diff
@@ -0,0 +1,255 @@
...
+        imageView.accessibilityLabel = L10n.accountLogin
+        controller.navigationItem.titleView = imageView
+
+        controller.navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Not Now", style: .plain, target: coordinator, action: #selector(LoginCoordinator.dismissTapped))
+    }
+}
+
+private struct LoginLandingContent: View {
+    @EnvironmentObject var theme: Theme
+    @ObservedObject var coordinator: LoginCoordinator
```
leandroalonso:
```suggestion
    var coordinator: LoginCoordinator
```

You don't need this to be an ObservedObject because it doesn't publish any changes.

podcasts/Onboarding/Login/LoginLandingView.swift
```diff
@@ -0,0 +1,255 @@
...
+                .padding([.leading, .trailing], Config.padding)
+                .padding(.top, calculatedHeight + (Config.padding * 2))
+                .padding(.bottom)
+            }
+        }
+    }
+
+    // Cover Models 📸
+    private let models: [CoverModel] = [
+        CoverModel(imageName: "login-cover-1", size: 0.26, x: -0.1973, y: 0.2606),
```
leandroalonso, line 57: Do you think it would be bad if we loaded those covert arts?

The reason I ask is that these 7 covers will probably add some MBs into the final app version for something that is not critical.

We also need to take into account that if we load them from remote they might change (not sure if that would be an issue).

podcasts/Onboarding/LoginCoordinator.swift
leandroalonso:
```suggestion
class LoginCoordinator {
```

(maybe you're planning having states published here, if that's the case just ignore my comment)

podcasts/PodcastListViewController.swift
```diff
@@ -219,10 +219,7 @@ class PodcastListViewController: PCViewController, UIGestureRecognizerDelegate,
 
     @objc private func createFolderTapped(_ sender: UIBarButtonItem) {
         if FeatureFlag.onboardingUpdates {
-            let hostingController = PCHostingController(rootView: PlusLandingView(dismissAction: {
-                self.dismiss(animated: true)
-            }))
-            present(hostingController, animated: true, completion: nil)
+            self.present(LoginCoordinator.make(), animated: true)
```
leandroalonso: A suggestion:

I think we can have 3 different parts:

1. The `LoginCoordinator` would just handle the taps and routing
2. You can create a `LoginNavigationController` in which you configure the navigation bar (removing that from the view)
3. Keep the view as just the view

leandroalonso (commented):
@emilylaguna left a few suggestions to improve the code, let me know what you think.

Also, the current "Configure account" from Profile doesn't trigger this new screen. Not sure if you're planning for a subsequent task but it's worth mentioning.

podcasts/Onboarding/Login/LoginLandingView.swift
emilylaguna: My plan is to talk to the designer about the intention of this and make a PR to change it in the future. 

Also since the login view is planned to be shown when the app first launches it would make sense to bundle them in the app, or at least lower res versions of them since right now I'm loading the 600x600 but only using the 200x200 at max. 

I'll address this in a future task.

podcasts/PodcastListViewController.swift
emilylaguna: Makes sense. Done in: [fe558bb00cfc347b962e4021030a0ca507e25e2e](https://github.com/Automattic/pocket-casts-ios/pull/503/commits/fe558bb00cfc347b962e4021030a0ca507e25e2e)

I'm planning on removing the `make` methods from the coordinators and moving the logic into a navigation manager in the future.

podcasts/Onboarding/LoginCoordinator.swift
emilylaguna: I'll need this once I do the social login logic but it's not needing right now. Removed: fe38ec0b37e48fbd089d5c78844d10507e2e2f89

podcasts/Onboarding/Login/LoginLandingView.swift
emilylaguna: Way better idea, done here: 6e4c1237f9eeea68f4c414b532c9fe092913136b

podcasts/Onboarding/Login/LoginLandingView.swift
emilylaguna: Done.

podcasts/Onboarding/LoginCoordinator.swift
emilylaguna: Removed: fe38ec0b37e48fbd089d5c78844d10507e2e2f89

leandroalonso (approved):

emilylaguna:
> Also, the current "Configure account" from Profile doesn't trigger this new screen. Not sure if you're planning for a subsequent task but it's worth mentioning.

Yeah, all the view displaying logic is placeholder for now. I'll address this in a future PR.

PR #509, opened 2022-11-12
Make custom time stepper more accessible

adincebic:
Improves the usability of time stepper through out the app and makes it way easier to navigate through screens like settings.

## To test

1. Launch the app
2. Go to profile tab
3. Tap on settings
4. Tap on general
5. Turn voiceover on if you didn't turn it on already
6. Use swipe right gesture to get to the cell with time stepper (i.e. skip forward cell in player section)
7. When you hear current skip forward value (i.e. 45 seconds) swipe right once more and you should hear voiceover pronounce "adjustable" or "slider"
8. Use swipe up or down voiceover gestures to adjust the value

This significantly improves the experience for voiceover users and makes it much easier to navigate through settings screen.

### DIFF
--- a/podcasts/CustomTimeStepper.swift
+++ b/podcasts/CustomTimeStepper.swift
@@ -36,7 +36,11 @@ class CustomTimeStepper: UIControl {
     var bigIncrements: TimeInterval = 5.minutes
     var smallIncrements: TimeInterval = 1.minute
     var smallIncrementThreshold: TimeInterval = 5.minutes
-    var currentValue: TimeInterval = 1.hour
+    var currentValue: TimeInterval = 1.hour {
+        didSet {
+            accessibilityValue = L10n.timeShorthand(Int(currentValue))
+        }
+    }
 
     override init(frame: CGRect) {
         super.init(frame: frame)
@@ -176,15 +180,9 @@ class CustomTimeStepper: UIControl {
 
     override func accessibilityIncrement() {
         performIncrementAction()
-        setAccessibilityValue()
     }
 
     override func accessibilityDecrement() {
         performDecrementAction()
-        setAccessibilityValue()
-    }
-
-    private func setAccessibilityValue() {
-        accessibilityValue = L10n.timeShorthand(Int(currentValue))
     }
 }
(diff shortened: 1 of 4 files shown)


adincebic:
@emilylaguna  could you take a look at this PR?

jgcaruso:
This tested really well, and does feel like a big improvement over the previous VoiceOver experience.

I'm not in the code owner's group on this repo, so I'll leave it to @emilylaguna for final review.

podcasts/CustomTimeStepper.swift
```diff
@@ -36,7 +36,11 @@ class CustomTimeStepper: UIControl {
     var bigIncrements: TimeInterval = 5.minutes
     var smallIncrements: TimeInterval = 1.minute
     var smallIncrementThreshold: TimeInterval = 5.minutes
-    var currentValue: TimeInterval = 1.hour
+    var currentValue: TimeInterval = 1.hour {
+        didSet {
+            accessibilityValue = L10n.timeShorthand(Int(currentValue))
```
emilylaguna: We can use the `localizedTimeDescription` to read out the full time value instead of the shorthand. 

```suggestion
            accessibilityValue = currentValue.localizedTimeDescription ?? ""
```

emilylaguna (changes requested):
👋 Sorry for the delay in my review!

I took a look at this it's working well except for 1 thing. When I swipe up to have a value that is more a minute or more, VO reads it out as "1 meter" and if it has seconds it reason "1 meter 5 ess" or "1 meter 10's"

I left a comment below on how we can fix this.

podcasts/CustomTimeStepper.swift
emilylaguna: Here's a video of the values VO is reading out:

https://user-images.githubusercontent.com/793774/203174415-6792c994-2e5d-4d49-a504-f0d0a4bc7012.mov

adincebic: Hey @emilylaguna good that you caught it. I will get on it probably in a few days. This is something that I really want to get right.

CookieyedCodes:
@adincebic want to come back to this?

adincebic:
Yes, thanks for reminding me.> @adincebic want to come back to this?

podcasts/CustomTimeStepper.swift
adincebic: Hey @emilylaguna I finally got around to fixing this. Sorry for the huge delay. Please check it out.

adincebic:
> @adincebic want to come back to this?

Finally finished it. Again thanks for the reminder.

emilylaguna (approved):
Nice work! Sounds good to me now!

PR #515, opened 2022-11-15
End of Year: exclude empty categories from the top categories

leandroalonso:
We had multiple reports of the title of the category not appearing on the top category's story. Eg.:

With the help of @emilylaguna, we discovered that some podcasts have their category set to empty or null. A lot of private feeds have this issue, but we cannot assume that all of them are private feeds.

Due to the lack of time, we'll ignore those podcasts from the calculation for now and just display data about the categories we have set.

cc @ashiagr  

## To test

1. Run `release/7.27` (or TestFlight version) and save your top categories story
2. Run this branch
3. Check your stories
4. Make sure it's the same as the one you saved

If you had any "empty" category before, it should not be shown anymore.

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -130,7 +130,7 @@ class EndOfYearDataManager {
                             SELECT COUNT(DISTINCT podcastUuid) as numberOfPodcasts,
                                 SUM(playedUpTo) as totalPlayedTime,
                                 \(DataManager.podcastTableName).*,
-                                replace(IFNULL( nullif(substr(\(DataManager.podcastTableName).podcastCategory, 0, INSTR(\(DataManager.podcastTableName).podcastCategory, char(10))), '') , \(DataManager.podcastTableName).podcastCategory), CHAR(10), '') as category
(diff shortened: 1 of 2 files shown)


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -290,9 +290,14 @@ public struct ListenedCategory {
 
     public init(numberOfPodcasts: Int, categoryTitle: String, mostListenedPodcast: Podcast, totalPlayedTime: Double) {
         self.numberOfPodcasts = numberOfPodcasts
-        self.categoryTitle = categoryTitle
         self.mostListenedPodcast = mostListenedPodcast
         self.totalPlayedTime = totalPlayedTime
+
+        if categoryTitle == "" || categoryTitle.isEmpty {
```
emilylaguna: Aren't these checks doing the same thing? Couldn't we just do `categoryTitle.isEmpty`?

emilylaguna (changes requested):
We chatted in Slack, but to follow up my database has private feeds with empty and null categories and I am still being returned an empty top category (ignore the played time)

Screen Shot 2022-11-15 at 2 37 11 PM

I think we can simplify and just group empties into an other category for now.

leandroalonso:
@emilylaguna updated the description based on what we talked about.

emilylaguna (approved):
Works and excludes the empty results in my database.

PR #523, opened 2022-11-16
End of Year: improve listening history sync

leandroalonso:
| 📘 Project: #376 |
|:---:|

This PR improves and speeds up the process of syncing the 2022 listening history. I tested with an account that was taking ~6 minutes to sync. This has been reduced to ~34s. There's still room for improvement but I'll leave that for another PR.

Before the change: sync took 296.6710090637207
After the change: sync took 34.58926796913147

 ## Improvements

The way we were doing before, we would check the database once per episode, it doesn't matter if the user had them or not. For users with a big list of episodes, the performance was very poor.

Instead, what we do now is to check the episodes that are missing and request and add only them.

## To test

1. Run the TestFlight version
2. Check your stories
3. Run this branch
4. ✅  Check your stories and make sure both versions match

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -205,7 +205,7 @@ class EndOfYearDataManager {
                             WHERE `\(DataManager.podcastTableName)`.uuid = `\(DataManager.episodeTableName)`.podcastUuid and
                                 \(listenedEpisodesThisYear)
                             GROUP BY podcastUuid
-                            ORDER BY played_episodes DESC
+                            ORDER BY totalPlayedTime DESC
                             LIMIT \(limit)
                             """
                 let resultSet = try db.executeQuery(query, values: nil)
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -263,7 +263,7 @@ class EndOfYearDataManager {
         dbQueue.inDatabase { db in
             do {
                 let query = """
-                            SELECT uuid FROM \(DataManager.episodeTableName) WHERE \(DataManager.episodeTableName).uuid IN \(DBUtils.valuesQuestionMarks(amount: uuids.count)) and
+                            SELECT DISTINCT uuid FROM \(DataManager.episodeTableName) WHERE \(DataManager.episodeTableName).uuid IN \(DBUtils.valuesQuestionMarks(amount: uuids.count)) and
                                 \(listenedEpisodesThisYear)
                             """
                 let resultSet = try db.executeQuery(query, values: uuids)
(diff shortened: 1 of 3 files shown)


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -205,7 +205,7 @@ class EndOfYearDataManager {
                             WHERE `\(DataManager.podcastTableName)`.uuid = `\(DataManager.episodeTableName)`.podcastUuid and
                                 \(listenedEpisodesThisYear)
                             GROUP BY podcastUuid
-                            ORDER BY played_episodes DESC
+                            ORDER BY totalPlayedTime DESC
```
leandroalonso, line 208: I changed that because if a person listened to, let's say, 40 episodes of a podcast that has 1-minute episodes, that would be their top podcast. Which is weird (and we already received feedback about that).

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
```diff
@@ -933,4 +933,8 @@ public extension DataManager {
     func longestEpisode() -> Episode? {
         endOfYearManager.longestEpisode(dbQueue: dbQueue)
     }
+
+    func episodesThatExists(uuids: [String]) -> [String] {
```
emilylaguna:
```suggestion
    func episodesThatExist(uuids: [String]) -> [String] {
```

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -256,6 +256,32 @@ class EndOfYearDataManager {
         return episode
     }
 
+    /// Given a list of UUIDs, return which UUIDs are present on the database
+    func episodesThatExists(dbQueue: FMDatabaseQueue, uuids: [String]) -> [String] {
```
emilylaguna:
```suggestion
    func episodesThatExist(dbQueue: FMDatabaseQueue, uuids: [String]) -> [String] {
```

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -256,6 +256,32 @@ class EndOfYearDataManager {
...
+        var episodes: [String] = []
+
+        dbQueue.inDatabase { db in
+            do {
+                let query = """
+                            SELECT uuid FROM \(DataManager.episodeTableName) WHERE \(DataManager.episodeTableName).uuid IN \(DBUtils.valuesQuestionMarks(amount: uuids.count)) and
```
emilylaguna: When I ran the query on my local DB I ended up with multiple results for the same episode (not sure how). I think we can limit the chance of that happening by doing:

```suggestion
                            SELECT DISTINCT uuid FROM \(DataManager.episodeTableName) WHERE \(DataManager.episodeTableName).uuid IN \(DBUtils.valuesQuestionMarks(amount: uuids.count)) and
```

emilylaguna (changes requested):
Left a few comments.

emilylaguna (approved):

PR #530, opened 2022-11-17
End of Year: show progress indicator when syncing listening history

leandroalonso:
| 📘 Project: #376 |
|:---:|

Adds a progress indicator when syncing history. For power users with a long history, this indicates that loading is in progress so they don't think something is broken.

https://user-images.githubusercontent.com/7040243/202543508-2ea85227-a3ad-410e-a643-9db4d38918b3.mp4

## To test

1. Run the app
2. Check your stories
3. ✅ Check that the progress indicator appears

You can force a sync by replacing line 38 in EndOfYearStoriesBuilder to `if true {`

## Checklist

### DIFF
--- a/podcasts/End of Year/Views/CircularProgressView.swift
+++ b/podcasts/End of Year/Views/CircularProgressView.swift
@@ -1,0 +1,26 @@
+import SwiftUI
+import PocketCastsServer
+
+struct CircularProgressView: View {
+    @ObservedObject private var model = SyncYearListeningProgress.shared
+
+    var body: some View {
+        ZStack {
+            Circle()
+                .stroke(
+                    Color.white.opacity(0.5),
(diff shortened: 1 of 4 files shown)


podcasts/End of Year/Views/CircularProgressView.swift
```diff
@@ -0,0 +1,26 @@
+import SwiftUI
+import PocketCastsServer
+
+struct CircularProgressView: View {
+    @ObservedObject private var model = SyncYearListeningProgress.shared
```
emilylaguna, line 5: Not blocking but in the future we should make this more generic so it can be used elsewhere.

emilylaguna (approved):
Works as intended. I had to manually inject some UUIDs to get the progress to appear

PR #540, opened 2022-11-22
Onboarding: Adds some UI improvements for small devices and accessibility

emilylaguna:
| 📘 Project: #500 |
|:---:|

- The card heights on the plus landing view are now calculated dynamically based on the content
- Fixed an issue where plus modal pill wasn't appearing
- Updated any static strings with localized strings
- Added a missing public.opml to the plist
- Reduces the login landing header height for smaller devices
- The login landing header will now only show podcasts that we have cached images for
   - This removes images "popping" in after loading
   - This removes an issue where images were appearing blank
- Sets max accessibility font sizes for the onboarding flows

### Login Landing

| iPhone SE - Default Font Size | iPhone SE - Max Font Size |
|:---:|:---:|
|||

### Plus Landing

| iPhone SE - Default Font Size | iPhone SE - Max Font Size |
|:---:|:---:|
|||

### Plus Modal
| iPhone SE - Default Font Size | iPhone SE - Max Font Size |
|:---:|:---:|
|||

### Login Flow
1. Set your font size to the max option 
   - Settings > Accessibility > Display & Text Sizes > Larger Text > Enable Larger text > Set slider to the far right
2. Launch the app with the `onboardingUpdates` FeatureFlag enabled
   - In Xcode: `FeatureFlag.swift` -> `onboardingUpdates = true`
3. Sign out if you're signed in
4. Go to the Profile Tab
5. Tap on the "Set up an account" bubble
6. ✅ Verify the images in the header appear immediately and there are no blanks
7. ✅ Verify the text is enlarged but the view is usable
8. Test using a smaller device such as the iPhone SE

### Plus Flow
1. Set your font size to the max option 
   - Settings > Accessibility > Display & Text Sizes > Larger Text > Enable Larger text > Set slider to the far right
2. Launch the app with the `onboardingUpdates` FeatureFlag enabled
   - In Xcode: `FeatureFlag.swift` -> `onboardingUpdates = true`
3. Sign into an account that does not have Plus
9. Go to the Podcasts Tab
10. Tap on the Create Folder button in the top left corner
11. ✅ Verify the feature cards all have the same height, and are equal to the card with the longest text
12. Tap the Unlock All Features button
13. ✅ Verify the "pill" at the top of the modal appears
14. ✅ Verify the modal text is enlarged but is still usable.
15. Test using a smaller device such as the iPhone SE

## Checklist

### DIFF
--- a/podcasts/ImageManager.swift
+++ b/podcasts/ImageManager.swift
@@ -140,6 +140,11 @@ class ImageManager {
         }
     }
 
+    func hasCachedImage(for uuid: String, size: PodcastThumbnailSize) -> Bool {
+        let url = podcastUrl(imageSize: size, uuid: uuid)
+        return subscribedPodcastsCache.isCached(forKey: url.absoluteString)
+    }
+
     func cachedImageFor(podcastUuid: String, size: PodcastThumbnailSize) -> UIImage? {
         let url = podcastUrl(imageSize: size, uuid: podcastUuid)
 
@@ -440,7 +445,7 @@ class ImageManager {
         }
     }
 
-    private func podcastUrl(imageSize: PodcastThumbnailSize, uuid: String) -> URL {
+    func podcastUrl(imageSize: PodcastThumbnailSize, uuid: String) -> URL {
         let sizeRequired = ImageManager.sizeFor(imageSize: imageSize)
         let closestSize = closestImageSize(sizeRequired: sizeRequired)
 
--- a/podcasts/Onboarding/Login/LoginLandingCoverImage.swift
+++ b/podcasts/Onboarding/Login/LoginLandingCoverImage.swift
@@ -16,7 +16,7 @@ struct LoginLandingCoverImage: View {
         self.podcastUuid = podcastUuid
         self.viewBackgroundStyle = viewBackgroundStyle
 
-        let cachedImage = Self.cachedImage(for: podcastUuid)
+        let cachedImage = ImageManager.sharedManager.cachedImageFor(podcastUuid: podcastUuid, size: .grid)
 
         if let cachedImage {
             self.image = cachedImage
@@ -25,36 +25,6 @@ struct LoginLandingCoverImage: View {
         } else {
             self.image = UIImage()
         }
-    }
-
-    /// Helper function to return whether or not we will have a cached image for the podcast
-    static func hasCache(for uuid: String) -> Bool {
-        if ImageManager.sharedManager.hasCachedImage(for: uuid, size: .grid) {
-            return true
-        }
-
-        let key = ImageManager.sharedManager.podcastUrl(imageSize: .grid, uuid: uuid).absoluteString
-        return ImageCache.default.isCached(forKey: key)
-    }
-
-    /// Attempt to find a cached image for the given podcast UUID
-    private static func cachedImage(for uuid: String) -> UIImage? {
-        // Check the subscribed cached
-        if let cachedImage = ImageManager.sharedManager.cachedImageFor(podcastUuid: uuid, size: .grid) {
-            return cachedImage
-        }
-
-        // The key is the URL string
-        let key = ImageManager.sharedManager.podcastUrl(imageSize: .grid, uuid: uuid).absoluteString
-
-        // Check the default cache
-        guard ImageCache.default.isCached(forKey: key) else {
-            return nil
-        }
-
-        // Get the path of the potentially cached image
-        let path = ImageCache.default.cachePath(forKey: key)
-        return UIImage(contentsOfFile: path)
     }
 
     var body: some View {
--- a/podcasts/Onboarding/Login/LoginLandingCoverImage.swift
+++ b/podcasts/Onboarding/Login/LoginLandingCoverImage.swift
@@ -1,0 +1,66 @@
+import SwiftUI
+import Kingfisher
+import PocketCastsServer
+
+struct LoginLandingCoverImage: View {
+    /// UUID of the podcast to load the cover
+    let podcastUuid: String
+
+    /// The color of the view the cover will appear on
+    /// Prevents a flickering issue on dark backgrounds
+    let viewBackgroundStyle: ThemeStyle?
+
+    private let image: UIImage
+
+    init(podcastUuid: String, viewBackgroundStyle: ThemeStyle? = nil, placeholderImage: String) {
+        self.podcastUuid = podcastUuid
+        self.viewBackgroundStyle = viewBackgroundStyle
+
+        let cachedImage = Self.cachedImage(for: podcastUuid)
+
+        if let cachedImage {
+            self.image = cachedImage
+        } else if let image = UIImage(named: placeholderImage) {
+            self.image = image
+        } else {
+            self.image = UIImage()
+        }
+    }
+
+    /// Helper function to return whether or not we will have a cached image for the podcast
+    static func hasCache(for uuid: String) -> Bool {
+        if ImageManager.sharedManager.hasCachedImage(for: uuid, size: .grid) {
+            return true
+        }
+
+        let key = ImageManager.sharedManager.podcastUrl(imageSize: .grid, uuid: uuid).absoluteString
+        return ImageCache.default.isCached(forKey: key)
+    }
+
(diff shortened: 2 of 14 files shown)


podcasts/ImageManager.swift
```diff
@@ -140,6 +140,11 @@ class ImageManager {
         }
     }
 
+    func hasCachedImage(for uuid: String, size: PodcastThumbnailSize) -> Bool {
+        let url = podcastUrl(imageSize: size, uuid: uuid)
+        return subscribedPodcastsCache.isCached(forKey: url.absoluteString)
```
emilylaguna, line 145: `subscribedPodcastsCache` is private, so this is just to interface with it.

podcasts/Onboarding/Login/LoginLandingCoverImage.swift
```diff
@@ -0,0 +1,66 @@
+import SwiftUI
+import Kingfisher
+import PocketCastsServer
+
+struct LoginLandingCoverImage: View {
```
emilylaguna, line 5: This was originally built in `PodcastCover`, but the amount of changes I wanted to make to it to support the caching was too much so I moved it into its own view.

podcasts/Onboarding/LoginCoordinator.swift
```diff
@@ -53,18 +63,14 @@ class LoginCoordinator {
...
 extension LoginCoordinator {
-    func signInWithAppleTapped() {
-
-    }
-
-    func signInWithGoogleTapped() {
-
-    }
+    func signInWithAppleTapped() { }
+    func signInWithGoogleTapped() { }
```
emilylaguna, line 73: NOOP for now.

podcasts/podcasts-Info.plist
```diff
@@ -24,6 +24,7 @@
 			<key>LSItemContentTypes</key>
 			<array>
 				<string>unofficial.opml</string>
+				<string>public.opml</string>
```
emilylaguna, line 27: This was added in a previous PR, but Xcode decided to remove it at some point. So I'm adding it back in.

podcasts/Onboarding/Login/LoginLandingCoverImage.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /// Helper function to return whether or not we will have a cached image for the podcast
+    static func hasCache(for uuid: String) -> Bool {
+        if ImageManager.sharedManager.hasCachedImage(for: uuid, size: .grid) {
+            return true
+        }
+
+        let key = ImageManager.sharedManager.podcastUrl(imageSize: .grid, uuid: uuid).absoluteString
+        return ImageCache.default.isCached(forKey: key)
```
jgcaruso: I was wondering if this should be done inside `ImageManager.sharedManager.hasCachedImage` instead?

I'm not very clear on what the `default` cache would have over the specific `subscribedPodcastsCache`, but if its useful to check here it might be useful elsewhere as well?

podcasts/Onboarding/Login/LoginLandingCoverImage.swift
```diff
@@ -0,0 +1,66 @@
...
+        let key = ImageManager.sharedManager.podcastUrl(imageSize: .grid, uuid: uuid).absoluteString
+
+        // Check the default cache
+        guard ImageCache.default.isCached(forKey: key) else {
+            return nil
+        }
+
+        // Get the path of the potentially cached image
+        let path = ImageCache.default.cachePath(forKey: key)
+        return UIImage(contentsOfFile: path)
```
jgcaruso: Similar comment as above, might this be helpful in `cachedImageFor` as well (or to the other image cache checks elsewhere in the codebase?

jgcaruso (commented):
This tested well in general, but I found a few quirks on different devices:

1. On iPhone Mini, there are a few cases of text/buttons being too close to the bottom of the screen

| login too close to home bar | "Not Now" cut off, view is scrollable but feels like if some spacing is adjusted it doesn't need to be |
|--|--|
|  | 12-mini-not-now-too-close-bottom-corners-cut |

2. When I was signing in for the first time, the Plus sheet slid in after I entered my email address but before I entered my password. This only happened once on first run in an iPhone 14 simulator, I wasn't able to reproduce it again in any other simulator tests
3. In iOS 14.1 simulators, when tapping `Unlock All Features` (or the `Upgrade to Plus` in the Profile screen), I got an "Unable to load, check your internet connection" Alert, even though other network activities were working fine. This only happened on iOS 14.1 simulators, not iOS 15 or 16 simulators.
4. On iPhone SE 2nd generation, in the Plus features sheet, when switching font sizes from small to large while the app was already on the Plus features sheet, the fonts changed size but something with the view wasn't recalculated and I was unable to scroll to get access to the buttons which were now offscreen (this is maybe not a very likely thing for a user to do, and force quit -> reopen app builds the view correctly again)

podcasts/Onboarding/Login/LoginLandingCoverImage.swift
emilylaguna: > I was wondering if this should be done inside ImageManager.sharedManager.hasCachedImage instead?

> I'm not very clear on what the default cache would have over the specific subscribedPodcastsCache, but if its useful to check here it might be useful elsewhere as well?

Good question, the `default` cache references is actually a holdover from when I had the cover image load any missing images from the network. I have since removed that ability but the code still exists. 

I removed that code here: [822ed7b](https://github.com/Automattic/pocket-casts-ios/pull/540/commits/822ed7b6762a77a7fa3ad9f1d9b65262e3de8624)

podcasts/Onboarding/Login/LoginLandingCoverImage.swift
emilylaguna: Replied here: https://github.com/Automattic/pocket-casts-ios/pull/540/files/c23f58951a89e117bcd6441f4c5467a808caa7d9#r1033746306

tl;dr I removed this code.

emilylaguna:
> On iPhone Mini, there are a few cases of text/buttons being too close to the bottom of the screen

I fixed the login here: e43144900eef60a483dbc482f9cde082f184cc85

> "Not Now" cut off, view is scrollable but feels like if some spacing is adjusted it doesn't need to be

When it comes to the accessibility sizes the fallback handling for any extra spacing needed is to scroll the view. I don't think it's necessary to also try and reduce spacing in these instances since the number of users it will effect is very low.

emilylaguna:
> When I was signing in for the first time, the Plus sheet slid in after I entered my email address but before I entered my password. This only happened once on first run in an iPhone 14 simulator, I wasn't able to reproduce it again in any other simulator tests

That's a weird one. I haven't been able to reproduce that yet. Once I can repro I can fix it in a future PR.

> In iOS 14.1 simulators, when tapping Unlock All Features (or the Upgrade to Plus in the Profile screen), I got an "Unable to load, check your internet connection" Alert, even though other network activities were working fine. This only happened on iOS 14.1 simulators, not iOS 15 or 16 simulators.

There is a fun bug with the iOS simulators in general that for some reason causes the app to not connect to the internet at all. Restarting them usually fixes this, and I suspect it has something to do with the PAC since I random errors for it when this happens:

```
Received XPC error Connection interrupted for message type 3 kCFNetworkAgentXPCMessageTypePACQuery
Received XPC error Connection invalid for message type 3 kCFNetworkAgentXPCMessageTypePACQuery
```

> On iPhone SE 2nd generation, in the Plus features sheet, when switching font sizes from small to large while the app was already on the Plus features sheet, the fonts changed size but something with the view wasn't recalculated and I was unable to scroll to get access to the buttons which were now offscreen (this is maybe not a very likely thing for a user to do, and force quit -> reopen app builds the view correctly again)

I'm not too worried about this case, as you said normal users probably won't experience this and I'm worried adding additional handling for this may cause other issues. I think if this gets reported in the future we can address it then.

emilylaguna:
@jgcaruso another look please?

jgcaruso (approved):
Changes look good and test well:

- `Login` button on iphone mini no longer touches the bottom home bar
- podcast images still appear as expected (no flicker or blanks) after refactor

PR #550, opened 2022-11-29
Onboarding: Fixes some image assets

emilylaguna:
| 📘 Project: #500 |
|:---:|

On the TestFlight build the plus confetti and the selected icon are not being rendered correctly. I suspect this has something to do with the SVG so I've swapped those assets to PDFs to see if that helps. 

I've also disabled the keyboard detection for the modal because it was causing a weird issue if the user needed to enter their password during the checkout process. 

## To test

- The image asset change isn't really testable until it's released on TestFlight.

You can test the purchase issue on a real device without the storekit config enabled:
1. Launch the app
2. Sign into an account without plus
3. Tap the Podcasts Tab > Create Folder button
4. Tap Unlock All
5. Tap Subscribe
6. When asked for your password
7. Swipe horizontally on the empty space at the bottom of the view (as if you were going to switch apps)
8. The keyboard should hide and the password prompt should animate down a bit but not fully
9. Tap the X and verify the modal appears the same.

## Checklist


jgcaruso (approved):
Testing on simulator: Plus icon confetti continues to work as expected

Testing on real device: The modal now scrolls within itself instead of being pushed up when the keyboard appears for user login to complete the purchase.

PR #553, opened 2022-11-29
Support hidden chapters

maarut:
Fixes #

Added support for hidden chapters (hidden chapters don't show up in the Table Of Contents).
When scrubbing through the timeline, chapter titles, URLs and images are updated based on the position of the playhead.

## To test

I tested using the Stratechery Podcast episode dated 25th October 2022. It has multiple hidden chapters with images.

## Checklist


maarut:
[This is the first time I've submitted a PR for an open source project. Please advise if I haven't correctly followed your processes. I can close this PR and re-create a new one following the proper process if necessary]

podcasts/Chapters.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  Chapters.swift
+//  podcasts
+//
+//  Created by Maarut Chandegra on 11/11/2022.
+//  Copyright © 2022 Shifty Jelly. All rights reserved.
+//
+
```
emilylaguna: Can you remove the header from this file?

podcasts/Chapters.swift
```diff
@@ -0,0 +1,60 @@
...
+import Foundation
```
emilylaguna:
```suggestion
```

Since UIKit already includes Foundation we don't need to double the imports up

podcasts/Chapters.swift
```diff
@@ -0,0 +1,60 @@
...
+        visibleChapter()?.startTime ??
+        chapters.min(by: { $0.startTime < $1.startTime} )?.startTime ??
+        CMTime()
+    }
+
+    func duration() -> TimeInterval {
+        visibleChapter()?.duration ??
+        chapters.max(by: { $0.duration < $1.duration })?.duration ??
+        1
+    }
```
emilylaguna: We can reduce the overall code changes by making these all be read-only computed properties.

podcasts/Chapters.swift
```diff
@@ -0,0 +1,60 @@
...
+//  Copyright © 2022 Shifty Jelly. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+class Chapters: Equatable {
+    private var chapters = [ChapterInfo]()
+
+    init(chapters: [ChapterInfo]) {
```
emilylaguna: I see elsewhere in the code you're init'ing with an empty array. I think we can clean up the code by adding a default value, that way you can just do `Chapters()`

```suggestion
    init(chapters: [ChapterInfo] = []) {
```

emilylaguna (changes requested):
👋 Hi @maarut, Thank you so much for your contribution! We are super thrilled that you chose to contribute to Pocket Casts for your **first ever** OSS project! 

Since this PR has a lot of changes we will need to do some extra testing on our end so our reviews may be a bit delayed. However I did a quick review and left some comments below. 

In addition to the comments below, I think these changes could benefit from adding some unit tests to make sure everything works as intended.

Let me know what you think.

podcasts/Chapters.swift
maarut: I've removed all copyright notices.

podcasts/Chapters.swift
maarut: I've removed superfluous imports

podcasts/Chapters.swift
maarut: I've reworked this code to be computed properties.

podcasts/Chapters.swift
maarut: Default value defined on init.

maarut:
@emilylaguna Thanks for providing review comments and suggestions for improvements. Sorry for the delay, but I think I have addressed all outstanding comments. I hope this gets us a step closer to merging the PR.

podcasts/ChapterManager.swift
```diff
@@ -12,7 +12,7 @@ class ChapterManager {
 
     private var lastEpisodeUuid = ""
 
-    var currentChapters = Chapters(chapters: [])
+    var currentChapters: Chapters = Chapters()
```
emilylaguna: We can remove the Chapters type here since it's inferred from the `Chapters()` init.

```suggestion
    var currentChapters = Chapters()
```

podcasts/Chapters.swift
```diff
@@ -1,60 +1,83 @@
...
+        }
+        self.chapters = chapters
+    }
+
     static func == (lhs: Chapters, rhs: Chapters) -> Bool {
         lhs.chapters.elementsEqual(rhs.chapters)
     }
 }
+
+private func chaptersOverlap(_ chapters: [ChapterInfo]) -> Bool {
```
emilylaguna: While this does work, the preferred way to accomplish this would be to use a private extension like:

```suggestion
private extension Chapters {
    func chaptersOverlap(_ chapters: [ChapterInfo]) -> Bool {
```

podcasts/Chapters.swift
```diff
@@ -1,60 +1,83 @@
...
+    }
+#endif
+
+    init() {
+        self.chapters = []
+    }
+
+    init?(chapters: [ChapterInfo]) {
+        if !chaptersOverlap(chapters) {
+            return nil
```
emilylaguna: Maybe a dumb question, but could you explain the purpose of doing the chapters overlap check and why we want to return nil if they don't overlap?

podcasts/Chapters.swift
```diff
@@ -1,60 +1,83 @@
...
+    }
+
     static func == (lhs: Chapters, rhs: Chapters) -> Bool {
         lhs.chapters.elementsEqual(rhs.chapters)
     }
 }
+
+private func chaptersOverlap(_ chapters: [ChapterInfo]) -> Bool {
+    let ranges = chapters.compactMap { $0.duration > 0 ? $0.startTime.seconds ... ($0.startTime.seconds + $0.duration) : nil }
+    var ranegsOverlap = true
```
emilylaguna: Small typo here. 

```suggestion
    var rangesOverlap = true
```

podcasts/Chapters.swift
```diff
@@ -1,60 +1,83 @@
...
+#endif
+
+    init() {
+        self.chapters = []
+    }
+
+    init?(chapters: [ChapterInfo]) {
+        if !chaptersOverlap(chapters) {
+            return nil
+        }
```
emilylaguna: Since the `chapters` property already sets itself to an empty array we can remove the need for an optional and just only set the property if the overlap check succeeds like so:

```suggestion
    init(chapters: [ChapterInfo] = []) {
        if chaptersOverlap(chapters) {
            self.chapters = chapters
        }
    }
```

emilylaguna (changes requested):
Hey! Thanks for making the edits, the code is shaping up nicely. 

I left a few more comments below, along with some questions. 

Please let me know what you think.

podcasts/Chapters.swift
maarut: The purpose of the class is to represent the chapters for an audio file at a given playback position. Given this purpose, if the chapters don't overlap over that given playback position, then the query used to extract the chapters isn't valid for the use of this class, and there is a potential for returning incorrect data to users of the class. I think a reasonable case at this point is therefore to return nil.

Given your later comment regarding an initialiser returning nil, I have changed my opinion on returning nil. A Chapters object created with no chapters can operate correctly by users of the class. They can simply check if there are a non-zero number of chapters contained in the object prior to querying it for title, image, url, etc.

podcasts/ChapterManager.swift
maarut: Removed type specification

podcasts/Chapters.swift
maarut: Moved code in to a private class extension as suggested.

podcasts/Chapters.swift
maarut: Typo corrected.

podcasts/Chapters.swift
maarut: Adopted suggested code.

maarut:
Sorry for the delay in submitting a summary comment, but I have addressed the simple code related improvements you suggested, and responded to your query around the Chapters class.

CookieyedCodes:
@emilylaguna should this be merged now?

emilylaguna (approved):
Hey @maarut! I apologize, I didn't see your reply and this slipped off my radar. 

I've retested and all looks good. 

I have [made a PR](https://github.com/Automattic/pocket-casts-ios/pull/839) that merges trunk into the PR and moves the changes in the CHANGELOG to the new version, once that is 🟢 on the CI I'll merge it in. 

Thank you for your contribution and patience.

maarut:
Thanks for accepting my contribution! No worries about the delay.

leandroalonso:
@maarut JFYI, we're reverting the changes made here for the next release: https://github.com/Automattic/pocket-casts-ios/pull/883

While there isn't a problem per se with the code, we had a feature regression and there are a few reports coming in about the missing chapter information — that used to be available.

What I think we can do is reopen this PR making sure that:

1. Episodes that show chapters still show them even after this change;
2. The "Chapters" tab shows the chapter progress
3. The full-player controller has the UI for navigating through chapters

I am more than happy to test and review these changes if you create a PR. :)

PR #557, opened 2022-11-30
End Of Year: Implement Design Notes Feedback for the Intro Story

emilylaguna:
| 📘 Project: #376 |
|:---:|

#### Text Changes:
- Increases line height to 2.5 to better match the design
- Replaces typewriter apostrophes `'` with typographic ones `’`
- Replaces `...` with `!`
- Inserts a newline between `...your year` and `of ...` to prevent widows

#### UI Changes
- Moves the 2022 logo down, it is now 10% from the top of the view
- Reduces the logo bottom padding from 40 to 24
   - This applies to all stories
- Increased share button leading/trailing padding to 20 from 5
    - This applies to all stories, not seen in below screenshots

#### Code changes
- Created a new common view for the logo and applied it to all stories

### Design vs Implementation Comparison
<p align="center"></p>

### Devices
| iPhone SE - 14.1 | iPhone 13 | iPhone 14 Pro Max |
|:---:|:---:|:---:|
||||

## To test

1. Launch the app
2. Go to the Profile Tab
3. Tap on the End of Year Card
4. ✅ Verify the Intro Story matches the designs (See comparison above)
7. Tap through to the other stories
8. ✅ Verify the Logo appears in a consistent location on each of them (except for the final view, will be updated later)
5. Launch the app on various device sizes
6. ✅ Verify the view still looks good on all of them

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #558, opened 2022-12-01
End of Year: Implement design notes for the listening time

emilylaguna:
| 📘 Project: #376 | 🛫 Depends on: #557 |
|:---:|:---:|

This implements the following notes:
- Always show the full time unit string
   - This changes `1 hour 2 min` to `1 hour 2 minutes`
- Adjusts the font sizes / weights to match the designs
- Adjusts the line heights
- Adjusts the spacing between the diagonal stack and the labels to be inline with the designs
- I also tried adjusting the centering of the stack for larger devices, but it wasn't working on so I skipped it. 

<p align="center"></p>

### Devices
|:---:|:---:|:---:|
||||

## To test

1. Launch the app
2. Tap on the Profile Tab > End of Year Card
3. Tap to the second screen
4. ✅ Verify the view looks like the designs and looks good on all screen sizes

## Checklist


Modules/Utils/Sources/PocketCastsUtils/Extensions/Double+formatTime.swift
```diff
@@ -5,6 +5,17 @@ extension Double {
     ///
     /// Eg.: 5400 will return "1 hour 30 min"
     public var localizedTimeDescription: String? {
+        calculateDescription(useFullUnits: false)
+    }
+
+    /// Returns a localized time description using the full units for all description
+    ///
+    /// Eg.: 88479 will return "1 day 34 minutes 29 seconds"
+    public var localizedTimeDescriptionFullUnits: String? {
```
emilylaguna, line 14: This code is actually reverted in the next PR.

leandroalonso (commented):
Tested and it works as described. There are two issue though, one I think we can fix in another PR the other might be better to tackle here:

### Shadow

The podcast shadow is super intense now, almost feels like a border. If you look at the designs and how it was before it's more subtle:



(The same happens on top 5 podcasts)

### Login and EoY prompt

This one can be tackled in another PR, but when logging in within the new flow the EoY modal is not shown (because it relied on the login always starting on the Profile screen).

leandroalonso (approved):
Given the shadow is fixed on a subsequent PR I'm approving this one.

emilylaguna:
> This one can be tackled in another PR, but when logging in within the new flow the EoY modal is not shown (because it relied on the login always starting on the Profile screen).

I have a fix for this, will make a new PR for it.

PR #559, opened 2022-12-01
End of Year: Remove orphaned/widowed text from all stories

emilylaguna:
| 📘 Project: #376 | 🛫 Depends on: #558 |
|:---:|:---:|

> **Warning** Since I made changes to all the story labels not all the screens are currently rendering correctly on smaller devices (some text is too low, or way too small), which I will be fixing in a future PR.

### Story label text changes
This implements the design feedback to adjust remove text that is appearing "widowed" which means the last line is a single word. 

I originally approached this by injecting new lines into the text widowed text itself, but this doesn't work for various screen sizes, so I ultimately landed on preprocessing the text and replacing the space between the last 2 words to a non-breaking space (`&nbsp;`) which looks like:

<table>
<thead>
<tr>
<th align="center" width="300">Text with a widow</th>
<th align="center"  width="300">Same text, with &amp;nbsp; as the last space</th>
</tr>
</thead>
<tr>
<td align="center" width="300">Let's take a look at some of your favorites...</td>
<td align="center"  width="300">Let's take a look at some of your&nbsp;favorites...</td>
</tr>
</table>

☝️ **Fun Fact**: This is a real example using `&nbsp;` in HTML

In addition to injecting nbsp's, I am also:
- Replacing the `'` with `ʼ` so we don't need to adjust the translated text for this
- Injecting an `&nbsp;` between Pocket and Casts to ensure it doesn't get broken up
   - This is already done elsewhere in the EOY, but now it's done here instead

I've also updated all existing Text labels to use StoryLabel, and moved common formatting into it. 

### Time Labels
While fixing orphans I also noticed that in various places in the labels the formatted time (ie: 1 minute 10 seconds) was also being split up which resulted in odd looking labels, so I created a way to also inject non-breaking spaces between the units to ensure they aren't broken up:

<table>
<thead>
<tr>
<th align="center" width="300">Original</th>
<th align="center"  width="300">Same text, with &amp;nbsp; between the time components</th>
</tr>
</thead>
<tr>
<td align="center" width="300">You listened to 5 episodes for a total of 2 hours 15 minutes</td>
<td align="center"  width="300">You listened to 5 episodes for a total of 2&nbsp;hours&nbsp;15&nbsp;minutes</td>
</tr>
</table>

However this doesn't work for the category pillars and since we started using full units they too where also being split up incorrectly. So, I split the components up using a new line, and include a non-breaking space between the value and unit:

<table>
<thead>
<tr>
<th align="center">Split text</th>
<th align="center">Fixed text</th>
</tr>
</thead>
<tr>
<td align="center" width="90">2 hours 15 minutes</td>
<td align="center" width="90">2&nbsp;hours<br/>15&nbsp;minutes</td>
</tr>
</table>

### Conclusion
Making these changes ensures that all the text regardless of screen size is rendered the same and matches the typographic requirements from the design notes.

## To test

1. Launch the app
2. Tap on the Profile tab
3. Tap on the End Of Year card
4. Tap through the stories
5. ✅ Verify that each label on the story includes no widowed text
6. Repeat the steps on various device sizes

## Checklist


Modules/Utils/Sources/PocketCastsUtils/Extensions/Double+formatTime.swift
```diff
@@ -31,11 +20,11 @@ extension Double {
         }
 
         let secondsForDisplay = hours < 1 ? secs : 0
-        if let minsSeconds = formatMinsSeconds(mins: mins, secs: secondsForDisplay, unitStyle: useFullUnits ? .full : .short) {
+        if let minsSeconds = formatMinsSeconds(mins: mins, secs: secondsForDisplay) {
```
emilylaguna, line 23: I reverted the changes from here since I moved all this logic into a custom formatter

podcasts/End of Year/Stories/Views/CommonStoryViews.swift
```diff
@@ -1,16 +1,166 @@
...
+// MARK: - Story time formatter
+extension Double {
+    var storyTimeDescription: String {
+        // Prevent the time from being split across paragraphs by replacing the spaces with non breaking ones
+        calculateStoryTimeDescription(unitSeparator: "\u{00a0}", componentSeparator: "\u{00a0}")
+    }
+
+    /// Return normal text when displaying the time for sharing
+    var storyTimeDescriptionForSharing: String {
+        calculateStoryTimeDescription()
```
emilylaguna, line 91: This is to make sure we have don't break any sharing services text

leandroalonso (approved):

PR #560, opened 2022-12-02
End of Year: Add highlighting of key values

emilylaguna:
| 📘 Project: #376 | 🛫 Depends on: #559 |
|:---:|:---:|

This addresses the feedback of making the numbers and other key values appear more prominently within the story. This adds a way to highlight those values within the text. 

The implementation uses AttributedString which is only available on iOS 15+ so I included a fallback for iOS 14 devices to just display as it did before. 

<p align="center"></p>

### Devices
|:---:|:---:|:---:|
||||

## To test

1. Run the app before applying this PR, and take screenshots of each story to compare against
2. Launch the app after applying this PR
3. Tap on the Profile > End of Year Card
4. Tap through the stories and ✅ verify the key values appear bold and the rest of the text appears at a regular weight
5. ✅ Verify the wording appears the same for most stories (there are a couple that have updated wording from [the designs](https://www.figma.com/file/lH66LwxxgG8btQ8NrM0ldx/End-of-Year-2022?node-id=958%3A23682&t=JZB2eKQAaJfdoIvP-4))
6. ✅ Verify the highlighted values appear according to [the designs](https://www.figma.com/file/lH66LwxxgG8btQ8NrM0ldx/End-of-Year-2022?node-id=958%3A23682&t=JZB2eKQAaJfdoIvP-4)]
7. Repeat the steps on various device sizes
8. Test on iOS 14 and ✅ Verify the labels all appear correctly and have a bold weight

## Checklist

### DIFF
--- a/podcasts/en.lproj/Localizable.strings
+++ b/podcasts/en.lproj/Localizable.strings
@@ -3361,7 +3361,7 @@
 "eoy_story_listened_to_podcast_text" = "%1$@ different podcasts";
 
 /* String telling the user how many episodes they listened to this year, %1$@ is a placeholder for the number number of episodes. */
-"eoy_story_listened_to_episode_text" = "%1$@ episodes";
+"eoy_story_listened_to_episodes_text" = "%1$@ episodes";
 
 /* Subtitle for the story containing number of podcasts and episodes played this year. Segway for the next story. */
 "eoy_story_listened_to_numbers_subtitle" = "But there was one that you kept coming back to...";
--- a/podcasts/en.lproj/Localizable.strings
+++ b/podcasts/en.lproj/Localizable.strings
@@ -3342,8 +3342,11 @@
 /* Text that appears when someone shares the listened to story to Twitter, for example. %1$@ is a placeholder for the amount of time. */
 "eoy_story_listened_to_share_text" = "I spent %1$@ listening to podcasts in 2022";
 
-/* String telling the user how much podcast categories they listened to podcasts in 2022, %1$@ is a placeholder for the number of categories. */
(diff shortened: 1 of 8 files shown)


podcasts/en.lproj/Localizable.strings
```diff
@@ -3381,6 +3390,12 @@
 /* Title for the story showing the longest episode listened for the user in the current year. %1$@ is a placeholder for the episode title and %2$@ is a placeholder for the podcast title. */
 "eoy_story_longest_episode" = "The longest episode you listened to was %1$@ from the podcast %2$@";
 
+/* Title for the story showing the longest episode listened for the user in the current year. %1$@ is a placeholder for the duration of the episode. */
```
leandroalonso, line 3393: I think adding new strings at this point will affect all the non-English languages.

While I think it's ok to have strings in English, the problem with the stories is that everything else is in their main language already, but those will appear in English, which kinda feels broken. I'm testing the PRs in pt-br and some stories now have content in English, some others in Portuguese:

Simulator Screen Shot - iPhone 14 - 2022-12-05 at 09 04 30

It's up to us to decide if we want to move ahead with this. But we won't be able to have translations in time and it feels broken.

leandroalonso (changes requested):
In Portuguese the end result for the number of podcasts/episodes listened to is broken:



I think this has to do with translations and it will probably affect all other non-English languages.

podcasts/en.lproj/Localizable.strings
emilylaguna: This makes sense. I'll add a check in a new PR that restricts the localization to english only to prevent any issues.

leandroalonso (approved):

PR #561, opened 2022-12-02
End of Year: Implements design feedback for the Top Categories story

emilylaguna:
| 📘 Project: #376 | 🛫 Depends on: #559 |
|:---:|:---:|

This addresses the following items:
- Columns are missing the dark shadow on the left
- Columns appear too low on the view
- Perspective of the column header is a bit off compared to the other perspectives used in the stories
- The column header's corners aren't rounded
- This also adjusts some fonts and weights to match the designs

### Design vs Implementation Comparison
Note: the designs don't reflect all of the feedback notes. 

<p align="center"></p>

### Devices
| iPhone SE - 14.1 | iPhone 13 | iPhone 14 Pro Max |
|:---:|:---:|:---:|
||||

## To test

1. Launch the app
2. Go to Profile Tab > End of Year Card
3. Tap through to the 4th story - Top Categories
4. ✅ Verify the view appears correctly
5. Repeat the steps on multiple device sizes
6. Also please test only showing a couple of the pillars (ie: you only have 1 or 2 top categories)
   - You can test this by making a new account then listening to a couple episodes from 1 or 2 categories (scrub ahead to the end and you'll get credit for listening to the entire episode)

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #562, opened 2022-12-02
End of Year: Implements design feedback for the listened numbers story

emilylaguna:
| 📘 Project: #374 | 🛫 Depends on: #561 |
|:---:|:---:|

This addresses the following items:
- Moves "kept" to a new line
- Updates the fonts
- Adjusts spacing between the covers
- I did not scale the covers up in size because the image assets we have became blurry.

<p align="center"></p>

### Devices
|:---:|:---:|:---:|
||||

## To test

1. Launch the app
2. Tap on Profile > End of Year card
3. Tap to the 5th screen
4. ✅ Verify it looks good and matches the designs
5. ✅ Repeat the steps of various screen sizes

## Checklist


podcasts/End of Year/Stories/Views/CommonStoryViews.swift
```diff
@@ -194,3 +195,71 @@ private extension Double {
...
+    }
+
+    func body(content: Content) -> some View {
+        content
+            .rotationEffect(Self.rotationAngle, anchor: .center)
+            .scaleEffect(Self.scale, anchor: scaleAnchor)
+    }
+}
+
+struct PodcastPerspectiveRotator<Cover: View>: View {
```
emilylaguna, line 216: This solves a problem when applying rotation effects which messes up the view's frame. This instead will calculate the proper height for it and reapply it after the rotation effect takes place.

podcasts/End of Year/Stories/ListenedNumbersStory.swift
```diff
@@ -15,62 +15,52 @@ struct ListenedNumbersStory: ShareableStory {
...
+
+                    podcastCover(3)
+                        .frame(width: geometry.size.width * 0.35, height: geometry.size.width * 0.35)
+                        .padding(.leading, -(geometry.size.width / 4))
                 }
+                .applyPodcastCoverPerspective()
+                .padding(.top, geometry.size.height * 0.10)
+
+                let podcasts = L10n.eoyStoryListenedToPodcastText(listenedNumbers.numberOfPodcasts)
+                let episodes = L10n.eoyStoryListenedToPodcastText(listenedNumbers.numberOfEpisodes)
```
leandroalonso:
```suggestion
                let episodes = L10n.eoyStoryListenedToEpisodeText(listenedNumbers.numberOfEpisodes)
```

leandroalonso (changes requested):
Tested and it looks good, but we're using the podcast text for the episodes.

podcasts/End of Year/Stories/ListenedNumbersStory.swift
emilylaguna: Ah I made a change to fix that, but I don't see it anywhere in my commits. Fixed here: d7aa9e909552b97ec5f26e04d7b7a0deb9accc42

leandroalonso (approved):

PR #567, opened 2022-12-03
End of Year: Epilogue Grand Finale Idea

emilylaguna:
| 📘 Project: #376 | 🛫 Depends on: #566 |
|:---:|:---:|

This implements a bit more pizazz on the epilogue view based on the design review notes:
- Changes background to black
- Adds confetti
- Adds a shiny effect to the heart to add a pop of interactive color
- Adds haptics when the confetti appears

## Video

https://user-images.githubusercontent.com/793774/205430061-8b2dbb6f-cfcd-4e9e-ab4d-39ed6857e6cc.MP4

## To test

1. You should test this on a real device
2. Go to Profile > End of Year Card
3. Tap through to the second to last view
4. Let the last story play out, and hold your device in your hand
5. ✅ Verify the epilogue appears with some extra flare
7. ✅ Verify the haptics play and feel cool
6. ✅ Tilt your device around for extra effects

## Checklist


podcasts.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -163,6 +163,15 @@
           "version": "7.25.1"
         }
       },
+      {
```
leandroalonso: I would really, really avoid adding a new library just for this effect (and at this point).

I already plan to remove the EoY files from the final app build once we remove the feature in January. Of course, we can also remove that but I wonder if it really worth it.

leandroalonso (commented):
@emilylaguna I left a comment about adding a library to have a specific effect.

I think the confetti is fine but I wouldn't consider these changes to have a high impact.

podcasts.xcworkspace/xcshareddata/swiftpm/Package.resolved
emilylaguna: I agree. I did this for a quick proof of concept idea.

I asked the designers for input on this, and if it's something they want to do I can update the package reference to my own version of it.

leandroalonso (approved):
Nice work on this!

PR #393, opened 2022-10-13
End of Year: add badge and card to Profile

leandroalonso:
| 📘 Project: #376 |
|:---:|

Add a badge and the End of Year card structure to the Profile screen.

| Card | Badge |
| ---- | ------- |
|  | Simulator Screen Shot - iPhone 14 - 2022-10-14 at 10 06 27 |

Design is not on the scope of this task.

### Badge

1. Enable the `endOfYear` flag in `FeatureFlag.swift`
2. Run the app
3. ✅ Check that there's a badge on the Profile icon at the tab bar
4. Tap Profile
5. ✅ Check that the badge disappears
6. Re-run the app
7. ✅ Check that the badge doesn't appear anymore

### Card

1. Enable the `endOfYear` flag in `FeatureFlag.swift`
2. Run the app
3. ✅ Check that there's a card at the top of Stats about the End of Year
4. Tap it
5. ✅ Check that the stories screen is shown

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #577, opened 2022-12-06
Onboarding: Make sure the end of year modal is triggered after a user logs in

emilylaguna:
| 📘 Project: #500 |
|:---:|

This fixes an issue where the onboarding flow doesn't trigger the viewDidAppear on the MainTabBarController which means a user who is logging in may not see the modal when the should.

## To test

1. Launch the app
2. Logout if you're logged in
3. Go to Profile > Tap the login button
4. Login to account that has listening history in 2022
5. ✅ Verify you see the end of year modal after logging in
6. ✅ Verify you see the badge under Profile
7. Tap on Profile
7. ✅ Verify the badge is gone
8. Re-run the app
9. ✅ The badge is not shown

## Checklist

### DIFF
--- a/podcasts/MainTabBarController.swift
+++ b/podcasts/MainTabBarController.swift
@@ -389,7 +389,6 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
 
     func showOnboardingFlow(flow: OnboardingFlow.Flow?) {
         let controller = OnboardingFlow.shared.begin(flow: flow ?? .initialOnboarding)
-
         guard let presentedViewController else {
             present(controller, animated: true)
             return
@@ -440,6 +439,10 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
 
         NotificationCenter.default.addObserver(forName: .userSignedIn, object: nil, queue: .main) { notification in
             self.endOfYear.resetStateIfNeeded()
+        }
+
+        NotificationCenter.default.addObserver(forName: .onboardingFlowDidDismiss, object: nil, queue: .main) { notification in
+            self.endOfYear.showPromptBasedOnState(in: self)
         }
 
         // If the requirement for EOY changes and registration is not required anymore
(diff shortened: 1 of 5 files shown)


podcasts/MainTabBarController.swift
```diff
@@ -442,6 +441,10 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
             self.endOfYear.resetStateIfNeeded()
         }
 
+        NotificationCenter.default.addObserver(forName: .onboardingFlowDidDismiss, object: nil, queue: .main) { notification in
+            self.endOfYear.showPromptBasedOnState(in: self)
```
leandroalonso, line 443: We are missing displaying the red dot on the profile button in the tab bar (ref #393).

I tried adding that here but I'm facing this problem:



For some reason `updateTabBarColor` is not working. 🤔 

This is the code I used:

```swift
            if EndOfYear.isEligible && Settings.showBadgeFor2022EndOfYear {
                self.updateTabBarColor()
                self.profileTabBarItem.badgeValue = "●"
            }
```

leandroalonso: Oh I managed to fix that. The code using your notification (we might want to extract the badge assignment cause it's the same on `viewDidLoad`:

```swift
        NotificationCenter.default.addObserver(forName: .onboardingFlowDidDismiss, object: nil, queue: .main) { notification in
            self.endOfYear.showPromptBasedOnState(in: self)

            if EndOfYear.isEligible && Settings.showBadgeFor2022EndOfYear {
                self.profileTabBarItem.badgeValue = "●"
            }
        }
```

On `updateTabBarColor` remove the `if EndOfYear.isEligible {` and it works. :)

leandroalonso (changes requested):
@emilylaguna the modal works just fine! We also need to tackle the profile badge, I managed to make it work and added the info on a comment.

leandroalonso:
@emilylaguna I just went ahead and committed the badge fix in 6d08175129437b590a1f1b3a78e2e6c73deed2f9

leandroalonso (approved):

emilylaguna:
@leandroalonso Ah good catch. I tested this and it works well.

PR #580, opened 2022-12-06
End of Year: fixes Index out of range

leandroalonso:
[There is a crash affecting 11 users](https://sentry.io/share/issue/08935f64c9e341ad805ce7faa48accfc/) when showing `ListeningTimeStory`. (You can easily reproduce that by changing `EndOfYearStoriesDataSource` to instantiate the `ListeningTimeStory` with an empty array: `return ListeningTimeStory(listeningTime: data.listeningTime, podcasts: [])`).

It's not clear to me how a user might have listening time but not any podcasts to display. Anyway, it's happening and this PR prevents that.

## To test

1. Run the app on the simulator and on your device (TestFlight or App Store version and this branch)
2. ✅ Compare the stories, they should be the same on both device and simulator

## Checklist

### DIFF
--- a/podcasts/End of Year/EndOfYearStoriesBuilder.swift
+++ b/podcasts/End of Year/EndOfYearStoriesBuilder.swift
@@ -46,9 +46,16 @@ class EndOfYearStoriesBuilder {
                 }
             }
 
+            // First, search for top 10 podcasts to use throughout different stories
+            let topPodcasts = dataManager.topPodcasts(limit: 10)
+            if !topPodcasts.isEmpty {
+                data.topPodcasts = Array(topPodcasts.prefix(5))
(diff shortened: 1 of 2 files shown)


podcasts/End of Year/EndOfYearStoriesBuilder.swift
```diff
@@ -90,10 +95,6 @@ class EndOfYearStoriesBuilder {
                 stories.append(.longestEpisode)
             }
 
-            // TODO: the color of podcasts is downloaded when needed
```
leandroalonso, line 93: This TODO is not valid anymore, removed it to prevent any confusion.

emilylaguna (approved):
Working for me, all my stories were the same.

PR #590, opened 2022-12-09
Adds a SwiftUI view to allow for highlighting of text 

emilylaguna:
This extracts the highlighting from the `StoryLabel` and adds some new features such as:
- Matching all occurrences, not just the first
- Applying custom styles to the highlight
- Supporting font, color, weight, and background color styles

## To test

1. Open HighlightedText
2. Run the SwiftUI Preview
3. Verify it works
4. Review the code

## Checklist

### DIFF
--- a/podcasts/SwiftUI/HighlightedText.swift
+++ b/podcasts/SwiftUI/HighlightedText.swift
@@ -3,12 +3,12 @@ import PocketCastsUtils
 
 /// Allows applying to styles to subtext located within a string
 struct HighlightedText: View {
-    typealias HighlightStyleBlock = (HighlightTokenizer.Token) -> HighlightStyle?
+    typealias HighlightStyleBlock = (Highlight) -> HighlightStyle?
 
-    /// Internal configuration to allow the view to "configure itself" without
-    /// needing to pass everything through the init
-    @ObservedObject private var config = Configuration()
-
+    // Internal config
+    private var highlights: Set<String> = []
+    private var font: Font = .body
+    private var highlightBlocks: [String: HighlightStyleBlock] = [:]
     private let text: String
 
     init(_ text: String) {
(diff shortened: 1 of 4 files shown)


podcasts/SwiftUI/HighlightedText.swift
```diff
@@ -0,0 +1,275 @@
...
+        } else {
+            Text(text)
+        }
+    }
+
+    // MARK: - View Configuration
+
+    /// Sets the font for the entire text
+    func font(_ font: Font) -> Self {
+        var mutableSelf = self
```
emilylaguna, line 31: This `var mutableSelf = self` to allow mutating of the structs properties looks weird, but I actually got this straight from an Apple example 😬

podcasts/SwiftUI/HighlightedText.swift
```diff
@@ -0,0 +1,275 @@
...
+        string.nonBreakingSpaces()
+    }
+
+    /// Format the text with markdown
+    private func tokenize() -> String {
+        var result = text
+
+        // Use custom start and end delimeters to prevent replacing token text
+        let startToken = "\u{001A}"
+        let endToken = "\u{001D}"
```
emilylaguna, line 126: These are non-rendering characters:
- https://www.ascii-code.com/26
- https://www.ascii-code.com/29

Since the resulting markdown text uses real words this prevents accidentally overwriting the markdown when trying to tokenize.

podcasts/SwiftUI/HighlightedText.swift
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /// Defines a style for a highlight
+    struct HighlightStyle {
+        static let defaultStyle: HighlightStyle = .init(weight: .bold)
+
+        let font: Font?
+        let weight: Font.Weight?
+        let color: Color?
+        let backgroundColor: Color?
```
emilylaguna, line 169: What we can set is limited by [`SwiftUIAttributes`](https://developer.apple.com/documentation/foundation/attributescopes/swiftuiattributes)

podcasts/SwiftUI/HighlightedText.swift
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    // MARK: - Public structs
+    struct Highlight {
+        let string: String
+        let matchNumber: Int
+    }
+
+    /// Defines a style for a highlight
+    struct HighlightStyle {
```
emilylaguna, line 163: We define a custom struct instead of allowing the style block to set the styles directly because we need to support iOS 14 and AttributedString isn't available there.

podcasts/SwiftUI/HighlightedText.swift
leandroalonso: 😅

podcasts/SwiftUI/HighlightedText.swift
```diff
@@ -0,0 +1,275 @@
...
+    private var font: Font = .body
+    private var highlightBlocks: [String: HighlightStyleBlock] = [:]
+    private let text: String
+
+    init(_ text: String) {
+        self.text = text.preventWidows()
+    }
+
+    var body: some View {
+        if #available(iOS 15, *), let attributedString {
```
leandroalonso, line 19: I think for 7.29 we can finally ditch iOS 14 support. Wdyt?

leandroalonso (approved):
**Nice** _job_!

PR #633, opened 2023-01-10
Fix distorted playback when playback speed is bigger than 1.0

### LINKED ISSUE #54: Playback: Distorted Playback or Audio Quality Issues When Listening Over Bluetooth
This is for compiling any cases related to distorted playback when listening with Bluetooth devices, in case it helps us find a common cause.

### #5248291-zen

- App Version: 7.19.3
- Device: iPhone 11
- OS: 15.5
- Bluetooth device: Bose QC35 II
- Enabled Playback Effects:
  - Trim Silence
  - Volume Boost

### #5278914-zen

- App Version: 7.19.3
- Device: iPhone13,2
- OS: 15.5
- Bluetooth device: boAt Airdopes 141
  - Enabled Playback Effects:
  - Speed > 1.0x

leandroalonso:
Fixes #54 

We've been having issues with distorted sounds when the playback speed is more than 1.0. This is not due to Bluetooth, but because of the pitch algorithm, we've been using on the default and effects player.

This PR changes the used algorithm thus solving the problem. Although it seems we don't have too many complaints about this issue, when compared to other apps this is very noticeable.

## To test

This needs to be tested on a real device. I couldn't reproduce the issue on simulator.

### Reproducing the problem

Before testing the solution, I'd recommend seeing the problem in place.

1. Run the app on your phone
2. Play a podcast and change the speed to 1,1 or bigger ([you can use this one](https://pca.st/episode/c9a7667e-c22b-4271-9397-f48f01c046a1))
3. ✅ Notice that after a few seconds the sound is a little bit distorted
4. Download the episode and test again (this will ensure effects player is used)

### Testing the fix

1. Run the app on your phone
2. Play a podcast without downloading it (so the default player is used)
3. Increase the speed to 1,1 or bigger if it's not already
4. ✅ Make sure the sound is ok and there isn't any distortion
5. Download the episode (so effects player is used)
6. Play the episode
7. ✅ Make sure the sound is ok and there isn't any distortion

## Checklist

### DIFF
--- a/podcasts/DefaultPlayer.swift
+++ b/podcasts/DefaultPlayer.swift
@@ -439,11 +439,8 @@ class DefaultPlayer: PlaybackProtocol, Hashable {
         }
 
         player?.rate = Float(requiredPlaybackRate)
-        if requiredPlaybackRate < 1.0 || requiredPlaybackRate > 1.0 {
-            player?.currentItem?.audioTimePitchAlgorithm = .timeDomain
-        } else {
-            player?.currentItem?.audioTimePitchAlgorithm = .lowQualityZeroLatency
-        }
+
+        player?.currentItem?.audioTimePitchAlgorithm = .lowQualityZeroLatency
     }
 
     private func jumpToStartingPosition() {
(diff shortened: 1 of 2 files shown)


emilylaguna (approved):
Nice work! I tested this with a custom episode and the difference is very noticeable.

PR #641, opened 2023-01-12
Update the import/export in settings to show the new import flow

emilylaguna:
## Description
This applies a quick replacement of the import steps in the settings to use the new ones from the onboarding flow. 

I am presenting the view instead of pushing because the mini player was blocking the view and the fix for that was a bit more than I wanted to do. I think this also looks nicer. 

I also didn't modify the old importexport view, I figured we could do that in the future. 

## Screenshots

| Settings | Import |
|:---:|:---:|
|||

## To test
1. Launch the app
2. Tap on the Profile tab
3. Tap on the Settings icon
4. Swipe down
5. ✅ Verify you see the new import / export section
6. Tap on Import
7. ✅ Verify you are presented with the new import flow and there is a dismiss button
8. ✅ Verify the dismiss button closes the view

## Checklist


leandroalonso (approved):
[nit]

Shall we hide the subtitle? It says "You can always do this later in settings." which seems weird, given I'm opening it from Settings.

emilylaguna:
@leandroalonso good catch, updated here: ab4d7093e8ff01ba8881ad2dbba38197a408f985

leandroalonso (approved):

PR #649, opened 2023-01-17
SSO: Add initial Google Sign In flow

leandroalonso:
| 📘 Project: #381 |
|:---:|

This PR adds the initial structure for Google Sign In. It only allows account creation and log in. Token refresh logic is not implemented yet.

The implementation is not generic and my plan is to make it a little bit more generic once we add SIWA.

## To test

Before everything:

1. Run `bundle exec fastlane run configure_apply`
3. Go to Settings > Beta Features > enable `signInWithApple`

### Account creation

1. Go to Staging console and remove the account with your Google email (if there's any)
2. Run the app
3. Profile > tap the login bubble
4. Select "Continue with Google"
5. Select an account
6. 💥 You should see "Sign in failed. Please try again"

This is not the desired outcome. However, there's a bug on the backend for now that returns a 401 when the account is created. Once this is fixed the normal account-created flow should work.

### Login

1. Run the app
2. Go to Profile and tap the login bubble
4. Select "Continue with Google"
5. Continue with the flow using the same account you used in the previous step
5. ✅ You should see the plus screen
6. Dismiss it
7. Go to Staging and gift yourself plus
8. Sign out
9. Sign In again using a Google account
10. ✅ The login flow should just be dismissed without presenting the Plus screen

## Checklist


Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
```diff
@@ -173,6 +173,28 @@ public extension ApiServerHandler {
         }
     }
 
+    func obtainToken(request: URLRequest, provider: SocialAuthProvider) async throws -> AuthenticationResponse {
```
leandroalonso, line 176: A lot of code duplication here. But I believe at this stage is better to duplicate versus to try to be smart and change the previous one.

podcasts/Login/GoogleSocialLogin.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+import PocketCastsServer
+import GoogleSignIn
+
+enum GoogleSocialLoginError: Error {
+    case emptyIdToken
+}
+
+class GoogleSocialLogin {
```
leandroalonso, line 9: This is super specific for now.

My plan when adding Apple is that we'll have a protocol and all login providers will implement the necessary methods.

podcasts/Onboarding/LoginCoordinator.swift
```diff
@@ -93,10 +104,45 @@ extension LoginCoordinator {
...
+        }
+
+        Task {
+            progressAlert = SyncLoadingAlert()
+            do {
+                // First get token
+                try await self.googleSocialLogin.getToken(from: navigationController)
+
+                // If token is returned, perform login on our servers
+                progressAlert?.showAlert(navigationController, hasProgress: false, completion: nil)
```
emilylaguna: When my request fails, instead of seeing an error I see the loading alert that says "Logged in..." and it never goes away.

podcasts/Onboarding/Login/SyncLoadingAlert.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    override func showAlert(_ presentingController: UIViewController, hasProgress: Bool, completion: (() -> Void)?) {
+        super.showAlert(presentingController, hasProgress: hasProgress, completion: completion)
+        subscribeToSyncChanges()
+    }
+
+    override func hideAlert(_ animated: Bool, completion: (() -> Void)? = nil) {
+        super.hideAlert(animated, completion: completion)
+        unSubscribeToSyncChanges()
```
emilylaguna: nit: 

```suggestion
        unsubscribeToSyncChanges()
```

podcasts/AuthenticationHelper.swift
```diff
@@ -67,7 +74,9 @@ class AuthenticationHelper {
             ServerSettings.setSyncingEmail(email: response.email)
         }
 
-        NotificationCenter.default.post(name: .userLoginDidChange, object: nil)
+        DispatchQueue.main.async {
+            NotificationCenter.default.post(name: .userLoginDidChange, object: nil)
+        }
```
emilylaguna: There's a helper extension that handles this. 
```suggestion
        NotificationCenter.postOnMainThread(notification: .userLoginDidChange)
```

emilylaguna (changes requested):
I gave this a test using my local server that includes the account creation fix and all worked well. 

I left a few comments, and 1 bug that I noticed.

podcasts/Onboarding/LoginCoordinator.swift
leandroalonso: @emilylaguna I wasn't able to reproduce this. Can you please share the steps? Which request fails? The sync one? Or the login?

emilylaguna: Heya, the `self.googleSocialLogin.login` call is failing. Try doing this:

1. Blocklist api.pocketcasts.net in Charles
2. Login using Google
3. You should see this issue

leandroalonso: @emilylaguna fixed in 36242b19bd5d2a6912b77f5e86ff2cbb534568b9

While I think this won't affect users it's good to have this in place just in case. What was happening was a race condition when presenting/dismissing the alert.

emilylaguna (approved):
Issue is fixed now.

PR #652, opened 2023-01-18
SSO: Add Apple Login

leandroalonso:
| 📘 Project: #381 |
|:---:|

* Adds Apple login
* Refactor the code so adding other providers in the future is easier

## To test

Before everything:

3. Go to Settings > Beta Features > enable `signInWithApple`

### Account creation

1. Go to Staging console and remove the account with your Apple email (if there's any)
2. Run the app
3. Profile > tap the login bubble
4. Select "Continue with Apple"
5. Select an account
6. ✅ You should see the Plus screen and the confetti screen
7. Log out

### Login

1. Run the app
2. Go to Profile and tap the login bubble
4. Select "Continue with Apple"
5. Finish the flow
5. ✅ You should see the plus screen
6. Dismiss it
7. Go to Staging and gift yourself plus
8. Sign out
9. Sign In again using an Apple account
10. ✅ The login flow should just be dismissed without presenting the Plus screen

### Token refresh

1. While logged in with an account initiated through Apple Sign In go to Settings -> Developer -> Corrupt Sync Login Token
5. Go back to Profile and tap "Refresh Now"
6. ✅ The app should refresh (you can check the console for a "Sync succeeded")

### Cancel flow

1. Run the app
2. Go to Profile and tap the login bubble
4. Select "Continue with Apple"
4. Dismiss the bottom sheet
4. ✅ Nothing should be shown
4. Repeat the same with Googlem when the webview appears, dismiss it
4. ✅ Nothing should be shown

## Checklist

### DIFF
--- a/podcasts/Onboarding/LoginCoordinator.swift
+++ b/podcasts/Onboarding/LoginCoordinator.swift
@@ -1,5 +1,4 @@
 import Foundation
-import AuthenticationServices
 import PocketCastsServer
 import SwiftUI
 import PocketCastsDataModel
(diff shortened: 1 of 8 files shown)


podcasts/Onboarding/LoginCoordinator.swift
```diff
@@ -176,6 +171,7 @@ extension LoginCoordinator: SyncSigninDelegate, CreateAccountDelegate {
     }
 
     func showError(_ error: Error) {
+        navigationController?.presentedViewController?.dismiss(animated: true)
```
emilylaguna: If I open the SSO flow and then cancel it, the error is still shown.

podcasts/Onboarding/LoginCoordinator.swift
```diff
@@ -95,37 +90,31 @@ class LoginCoordinator: NSObject, OnboardingModel {
...
                         continuation.resume()
                     }
                 }
-                let response = try await self.googleSocialLogin.login()
-                newAccountCreated = response.isNewAccount ?? false
+
+                listenToSync()
+
+                let response = try await self.socialLogin?.login()
+                newAccountCreated = response?.isNewAccount ?? false
```
emilylaguna: What about moving this at the end since the refresh won't start until the next runloop, and this will ensure that we only add the listeners once the user has actually successfully signed in. 

```suggestion
                let response = try await self.socialLogin?.login()
                newAccountCreated = response?.isNewAccount ?? false
                listenToSync()
```

emilylaguna (changes requested):
Works nicely, I was able to sign in using an existing and new account. 

I left a few comments below. 

Also, since the notification observer logic has been changed do we still need the `loginFinished` code in the LoginCoordinator?

podcasts/Onboarding/LoginCoordinator.swift
leandroalonso: I addressed this in b80f2a00105c23f619ba665effd8610c074636cd

Let me know what you think of the solution.

emilylaguna: Looks good.

emilylaguna (approved):
Nice work!

PR #656, opened 2023-01-19
SSO: add support for logins made through Apple/Google on Apple Watch

leandroalonso:
| 📘 Project: #381 |
|:---:|

* Adds support for logins made with Apple/Google on Apple Watch
* Remove a lot of code related to the previous Apple SSO implementation

## To test

1. You can test this on Simulator.
3. Do the same for Pocket Casts Watch App (if you don't see them, tap "Autocreate Schemes Now" in the Schemes screen)
3. When running, select the `Pocket Casts Watch App` scheme, not the `pocketcasts`

### Regular username/password login

1. On the iOS app, login using a username/password
2. Go to the watch app and tap "Refresh Data"
3. ✅ You should see your email on the Apple Watch app and a `Sync succeeded` on the console
4. For the Watch simulator, go to Device > Erase All Content And Settings... and erase it ¶
5. Remove the app from the iPhone simulator and close all the simulators (the Watch loses the connection with the device)

¶ This is needed because some Keychain keys are kept on the Watch app when you change your login. Fixing that is out of the scope of this PR.

### Apple login

1. On the iOS app, go to Settings > Beta Features > enable `signInWithApple`
1. Login with an Apple account on the main app
1. 2. Go to the watch app and tap "Refresh Data"
3. ✅ You should see your email on the Apple Watch app and a `Sync succeeded` on the console

### Apple login token refresh

3. Go to `SourceInterfaceController.swift` and add this as the first line of `refreshAccountTapped()`: `ServerSettings.syncingV2Token = "corrupt"`
3. Re-run the app
3. Tap "Refresh Data"
3. ✅ You should see your email on the Apple Watch app and a `Sync succeeded` on the console
4. For the Watch simulator, go to Device > Erase All Content And Settings... and erase it ¶
5. Remove the app from the iPhone simulator and close all the simulators (the Watch loses the connection with the device)

### Google login

1. On the iOS app, go to Settings > Beta Features > enable `signInWithApple`
1. Login with a Google account on the main app
1. 2. Go to the watch app and tap "Refresh Data"
3. ✅ You should see your email on the Apple Watch app and a `Sync succeeded` on the console

### Google login token refresh

3. Go to `SourceInterfaceController.swift` and add this as the first line of `refreshAccountTapped()`: `ServerSettings.syncingV2Token = "corrupt"`
3. Re-run the app
3. Tap "Refresh Data"
3. ✅ You should see your email on the Apple Watch app and a `Sync succeeded` on the console

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
@@ -20,7 +20,7 @@ public extension ApiServerHandler {
             throw APIError.UNKNOWN
         }
 
-        return try await obtainToken(request: request)
+        return try await obtainToken(request: request, usingRefreshToken: false)
     }
 
     func validateLogin(username: String, password: String, completion: @escaping (_ success: Bool, _ userId: String?, _ error: APIError?) -> Void) {
@@ -143,7 +143,7 @@ public extension ApiServerHandler {
     func obtainToken(request: URLRequest, completion: @escaping (Result<AuthenticationResponse, APIError>) -> Void) {
         Task {
             do {
-                let response = try await obtainToken(request: request)
+                let response = try await obtainToken(request: request, usingRefreshToken: false)
                 completion(.success(response))
             } catch {
                 completion(.failure((error as? APIError) ?? .UNKNOWN))
@@ -151,7 +151,7 @@ public extension ApiServerHandler {
         }
     }
 
-    func obtainToken(request: URLRequest) async throws -> AuthenticationResponse {
+    func obtainToken(request: URLRequest, usingRefreshToken: Bool) async throws -> AuthenticationResponse {
         try await withUnsafeThrowingContinuation { continuation in
             URLSession.shared.dataTask(with: request) { data, response, error in
                 guard let responseData = data, error == nil, response?.extractStatusCode() == ServerConstants.HttpConstants.ok else {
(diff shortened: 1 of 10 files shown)


Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
```diff
@@ -162,11 +162,9 @@ public extension ApiServerHandler {
...
 
                 do {
-                    // Response that contains the refresh token
                     if let response = try? Api_TokenLoginResponse(serializedData: responseData) {
                         continuation.resume(returning: AuthenticationResponse(from: response))
                     } else {
-                        // Old response for username/password without refresh token
                         let userLoginResponse = try Api_UserLoginResponse(serializedData: responseData)
                         continuation.resume(returning: AuthenticationResponse(from: userLoginResponse))
                     }
```
emilylaguna: I gave this a test and hit a really weird bug where the login token would appear on the app as the email. 

I did some digging and found that this code doesn't do what we expected it to. 

In the event that the `UserLoginResponse` is sent back, the first check for `TokenLoginResponse` always succeeds, but the values are incorrectly mapped so you end up with something like this:

Screen Shot 2023-01-23 at 3 01 36 PM

where the email value is set to the login token. 

I think we need to add flag to determine which token request type is being used so we can decode it accordingly.

emilylaguna (changes requested):

Modules/Server/Sources/PocketCastsServer/Public/API/ApiServerHandler+Account.swift
leandroalonso: Thanks for checking that @emilylaguna, changed in 988815b4bfcaa3fb0c7fe8218674c064cf879b19

emilylaguna (approved):
I gave this a test using the steps and all appears to be working. 

I wasn't able to test on my real device because the app would never install, however I did test using the simulators and that is working.

PR #660, opened 2023-01-20
Logged out users can now star episodes from Now Playing screen without being prompted to log in

### LINKED ISSUE #653: Now Playing Screen Prevents Starring Episodes for Logged Out Users
### Expected behavior
I can star an episode any where in Pocket Casts with the same effect. If I star on the episode page from the podcast page or in the Now Playing screen the results should be the same

### Actual behavior
When logged out of an account I can star an episode from the podcast episode screen, but not from the now playing screen.

1. Play an unstarred epside
2. Tap the player to open the Now Playing screen
3. Tap the star icon

https://user-images.githubusercontent.com/7907292/213487922-755f422d-4f07-4efa-8d15-446c292f0e76.MP4

##### Tested on iPhone SE, iOS 16.2, Pocket Casts 7.29
Reported 5864335-zen

rviljoen:
Fixes #653 

`performStarAction` updated the episode and always requested a server sync, regardless of the user login status. This incorrectly triggered the login flow. We now first check the user login status and pass that to the update method, similar to other parts of the code (e.g. `EpisodeDetailViewController`)

## To test

1. Log out
2. Go to fullscreen player and star en episode
3. While the star (or un-star) is performed correctly, the user is also asked to log in, which should not be the case
4. Install the update and repeat steps above. Logged out user should no longer be prompted to sign in when starring an episode on the player screen

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fixed an issue where the close button was rendered as a solid white circle when the Dark Contrast theme was active (#552)
 - Updated the import process in Settings (#641)
 - Fixed shownotes not always scrolling back to the beginning when new episode is loaded (#651)
+- Fixed an issue where logged out users were incorrectly prompted to sign in when starring episodes on the Now Playing screen (#653)
 
 7.30
 -----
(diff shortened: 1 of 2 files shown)


leandroalonso (approved):

leandroalonso (approved):
:shipit: 

Thanks for fixing that @rviljoen!

PR #661, opened 2023-01-21
Up Next Queue Syncing: Attempt to merge the local and remote device changes to prevent items from the queue being removed or cleared

### LINKED ISSUE #24: Up Next: Queue Emptied Unexpectedly
### What's happening:

Multiple reports of Up Next lists disappearing:

> my up next list just wiped out without no reason

>However recently I’ve experienced a problem with the up next queue - one day most of the episodes vanished from the queue and only about the top 20 episodes were left. 

User Reports:

#105828-zen
#106235-zen
#106088-zen

Pocket Casts (7.18), running on an iPad 6th generation and on an iPhone SE (2020) both on iOS 14.6

emilylaguna:
Part of #24

> **Note**: The majority of the line changes are actually me adding a ton of documentation and logging. 

## Description

This attempts to solve the instances where a users up next queue can be incorrectly cleared, or replaced with the contents of an outdated account. For more info on the scenarios that this could occur [see this comment](https://github.com/Automattic/pocket-casts-ios/issues/24#issuecomment-1397858597).

In order to prevent a users queue from being removed we now apply different logic based on the context ("reason") that a sync is occurring.

#### Account Creation
If we are syncing because due to a new account being created, we treat the local queue as the source of truth and it is then persisted to the server. 

#### Logging In
If we are syncing because the user is logging into an existing account we apply a non-destructive merge:
- Retrieve the latest up next queue from the server
- Add all the remote items to the users local queue
- Ensure any local episodes that were not on the server are not removed
- Persist the "merged" changes to the server

#### Everything else 
We apply the same logic as we did before. 

#### Removed episodes reappearing
Since during a login we only add items to the queue, and never remove them a user who has previously removed episodes from their queue may have them added back in if they login on a device that is not up to date. 

However, I think this is a worthy trade off to deleting or replacing users queues entirely. 

#### Sort order not persisting for logged out users

During a login the latest server changes are applied to the episodes in the queue which means any local episodes that were added may appear out of their original order. 

Again, I think this is an okay trade off. 

## Demo Videos

<details>
  <summary><h3>Installing the app from scratch and logging into an existing account that has a queue</h3>
</summary>

---
> ✅ **Expectation**: The existing users queued will be persisted on the new installation.
---

https://user-images.githubusercontent.com/793774/213831326-3cb5fa0a-98b2-4cf7-ab6a-13809f67a1a4.mov

</details>

<details>
<summary><h3>Logged out user with queued items, logging into existing account with a queue</h3></summary>

---
> ✅ **Expectation**: The logged out and existing queued items will be merged.
---

https://user-images.githubusercontent.com/793774/213831680-07b4e2b2-622c-4c71-b1ec-4d5894f4a679.mov

</details>

<details>
<summary><h3>Logged out user with queued items, logging into an account that has no queue</h3></summary>

---
> ✅ **Expectation**: The logged out queued items will appear on the account
---

https://user-images.githubusercontent.com/793774/213831935-83edc9af-9ce9-4e8d-9130-b758ba93023c.mov

</details>

<details>
<summary><h3>Logged out user with items that were removed on the other device</h3></summary>

---
> ✅ **Expectation**: The deleted items will be restored when the logged out device logs back in 
---

https://user-images.githubusercontent.com/793774/213832139-dbc25743-8ec7-495f-bbe8-7915d97927db.mov

</details>

### Account Creation

1. Reinstall the app
2. Dismiss the login prompt
3. Add some items to your up next queue
4. Create a new account
5. ✅ Verify your local queue remains the same as it was
6. Sign into the new account on another device
7. ✅ Verify the correct queue is sync'd

### Installing the app from scratch and logging into an existing account that has a queue

1. Sign into an account on device 1
2. Add some items to your up next queue and sync them
3. Reinstall the app on device 2
4. When prompted tap login
5. Sign into the device 1 account
6. ✅ Verify the correct queue is added to your device
7. Go to Device 1, and perform a sync
8. ✅ Verify the queue remains the same 

### Being a logged out user with items added to the up next, and logging into an account that has no items in the queue

1. Sign into an account on device 1
2. Clear your queue
3. Reinstall the app on device 2
4. Dismiss the login prompt
5. Add some items to the queue
7. Sign into the device 1 account
8. ✅ Verify your local queue remains the same
9. Go to Device 1, and perform a sync
10. ✅ Verify the device 2 queue is moved over

### Being a logged out user with items added to the up next, and logging into an account that has existing items in the queue

1. Sign into an account on device 1
2. Add some items to your up next queue and sync them
3. Reinstall the app on device 2
4. Dismiss the login prompt
5. Add some items to the queue
7. Sign into the device 1 account
8. ✅ Verify the queue consists of the items you added and the items from device 1
9. Go to Device 1, and perform a sync
10. ✅ Verify you see both device 1 and 2's queue

### Being a logged out user who removed items from their queue, and logging into an account where those changes weren't persisted

1. Sign into an account on device 1
2. Add some items to your up next queue and sync them
3. Reinstall the app on device 2
4. Sign into device 1's account
5. Then sign out
8. Remove some items from the queue
9. Sign back into device 1's account
10. ✅ ⚠️ The items that you removed will be added back to device 2 (See known issues above)

### Being a logged in user both devices and syncing changes
1. Sign into an account on device 1
2. Add some items to your up next queue and sync them
4. On Device 2, Sign into device 1's account
5. On Device 1, make some changes to the queue (add, remove, move, etc)
8. Sync the changes
9. On device 2, refresh
10. ✅ The queue is updated correctly
11. On device 2, make some changes to the queue. 
12. Sync them
13. On device 1, refresh
14. ✅ Verify the changes are sync'd back correctly.

## Checklist


Modules/Server/Sources/PocketCastsServer/Public/Sync/UpNextSyncTask.swift
```diff
@@ -50,34 +50,56 @@ class UpNextSyncTask: ApiBaseTask {
...
+        //
+        // However if the user is logging into an existing account and has local changes that are not fully up to
+        // sync with the latest server changes (Ie: a user who is installing the app on another device), then sending
+        // the devices changes could result in the remote data being replaced unintentionally.
+        //
+        // Instead, during a login we don't send the changes to the server which will return the latest up next queue
+        // to us, and we will attempt to merge the changes in the `applyServerChanges` call below
+        //
+        if ServerSettings.syncReason != .login {
+            // replace action
```
emilylaguna, line 68: The code block here is the same as it was before, it's just been wrapped in the if check

podcasts/PlaybackQueue.swift
```diff
@@ -126,7 +126,9 @@ class PlaybackQueue: NSObject {
 
     func persistLocalCopyAsReplace() {
         saveReplaceIfRequired()
-        startSyncTimer()
+
+        // Persist the copy on the server immediately.
+        RefreshManager.shared.syncUpNext()
```
emilylaguna: Normally this would fire a sync after a delay, however since this is only used while we're already syncing I think we should trigger it immediately.

podcasts/AuthenticationHelper.swift
```diff
@@ -56,6 +56,7 @@ class AuthenticationHelper {
         // we've signed in, set all our existing podcasts to be non synced
         DataManager.sharedManager.markAllPodcastsUnsynced()
 
+        ServerSettings.syncReason = .login
```
emilylaguna: Will need to update this for SSO new accounts / logged in accounts as well.

Modules/Server/Sources/PocketCastsServer/Public/Sync/UpNextSyncTask.swift
```diff
@@ -50,34 +50,51 @@ class UpNextSyncTask: ApiBaseTask {
...
+        // to apply those changes to the stored queue in the database. Once complete, the modified queue is saved in the
+        // database and sent back to us. The changes are then applied to the local queue by the "applyServerChanges" logic.
+        //
+        // This works well when each device keeps up to date and tracks its own changes to the queue, but can be problematic if a user
+        // signs into an existing account using a device whose queue is completely different from what is stored on the server,
+        // such as a user who installs the app, adds items to their queue, and then signs into an existing account. This
+        // can result in the queue being completely replaced by the new devices changes.
+        //
+        // To prevent this, during a login we no longer send any local changes to the server, instead we pull the latest sync'd
+        // up next queue and attempt to merge the changes later in the "applyServerChanges" call.
```
mchowning, line 66: > ...we no longer send any local changes to the server, instead we pull the latest sync'd up next queue and attempt to merge the changes

FWIW, it looks like Android already wasn't sending the local changes to the server until after it merged the queues, so this aligns well (Android's problem was that it was "merging" the server and local queues [by removing any locally queued episodes that weren't in the server queue](https://github.com/Automattic/pocket-casts-android/blob/d1502d9bf89fec3e19098f3ec78bc41e1473b447/modules/services/model/src/main/java/au/com/shiftyjelly/pocketcasts/models/db/dao/UpNextDao.kt#L182-L185) 😬 ).

Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerConstants.swift
```diff
@@ -143,4 +143,11 @@ public enum ServerConstants {
         static let maxHistoryItems = 100
         static let maxEpisodesToSync = 2000
     }
+
+
+    /// Defines a reason why a sync is being performed
+    public enum SyncingReason: String {
+        case accountCreated
+        case login
+    }
```
leandroalonso: I don't think this belongs to `ServerConstants`. This is used only by the client, right?

This induces me to think that the sync reason is also given to the server, which is not true.

I believe we can declare that outside of the `ServerConstants` and stating that this is for internal use.

leandroalonso (approved):
@emilylaguna I have a concern regarding the known issues. When we think about data loss, it indeed pays off. But what if those issues start happening much more than the current data loss issue?

I believe there's no way to know that and we'll need to test and see how it goes. In the meantime, it would be interesting to give support all this context beforehand.

If we start receiving complaints about those issues we might need to rethink this approach.

Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerConstants.swift
emilylaguna: I moved it into the SyncManager here: 9ecdec147731c790940ae0b0f13da22011559095

I figured this was a better choice since contextually it goes hand in hand, and it already has a `isFirstSyncInProgress` helper function.

emilylaguna:
> When we think about data loss, it indeed pays off. But what if those issues start happening much more than the current data loss issue?

This is a a very valid point. I tried to mitigate this by reducing the instances where the merging takes place to just a login action. However like you said, it is impossible to know before hand how many people will encounter this and if it's more than the people who were losing their data. 

Once this is merged in I'll write a P2 describing the issue, the fix, and make sure that support is also aware of this.  

> If we start receiving complaints about those issues we might need to rethink this approach.

100% agree.

PR #655, opened 2023-01-19
put podcast author in Artist "now playing" fields

jgcaruso:
Fixes #48 

This PR changes the "Now Playing" field mappings to the proper fields to ensure Siri says the correct details, and Siri Suggestions has accurate information to make meaningful suggestions (as suggested in the [MPNowPlayingInfoCenter](https://developer.apple.com/documentation/mediaplayer/mpnowplayinginfocenter) doc)

The date was put in place of the author as a convenience in the past. Quote from an old issue "gives you more relevant info in that small space". I think if we want this information in the Now Playing widget it might be worth asking Apple for a "PublishedDate" field in the api (not sure how likely this is though).

I also compared Now Playing on an android device (just visually) and date doesn't appear to be pushed to the Now Playing widget there either (at least on a OnePlus device). cc @geekygecko maybe you can confirm if it does and if this should be consistent between the 2 platforms?

| before | after |
| --- | --- |
| IMG_0015 | IMG_0014 |

_Screenshots are from iOS 15. I don't have an iOS 16 device to dev with but from seeing the Now Playing widget in iOS 16 it seems to show less info on the lock screen, so I think this change is even more relevant now since you only get `Artist` showing in the widget._

## To test

1. pull pr and deploy on a real device (simulator doesn't show now playing info)
2. start playing a podcast
3. ask siri "what is playing", it should say something like "[podcast title] by [podcast author]" instead of "... by [podcast date]"
4. if you have a car with bluetooth, pair your phone with the car and check that your car's "now playing" display info shows the podcast author instead of the date

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #668, opened 2023-01-24
Show episode date on album property so it's displayed on Carplay

### LINKED ISSUE #48: Siri: with "Hey Siri, what's playing?", Siri responds with "This is [episode title] by [day of the week published]."
When using "Hey Siri, what's playing?" with Siri, Siri answers with "This is [episode title] by [day of the week published]."

The expected response should be "This is [episode title] by [podcast author]."

### Steps to Reproduce

1. Play an episode
2. Ask Siri, "Hey Siri, what's playing?"
3. Siri will respond with "This is [episode title] by [day of week published]."

### Additional Information

- Reported in #5263803-zen

leandroalonso:
Fixes #48

This is a follow-up from #655.

In this PR I added back the episode date but as a `MPMediaItemPropertyAlbumTitle`. This guarantees that:

* The Now Playing on iOS 16 shows the correct information
* Siri responds correctly to "what's playing"
* We show episode date on Carplay

## To test

Test on a device.

2. Start playing a podcast
3. Ask siri "what is playing", it should say something like "[podcast title] by [podcast author]"
4. If you have a car with bluetooth, pair your phone with the car and check that your car's "now playing" display info shows the podcast author and on Carplay it displays the date too (you can use Simulator to test that)

## Checklist

### DIFF
--- a/podcasts/NowPlayingHelper.swift
+++ b/podcasts/NowPlayingHelper.swift
@@ -87,7 +87,11 @@ class NowPlayingHelper {
             nowPlayingInfo[MPMediaItemPropertyArtist] = safeCharacterPodcastAuthor as NSString
             nowPlayingInfo[MPMediaItemPropertyComposer] = safeCharacterPodcastAuthor as NSString
 
-            nowPlayingInfo[MPMediaItemPropertyAlbumTitle] = safeCharacterPodcastTitle as NSString
+            // we purposely show the date here instead, but as with the above there's a car stereo bug we need to work around as well where we don't show the word "Wednesday" in the artist field
+            // because on some car stereos that have embedded image databases, this comes up with a really grotesque image (more info: https://github.com/shiftyjelly/pocketcasts-ios/issues/3874)
+            let publishedDate = DateFormatHelper.sharedHelper.tinyLocalizedFormat(episode.publishedDate).replacingOccurrences(of: "Wednesday", with: "Wed", options: .caseInsensitive)
+            nowPlayingInfo[MPMediaItemPropertyAlbumTitle] = publishedDate as NSString
+
             nowPlayingInfo[MPMediaItemPropertyPodcastTitle] = safeCharacterPodcastTitle as NSString
 
             // genre


leandroalonso:
@jgcaruso @rviljoen would love to hear your thoughts about that, thanks!

rviljoen:
@leandroalonso I think this is a good compromise. The only risk I see is that it may break again down the line if Apple uses that field for something we are not considering now (e.g. adding it to Siri dictation as it did with artist). We could be playing _"Whac-A-Mole"_ forever trying to trick iOS into displaying what we want ☺️

Other alternatives (some of which we may have discussed):
1. Make the title configurable. We already add episode in brackets for some conditions, so we may as well add published date as well. The downside here is that the episode number (and date if we add it) is not immediately visible and you have to wait for it to scroll. The upside is that it would work across all platforms. Here is the section of code I am referring to:
https://github.com/Automattic/pocket-casts-ios/blob/247c7307d5ef60c66067bf5a1afc3d4725103fb2/podcasts/NowPlayingHelper.swift#L47-L58

2. The other, potentially cleaner, option is to make an explicit user setting to replace Artists with date, just like we do for Chapter:
image
In this approach, we can warn the user that setting artist to date could cause issues with Siri. Also, we could leave it off by default, which makes sense for the average user. This approach should also work nicely across all platforms.

While I'm always cautious to introduce too many settings (which creates many permutations to test down the line), I'm leaning towards option 2 in this case. It essentially restores the old functionality for power users, but behind a setting while giving the average user the correct information as Apple intended.

emilylaguna (approved):
This tested well, was able to confirm Siri reads out correctly, and that the release date appears as the album on CarPlay.

leandroalonso:
@rviljoen This approach is similar to what other apps do.

I also like the proposal of adding an option but I fear for the number of different options we might end up with. I think there's a thin line between making the app customizable versus having so many options that most users wouldn't even know what it does.

In the meantime I believe we can go ahead with this change which 1) solves the original issue while 2) maintains the date that is useful for some users.

Of course we can run into the "Whac-A-Mole" thing, but if that happens then we can reiterate this decision. And, of course, let's listen to the users and make changes accordingly.

PR #672, opened 2023-01-25
SSO: hide change password/email for SSO logins

leandroalonso:
| 📘 Project: #381 |
|:---:|

The current endpoint doesn't support changing password/email for SSO login. This PR hides this option in case the user is logged in with either Google or Apple.

## To test

4. Login with username/password
4. Go to Profile > Tap the bubble account
4. ✅ You should see change password and change email options
4. Log out
4. Login with Google
4. Go to Profile > Tap the bubble account
4. ✅ You should **not** see change password and change email options
4. Log out
4. Login with Apple
4. Go to Profile > Tap the bubble account
4. ✅ You should **not** see change password and change email options

## Checklist


podcasts/AccountViewController.swift
```diff
@@ -9,6 +9,10 @@ class AccountViewController: UIViewController, ChangeEmailDelegate {
     static let newsletterCellId = "NewsletterCellId"
     static let actionCellId = "AccountActionCellId"
 
+    private var isUsernamePasswordLogin: Bool {
```
emilylaguna, line 12: I wonder if instead of just doing a simple, we could instead save the auth provider they used to sign in so we can improve messaging if needed in the future such as: 

"Your login is managed by Apple|Google" or something.

emilylaguna (changes requested):
This tested well, the options are not shown for SSO accounts and shown for email accounts. 

I left a comment about an improvement we could make to allow us to be more contextual in the future. 

Let me know what you think.

podcasts/AccountViewController.swift
leandroalonso: I thought about that. There are a few reasons why I decided to go with this simple approach:

* I would need to persist that, which seems a little bit redundant because we already know if a login was done using username/password with what is currently persisted
* I would have to change the login flow to add this there. While this is not bad, it's more code
* I believe the near-term plan is to have at least a password reset so this code might die soon 💀 

I think we can wait until the need appears before doing something that may not be used?

emilylaguna: Sounds good. ✅

emilylaguna (approved):

PR #674, opened 2023-01-26
SSO: when logged out show the onboarding flow

leandroalonso:
| 📘 Project: #381 |
|:---:|

Changes the logged-out flow to show the new onboarding screen with Apple and Google buttons.

## To test

2. Go to `podcasts` scheme and change Build Configuration to `Debug`
3. Login (using username/password)
4. Go to `podcasts` scheme and change Build Configuration to `Staging`
4. Re-run the app
4. Go to Profile > tap "Refresh Now"
4. You should see the logged-out modal, tap "Log in"
4. Login with username/password, Google or Apple
4. ✅ The login should work successfully and the onboarding flow should dismiss

## Checklist

### DIFF
--- a/podcasts/Onboarding/LoginCoordinator.swift
+++ b/podcasts/Onboarding/LoginCoordinator.swift
@@ -143,7 +143,7 @@ extension LoginCoordinator: SyncSigninDelegate, CreateAccountDelegate {
      }
 
     func signingProcessCompleted() {
-        let shouldDismiss = OnboardingFlow.shared.currentFlow == .sonosLink || (SubscriptionHelper.hasActiveSubscription() && !presentedFromUpgrade)
+        let shouldDismiss = OnboardingFlow.shared.currentFlow.shouldDismiss || (SubscriptionHelper.hasActiveSubscription() && !presentedFromUpgrade)
 
         if shouldDismiss {
             handleDismiss()
--- a/podcasts/Onboarding/LoginCoordinator.swift
+++ b/podcasts/Onboarding/LoginCoordinator.swift
@@ -154,7 +154,7 @@ extension LoginCoordinator: SyncSigninDelegate, CreateAccountDelegate {
     }
 
     func handleAccountCreated() {
-        let shouldDismiss = OnboardingFlow.shared.currentFlow == .sonosLink
+        let shouldDismiss = OnboardingFlow.shared.currentFlow.shouldDismiss
 
         if shouldDismiss {
             handleDismiss()
--- a/podcasts/Onboarding/OnboardingFlow.swift
+++ b/podcasts/Onboarding/OnboardingFlow.swift
@@ -87,13 +87,16 @@ struct OnboardingFlow {
         /// continue with the Sonos connection process
         case sonosLink = "sonos_link"
 
+        /// When the user was logged in but was forced logged out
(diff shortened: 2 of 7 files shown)


podcasts/Onboarding/OnboardingFlow.swift
```diff
@@ -88,5 +88,16 @@ struct OnboardingFlow {
         case sonosLink = "sonos_link"
 
         var analyticsDescription: String { rawValue }
+
+        /// If after a successful sign in or sign up the onboarding flow
+        /// should be dismissed right away
+        var shouldDismiss: Bool {
+            switch self {
+            case .sonosLink, .loggedOut:
```
emilylaguna: Unfortunately reusing the `loggedOut` flow here causes the normal logged out flow to no longer prompt for Plus. You can test this by:

1. Launch the app
2. Stay signed out
3. Tap the account button on the Profile tab
4. Sign in using username/password that doesn't have plus
5. 💥 The plus prompt is no longer shown

podcasts/Onboarding/LoginCoordinator.swift
```diff
@@ -143,7 +143,7 @@ extension LoginCoordinator: SyncSigninDelegate, CreateAccountDelegate {
      }
 
     func signingProcessCompleted() {
-        let shouldDismiss = OnboardingFlow.shared.currentFlow == .sonosLink || (SubscriptionHelper.hasActiveSubscription() && !presentedFromUpgrade)
+        let shouldDismiss = OnboardingFlow.shared.currentFlow.shouldDismiss || (SubscriptionHelper.hasActiveSubscription() && !presentedFromUpgrade)
```
emilylaguna, line 146: I think we should also apply the `OnboardingFlow.shared.currentFlow.shouldDismiss` to the `handleAccountCreated` check below as well.

emilylaguna (changes requested):
Nice work. I found a bug with reusing the existing `loggedOut` flow causing the upsell prompt to no longer show when logging in via the Profile tab. 

Instead of reusing an existing flow, we should add a new one such as `loggedOutNeedsLogin` so we can:
- Add specific handling for this state
- Have better context of how people sign in via this flow.

I also left a comment about using the same logic on the account created handler as well.

podcasts/Onboarding/LoginCoordinator.swift
leandroalonso: Ops I forgot to push 😅 2c61d832c41f27c38f47d4b6b54ca01a6270a02f

podcasts/Onboarding/OnboardingFlow.swift
```diff
@@ -87,6 +87,20 @@ struct OnboardingFlow {
         /// continue with the Sonos connection process
         case sonosLink = "sonos_link"
 
+        /// When the user was logged in but was forced logged out
```
emilylaguna: Wdyt about adding some more info to the comment?
```suggestion
        /// When the user was logged out due to a server or token issue, not as a result of user interaction and is
        /// asked to sign in again. See the `BackgroundSignOutListener`
```

emilylaguna (approved):
Nice work, this is working nicely for me now. I left 1 comment about improving the documentation comment, but it's not blocking.

Issue #498, opened 2022-11-09
CarPlay: Playback Pauses on the First Attempt


thabotswana:
### Expected behavior

Playback should work as expected when you tap Play.

### Actual behavior

Playback pauses when the user taps Play. However, it will work when tapping Play the second time.

### Steps to reproduce the behavior

Unable to repro. This only happens in their Kia. It works properly in other cars. 

##### Info:

Kia Sorento 2020
App Version: 7.24.1
Device: iPhone12,1
OS: 15.6.1

Reported in 5686326-zen

erricgunawan:
A similar report from 5717520-zen:

> I’m having few troubles with podcast reproduction on my device (iPhone 14 Pro, iOS 16.1.1, PocketCast 7.26). For some reason, podcast reproduction is suddenly interrupted. Pressing play button won’t restart reproduction in most cases, even though time counter still runs, along with episode position. Player resumes after 2-3 tries. 

> It happens a lot more often when on CarPlay (to exclude car-related issues, I’ve tested and reproduced the bug on 3 different cars models / brands, using different cables to connect). Never had the bug when using Bluetooth devices. Had only one event using wired earbuds. Couldn’t test it on wireless CarPlay (not available to me).

> It was already happening before last update, but it looks slightly more frequent now.

App Version: 7.26
Device: iPhone15,2
OS: 16.1.1

I noticed a couple of these instances during the period this happens:

```
Remote control: playCommand, treating as playPause
```

Full Logs provided in the original report.

erricgunawan:
New report from 6934882-zen:

> Pocket Casts attached to my CarPlay will never start with the first tapping of the start button. It immediately stops. Only when you tap it for the second time the podcast will play. It’s been like that for a really long time, and I hoped with each update that it would be fixed. But since it is still an issue, I thought I’d contact you.

> App Version: 7.46
Device: iPhone15,2
OS: 16.6.1

Asked them to first update to 7.47 to see if that helps.

mdrockwell:
Additional details from 6934882-zd-a8c:

> It’s a Kona 1.0 T-GDI, Premium MY 19. The stereo was included in the car. Hope that helps!

They've also said that the problem persists after updating to the latest version and installing iOS 17.

erricgunawan:
1-star iOS App Review:

[...]

PR #345, opened 2022-10-03
Set the data protection to none when connecting to the database

emilylaguna:
Fixes: https://github.com/Automattic/pocket-casts-ios/issues/346

This attempts to fix the DataManager crash that happens early in the app lifecycle. This is a different crash than: https://github.com/Automattic/pocket-casts-ios/issues/34

This is a similar fix to: https://github.com/Automattic/pocket-casts-ios/pull/344

### Verify the database data is the same:
1. Launch the app 
2. Tap around to your subscribed Podcasts and verify they are all there, and you are able to interact / play them, etc
3. Verify your filters are all still there

### Reproducing the crash and verifying the fix
1. Open DataManager.swift
5. Go to Line 31 and replace `SQLITE_OPEN_FILEPROTECTION_NONE` with `SQLITE_OPEN_FILEPROTECTION_COMPLETE`
6. In Xcode, go to: Window > Devices and Simulators > Open Console
7. Click the Start Streaming action
8. In the search field in the top right corner search for `EXC_BREAKPOINT`
9. Click the Clear button
10. Run the app on your phone. 
11. Play any episode
12. Quit the app, and remove it from your background apps
13. Lock your device
14. Wait about 10 seconds, this step is required because the data protection expires after about 7 - 10 seconds
15. Cover your FaceID if you have it, to prevent unlocking the device
16. Use the play action on your AirPods
    - This forces the app to open in the background while the device is locked
17. In Console.app on your Mac
18. You should see a few new logs appear, clicking on one you should see something like:

```Aggregated. Transform: StabilityCrashNumerator3 Dirty: 1 Event: com.apple.stability.crash {"appVersion":"7.24","bundleID":"au.com.shiftyjelly.podcasts","bundleVersion":"7.24.0.0","exceptionCodes":"0x0000000000000001, 0x00000001c793df54(\n    1,\n    7643324244\n)EXC_BREAKPOINTSIGTRAP","incidentID":"EC1655FC-23B6-4D1D-ACF7-12791D370D4F","logwritten":0,"process":"podcasts","terminationReasonExceptionCode":"0x5","terminationReasonNamespace":"SIGNAL","timestamp":1664838276464015}```

19. This is the same error that appears in the crash logs
20. Replace `SQLITE_OPEN_FILEPROTECTION_COMPLETE` with `SQLITE_OPEN_FILEPROTECTION_NONE`
21. Relaunch the app and repeats the steps
22. Verify you don't see any crash logs appear in console

## Checklist

[...]

PR #675, opened 2023-01-26
Idea: Set the FileProtection.None for all files and folders Pocket Casts creates

emilylaguna:
### Description
This is a continuation of the following PRs:
- https://github.com/Automattic/pocket-casts-ios/pull/344
- https://github.com/Automattic/pocket-casts-ios/pull/345

This is an attempt to mitigate some potential audio playback issues by setting the default file protection to None instead of using the default which is `CompleteUntilFirstUserAuthentication` for any files and folders that are created by the app.

### What issues could this fix?
While this doesn't target one specific issue I hope that by changing the file protection we can reduce some of the random playback failures for downloaded or custom files that do not trigger crashes such as:
- User Report: After locking my device and pausing using my Airpods the playback will not resume and I have to unlock my phone then it will work. 
- [CarPlay: Playback Pauses on the First Attempt](https://github.com/Automattic/pocket-casts-ios/issues/498)

Streaming audio files are managed by the AVPlayer cache and I don't see a way to change the protections for them. 

### Why would this fix things?

I suspect there is a bug with the default file protection that for some reason causes the system to reject us permission to access files while the app is in the background. Despite the documentation saying that we'll have access after the user unlocks their phone for the first time. 

Because of this while the app is running in the background, the app may no longer have permissions to a file and when someone attempts to play it via Siri, Now Playing, or CarPlay the playback will fail with an error such as:

```
activating audio session succeeded
[default]          ExtAudioFile.cpp:210   about to throw -54: open audio file
[avae]            AVAEInternal.h:109   [AVAudioFile.mm:135:AVAudioFileImpl: (ExtAudioFileOpenURL((CFURLRef)fileURL, &_extAudioFile)): error -54
playbackDidFail: The operation couldn’t be completed. (com.apple.coreaudio.avfaudio error -54.)
```

We have also seen success with making this change in the above PRs and have completely removed some of our top crashes for DataManager and Tracks DB creation. 

## To test

1. In Xcode, Go to StorageManager.swift line 66
2. Change `FileProtectionType.none` to `FileProtectionType.complete`
   - This is the only way I know how to simulate this issue
6. In Xcode, go to: Window > Devices and Simulators > Open Console
7. Click the Start Streaming action
8. In the search field in the top right corner search for `process:podcasts`
10. Run the app on your phone. 
3. Run the App
4. Download any episode of a podcast
5. Start playing it
6. Pause it
7. Then quit the app
8. In the Console.app: Click the Clear button
9. Wait about 10 seconds, this step is required because the data protection expires after about 7 - 10 seconds
10. Cover your FaceID if you have it, to prevent unlocking the device
16. Use the play action on your AirPods
    - This forces the app to open in the background while the device is locked
17. In Console.app on your Mac
18. ✅ Verify you see the following error: `ExtAudioFile.cpp:210   about to throw -54: open audio file`

Screen Shot 2023-01-26 at 1 40 09 PM

18. You may need to scroll a bit to find it
20. Revert the change to StorageManager.swift, and open AppDelegate+Defaults
21. Go to line 103, and change `UpdateFileProtection` to `UpdateFileProtection1`
    - This forces the "update all" action to be fired again which will reset the protection for the downloaded files
23. Rerun the app
24. Play the same downloaded episode again
25. Pause it and quit the app
26. Wait about 10 seconds, cover your FaceID and then use the play action 
27. ✅ Verify the audio plays, and you do not see any errors in console
28. Download a new episode and repeat the steps and verify they play in the background
29. Add a new local file and repeat the steps and verify they play in the background

### Testing the existing app functions
1. Launch the app
2. Play an episode via streaming and verify it plays correctly in the foreground and background
3. Download an episode and verify it plays correctly in the foreground and the background
4. Add a local file and verify it plays correctly in the foreground and the background
5. Reinstall the app and repeat the steps to make sure it all works correctly

## Checklist

### DIFF
--- a/StorageManager.swift
+++ b/StorageManager.swift
@@ -1,0 +1,78 @@
+import Foundation
+import PocketCastsUtils
+
+struct StorageManager {
+    typealias Attributes = [FileAttributeKey: Any]
+
+    private static var fileManager: FileManager = .default
+
+    @discardableResult
+    static func moveItem(at fromURL: URL, to toURL: URL, attributes: Attributes? = nil, options: Options? = nil) throws -> Bool {
+        if let options, options.contains(.overwriteExisting) {
+            removeItem(at: toURL)
+        }
+
+        try moveItem(at: fromURL, to: toURL)
+
+        let attrs = (attributes ?? [:]).merging(Constants.defaultAttributes) { current, _ in current }
+        setAttributes(attrs, of: toURL)
+
+        #if DEBUG
+        FileLog.shared.addMessage("🟢 [StorageManager] moveItem: fromURL: \(fromURL.path) to: \(toURL.path)")
+        #endif
+        return true
+    }
+
+    @discardableResult
+    static func createDirectory(atPath path: String, withIntermediateDirectories createIntermediates: Bool, attributes: Attributes? = nil) -> Bool {
+        let attrs = (attributes ?? [:]).merging(Constants.defaultAttributes) { current, _ in current }
+
+        return tryLog(try fileManager.createDirectory(atPath: path, withIntermediateDirectories: createIntermediates, attributes: attrs), operation: "createDirectory")
+    }
+
+    @discardableResult
+    static func removeItem(at url: URL) -> Bool {
+        guard fileManager.fileExists(atPath: url.path) else {
+            return true
+        }
+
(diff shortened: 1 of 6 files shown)


leandroalonso (approved):
:shipit:

PR #676, opened 2023-01-28
Fix the next episode in the up next queue not playing when streaming on AirPlay

### LINKED ISSUE #47: AirPlay/HomePod: Up Next Queue Doesn't Continue Playing
**What the listener said:**

>Up Next won’t play next podcast in queue if connected to HomePod speakers. I can only listen to my queue if I don’t connect to speakers. Otherwise it stops after the currently playing episode.

>I’m connecting to my HomePod via Airplay. Pocketcasts is the only app that has this issue. Only the episode I actively press play will play. 

This has been happening for weeks with every podcast.

5253618-zen

emilylaguna:
Fixes #47 

## Description

This took a long time to figure out and I'm not 100% sure the fix I applied the correct one. 

So for some reason, when you're connected to AirPlay the AVPlayer will pause itself once the episode resumes. If you monitor its various statuses:
- rate (0 means it's paused)
- timeControlStatus
- reasonForWaitingToPlay

You'll see that the app sets the rate as it does normally, and the rate is confirmed by the AVPlayer, but after a second or 2 the rate gets set to `0.0` and then `timeControlStatus` gets set to `paused` and the `reasonForWaitingToPlay` is nil. Which is a bit odd. 

I tried a lot of things to fix this:
- Tried using the iOS 16 `defaultRate` value, didn't work. 
- Tried changing `actionAtItemEnd` to `none`, didn't work
- Tried changing the rate, didn't work
- and about 200 other things, heh. 

So instead I opted to try and monitor the paused state and resume if needed. I'm using the existing `shouldKeepPlaying` flag that is used elsewhere in the player to auto resume. 

In my testing this works really well. 

## To test

1. On your Mac, go to System Preferences > Sharing > Enable Airplay Receiver
   - This will allow your Mac to act as an Airplay speaker
2. On your device add many items to your up next
3. Tap the Airplay icon and select your Mac as the receiver
4. Play an episode
   - To speed things up you can skip to the last 20 seconds or so
6. Put the app in the background
13. Let the episode play out
14. ✅ Verify the playback continues as planned
15. ✅ Verify you see: `[DefaultPlayer] Detected that playback was paused while trying to play the next item. Attempting to resume playback...` in the logs to indicate that the fix was applied

Please also make sure to test:
- Normal playback
- Playback Effects
- Local Files
- Streaming to AirPods
- etc

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 7.32
 -----
-
+- Fixed an issue where the Up Next queue doesn't continue playing the next episode when connected to AirPlay (#676)
 
 7.31
 -----
--- a/Modules/Utils/Sources/PocketCastsUtils/Logging/FileLog.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Logging/FileLog.swift
@@ -45,7 +45,7 @@ public class FileLog {
         guard let message = message, message.count > 0 else { return }
 
         // if it's important enough to log to file, write it to the debug console as well
-        Self.logger.log("\(message)")
+        Self.logger.log("🦄 \(message)")
         let dateFormatter = DateFormatHelper.sharedHelper.localTimeJsonDateFormatter
         appendStringToLog("\(dateFormatter.string(from: Date())) \(message)\n")
     }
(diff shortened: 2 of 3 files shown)


leandroalonso (approved):
@emilylaguna I tested it and it works as described.

I faced another issue that seems to be unrelated to this change. In order to reproduce:

1. Device needs to be running iOS 15
1. Add at least two items to the Up Next queue
1. Play the first and lock the phone
1. Change the source to play over AirPlay
1. Skip to the last 20 seconds or so (from the lock screen)
1. Wait for the episode to end
1. The episode will end and nothing happens, triggering the next episode never happens¶

¶ It worths nothing that sometimes it works just as intended, but most of the time not. Also, if the episode is downloaded this never happen. It seems that is something related to `DefaultPlayer`.

PR #683, opened 2023-01-31
Switch to EffectsPlayer if required when disconnecting Airplay

### LINKED ISSUE #678: Trim Silence does not work properly when disconnected from AirPlay
### Expected behavior

Trim Silence is automatically enabled when the audio is played via the phone speakers or Bluetooth. 

### Actual behavior

> I frequently will connect and disconnect to my HomePod mini as I move through my apartment listening to podcasts as it’s the easiest way to hear them. I also always have trim silence on with Mad Max mode.

I know that silence trimming is not supported on the Airplay stream to the HomePod. But my main issue is that when I do disconnect from the HomePod and start playing on my phone’s speakers or a Bluetooth speaker again, silence trimming is always off *even though still shows as on.*

This means every time I disconnect from my HomePod I have to pause and restart the audio stream.

With the way that the effects player works, it seems normal for this behaviour but I wonder if this can be improved. 

- Use Airplay to play audio from a podcast with Mad Max Trim Silence enabled
- You should hear that Trim Silence does not work (which is normal since this is a technical limitation with the Apple iOS library)
- Disconnect from Airplay and start playing the audio on your phone speakers
- You should notice Trim Silence is enabled in the app, but still not active in the audio

##### Tested on [iPhone 14 Pro Max], iOS [16.2], Pocket Casts [7.31]

#5906111-zen

rviljoen:
Fixes #678 

When the audio route changes to an Airplay device, we switch to `DefaultPlayer`. When the route changes back, the same check is not performed and we keep using `DefaultPlayer`, even when the user has effects enabled.

We now always check if a player switch is required when the audio route changes.

## To test

- Use Airplay to play audio from a podcast with Mad Max Trim Silence or Volume Boost enabled (you can use your Mac as Airplay receiver)
- You should hear that the effects does not work (which is expected for Airplay, which uses `DefaultPlayer`)
- Disconnect from Airplay and start playing the audio on your phone speakers
- You should notice that the effects are enabled in the app UI, but still not active in the audio
- Apply the fix and repeat steps above. Effects will now be audible after switching back from Airplay.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 7.32
 -----
-
+- Fixed an issue where effects were not re-enabled when switching from an Airplay device back to phone output (#678)
 
 7.31
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 7.32
 -----
 - Fixed an issue where effects were not re-enabled when switching from an Airplay device back to phone output (#678)
+- Fixed an issue where the Up Next queue doesn't continue playing the next episode when connected to AirPlay (#676)
 
 7.31
 -----
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -1670,7 +1670,7 @@ class PlaybackManager: ServerPlaybackDelegate {
         guard let userInfo = notification.userInfo, let changeReason = userInfo[AVAudioSessionRouteChangeReasonKey] as? NSNumber else { return }
 
         let reason = changeReason.uintValue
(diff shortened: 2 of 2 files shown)


rviljoen (commented):
When the route changes to an Airplay device, `playingOverAirplay()` returns true and the `playerSwitchRequired()` check is performed. When switching back from Airplay to phone output, `playingOverAirplay()` returns false and hence the player check is skipped. This results in `DefaultPlayer` being used until the user pause/resumes or changes effect settings again.

leandroalonso (approved):

leandroalonso:
@rviljoen FYI: #686

rviljoen:
> @rviljoen FYI: #686

@leandroalonso Thanks, I'll have a look.

PR #688, opened 2023-02-02
Single Sign On

leandroalonso:
| 📘 Project: #381 |
|:---:|

* Adds sign-in with Google and Apple
* This is still hidden behind a feature flag

## To test

In order to merge we just need to make sure that:

1. Currently login works just fine for new logins
2. User migrating from an older version to this code doesn't see any changes
3. Token refresh works as intended

So:

1. Run `trunk` (without this branch code)
2. Login
3. Checkout this branch and run the app
5. Go to Profile > tap "Refresh Now"
6. ✅ The sync should succeed
7. Go to Profile > Settings > Developer > Corrupt Sync Login Token
4. Go to Profile > tap "Refresh Now"
5. ✅ The sync should succeed
5. Log out
5. Login again
5. ✅ You should login into your account

## Checklist


emilylaguna (approved):
I gave this a test and went through a bunch of flows:

**Account Creation:**
✅ New account using email
✅ New account using Apple
❌ There is a case where if the user signs up, deletes their account and then later tries to sign up again using the same account Apple may not include the email field which will cause the sign up to fail. 
    - https://developer.apple.com/forums/thread/121496
    - I don't think we can really handle for this, and not sure how many real users will encounter this. The fix is to reset SIWA for PC in Settings.app
✅ New account using Google
✅ Trying to create a new account using an email that already has an SSO account

**Login:**
✅ Logged in using email
✅ Logged in using Apple
✅ Logged in using Google

**Token Refreshing:**
✅ Refreshed using email
✅ Refreshed using Apple
✅ Refreshed using Google

**App Upgrade:**
✅ Upgraded the app using a logged in account
✅ Verified token refreshing works

**App Downgrade:**
✅ Downgraded the app from the SSO to the 7.31 (current beta)
   - The user stays signed in
   - If their token expires they will see the the signed out alert
   - Probably won't happen in production, but may effect beta testers

**Signed out alert:**
✅ Verified it works with an email account
✅ Verified it works with an sso account

**Migrations:**
✅ Created an account using email, then later migrated it to an SSO account
✅ Used an old version of the app with the email migrated SSO account and made sure syncing works

PR #692, opened 2023-02-06
Onboarding: fix retain cycle

leandroalonso:
| 📘 Project: #381 |
|:---:|

Fixes a retain cycle that was making `PlusLandingViewModel`, `PlusHostingViewController`, and `WelcomeViewModel` not deallocated.

## To test

1. Run a clean install of the app
2. Open Instruments > Allocations
3. Select the running instance of Pocket Casts
4. At the bottom, on the filter, add: `SyncSignInViewController` `PlusLandingViewModel` `NewEmailViewController` `PlusHostingViewController` `WelcomeViewModel` (they need to be added one by one, unfortunately)
5. Go through the whole Sign Up flow
6. After finishing it, check that no one of the filtered entities was left allocated ¶

¶ There's a SwiftUI observable that doesn't seem to go away. However, it's just a Bool so it's not a big issue and I think it's ok to leave it as it is for now.

## Checklist

### DIFF
--- a/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
+++ b/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
@@ -3,7 +3,7 @@ import PocketCastsServer
 import SwiftUI
 
 class PlusLandingViewModel: PlusPricingInfoModel, OnboardingModel {
-    var navigationController: UINavigationController? = nil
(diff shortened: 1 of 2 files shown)


emilylaguna (approved):
Nice work. This fixes the retain cycles.

PR #705, opened 2023-02-08
CarPlay: Improve the speed and reduce overall memory usage of CarPlay

### LINKED ISSUE #64: CarPlay: crash reports on iOS 15.4
Original report in Slack: p1647503453744539-slack-C02A333D8LQ

TL;DR: Pocket Casts app crashes on CarPlay after updating iOS to 15.4

> When I click on the icon, the area on the screen where the app should be is blank. It stays like that for a few seconds and then goes back to the main screen of CarPlay.

**List of tickets so far**

* #4865900-zen
* #4873212-zen
* #4898932-zen

One of the users tried with Airplane mode and reported it is still happening:

> I fiddled around with airplane mode again this morning, and I managed to get the app working in CarPlay for 1 hour / 2 connections. Then it started crashing again, and I couldn’t get it to work again.

emilylaguna:
| 🚧 Fixes #64 | 
|:---:|

I am having this mark #64 as complete because I think that based on the changes in this pr and the other improvements we should see a dramatic reduction of CarPlay related crashes. 

## Description

When reviewing crash logs for CarPlay the top crash is `Unable to allocate enough memory to encode data` which means the users device is running out of memory and the CarPlay app doesn't have enough free memory to encode the lists/images/etc. 

When reviewing the memory usage of the app:
1. On first load, we load the content of every initial tab
2. The images are loaded from Cache, but then processed in `adjustImageIfRequired` which creates a new image instance for each image for every reload.
3. The images that are passed to CarPlay are too big, CarPlay has 2 properties for: `CPListImageRowItem.maximumImageSize` and `CPListItem.maximumImageSize` which are set to 40x40 and 60x60, however we are passing 210x210 images.

Each refresh increases the memory footprint of the CarPlay app which can cause out of memory issues.

This introduces the following:

1. Only the podcasts tab data is loaded when the CarPlay app first launches
2. The data loading is deferred until the app becomes fully active
3. The data is only loaded once when the app first launches (before it would load at minimum twice)
4. Each tab now shows a loading message before loading the data to prevent hanging
5. The images:
   - Are now properly resized to the max size CarPlay expects
   - The image scale is set using the traits that CarPlay provides
   - The images are cached on disk and in memory
       - If a memory warning happens the memory cache is cleared defaulting to loading from the disk again

In my testing on my very old iPhone 8 Plus this has:
- Reduced the memory footprint **27.4%** (62 MB from 85.4 MB)
- Reduced the time to first launch by **61.83%** (645ms from 1.69s)

## Demo Videos

<details>
  <summary>Before 2</summary>

https://user-images.githubusercontent.com/793774/217630192-b35b9948-e4e6-49da-bda7-d8399d6b62c4.mov

</details>

<details>
  <summary>After</summary>

https://user-images.githubusercontent.com/793774/217630276-eaedf30d-ee36-4977-bfcc-5d1f005befb1.mov

</details>

## To test

> **Note**
> You'll want to test on a real device 📱

### Setup
1. Download the CarPlay Simulator from the [Additional Tools for Xcode](https://developer.apple.com/download/all/?q=Additional%20Tools%20for%20Xcode) package
2. Launch the CarPlay Simulator
3. Connect your phone via USB
4. Launch the app in the CarPlay Simulator
6. ✅ Verify you see a loading message and then the list loads
7. ✅ Verify the images appear correctly (not weird cropping or pixelations)
8. Tap on the Up Next bubble
9. ✅ Verify you see a loading indicator and the images load correctly
10. Tap back
11. ✅ Verify you don't see a loading indicator
12. Tap on the filters tab
13. Select a filter
14. ✅ Verify you see a loading indicator and the images load correctly
15. Tap on an Episode to play it
16. Tap on the Playing Next item
17. ✅ Verify you see a loading indicator and the images load correctly

## Checklist

### DIFF
--- a/podcasts/CarPlayImageHelper.swift
+++ b/podcasts/CarPlayImageHelper.swift
@@ -51,3 +51,56 @@ class CarPlayImageHelper {
         static let folderPreviewSize: CGRect = .init(x: 0, y: 0, width: 240, height: 240)
     }
 }
+
+// MARK: - CarPlay Resizing
+
+private extension UIImage {
+    /// This will process the image for us and return an image that CarPlay expects for its current traits (ie: scaling)
+    func carPlayImage(with traits: UITraitCollection, maxSize: CGSize) -> UIImage {
+        let imageAsset = UIImageAsset()
+        imageAsset.register(self, with: traits)
+        let processedImage = imageAsset.image(with: traits)
+
+        // Don't resize if we don't need to
+        if processedImage.size == maxSize {
+            return processedImage
+        }
+
+        // Scale the image to the max size CarPlay expects
+        return processedImage.scaleTo(maxSize: maxSize, traits: traits)
+    }
+
+    /// Resize the image to the CPListItem.maximumImageSize
+    func scaleTo(maxSize: CGSize, traits: UITraitCollection) -> UIImage {
+        let displayScale = traits.displayScale
+
+        let widthRatio = maxSize.width / size.width
+        let heightRatio = maxSize.height / size.height
+
+        let scaleFactor = min(widthRatio, heightRatio)
+        let scaledImageSize = CGSize(
+            width: size.width * scaleFactor,
+            height: size.height * scaleFactor
+        )
+
+        guard let resized = resized(to: scaledImageSize, scale: displayScale) else {
+            return self
+        }
+
+        return resized
+    }
+}
+
+private extension BaseEpisode {
+    var cacheKey: String {
+        if let episode = self as? Episode {
+            return episode.podcastUuid
+        }
+
+        if let userEpisode = self as? UserEpisode {
+            return userEpisode.urlForImage().absoluteString
+        }
+
+        return uuid
+    }
+}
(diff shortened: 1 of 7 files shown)


pocketcasts:
### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
CarPlayImageHelper.swift | 77 | Lines should not have trailing whitespace. (trailing_whitespace)


</p>

podcasts/CarPlayImageHelper.swift
```diff
@@ -25,27 +38,121 @@ class CarPlayImageHelper {
...
+        if processedImage.size == maxSize {
+            return processedImage
+        }
+
+        // Scale the image to the max size CarPlay expects
+        return processedImage.scaleTo(maxSize: maxSize, traits: traits)
+    }
+
+    /// Resize the image to the CPListItem.maximumImageSize
+    func scaleTo(maxSize: CGSize, traits: UITraitCollection) -> UIImage {
```
leandroalonso: Maybe this can go to `Utils`? I can easily see this being used somewhere else.

leandroalonso (approved):

podcasts/CarPlayImageHelper.swift
emilylaguna: 👍 Moved here: 69a72a25c20dfd528d64f85e31801b13e79f1656

PR #718, opened 2023-02-14
Bookmarks: Adds the FeatureFlag and table creation

emilylaguna:
| 📘 Project: #717 |
|:---:|

This adds the FeatureFlag and creates the table when the app launches. 

Since the FeatureFlag is not available in the Packages right now I have instead wrapped the table creation code in a DEBUG check along with a warning to remember to remove it. 

## To test

1. Launch the app
2. ✅ Verify there are no database errors in console
3. Go to Profile > Cog > Beta Features
4. ✅ Verify the Feature Flag is listed there

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

Issue #717, opened 2023-02-14
Project: Bookmarks


emilylaguna:
### Setup

- [x] Add feature flag: https://github.com/Automattic/pocket-casts-ios/pull/718
- [x] Create a new table in the database: https://github.com/Automattic/pocket-casts-ios/pull/718
- [ ] Add a data manager that:
   - [ ] Adds new bookmarks: https://github.com/Automattic/pocket-casts-ios/pull/719
   - [ ] Retrieves bookmarks: https://github.com/Automattic/pocket-casts-ios/pull/719
   - [ ] Deletes bookmarks
   - [ ] Updates bookmarks

### Adding Bookmarks

- [ ] Add a way to add via the Player: https://github.com/Automattic/pocket-casts-ios/pull/720
- [ ] Add a way to add via headphones
- [ ] Add a way to add via Siri
- [ ] Add a way to add via CarPlay
- [ ] Add a way to add via the Watch App
- [ ] Add a sound notification when the user creates a bookmark via the headphone gesture: https://github.com/Automattic/pocket-casts-ios/pull/720

### Viewing Bookmarks
- [ ] Add a way to view a list of the bookmarks from the player
- [ ] Add a way to view a list of the bookmarks on the profile tab
- [ ] Add a tap handler to play the episode from the bookmarked timestamp
- [ ] Show a visual indicator on the player timeline of where a bookmark is

### Managing Bookmarks
- [ ] Add a way to delete bookmarks
- [ ] Add a new settings section to control how bookmarks work (headphone gesture, length?, etc)

### Education
- [ ] Add a method of informing the user of the new feature, and educating the user on how to create bookmarks using the various methods
- [ ] Add a sound clip that educates the user when if they make a bookmark using a non-visual method (headphones, Siri, etc) for the first time

### Final touches

- [ ] Analytics
- [ ] iPad Testing
- [ ] iOS 14 Testing

BKrnk1337:
This project sounds Promising

PR #719, opened 2023-02-15
Bookmarks: Add BookmarkDataManager

emilylaguna:
| 📘 Project: #717 | 🛫 Depends on: #718 |
|:---:|:---:|

This adds the data layer to add, and retrieve bookmarks from the DB. 

I also added a few helpful extensions to the DBUtils. 

## To test
There are no changes to actually test in this PR. Just a code review. 

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/BookmarkDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/BookmarkDataManager.swift
@@ -1,0 +1,11 @@
+import FMDB
+import PocketCastsUtils
+
+public struct BookmarkDataManager {
+    static let tableName = "Bookmark"
+    private let dbQueue: FMDatabaseQueue
+
+    init(dbQueue: FMDatabaseQueue) {
+        self.dbQueue = dbQueue
+    }
+}
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/BookmarkDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/BookmarkDataManager.swift
@@ -30,7 +30,7 @@ public struct BookmarkDataManager {
                 let uuid = UUID().uuidString.lowercased()
                 let now = Date().timeIntervalSince1970
 
-                let columns = Column.allCases
+                let columns = Column.insertColumns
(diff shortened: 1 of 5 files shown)


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/BookmarkDataManager.swift
```diff
@@ -0,0 +1,192 @@
...
+        dbQueue.inDatabase { db in
+            do {
+                try db.insert(into: Self.tableName, columns: columns.map { $0.rawValue }, values: values)
+            } catch {
+                FileLog.shared.addMessage("BookmarkManager.add failed: \(error)")
+            }
+        }
+    }
+
+    /// Retrieves a single Bookmark for the given UUID
```
leandroalonso, line 47: Would that return the first bookmark? Last? Random?

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/BookmarkDataManager.swift
```diff
@@ -0,0 +1,192 @@
+import FMDB
+import PocketCastsUtils
+
+public struct BookmarkDataManager {
```
leandroalonso, line 4: I wonder if we can add a few unit tests for this?

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/BookmarkDataManager.swift
emilylaguna: The UUID should be fully unique to the bookmark and will be used to identify it across platforms for syncing, so this should only ever return a single bookmark regardless of order. 

The select query however defaults to ordering the newest items first.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/BookmarkDataManager.swift
emilylaguna: I originally wasn't able to add tests because FMDB wasn't easily mockable, however I learned you can make an in memory only DB so I've done that. 

Tests here: 9b733021c77d95f0711ed82aec89d94a10f207c9

leandroalonso (approved):

PR #720, opened 2023-02-15
Bookmarks: Add bookmarking when the previous track command is fired

emilylaguna:
| 📘 Project: #717 | 🛫 Depends on: #719  |
|:---:|:---:|

This adds the BookmarkManager which acts as the app interface to the BookmarkDataManager. It currently can:

1. Add a new bookmark for any episode
2. Play a test confirmation tone
3. Retrieve bookmarks

This also adds the basic action to create a bookmark using the previous track command via headphones or an external remote. 

### TODO Messages
We don't currently have a tone to play when the bookmark is created so I added TODO's to remind me to replace the temporary tone that I am using (which is just a system sound).

1. Connect your headphones to your device
2. Launch the app
3. Play any episode
5. Perform a previous track action (triple click on AirPods)
6. ✅ Verify the episode goes back to the beginning
7. Enable the Bookmarks FeatureFlag in Profile > Cog > Beta Features
8. Perform a previous track action again
9. ✅ Verify you hear a pleasant tone,
   - If you don't, you will need to bring your phone off silent because it won't play
10. ✅ Verify you see a log message for `[Bookmarks] Added bookmark for Episode_Title from Start_Time to End_Time`
11. ✅ Verify the End_Time is the point in time the episode was playing at
12. ✅ Verify the Start_Time is 60 seconds before that

## Checklist

### DIFF
--- a/podcasts/BookmarkManager.swift
+++ b/podcasts/BookmarkManager.swift
@@ -1,9 +1,31 @@
+#if DEBUG
+    #if canImport(AudioToolbox)
+    import AudioToolbox
+    #endif
+#endif
+
 import Foundation
 import PocketCastsDataModel
 import PocketCastsUtils
 
 struct BookmarkManager {
+    typealias Bookmark = BookmarkDataManager.Bookmark
+
     private let dataManager = DataManager.sharedManager.bookmarks
+
+    /// Plays the "bookmark created" tone
+    private lazy var tonePlayer: AVAudioPlayer? = {
+        guard
+            let url = Bundle.main.url(forResource: "TODO", withExtension: "TODO"),
+            let player = try? AVAudioPlayer(contentsOf: url)
+        else {
+            return nil
+        }
+
+        player.prepareToPlay()
+        return player
+    }()
+
     /// How long a bookmark clip is
     /// TODO: Make configurable
     private let clipLength = 1.minute
@@ -19,6 +41,7 @@ struct BookmarkManager {
 
         dataManager.add(episodeUuid: episode.uuid, podcastUuid: podcastUuid, start: startTime, end: time)
 
+        playTone()
 
         FileLog.shared.addMessage("[Bookmarks] Added bookmark for \(episode.displayableTitle()) from \(startTime) to \(time)")
     }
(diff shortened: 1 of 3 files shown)


podcasts/BookmarkManager.swift
```diff
@@ -0,0 +1,86 @@
...
+#endif
+
+import Foundation
+import PocketCastsDataModel
+import PocketCastsUtils
+
+struct BookmarkManager {
+    typealias Bookmark = BookmarkDataManager.Bookmark
+
+    private let dataManager = DataManager.sharedManager.bookmarks
```
leandroalonso: I think it would be better to inject that through the constructor. This way if we want to unit test it, it's easy.

podcasts/BookmarkManager.swift
```diff
@@ -0,0 +1,86 @@
...
+                }
+            }
+            #endif
+        #else
+        // Stop playing immediately and reset to 0
+        tonePlayer?.pause()
+        tonePlayer?.currentTime = 0
+
+        // Play
+        tonePlayer?.play()
```
leandroalonso, line 87: This never plays to me.

> If you don't, you will need to bring your phone off silent because it won't play

I'm assuming this means toggling Silent Mode off? I did that and still it didn't work (iOS 15).

podcasts/PlaybackManager.swift
```diff
@@ -1866,3 +1873,13 @@ class PlaybackManager: ServerPlaybackDelegate {
 
     private let commandCenterSource: AnalyticsSource = .nowPlayingWidget
 }
+
+extension PlaybackManager {
+    func bookmark() {
+        guard FeatureFlag.bookmarks.enabled else { return }
+        guard let episode = currentEpisode() else { return }
```
leandroalonso:
```suggestion
        guard FeatureFlag.bookmarks.enabled,
              let episode = currentEpisode() else {
            return
        }
```

podcasts/BookmarkManager.swift
emilylaguna: Done here: b681afde136c69d07d7b41b0affac855535fb2dc

podcasts/BookmarkManager.swift
emilylaguna: hm perhaps that sound isn't available on iOS 15, I've updated it here: 85ef0deb6b1de4ffe913d7906f2079d57e893920

podcasts/PlaybackManager.swift
emilylaguna: Done here: 3a8d4c83899fe0cbd296fe509a925084a562b31f

pocketcasts:
### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
PlaybackManager.swift | 1885 | Lines should not have trailing whitespace. (trailing_whitespace)


</p>

podcasts/BookmarkManager.swift
leandroalonso: Now it works!

I have just one last question tho: don't we want this to play even if the user is in silent mode?

leandroalonso (approved):

PR #727, opened 2023-02-20
Auto scroll Discover Featured collection view by adhering to new AutoScroll protocol

yaelirub:
As part of upcoming Discover updates, we want to allow the Featured collection view to auto scroll.

This PR:
1. Adds a new protocol `AutoScrollCollectionViewDelegate` and adheres ThemeableCollectionView to it. 
2. Defines featuredCollectionView as ThemeableCollectionView (which it already was in Xib) and initializes its timer to auto scroll in `FeaturedSummaryViewController`   
3. Adds a feature flag disabled by default to control activation of auto scroll

## To test
**No auto scroll on default**
1. Launch the app
2. Open Discover 
3. Notice the Featured Carousel doesn't auto scroll 

**Auto Scroll**
1. Go to profile -> Settings -> Beta Features 
2. Enable discoverFeaturedAutoScroll
3. Open Discover 
4. Watch the Featured  carousel auto scroll 
5. Watch that on the last item the carousel returns to the first item 

**Auto scroll and manual scroll**
1. Swipe to scroll the carousel forward or backwards
3. See the autoscroll continues to the next item as expected

https://user-images.githubusercontent.com/1335657/219984569-2b7799f1-73aa-4aef-9f72-5c133465e7d1.mov

Steps to test your PR.

## Checklist

### DIFF
--- a/podcasts/ThemeableCollectionView.swift
+++ b/podcasts/ThemeableCollectionView.swift
@@ -8,7 +8,7 @@ protocol AutoScrollCollectionViewDelegate: UICollectionView {
     func stopAutoScrollTimer()
 }
 
-class ThemeableCollectionView: UICollectionView {
+class ThemeableCollectionView: UICollectionView, AutoScrollCollectionViewDelegate {
     var style: ThemeStyle = .primaryUi04 {
         didSet {
             updateColor()
@@ -32,17 +32,9 @@ class ThemeableCollectionView: UICollectionView {
         backgroundColor = AppTheme.colorForStyle(style)
         indicatorStyle = AppTheme.indicatorStyle()
     }
-}
 
-extension ThemeableCollectionView: AutoScrollCollectionViewDelegate {
-    var timer: Timer? {
-        get {
-            return nil
-        }
-        set {
-        }
-    }
-
+   // MARK: - Auto scroll handling
+    var timer: Timer?
     func intializeAutoScrollTimer() {
         timer = Timer.scheduledTimer(timeInterval: 3.0, target: self, selector: #selector(scrolltoNextItem), userInfo: nil, repeats: true)
        }
--- a/podcasts/ThemeableCollectionView.swift
+++ b/podcasts/ThemeableCollectionView.swift
@@ -35,6 +35,7 @@ class ThemeableCollectionView: UICollectionView, AutoScrollCollectionViewDelegat
 
    // MARK: - Auto scroll handling
     var timer: Timer?
+
     func intializeAutoScrollTimer() {
         timer = Timer.scheduledTimer(timeInterval: 3.0, target: self, selector: #selector(scrolltoNextItem), userInfo: nil, repeats: true)
        }
--- a/podcasts/ThemeableCollectionView.swift
+++ b/podcasts/ThemeableCollectionView.swift
@@ -1,6 +1,13 @@
 
 import UIKit
 
+protocol AutoScrollCollectionViewDelegate: UICollectionView {
+    var timer: Timer? { get set }
+    func intializeAutoScrollTimer()
+    func scrolltoNextItem()
+    func stopAutoScrollTimer()
+}
+
 class ThemeableCollectionView: UICollectionView {
     var style: ThemeStyle = .primaryUi04 {
         didSet {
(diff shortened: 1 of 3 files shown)


podcasts/ThemeableCollectionView.swift
```diff
@@ -26,3 +33,32 @@ class ThemeableCollectionView: UICollectionView {
         indicatorStyle = AppTheme.indicatorStyle()
     }
 }
+
+extension ThemeableCollectionView: AutoScrollCollectionViewDelegate {
+    var timer: Timer? {
```
leandroalonso: We should move this to `ThemeableCollectionView`, otherwise, `timer` will always be `null`.

podcasts/ThemeableCollectionView.swift
```diff
@@ -26,3 +33,32 @@ class ThemeableCollectionView: UICollectionView {
...
+        let indexPath = IndexPath(item: nextIndex, section: 0)
+        if nextIndex < numberOfItems(inSection: 0) {
+            scrollToItem(at: indexPath, at: .left, animated: true)
+        } else if numberOfItems(inSection: 0) > 0 {
+            scrollToItem(at: IndexPath(item: 0, section: 0), at: .left, animated: false)
+        }
+    }
+
+    func stopAutoScrollTimer() {
+        timer?.invalidate()
```
leandroalonso, line 54: The `timer` is not invalidated because `timer` is always `null` (you can add a breakpoint here to check).

leandroalonso (commented):
@yaelirub works as described, I left one comment about an issue with the current `timer` property.

Also, I believe there's one improvement we can do that will make the experience slightly better:

When swiping manually, the timer doesn't stop, so occasionally you might swipe to a section, the timer is triggered, thus changing the section you swiped to right away:

https://user-images.githubusercontent.com/7040243/220394926-e8f6e684-2be7-49c2-b113-343f33c75ead.mov

If we stop the timer whenever there's a user interaction I believe we will solve this. :)

podcasts/ThemeableCollectionView.swift
yaelirub: Shouldn't the time be null till initialized from the instance of ThemeableCollectionView?

yaelirub: Ah! I thought `Timer.scheduledTimer` initializes it, but its indeed nil because in the get Im returning nil 🤦‍♀️ . How come it works then? 😬

leandroalonso: > How come it works then? 😬

I asked myself the same. 😬 I believe since we set `self` as the target it holds a reference to the timer. So it's there, in the memory 👻

yaelirub: I moved everything to ThemeableCollectionView and it still returns nil 🤔

yaelirub:
@leandroalonso , thanks for the suggestion for improvement with manual scroll. I can't seem to intercept a scroll on a `ThemableCollectionView`. I tried initializing a gesture recognizer:
``` let swipeGestureRecognizer = UISwipeGestureRecognizer(target: self, action: #selector(handleSwipeGesture(_:)))
        swipeGestureRecognizer.direction = .left
        swipeGestureRecognizer.delegate = self
        addGestureRecognizer(swipeGestureRecognizer)
```
with `handleSwipeGesture` being:
``` 
@objc func handleSwipeGesture(_ gestureRecognizer: UISwipeGestureRecognizer) {
        if gestureRecognizer.state == .began {
            stopAutoScrollTimer()
        } else if gestureRecognizer.state == .ended {
            intializeAutoScrollTimer()
        }
    }
```
handleSwipeGesture never gets called. 
I added:
```
func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
        return true
    }
```
but still doesn't get called. Any suggestions?

podcasts/ThemeableCollectionView.swift
yaelirub: updated

leandroalonso:
> but still doesn't get called. Any suggestions?

@yaelirub  It took me a while but I think I figured out how to do it. In `FeaturedSummaryViewController` add:

```swift
    func scrollViewDidEndDecelerating(_ scrollView: UIScrollView) {
        featuredCollectionView.intializeAutoScrollTimer()
    }

    func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
        featuredCollectionView.stopAutoScrollTimer()
    }
```

podcasts/ThemeableCollectionView.swift
```diff
@@ -25,4 +32,25 @@ class ThemeableCollectionView: UICollectionView {
         backgroundColor = AppTheme.colorForStyle(style)
         indicatorStyle = AppTheme.indicatorStyle()
     }
+
+   // MARK: - Auto scroll handling
+    var timer: Timer?
```
leandroalonso, line 37: [nit]

```suggestion
    var timer: Timer?

```

podcasts/ThemeableCollectionView.swift
```diff
@@ -25,4 +32,25 @@ class ThemeableCollectionView: UICollectionView {
...
+    func intializeAutoScrollTimer() {
```
leandroalonso: typo

```suggestion
    func initializeAutoScrollTimer() {
```

yaelirub:
@leandroalonso , updated to include the scroll delegate methods. As discussed in DM, this solution unfortunately only captures the improved functionality for `featuredViewController`. I tried to enforce the call to the scroll delegate methods through the protocol but it didn't work. will leave it like this for now.

leandroalonso (approved):
@yaelirub :shipit: 

As for the solution of detecting a swipe, I also looked at playing with the `delegate` somehow (by overriding it) but that is so much code that the solution is worst IMHO.

PR #729, opened 2023-02-20
Remove iOS 14 support

leandroalonso:
As we develop more new features using SwiftUI, iOS 14 has proven to be quite old and requires a lot of work for simple changes in UI.

iOS 14 user base is quite small so we're removing support for it.

## To test

1. Do a thorough code review.
2. Run the app and make sure everything is fine (including widgets)

## Checklist


emilylaguna (changes requested):
I gave this a test and everything seems to still work correctly. 

I tested:

1. Widget compact state correctly changes based on the dynamic type size
2. About view bubbles still appear
3. The name text field is focused when creating a new folder
4. The highlighted text view works
5. End of year stories still work correctly
6. 

While functionality wise everything seems good, there are a number of new warnings because of some iOS 15 deprecations which I think we should also fix. 

But since fixing some of the warnings will require more specific testing I am thinking we can fix them in another PR.

The only thing left for this PR is to also update the packages so their min is also iOS 15.

What do you think?

leandroalonso:
> But since fixing some of the warnings will require more specific testing I am thinking we can fix them in another PR.
> 
> The only thing left for this PR is to also update the packages so their min is also iOS 15.
> 
> What do you think?

Agree.

Packages updated in a063405a151de1789355cb6163dba5d3ec0c653d

emilylaguna:
@leandroalonso Ah I didn't think about the Pods. I was actually thinking of our [Swift Packages](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Modules/DataModel/Package.swift#L9)

leandroalonso:
@emilylaguna Done!

emilylaguna (approved):
:shipit:

Issue #709, opened 2023-02-13
Search Improvements


leandroalonso:
### Search History

- A list of recent searches will be displayed now when the user focus the search input box. 
- Whenever a user selects an item in a search results list, the selected item will be saved in the search history, and displayed in the list described in this task (p1665038946092499/1664975388.327359-slack-C03JCLLR9C3).

#### Tasks

- [x] Create the view #732
- [x] Save search terms and tapped podcasts #770
- [x] Populate the view with saved terms/podcasts/episodes #770
- [x] Open search when term is tapped / Open podcast/episode when tapped #770
- [x] Add Analytics #774

### Inline search results

The search results will be divided in two areas: podcasts and episodes. Each area will display a "Show all" link that will carry the user to the podcasts/episodes search result screen.

#### Tasks

- [x] Show podcasts and episode search result from the new APIs #766
- [x] When tapping a podcast/episode show it #769
- [x] Show all podcasts on a new screen #771 
- [x] Make sure both Discover and Podcasts search work as intended #772
- [x] Show local podcasts #773 
- [x] Add Analytics #774
- [x] UI Improvements #778 

#### Issues

- [x] When adding a URL, return the podcast #780
- [x] Folders do not appear on the list results correctly #780
- [x] On the list result, there's no loading indicator #780
- [x] The history is out of sync between screens #780
- [x] When swiping in the carousel after a local search, it gets reset #780
- [x] The History seems to be saved as the user types #795
- [x] From Emily: "I have a folder in my history but it's been deleted but still appears and tapping it doesn't do anything" #795

rviljoen:
Some good suggestions here https://github.com/Automattic/pocket-casts-ios/discussions/731

CameronMilliken:
I see several search improvements were merged. Does anyone know if any of them would help this user?

> Can’t search anything include Podcast station by Traditional Chinese input.
Like keyword "台東漫波電台"
Can’t find anything on Pocket Cast
but the keyword “台東慢波電台 taitung mambo radio” can find the right podcast.
The search is not friendly for Chinese user.
![image](https://user-images.githubusercontent.com/36011531/223216119-bb555a99-05ac-4012-a68f-4109b0a63c0d.jpg)
![image (1)](https://user-images.githubusercontent.com/36011531/223216278-da01bb16-0395-493f-9c8a-589b8f4c1e98.jpg)

ghost:
[...]

PR #732, opened 2023-02-21
Search Improvements: add "Search History" view

leandroalonso:
| 📘 Project: #709 |
|:---:|

Adds the "Search history" view. This view doesn't have any function and all the data displayed is mocked.

| Light mode | Dark Mode |
| ----------- | ----------- |
| Simulator Screen Shot - iPhone 14 - 2023-02-21 at 14 22 10 | Simulator Screen Shot - iPhone 14 - 2023-02-21 at 14 21 49 |

## To test

1. Run the app
1. In Settings > Beta Feature > enable `newSearch`
1. Under Podcasts or Discover tap the search
1. ✅ The search history should appear

You can test multiple themes and different font sizes in the Swift UI preview.

### Additional information

You'll notice that separators are not quite the same as the Figma designs. This is due to some limitations on iOS 14. Once we reach an agreement on whether to stop supporting it or not I'll make the appropriate changes.

## Checklist


yaelirub:
@leandroalonso , I realize this is a UI setup PR. I tested according to instructions and everything looks good, but when I started writing in the search field, I ran into a crash `Unexpectedly found nil while implicitly unwrapping an Optional value` in 
```
 self?.searchResultsTable.reloadData()
```
Is that expected?

podcasts/New Search/SearchHistoryView.swift
```diff
@@ -0,0 +1,130 @@
...
+                        print("remove tapped")
+                    }) {
+                        Image("close")
+                    }
+                    .buttonStyle(SecondaryButtonStyle())
+                    .frame(width: 48, height: 48)
+                }
+                Rectangle()
+                    .foregroundColor(.clear)
+//                    .foregroundColor(AppTheme.tableDividerColor(for: theme.activeTheme).color)
```
yaelirub: was this left here intentionally ?

podcasts/New Search/SearchHistoryView.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    init() {
+        UITableViewHeaderFooterView.appearance().backgroundView = UIView()
+    }
+
+    var body: some View {
+        VStack {
+            List {
+                HStack {
+                    Text("Recent searches")
```
yaelirub: Should this be localized? or does it not matter because its a mock and this won't be the text?

podcasts/New Search/SearchHistoryView.swift
```diff
@@ -0,0 +1,130 @@
...
+                        .font(style: .title2, weight: .bold)
+                    Spacer()
+                    Button("Clear all".uppercased()) {}
```
yaelirub: Same as to localized question

leandroalonso:
> Is that expected?

@yaelirub this only happens when the `newSearch` flag is enabled, but I fix it on a subsequent PR: https://github.com/Automattic/pocket-casts-ios/pull/733/files#diff-de7c3167399d649dfc90e4ba2d1a10e81e0b375e294d05516d4c318b8a49693dR135

podcasts/New Search/SearchHistoryView.swift
leandroalonso: It was, we need iOS 15+ support in order to use that. Given we decide to go ahead with stop supporting iOS 14 I'll update this PR once it gets merged and remove this comment. :)

podcasts/New Search/SearchHistoryView.swift
leandroalonso: It should! Even though is a mock it's the final string, done in a75b7fe4fcfa0a19fea1c4b235bfa455d6664b3e

podcasts/New Search/SearchHistoryView.swift
leandroalonso: Done in 9566d7d425c0a99cf89f7732f1d8aa817fd34a8b

podcasts/New Search/SearchHistoryView.swift
leandroalonso: Removed. :)

yaelirub (approved):
Leandro, looks good to me as a setup PR. I'm only getting my feet wet with SwiftUI, so feel free to get more 👀  on this if you want

PR #733, opened 2023-02-21
Search Improvements: create a search results delegate

leandroalonso:
| 📘 Project: #709  |
|:---:|

This PR adds a delegate so both the new and old search can be easily connected/disconnected from the app.

## To test

1. Run the app
1. In Profile > Settings > Beta features > make sure `newSearch` is disabled
1. ✅  Perform a search from both "Podcasts" and "Discover" and make sure it works as always
1. Go to Profile > Settings > Beta features > enable `newSearch`
1. ✅  When you tap the search, the search history should display
1. Tap any text into the search field
1. ✅ Search history should disappear and "Search results" should appear
1. Clear the text
1. ✅ Search history should appear again

## Checklist


podcasts/DiscoverPodcastSearchResultsController.swift
```diff
@@ -133,12 +133,16 @@ class DiscoverPodcastSearchResultsController: UIViewController, UITableViewDeleg
...
+    func clearSearch() {
         searchResults.removeAll()
         searchState = .notStarted
         searchResultsTable.reloadData()
     }
 
+    func performLocalSearch(searchTerm: String) {}
+
+    func performRemoteSearch(searchTerm: String, completion: @escaping (() -> Void)) { }
+
```
emilylaguna: Wdyt about using a default implementation in a protocol extension instead of adding stubs in the class?

podcasts/New Search/SearchResultsViewController.swift
```diff
@@ -1,12 +1,44 @@
 import Foundation
 import SwiftUI
 
-class SearchResultsViewController: OnboardingHostingViewController<AnyView> {
+protocol SearchResultsDelegate {
+    func clearSearch()
+    func performLocalSearch(searchTerm: String)
+    func performRemoteSearch(searchTerm: String, completion: @escaping (() -> Void))
+    func performSearch(searchTerm: String, triggeredByTimer: Bool, completion: @escaping (() -> Void))
```
emilylaguna, line 8: I'm curious if we need all of these since it seems like there's some overlap between them, like `performRemoteSearch` vs `performSearch`?

A thought that comes to mind is to have a single `performSearch` method that also includes additional information about the type of search / additional context that the delegate can act on, something like:

```
enum SearchType {
    case local, remote
}

enum SearchContext {
    case none, triggeredByTimer
}

---

func performSearch(searchTerm: String,
                       type: SearchType,
                       context: SearchContext,
                       completion: @escaping (() -> Void))
```

or we could have 2 methods for local / remote searching. 

Thoughts?

podcasts/PodcastListSearchResultsController.swift
```diff
@@ -131,10 +132,12 @@ class PodcastListSearchResultsController: UIViewController, UITableViewDelegate,
 
                 completion()
                 self?.remoteResults = results
-                self?.searchResultsTable.reloadData()
+                self?.searchResultsTable?.reloadData()
             }
         }
     }
+
+    func performSearch(searchTerm: String, triggeredByTimer: Bool, completion: @escaping (() -> Void)) {}
```
emilylaguna: Same question from above about using a default implementation instead of a stub here.

emilylaguna (changes requested):
Nice work, gave this a test and everything worked as intended. 🎉

I left a question about using a default implementation for the protocol methods, and the idea of limiting the number of methods in the protocol. 

Let me know what you think.

podcasts/DiscoverPodcastSearchResultsController.swift
leandroalonso: I think all in all the problem remains the same. 😬

My plan, for the future, is to have a single protocol with one clear and one search function. Is is this way now because of the current implementation, now the new one.

Anyway, I changed in b181532378451279d93e3a7a468d4b762b6a681f

podcasts/PodcastListSearchResultsController.swift
leandroalonso: Changed in b181532378451279d93e3a7a468d4b762b6a681f

podcasts/New Search/SearchResultsViewController.swift
leandroalonso: > I'm curious if we need all of these since it seems like there's some overlap between them, like performRemoteSearch vs performSearch?

The plan moving forward is to have one function to clear results and one to perform a search — independently from where it comes.

I just don't want to get ahead of myself and improve the current search without having the new one in place (given the current will be ditched anyway).

Plus, who should control remote results, or not should be the search entity, not the search bar as it is now.

Are you ok with changing that in the future?

emilylaguna: Makes sense, updating in the future works for me!

emilylaguna (approved):
Nice work, :shipit:

PR #737, opened 2023-02-23
Import OPML from URL

### LINKED ISSUE #442: Import OPML from a URL
### What problem does this address?

It would be great to add the ability for a user to enter a URL of an OPML and for Pocket Casts to import the podcast subscriptions.

Below is the user's suggestion. 
> it would be nice if they took a URL for importing an OPML file. 
https://twitter.com/davewiner/status/1583551708621217792

### What is your proposed solution?

The "Import/Export OPML" settings section could have another option to add by URL.

 

The option could be something like "By URL" with the description "Import your podcasts from an OPML file using a URL."

It might be best to follow a similar approach to the import from file code. The import by URL could download the content to the device and use the existing logic to parse the OPML file and subscribe to the podcasts.

Here is a URL that can be used for testing https://lists.pocketcasts.com/cdba0d8f-28ed-4654-bcb9-c65990f7399f.opml

This is the Android version of [the issue](https://github.com/Automattic/pocket-casts-android/issues/480) and [pull request](https://github.com/Automattic/pocket-casts-android/pull/482) of the change.

yaelirub:
Fixes #442 

Added the option to import podcasts from an opml file URL. 

https://user-images.githubusercontent.com/1335657/220860852-f56eab0b-fbab-4f9e-8840-0ec06d12f2f8.mov

A couple of notes 
- The icon for the "Import from URL" is temporary as I'm waiting on design. I will update it and the release notes once available.
- `opmlImporter` class logic for importing the podcasts from the opml file is strongly coupled with a progress view which is not ideal, especially now since we're calling it from SwiftUI view. That's why I had to make the progress view optional and change a few more things there. It would be good for us to refactor that at some point in the future. 

## To test
**from Profile**
1. Go to settings 
2. Go to import podcasts 
3. Tap import from URL
4. tap an invalid url 
5. see error "OPML import failed"
6. enter a valid url, but not opml 
7. see error "OPML import failed"
8. enter valid opml url
9. see "OPML import succeeded"
10. See button title change to "Done"
11. tap "Done"
12. see view dismissed 

**from Onboarding**
1. signup 
2. Tap Import podcasts from welcome screen 
3. repeat steps 1-12 from above section

## Checklist


yaelirub:
I probably need to update analytics

yaelirub:
Added the new icon and release notes.

podcasts/Onboarding/Import/ImportDetailsView.swift
```diff
@@ -1,7 +1,17 @@
 import SwiftUI
 
 struct ImportDetailsView: View {
+    enum OPMLImportResult {
+        case none
+        case success
+        case failure
+    }
```
leandroalonso, line 8: [nit] Wdyt about add breaklines to keep the `enum` separated from the `@Environment` and then another breakline separating the `@State`?

podcasts/Onboarding/Import/ImportDetailsView.swift
```diff
@@ -24,33 +34,99 @@ struct ImportDetailsView: View {
...
+        VStack {
+            TextField("https://...", text: $opmlURLText)
+                .textFieldStyle(RoundedBorderTextFieldStyle())
+                .autocapitalization(.none)
+
+            switch opmlImportResult {
+            case .none: Text("")
+            case .success:
+                Text(L10n.opmlImportSucceededTitle)
+                    .foregroundColor(Color.green)
```
leandroalonso: We should use a dynamic color here to avoid weirdness if the user is using any different theme, I believe `support02` would be the correct one here.

podcasts/Onboarding/Import/ImportDetailsView.swift
```diff
@@ -24,33 +34,99 @@ struct ImportDetailsView: View {
...
+            case .failure:
+                Text(L10n.opmlImportFailedTitle)
+                    .foregroundColor(Color.red)
```
leandroalonso: Same as above, here we should probably use `support05`

podcasts/Onboarding/Import/ImportDetailsView.swift
```diff
@@ -1,7 +1,17 @@
 import SwiftUI
 
 struct ImportDetailsView: View {
+    enum OPMLImportResult {
```
leandroalonso, line 4: Wdyt about renaming that to `OPMLURLImportResult`? Just so we make sure to convey that this is for external URLs.

podcasts/Onboarding/Import/ImportViewModel.swift
```diff
@@ -34,10 +34,11 @@ class ImportViewModel: OnboardingModel {
         .init(id: .castbox, displayName: "Castbox", steps: L10n.importInstructionsCastbox),
         .init(id: .overcast, displayName: "Overcast", steps: L10n.importInstructionsOvercast),
         .init(id: .other, displayName: "other apps", steps: L10n.importPodcastsDescription),
+        .init(id: .opmlFromURL, displayName: "URL", steps: L10n.importOpmlFromUrl)
     ]
 
     enum ImportAppId: String, AnalyticsDescribable {
```
leandroalonso: Now this is not only related to apps, maybe renaming it to `ImportSourceId` makes more sense? Wdyt?

leandroalonso (commented):
@yaelirub works as described, and left a few suggestions to improve the code.

Also, a question: when an OPML is successfully imported, we show a success message and a button. When I tap the button, I return to the previous screen (Take your podcasts with you...)

Wdyt about dismissing the modal as a whole?

yaelirub:
> when an OPML is successfully imported, we show a success message and a button. When I tap the button, I return to the previous screen (Take your podcasts with you...)

@leandroalonso , the reason I didn't want to dismiss the whole modal is because when coming from onboarding, I wanted them to go back to the welcome screen, in case they wanted to do something else there. Do you think its still worth coming back to the welcome screen from onboarding, or should we dismiss the whole modal? 

Actually, I'll just dismiss the full modal and see if you think it makes sense for both flows

podcasts/Onboarding/Import/ImportDetailsView.swift
yaelirub: updated

podcasts/Onboarding/Import/ImportDetailsView.swift
yaelirub: updated

podcasts/Onboarding/Import/ImportDetailsView.swift
yaelirub: updated

podcasts/Onboarding/Import/ImportDetailsView.swift
yaelirub: updated

podcasts/Onboarding/Import/ImportViewModel.swift
yaelirub: That lead me to replace all references to app to import source. Instead of installedApps, Im looking for AvailableResources

podcasts/Onboarding/Import/ImportLandingView.swift
```diff
@@ -5,11 +5,6 @@ struct ImportLandingView: View {
     let viewModel: ImportViewModel
 
     var body: some View {
-        // It's possible the user has no supported apps installed
```
yaelirub, line 8: No need to check that anymore. We will always have at least 2 available resources: others and opml from url

yaelirub:
@leandroalonso , ready for another look

leandroalonso (approved):
@yaelirub nice job! :shipit:

podcasts/Onboarding/Import/ImportDetailsView.swift
```diff
@@ -1,56 +1,134 @@
...
+            }
+        }
+    }
+
+    @ViewBuilder
+    private var opmlImportView: some View {
+        VStack {
+            TextField("https://...", text: $opmlURLText)
+                .textFieldStyle(RoundedBorderTextFieldStyle())
+                .autocapitalization(.none)
```
emilylaguna: I noticed the field isn't styled with theme and can have reduced readability depending on the system dark/light setting.

If we update to use the [`requiredStyle`](https://github.com/Automattic/pocket-casts-ios/blob/aee2844a9ed35524b897f75198999fd0cf9bf0be/podcasts/Themeable%2BSwiftUI.swift#L26) from Themeable+SwiftUI.swift we can apply the standard theme to the textfield.

<details>
  <summary>We also get highlighting the textfield in red if there's an error for free. </summary>
  

</details>

Wdyt?

| Before | After |
|:---:|:---:|
|||

```suggestion
            TextField("https://...", text: $opmlURLText)
                .requiredStyle(OPMLURLImportResult == .failure)
                .autocapitalization(.none)
```

podcasts/Onboarding/Import/ImportViewModel.swift
```diff
@@ -142,9 +154,40 @@ extension ImportViewModel {
...
+    func importFromURL(_ url: URL, completion: @escaping ((Bool) -> Void)) {
+        let task = URLSession.shared.dataTask(with: url) { data, response, error in
+            guard let data = data else {
+                print("Error downloading data: \(error?.localizedDescription ?? "Unknown error")")
+                completion(false)
+                return
+            }
+
+            let documentsDirectory = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
+            let fileURL = documentsDirectory.appendingPathComponent("feed.opml")
```
emilylaguna: [nit] I think it makes more sense to use a temporary directory here instead of the documents since we only need access to the file during the import process and without a cleanup step after the import, the feed.opml will continue to exist on the users device indefinitely. 

Using the temporary directory will let the system remove it whenever it needs to and we don't need to worry about cleaning it up. 

```suggestion
            let temporaryDirectory = FileManager.default.temporaryDirectory
            let fileURL = temporaryDirectory.appendingPathComponent("feed.opml")
```

podcasts/Onboarding/Import/ImportDetailsView.swift
emilylaguna, line 80: Wdyt about defaulting to the URL keyboard type?

```suggestion
                .autocapitalization(.none)
                .keyboardType(.URL)
```

emilylaguna (changes requested):
Excellent work on this, sorry I'm reviewing a bit later. I gave it a test and it imported like a charm!

I left a couple small comments, and one item about updating the text field to be consistent with the theme.

yaelirub:
Thank you @emilylaguna , updated and ready for another look

podcasts/Onboarding/Import/ImportDetailsView.swift
```diff
@@ -1,56 +1,135 @@
...
+
+            if opmlImportInProgress {
+                ProgressView(L10n.opmlImporting)
+            }
+        }
+    }
+
+    private var opmlViewButton: some View {
+        Button(action: {
+            if opmlURLImportResult == .success {
```
emilylaguna, line 102: I think we can add a return here to prevent the logic from falling through and kicking off the download again.

emilylaguna (approved):
Awesome work. 🥳 

I noticed one last thing while giving this another test, after I've finished my import and I tap the Done button the download kicks off again while the view is being dismissed and I get another log for 'File downloaded to...'

Approving now since after that's fixed we'll be good to merge :shipit:

PR #738, opened 2023-02-23
Search Improvements: add results view

leandroalonso:
Fixes #
| 📘 Project: #709 |
|:---:|

Add the search results view.

Please note that the podcast carousel is not exactly like what is contained in the Figma designs. I had a hard time reproducing that and decided to move on and tackle that later.

https://user-images.githubusercontent.com/7040243/220983844-ecf3d757-035c-459a-8499-ad6de2af888d.mp4

## To test

1. Run the app
3. Go to Discover
4. Tap the search input
5. ✅ Search results should appear
6. Type any search term on the search input
7. ✅ The search results should appear after a while
8. Tap "X" on search input
9. ✅ Search history should appear
10. Tap "Cancel"
11. ✅ The search is dismissed

## Checklist


pocketcasts:
### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
ThemeableListHeader.swift | 25 | Files should have a single trailing newline. (trailing_newline)


</p>

yaelirub:
Reviewed and works as expected. Is it possible that the separator line is a bit "weak" compared to designs in Figma ?

yaelirub (approved):
Looks like a setup PR. Reviewed and works as expected. Is it possible that the separator line is a bit "weak" compared to designs in Figma ?

leandroalonso:
> Is it possible that the separator line is a bit "weak" compared to designs in Figma ?

@yaelirub it is! Actually, it was "weak" compared to designs but stronger when compared to other areas of apps, I changed it in 1e0c872588bb1544ad66d928e30d681f16c69956 to be consistent with the rest of the app.

podcasts/New Search/Views/History/ThemeableSeparatorView.swift
```diff
@@ -0,0 +1,11 @@
+import SwiftUI
+
+struct ThemeableSeparatorView: View {
```
emilylaguna: It looks like this is doing the same thing as the [ThemedDivider](https://github.com/Automattic/pocket-casts-ios/blob/68353ce32559872ea1c9651d3f2555b89a34cf8e/podcasts/Styles.swift#L128), should we use that instead?

podcasts/New Search/Views/Results/PodcastCarouselView.swift
```diff
@@ -0,0 +1,75 @@
+import SwiftUI
+import PocketCastsDataModel
+
+struct PodcastsCarouselView: View {
+    @EnvironmentObject var theme: Theme
+
+    let size: Double = 0.48
```
emilylaguna: I don't see this used anywhere, should we remove this?

podcasts/New Search/Views/History/SearchHistoryCell.swift
```diff
@@ -0,0 +1,76 @@
...
+            .buttonStyle(ListCellButtonStyle())
+
+            VStack(spacing: 12) {
+                HStack(spacing: 12) {
+                    if let podcast {
+                        PodcastCover(podcastUuid: podcast.uuid)
+                            .frame(width: 48, height: 48)
+                            .allowsHitTesting(false)
+                        VStack(alignment: .leading, spacing: 2) {
+                            Text(podcast.title ?? "")
```
emilylaguna, line 43: Looking at the designs, for an episode result it shows the episode title. Based on this logic we'll always display the podcast title for both types of results.

podcasts/New Search/Views/History/SearchHistoryCell.swift
```diff
@@ -0,0 +1,76 @@
...
+    var podcast: Podcast?
+
+    var episode: Episode?
+
+    var searchTerm: String?
+
+    private var subtitle: String {
+        if let episode, let podcast {
+            let duration = TimeFormatter.shared.multipleUnitFormattedShortTime(time: TimeInterval(episode.duration))
+            return "\(L10n.episode) • \(duration) • \(podcast.title ?? "")"
```
emilylaguna: Looking at the designs, for an episode it looks like we should show the podcast author instead of the podcast title here. 

Also I left a comment below about hiding the trailing `•` if the title is empty.

podcasts/New Search/Views/History/SearchHistoryCell.swift
```diff
@@ -0,0 +1,76 @@
...
+import PocketCastsUtils
+
+struct SearchHistoryCell: View {
+    @EnvironmentObject var theme: Theme
+
+    var podcast: Podcast?
+
+    var episode: Episode?
+
+    var searchTerm: String?
```
emilylaguna, line 12: What do you think about simplifying this into a single struct that represents all the items we need to display the UI?

I think we could do something like this:

```
struct SearchHistoryItem {
    let title: String
    let subtitle: String?
    let image: AnyView

    init(episode: Episode, podcast: Podcast) {
        ...
    }

    init(podcast: Podcast) {
        ...
    }

    init(term: String) {
        ...
    }
}
```

podcasts/New Search/Views/History/SearchHistoryCell.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    private var subtitle: String {
+        if let episode, let podcast {
+            let duration = TimeFormatter.shared.multipleUnitFormattedShortTime(time: TimeInterval(episode.duration))
+            return "\(L10n.episode) • \(duration) • \(podcast.title ?? "")"
+        } else if let podcast {
+            return "Podcast • \(podcast.author ?? "")"
```
emilylaguna: I did some testing where the author isn't returned and it to me it looked quite odd to have a trailing ` • `

Screen Shot 2023-02-24 at 2 04 49 PM

wdyt about doing something like this?

```suggestion
            return [L10n.podcastSingular, podcast.author].compactMap { $0 }.joined(separator: " • ")
```

We can also use the `L10n.podcastSingular` here instead of the string directly.

podcasts/New Search/Views/History/SearchHistoryCell.swift
```diff
@@ -0,0 +1,76 @@
...
+                print("row tapped")
+            }) {
+                Rectangle()
+                    .foregroundColor(.clear)
+                    .frame(maxWidth: .infinity, maxHeight: .infinity)
+            }
+            .buttonStyle(ListCellButtonStyle())
+
+            VStack(spacing: 12) {
+                HStack(spacing: 12) {
```
emilylaguna: The right margin between the text and the close button is a bit too large which is resulting in the text being truncated a bit earlier. 

It looks like the `HStack` spacing is being applied between the text and the close button, but since the close button has a static width of 48 it includes its own padding it's adding too much visual padding. 

If we change the `HStack` spacing to 0, we can manually apply the padding between the image and the text and have a bit of extra space for the text.

| Before | After |
|:---:|:---:|
Screen Shot 2023-02-24 at 2 11 32 PM|Screen Shot 2023-02-24 at 2 13 50 PM|

podcasts/New Search/Views/Results/PodcastCarouselView.swift
```diff
@@ -0,0 +1,75 @@
...
+                            }
+                        }
+                    }
+                    .padding(.all, 10)
+                }
+                .frame(width: UIScreen.main.bounds.width, height: UIScreen.main.bounds.height * 0.3)
+                .tabViewStyle(PageTabViewStyle(indexDisplayMode: .never))
+            }
+            ThemeableSeparatorView()
+                .padding(EdgeInsets(top: 0, leading: 16, bottom: 0, trailing: 0))
```
emilylaguna: This is a bit of a nitpick, no need to specify all the edges here.

```suggestion
                .padding(.leading, 16)
```

podcasts/New Search/Views/Results/PodcastCarouselView.swift
```diff
@@ -0,0 +1,75 @@
...
+                }
+                Button(action: {
+                    print("subscribe")
+                }) {
+                    Image("discover_subscribe_dark")
+                }
+                .background(ThemeColor.veil().color)
+                .foregroundColor(ThemeColor.contrast01().color)
+                .cornerRadius(30)
+                .padding(EdgeInsets(top: 0, leading: 0, bottom: 6, trailing: 6))
```
emilylaguna: [nit]: 
```suggestion
                .padding([.trailing, .bottom], 6)
```

podcasts/New Search/Views/Results/SearchEpisodeCell.swift
```diff
@@ -0,0 +1,52 @@
...
+                        .allowsHitTesting(false)
+                    VStack(alignment: .leading, spacing: 2) {
+                        Text(DateFormatHelper.sharedHelper.tinyLocalizedFormat(episode.publishedDate).localizedUppercase)
+                            .font(style: .footnote, weight: .bold)
+                            .foregroundColor(AppTheme.colorForStyle(.primaryText02, themeOverride: theme.activeTheme).color)
+                        Text(podcast.title ?? "")
+                            .font(style: .subheadline, weight: .medium)
+                            .foregroundColor(AppTheme.colorForStyle(.primaryText01, themeOverride: theme.activeTheme).color)
+                            .lineLimit(2)
+                        Text(TimeFormatter.shared.multipleUnitFormattedShortTime(time: TimeInterval(episode.duration)))
```
emilylaguna: There's a few places we're showing this value I wonder if we should add a `displayableDuration` to the [BaseEpisode+Formatting](https://github.com/Automattic/pocket-casts-ios/blob/c9a937af6f4d017e337fad91d22915edea4aac57/podcasts/BaseEpisode%2BFormatting.swift#L18) extension?

podcasts/New Search/Views/SearchView.swift
```diff
@@ -9,7 +9,7 @@ struct SearchView: View {
 
     var body: some View {
         if displaySearch.isSearching {
-            Text("Search results")
+            SearchResultsView()
```
emilylaguna, line 19: If the mini player is open the last few rows of the results aren't visible. 

This can be fixed by applying the `Constants.Values.miniPlayerOffset` as a bottom padding if there's a current episode playing:

`.padding(.bottom, (PlaybackManager.shared.currentEpisode() != nil) ? Constants.Values.miniPlayerOffset : 0)`

https://user-images.githubusercontent.com/793774/221279420-284d763b-a464-4274-87a6-9bc62d8bd221.mp4

emilylaguna (changes requested):
Excellent work on this!

I left some comments below, let me know what you think. 

One last [nit] that I wanted to mention is we can replace all the:

`AppTheme.colorForStyle(style, themeOverride: theme.activeTheme).color`

with: 

`AppTheme.color(for: .primaryText01, theme: theme)`

podcasts/New Search/Views/History/SearchHistoryCell.swift
leandroalonso: Ah, this is just a mock. We won't even use `Podcast` and/or `Episode` for that.

podcasts/New Search/Views/History/SearchHistoryCell.swift
leandroalonso: Same thing as before, just a mock. :D

podcasts/New Search/Views/History/SearchHistoryCell.swift
leandroalonso: As mentioned previously, `Episode` and `Podcast` won't be used for displaying the results/history. It will be another entity. When we have this in place then we can make this change. Wdyt?

podcasts/New Search/Views/History/ThemeableSeparatorView.swift
leandroalonso: Thanks for that! Removed in 74003f037158b9caef9316bc8ce3a27349e1b3a8

podcasts/New Search/Views/Results/PodcastCarouselView.swift
leandroalonso: Done in d283b8143b0879483a190d25b21d2b0bb87b6054

podcasts/New Search/Views/History/SearchHistoryCell.swift
emilylaguna: Sounds good to me.

podcasts/New Search/Views/Results/SearchEpisodeCell.swift
leandroalonso: Done in a9c6649c09c47876e7da5805d76facae2fdfdb4f

podcasts/New Search/Views/SearchView.swift
leandroalonso: Done in 0f77f60d67e9116f6426dc27a1e55f377b881e1c

podcasts/New Search/Views/Results/SearchEpisodeCell.swift
leandroalonso: Done in a9c6649c09c47876e7da5805d76facae2fdfdb4f

podcasts/New Search/Views/SearchView.swift
leandroalonso: Done in 6dc8b956b053adedc6af2611ec30b7434b9005d5

leandroalonso:
@emilylaguna thank you for the thorough review! Ready for another look.

emilylaguna (changes requested):
Awesome work, the changes look good.

I did notice an issue with the miniplayer padding being applied when the keyboard is open which results in a bar above the keyboard. 



Once that's fixed we're good to :shipit: 

I played with an idea to make applying and adjusting the padding a bit more seamless in SwiftUI. Let me know what you think: 

https://gist.github.com/emilylaguna/2466dffe1a99d0b5d6dbc13b4aa21ce7

leandroalonso:
@emilylaguna fixed in 69a46c81175ae7ec9cf059d92c68ada5c7d58408

emilylaguna (approved):
:shipit: 

I'll note there is an issue with the longer list being clipped by the keyboard but based on our Slack discussion we'll be hiding the keyboard on scroll in the future so it's a non issue .

PR #739, opened 2023-02-23
Updates the AirPlay auto play logic to be more explicit about when it can be triggered

### LINKED ISSUE #735: Audio is no longer pausing during phone calls
### Expected behavior

The playback should pause during a call and resume once the call has ended. Seems to have started in v7.32

### Actual behavior

The playback continues while you're on a call. We may need to also look into how it works in CarPlay

- Play an episode
- Accept a phone call on the device playing the audio
- The podcast will continue playing while on the call

##### Tested on [iPhone 14 Pro Max], iOS [16.3], Pocket Casts [7.33]

5984411-zd-woothemes

emilylaguna:
Fixes #735

This updates the AirPlay play next episode logic that was added in (https://github.com/Automattic/pocket-casts-ios/pull/676) with more checks to ensure its only triggered when:
- We are waiting for the initial load of an episode
- We want to keep playing
- The previous playback status is "Waiting to play"
- The previous reason we are waiting is "no item" and there is no current reason.
- Only apply when we detect we're on AirPlay

## To test

**Note:** You'll need to test on a real device. 

### Verify the original issue is still fixed

1. On your Mac, go to System Preferences > Sharing > Enable Airplay Receiver
   - This will allow your Mac to act as an Airplay speaker
2. On your device add many items to your up next
3. Tap the Airplay icon and select your Mac as the receiver
4. Play an episode
   - To speed things up you can skip to the last 20 seconds or so
6. Put the app in the background
13. Let the episode play out
14. ✅ Verify the playback continues as planned
15. ✅ Verify you see: `[DefaultPlayer] Detected that playback was paused while trying to play the next item. Attempting to resume playback...` in the logs to indicate that the fix was applied

### Verify the new issue is fixed
1. Launch the app
2. Stream an episode with no effects so that the DefaultPlayer is used
3. Call any number on your phone
6. ✅ Verify the podcast stops playing
7. Hang up
8. ✅ Verify the podcast starts playing again
9. You can also test this by opening another app that plays such as a video on Reddit and playing an item

## Checklist

### DIFF
--- a/podcasts.xcodeproj/xcshareddata/xcschemes/pocketcasts.xcscheme
+++ b/podcasts.xcodeproj/xcshareddata/xcschemes/pocketcasts.xcscheme
@@ -82,6 +82,13 @@
             isEnabled = "YES">
          </CommandLineArgument>
       </CommandLineArguments>
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "CG_NUMERICS_SHOW_BACKTRACE"
+            value = "YES"
+            isEnabled = "YES">
+         </EnvironmentVariable>
+      </EnvironmentVariables>
    </LaunchAction>
    <ProfileAction
       buildConfiguration = "Debug"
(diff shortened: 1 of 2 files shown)


leandroalonso (approved):
:shipit:

PR #755, opened 2023-03-06
Discover Carousel: prepare it to receive Sponsored podcasts

leandroalonso:
This PR prepares the Discover Carousel to be able to show sponsored podcasts.

**Tracks is gonna be tackled in a separated task.**

## To test

1. First, run this PR with `Debug` Build Configuration (Manage Schemes > pocketcasts > Run > Info > Build Configuration: Debug)
2. ✅ Make sure the carousel appears just fine
3. Change the `Debug` Build Configuration to `Staging`
4. Go to `DiscoverServerHandler.swift` and change line 31 to `discoverRequest(path: ServerConstants.Urls.discover() + "ios/content-carousel.json", type: DiscoverLayout.self) { discoverItems, cachedResponse in` (so it will use the mock)
5. Run the app
6. ✅ The carousel should appear. All the items should display the "Featured" label except the first and the fourth, which displays "Sponsored"

## Checklist

### DIFF
--- a/podcasts/ArrayExtension.swift
+++ b/podcasts/ArrayExtension.swift
@@ -4,4 +4,11 @@ extension Array {
             Array(self[$0 ..< Swift.min($0 + size, count)])
         }
     }
+
+    @discardableResult
+    mutating func insert(_ element: Element, safelyAt at: Int) -> Int {
+        let indexToInsert = at <= count ? at : count
+        insert(element, at: indexToInsert)
+        return indexToInsert
+    }
 }
--- a/podcasts/ArrayExtension.swift
+++ b/podcasts/ArrayExtension.swift
@@ -7,7 +7,7 @@ extension Array {
 
     @discardableResult
     mutating func insert(_ element: Element, safelyAt at: Int) -> Int {
-        let indexToInsert = at <= count ? at : count
+        let indexToInsert = Swift.min(at, count)
         insert(element, at: indexToInsert)
         return indexToInsert
     }
--- a/podcasts/FeaturedSummaryViewController.swift
+++ b/podcasts/FeaturedSummaryViewController.swift
@@ -149,6 +149,9 @@ class FeaturedSummaryViewController: SimpleNotificationsViewController, GridLayo
             listType = delegate.replaceRegionName(string: title)
         }
 
+        let dispatchGroup = DispatchGroup()
+
+        dispatchGroup.enter()
         DiscoverServerHandler.shared.discoverPodcastList(source: source, completion: { [weak self] podcastList in
             guard let strongSelf = self, let discoverPodcast = podcastList?.podcasts else { return }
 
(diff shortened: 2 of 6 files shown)


podcasts/ArrayExtension.swift
```diff
@@ -4,4 +4,11 @@ extension Array {
             Array(self[$0 ..< Swift.min($0 + size, count)])
         }
     }
+
+    @discardableResult
+    mutating func insert(_ element: Element, safelyAt at: Int) -> Int {
+        let indexToInsert = at <= count ? at : count
```
emilylaguna: Should we just use `.min` instead of manually doing the check?

```suggestion
        let indexToInsert = Swift.min(at, count)
```

emilylaguna (changes requested):
Nice work, confirmed this is loading in the new items and displaying them in the correct positions. 

When I tested on my slower network I noticed that the normal featured items will display first, and then the sponsored items will pop in not too long after that which looks a bit odd. 

<details>
  <summary>Demo Video</summary> 

https://user-images.githubusercontent.com/793774/223515237-7620fd28-19fc-452c-92a2-90a5fb4d7c0c.mp4

</details>

I did some quick debugging and during the loading process the first request to get the featured items finishes first, and then modifies the podcasts property (`strongSelf.podcasts`), then elsewhere another request finishes which triggers a reload of the view the loaded featured items are displayed first. Then the sponsored podcast request finishes, and reloads again. 

I think changing from modifying the podcasts property to using a local variable, and setting `self.podcasts` once everything is loaded should fix this.

leandroalonso:
@emilylaguna fixed in 80515b96e9d122a415082931996fd9addadfdcef

podcasts/FeaturedSummaryViewController.swift
```diff
@@ -183,13 +181,26 @@ class FeaturedSummaryViewController: SimpleNotificationsViewController, GridLayo
...
+            guard let self else {
+                return
+            }
+
+            // Add featured podcasts
+            for (index, discoverPodcast) in podcastsToShow.enumerated() {
+                self.podcasts.append(discoverPodcast)
+
+                if index == (self.maxFeaturedItems - 1) { break }
+            }
```
emilylaguna: [Nit] I know this code isn't new, but this could be reduced down to a single line: 

```suggestion
            // Add featured podcasts
            self.podcasts = Array(podcastsToShow.prefix(self.maxFeaturedItems))
```

podcasts/FeaturedSummaryViewController.swift
```diff
@@ -149,20 +150,58 @@ class FeaturedSummaryViewController: SimpleNotificationsViewController, GridLayo
...
 
+        let dispatchGroup = DispatchGroup()
+
+        var podcastsToShow: [DiscoverPodcast] = []
+
+        var sponsoredPodcastsToAdd: [Int: DiscoverPodcast] = [:]
+
+        dispatchGroup.enter()
         DiscoverServerHandler.shared.discoverPodcastList(source: source, completion: { [weak self] podcastList in
             guard let strongSelf = self, let discoverPodcast = podcastList?.podcasts else { return }
```
emilylaguna: We can remove the self references since it's not being used anymore.

```suggestion
        DiscoverServerHandler.shared.discoverPodcastList(source: source, completion: { podcastList in
            guard let discoverPodcast = podcastList?.podcasts else { return }
```

emilylaguna (approved):
Nice work! Tested and it's all working well now. 

I left a couple comments, ~one about removing the self reference~, and another optional comment about refactoring some code down. 

~Once the warning is fixed we can :shipit:~

I see the warning was fixed in the [next PR](https://github.com/Automattic/pocket-casts-ios/pull/757)

PR #769, opened 2023-03-08
Search Improvements: add interactions for podcasts/episodes

leandroalonso:
| 📘 Project: #709 |
|:---:|

When showing search results, add interactions for podcasts/episodes.

### Before everything

1. Run the app

### Podcasts

1. Go to discover
1. Search for anything
1. Tap on the podcast image
1. ✅ The podcast details screen should show
1. Go back and tap on the title/author
1. ✅ The podcast details screen should show
1. Tap the "+" button
1. ✅ You should subscribe to the podcast

### Episodes

1. Go to discover
1. Search for anything
1. Tap on any episode
1. ✅ The episode should load
1. Interact with the buttons and make sure everything is working
1. Go back
1. ✅ Make sure "Show all" is not shown for episodes

## Checklist

### DIFF
--- a/podcasts/EpisodeLoadingController.swift
+++ b/podcasts/EpisodeLoadingController.swift
@@ -6,14 +6,20 @@ import PocketCastsUtils
 // MARK: - Test Code Below
 
 struct EpisodeLoadingView: View {
+    @EnvironmentObject var theme: Theme
+
     var body: some View {
         ZStack(alignment: .center) {
             ProgressView()
+                .tint(AppTheme.loadingActivityColor().color)
+                .scaleEffect(x: 2, y: 2, anchor: .center)
         }
+        .frame(maxWidth: .infinity, maxHeight: .infinity)
+        .applyDefaultThemeOptions()
     }
 }
 
-class EpisodeLoadingController: UIHostingController<EpisodeLoadingView> {
+class EpisodeLoadingController: UIHostingController<AnyView> {
     private let episodeUuid: String
     private let podcastUuid: String
 
@@ -21,7 +27,7 @@ class EpisodeLoadingController: UIHostingController<EpisodeLoadingView> {
         self.episodeUuid = episodeUuid
         self.podcastUuid = podcastUuid
 
-        super.init(rootView: EpisodeLoadingView())
+        super.init(rootView: AnyView(EpisodeLoadingView().setupDefaultEnvironment()))
     }
 
     // Do a quick check to see if we need to load this episode or not
--- a/podcasts/EpisodeLoadingController.swift
+++ b/podcasts/EpisodeLoadingController.swift
@@ -128,8 +128,7 @@ extension EpisodeLoadingController: UIViewControllerAnimatedTransitioning, UINav
         let duration = transitionDuration(using: transitionContext)
 
         // We delay the alpha transition here because the episode controller has a weird animation effect when it first loads
-        // 1 second is probably too long of a delay
-        UIView.animate(withDuration: duration, delay: 1.0) {
+        UIView.animate(withDuration: duration, delay: 0.3) {
             fromView.view.alpha = 0
         } completion: { _ in
             transitionContext.completeTransition(true)
--- a/podcasts/EpisodeLoadingController.swift
+++ b/podcasts/EpisodeLoadingController.swift
@@ -3,16 +3,29 @@ import PocketCastsServer
 import PocketCastsDataModel
 import PocketCastsUtils
 
-// MARK: - Test Code Below
+// MARK: - Async Episode Loader
+
+class EpisodeLoadingModel: ObservableObject {
+    @Published var error = false
+}
 
 struct EpisodeLoadingView: View {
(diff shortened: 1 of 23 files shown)


podcasts/New Search/Views/Results/SubscribeButtonView.swift
```diff
@@ -0,0 +1,50 @@
...
+                Image("discover_subscribe_dark")
+            }
+        }
+        .buttonStyle(SubscribeButtonStyle())
+        .onAppear {
+            isSubscribed = DataManager.sharedManager.findPodcast(uuid: podcastUuid) != nil
+        }
+    }
+
+    private func subscribe() {
```
leandroalonso, line 35: I'm debating whether I should create a model or not for the subscribe/isSubscribed logic.

While I think this is the correct approach, should we do something so simple? Open to feedback.

emilylaguna (changes requested):
I gave this a test and overall it works very well, but I did hit an issue. 

### Issue:

For some episodes I am getting stuck on the loading screen without anything else happening. 

<details>
  <summary>Demo Video</summary>
  
https://user-images.githubusercontent.com/793774/223874850-131fa14c-9531-46ec-a4c2-0764a67692c2.MP4
  
</details>

I did some digging and it turns out even after the add podcast code has ran the episode is still nil when it's being returned from the database. 

I think this issue happens when I load an episode for a podcast and then try to load another episode that's missing from the database for the same podcast. 

I think we need to change how we're adding the episode/podcast to use the:

`ServerPodcastManager.shared.addMissingPodcastAndEpisode` and the `ServerPodcastManager.shared.addMissingEpisode` to make sure it's actually added

Then display an error if the loading fails. 

### Idea

Looking through the changes in the PR to add async loading to the EpisodeDetailViewController has me a bit worried about some potential other issues that could pop up when the controller is being used elsewhere. While I don't have a specific example it's just a general concern. 

I did have a thought about how we could accomplish this but without effecting the existing code as much by having a loading view controller just for loading the episode, and then pushing to the episode controller once it's done. 

This would be done by wrapping the loading + episode controller in a navigation controller, and hiding the nav bar. 

I made a quick concept and pushed it here: https://github.com/Automattic/pocket-casts-ios/compare/issue/709-show-search-results...idea/episode-async-loading

Let me know what you think.

leandroalonso:
> Let me know what you think.

@emilylaguna  I agree. I applied your solution and changed it to handle episodes failing to load and supporting themes.

Ready for another 👀

podcasts/New Search/Views/Results/SubscribeButtonView.swift
emilylaguna: Hmm. While I think it's the correct approach it does feel like overkill for what this does. 

I'd say if we start using this in more places, or adding more functionality to it then we should make the model, but for now lets leave it as is.

podcasts/MainTabBarController.swift
```diff
@@ -170,26 +170,45 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
...
             // I know it looks dodgy, but the episode card won't load properly if you just dismissed another view controller. Need to figure out the actual bug...but for now:
             // (before you ask, using the completion block doesn't work above, regardless of whether animated is true or false
             DispatchQueue.main.asyncAfter(deadline: .now() + 0.5.seconds) {
-                let episodeController = EpisodeDetailViewController(episodeUuid: episodeUuid, source: .homeScreenWidget)
-                episodeController.modalPresentationStyle = .formSheet
+                var episodeController: UIViewController
+
+                if EpisodeLoadingController.needsLoading(uuid: episodeUuid), let podcastUuid {
+                    episodeController = EpisodeLoadingController(episodeUuid: episodeUuid,
+                                                                 podcastUuid: podcastUuid)
```
emilylaguna: We can remove the var here since it's not actually used. 

```suggestion
                if EpisodeLoadingController.needsLoading(uuid: episodeUuid), let podcastUuid {
                    let episodeController = EpisodeLoadingController(episodeUuid: episodeUuid,
                                                                 podcastUuid: podcastUuid)
```

podcasts/EpisodeLoadingController.swift
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        // Add the episode controller below the loading view so we can fade into it
+        transitionContext.containerView.insertSubview(toView.view, belowSubview: fromView.view)
+
+        let duration = transitionDuration(using: transitionContext)
+
+        // We delay the alpha transition here because the episode controller has a weird animation effect when it first loads
+        // 1 second is probably too long of a delay
+        UIView.animate(withDuration: duration, delay: 1.0) {
```
emilylaguna: I did some testing and 0.3 seconds seems to work well. 

```suggestion
        UIView.animate(withDuration: duration, delay: 0.3) {
```

emilylaguna (approved):
🎉 Nice work, gave this a test and all is working well now. 

I left a few small comments, then it's ready to 🚢

PR #772, opened 2023-03-10
Search Improvements: search from Podcasts and Discover tab

leandroalonso:
| 📘 Project: #709 |
|:---:|

Search now works correctly from both "Podcasts" tab and "Discover" tab.

### Before everything

1. Run the app

### Independently search

1. Go to Podcasts
1. Search for something
1. ✅ Make sure results are appearing
1. Without closing this search, go to Discover
1. Make another search
1. ✅ Check Podcasts and Discover, they should display independently results based on your search
1. Dismiss both searches

### Redoing a search from History

1. Make sure search is appearing under Podcasts
1. Tap the search field
1. In the history, select any search term
1. Go to Discover
1. ✅ The search input should be clear
1. Tap on it
1. Select any search term
1. Go back to Podcasts
1. Teh search term should display the search term you selected on step 3

## Checklist

### DIFF
--- a/podcasts/DiscoverViewController.swift
+++ b/podcasts/DiscoverViewController.swift
@@ -47,9 +47,9 @@ class DiscoverViewController: PCViewController {
 
         title = L10n.discover
 
-        setupSearchBar()
         handleThemeChanged()
         reloadData()
+        setupSearchBar()
 
         addCustomObserver(Constants.Notifications.chartRegionChanged, selector: #selector(chartRegionDidChange))
         addCustomObserver(Constants.Notifications.tappedOnSelectedTab, selector: #selector(checkForScrollTap(_:)))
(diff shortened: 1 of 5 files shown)


podcasts/DiscoverViewController.swift
```diff
@@ -47,9 +47,9 @@ class DiscoverViewController: PCViewController {
 
         title = L10n.discover
 
-        setupSearchBar()
         handleThemeChanged()
         reloadData()
+        setupSearchBar()
```
leandroalonso, line 52: Had to move this below because adding the search controller as a child wasn't working when the call was before.

No idea why. 🙃

emilylaguna (changes requested):
While testing I found that I am not able to get the search history to display again after my first search:

https://user-images.githubusercontent.com/793774/224797478-59eb0270-bdd6-4fff-a6d7-97c7c3b7c4b1.MP4

leandroalonso:
> While testing I found that I am not able to get the search history to display again after my first search:

I'm not able to reproduce that. 🤔

I've pushed a commit that maybe should fix that, but I'm not sure. Can you double-check, please?

emilylaguna (approved):
All working fine for me now. 🎉

PR #773, opened 2023-03-13
Search Improvements: show local podcasts and folders

leandroalonso:
| 📘 Project: #709 |
|:---:|

Show local podcasts and folders when searching from "Podcasts".

https://user-images.githubusercontent.com/7040243/224776262-de56c635-1704-43ce-889f-3dd595f42970.mp4

**UI is part of another task.**

### Before everything

1. Run the app
4. Login into an account with Plus, folders and subscribed podcasts

### Local podcasts and folders

1. Go to "Podcasts"
2. Search for a term that matches the name of your folders and podcasts
3. ✅ Your local podcasts and folders should appear at the beginning of "Podcasts" section. Also, after a while, a remote search is performed, and remote podcasts/folders are shown
4. Tap a folder
4. ✅ The folder should open
4. Clear the search
4. ✅ The folder should appear on the search history and tapping it open the folder
4. Go to Discover
5. Search for the same
6. ✅ Only remote podcasts should be returned, no local results and/or folders

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
@@ -1,7 +1,7 @@
 import Foundation
 import PocketCastsDataModel
 
-public struct PodcastSearchResult: Codable, Hashable {
+public struct PodcastFolderSearchResult: Codable, Hashable {
     public let uuid: String
     public let title: String
     public let author: String
@@ -32,7 +32,7 @@ public class PodcastSearchTask {
     public init(session: URLSession = .shared) {
         self.session = session
     }
-    public func search(term: String) async throws -> [PodcastSearchResult] {
+    public func search(term: String) async throws -> [PodcastFolderSearchResult] {
         let searchURL = URL(string: "\(ServerConstants.Urls.cache())discover/search")!
         var request = URLRequest(url: searchURL)
         request.httpMethod = "POST"
@@ -45,6 +45,6 @@ public class PodcastSearchTask {
 
         let (data, _) = try await session.data(for: request)
         let decoder = JSONDecoder()
-        return try decoder.decode([PodcastSearchResult].self, from: data)
+        return try decoder.decode([PodcastFolderSearchResult].self, from: data)
     }
 }
--- a/Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
@@ -5,14 +5,22 @@ public struct PodcastFolderSearchResult: Codable, Hashable {
     public let uuid: String
     public let title: String
     public let author: String
-    public let isFolder: Bool?
+    public let kind: Kind
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.uuid = try container.decode(String.self, forKey: .uuid)
+        self.title = try container.decode(String.self, forKey: .title)
+        self.author = try container.decode(String.self, forKey: .author)
+        self.kind = .podcast
+    }
 
     public init?(from podcast: Podcast) {
         if let title = podcast.title, let author = podcast.author {
             self.uuid = podcast.uuid
             self.title = title
             self.author = author
-            self.isFolder = nil
+            self.kind = .podcast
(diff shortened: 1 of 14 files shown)


Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
```diff
@@ -1,9 +1,29 @@
 import Foundation
+import PocketCastsDataModel
 
-public struct PodcastSearchResult: Codable, Hashable {
+public struct PodcastFolderSearchResult: Codable, Hashable {
```
leandroalonso, line 4: I don't like the idea of this representing a Podcast OR a Folder.

However, I tried encapsulating similar properties into a protocol and having both separated but didn't manage to succeed. Any ideas are appreciated.

podcasts/New Search/Model/SearchResultsModel.swift
```diff
@@ -34,4 +39,42 @@ class SearchResultsModel: ObservableObject {
             episodes = results ?? []
         }
     }
+
+    @MainActor
+    func searchLocally(term searchTerm: String) {
+        let allPodcasts = DataManager.sharedManager.allPodcasts(includeUnsubscribed: false)
```
emilylaguna, line 47: I think we should clear the remote search results when doing another local search since this can cause the remote episodes to stay visible.

```suggestion
        clearSearch()
        let allPodcasts = DataManager.sharedManager.allPodcasts(includeUnsubscribed: false)
```

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
emilylaguna: hmmm. I tried a few ideas but nothing really worked out. I think we could do something like this though:

```
public struct PodcastFolderSearchResult: Codable, Hashable {
    public let uuid: String
    public let title: String
    public let author: String?
    public let kind: Kind

    public init?(from podcast: Podcast) {
        guard let title = podcast.title, let author = podcast.author else {
            return nil
        }

        self.uuid = podcast.uuid
        self.title = title
        self.author = author
        self.kind = .podcast
    }

    public init?(from folder: Folder) {
        self.uuid = folder.uuid
        self.title = folder.name
        self.author = nil
        self.kind = .folder
    }

    public enum Kind: Codable {
        case podcast, folder
    }
}
```

That way we can just switch on that which cleans up some of the if/else's

emilylaguna (changes requested):
Nice work, I left a few comments. 

I tried coming up with a different solution for the podcast/folder item, but wasn't able to come up with anything better.

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
emilylaguna: So, I was thinking and since we have a lot of different kinds, I was thinking we could do something like this to generalize it even further:

```
public enum SearchItemType: String, Codable {
    case searchTerm = "search_term", episode, podcast, folder, unknown
}

protocol SearchItem {
    var kind: SearchItemType { get }
}

extension String: SearchItem {
    var kind: SearchItemType { .searchTerm }
}

extension EpisodeSearchResult: SearchItem {
    var kind: SearchItemType { .episode }
}

public struct PodcastFolderSearchResult: Codable, Hashable, SearchItem {
    ... 
    public var kind: SearchItemType
    ...
}

// Makes switching on the history easier too
struct SearchHistoryEntry: Codable, Hashable {
    var searchTerm: String?
    var episode: EpisodeSearchResult?
    var podcast: PodcastFolderSearchResult?

    var kind: SearchItemType {
        let items: [SearchItem?] = [searchTerm, episode, podcast]
        return items.compactMap { $0 }.first?.kind ?? .unknown
    }
}
```

Wdyt?

leandroalonso:
@emilylaguna thanks for the suggestion! I adapted the code to your idea.

emilylaguna (approved):
Code looks good now, I noticed a couple issues which I think we can tackle in a different PR, let me know what you think.

**Issue 1:**

If I tap on the Show All when a Folder is displayed the folder doesn't display correctly:



**Issue 2:** 

Also if I tap on the Show All while the remote results are loading I get this:



it updates correctly when the results load.

leandroalonso:
@emilylaguna I added those issues in #709

PR #774, opened 2023-03-14
Search Improvements: add tracks

leandroalonso:
| 📘 Project: #709 |
|:---:|

Add tracks to search history and search results.

### Before everything

1. Run the app
2. Go to Profile > Settings > Beta Features > enable `newSearch` and `tracksLogging`
4. Login into an account with Plus, folders and subscribed podcasts
5. Grab a coffee or tea

### Search History

Before starting that, do a few searches, and tap on podcast and folders, then:

1. Tap on the search bar in Discover or Podcasts
2. ✅ `search_history_item_tapped` should be tracked with the correct source (discover or podcast_list), type (search_term/folder/podcast/episode) and uuid (if available)
3. Make sure to tap on search term, folder, podcast and episode
4. Remove any item (by tapping the "X")
5. ✅ `search_history_item_delete_button_tapped` should be tracked with the correct source (discover or podcast_list), type (search_term/folder/podcast/episode) and uuid (if available)
3. Make sure to try on search term, folder, podcast and episode
4. Tap on "Clear all"
5. `search_history_cleared` should be tracked with correct `source` (discover or podcast_list)

### Search results

1. Tap on the search bar in Discover or Podcasts
1. ✅ `search_shown` should be tracked with correct `source` (discover or podcast_list)
1. Search for any term
1. ✅ `search_performed` should be tracked with correct `source` (discover or podcast_list)
1. Turn off wi-fi
1. Perform a search
1. ✅ `search_failed` should be tracked with correct `source` (discover or podcast_list)
1. Turn on wi-fi
1. Search for something
1. Tap "Show all" on Podcasts
1. ✅ `search_list_shown` should be tracked with correct `source` (discover or podcast_list) and `displaying: podcasts`
1. Go back, tap "Show all" for Episodes
1. ✅ `search_list_shown` should be tracked with correct `source` (discover or podcast_list) and `displaying: episodes`

### Tapping search results

For the next step, make sure to perform your search from "Podcasts" so local results and folders are returned.

1. Search for something that will return a local podcast
1. Tap the local podcast
1. ✅ `search_result_tapped` should be tracked with correct `source` (discover or podcast_list), `uuid` and `type: podcast_local_result`
1. Search and tap on a folder
1. ✅ `search_result_tapped` should be tracked with correct `source` (discover or podcast_list), `uuid` and `type: folder`
1. Search and tap on a podcast you're not subscribed to
1. ✅ `search_result_tapped` should be tracked with correct `source` (discover or podcast_list), `uuid` and `type: podcast_remote_result`
1. Search and tap on an episode
1. ✅ `search_result_tapped` should be tracked with correct `source` (discover or podcast_list), `uuid` and `type: episode`
1. Test the same by tapping on "Show all" and making sure these events are triggered there

## Checklist

### DIFF
--- a/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
+++ b/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
@@ -18,8 +18,8 @@ class SearchAnalyticsHelper: ObservableObject {
         Analytics.track(.searchPerformed, properties: ["source": source])
     }
 
-    func trackFailed() {
-        Analytics.track(.searchFailed, properties: ["source": source])
+    func trackFailed(_ error: Error) {
+        Analytics.track(.searchFailed, properties: ["source": source, "error_code": (error as NSError).code])
     }
 
     func trackResultTapped(_ searchResult: AnalyticsSearchResultItem) {
--- a/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
+++ b/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
@@ -17,6 +17,10 @@ class SearchAnalyticsHelper: ObservableObject {
         Analytics.track(.searchDismissed, properties: ["source": source])
     }
 
+    func trackSearchPerformed() {
+        Analytics.track(.searchPerformed, properties: ["source": source])
+    }
+
     // MARK: - Search History
 
     func trackHistoryCleared() {
--- a/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
+++ b/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
@@ -14,10 +14,6 @@ class SearchAnalyticsHelper: ObservableObject {
         Analytics.track(.searchShown, properties: ["source": source])
     }
 
-    func trackDismissed() {
-        Analytics.track(.searchDismissed, properties: ["source": source])
-    }
-
     func trackSearchPerformed() {
         Analytics.track(.searchPerformed, properties: ["source": source])
     }
--- a/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
+++ b/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
@@ -7,6 +7,12 @@ class SearchAnalyticsHelper: ObservableObject {
         self.source = source
     }
 
+    // MARK: - Search
+
+    func trackShown() {
+        Analytics.track(.searchShown, properties: ["source": source])
+    }
+
     // MARK: - Search History
 
     func trackHistoryCleared() {
--- a/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
+++ b/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
@@ -1,0 +1,15 @@
+import Foundation
+
+class SearchAnalyticsHelper {
+    let source: AnalyticsSource
+
+    init(source: AnalyticsSource) {
+        self.source = source
+    }
+
+    // MARK: - Search History
+
+    func trackHistoryCleared() {
+        Analytics.track(.searchHistoryCleared, properties: ["source": source])
+    }
+}
--- a/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
+++ b/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
@@ -60,7 +60,7 @@ class SearchAnalyticsHelper: ObservableObject {
 
 private extension SearchHistoryEntry {
     var type: String {
-        if podcast?.isFolder == true {
+        if podcast?.kind == .folder {
             return "folder"
         } else if podcast != nil {
             return "podcast"
@@ -74,7 +74,7 @@ private extension SearchHistoryEntry {
 
 private extension PodcastFolderSearchResult {
     var type: String {
-        if isFolder == true {
+        if kind == .folder {
             return "folder"
         } else if isLocal == true {
             return "podcast_local_result"
--- a/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
+++ b/podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
@@ -22,17 +22,8 @@ class SearchAnalyticsHelper: ObservableObject {
         Analytics.track(.searchFailed, properties: ["source": source])
     }
 
-    func trackResultTapped(_ searchResult: Any) {
-        var type = "unknown"
-        var uuid = ""
-        if let folderOrPodcast = searchResult as? PodcastFolderSearchResult {
-            type = folderOrPodcast.type
-            uuid = folderOrPodcast.uuid
-        } else if let episode = searchResult as? EpisodeSearchResult {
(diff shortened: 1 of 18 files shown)


leandroalonso:
@ashiagr FYI: I added a `search_list_shown` that is fired when "Show all" is tapped.

podcasts/New Search/Model/SearchResultsModel.swift
```diff
@@ -14,6 +16,16 @@ class SearchResultsModel: ObservableObject {
 
     @Published var isShowingLocalResultsOnly = false
 
+    init(analyticsHelper: SearchAnalyticsHelper) {
+        self.analyticsHelper = analyticsHelper
+    }
+
+    #if DEBUG
```
emilylaguna: Debug?

podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
```diff
@@ -0,0 +1,85 @@
...
+        var uuid = ""
+        if let folderOrPodcast = searchResult as? PodcastFolderSearchResult {
+            type = folderOrPodcast.type
+            uuid = folderOrPodcast.uuid
+        } else if let episode = searchResult as? EpisodeSearchResult {
+            type = "episode"
+            uuid = episode.uuid
+        }
+        Analytics.track(.searchResultTapped, properties: ["source": source, "uuid": uuid, "result_type": type])
+    }
```
emilylaguna: What do you think about specifying a type here?

We could do something like this:

```
protocol AnalyticsSearchResultItem: AnalyticsDescribable {
    var uuid: String { get }
}

extension EpisodeSearchResult: AnalyticsSearchResultItem {
    var analyticsDescription: String {
        "episode"
    }
}

extension PodcastFolderSearchResult: AnalyticsSearchResultItem {
    var analyticsDescription: String {
        if isFolder == true {
            return "folder"
        } else if isLocal == true {
            return "podcast_local_result"
        } else {
            return "podcast_remote_result"
        }
    }
}
```

Which would change this to: 
```suggestion
    func trackResultTapped(_ searchResult: AnalyticsSearchResultItem) {
        Analytics.track(.searchResultTapped, properties: ["source": source, "uuid": searchResult.uuid, "result_type": searchResult])
    }
```

podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    // MARK: - Search History
+
+    func trackHistoryCleared() {
+        Analytics.track(.searchHistoryCleared, properties: ["source": source])
+    }
+
+    func historyItemTapped(_ entry: SearchHistoryEntry) {
+        let uuid = entry.podcast?.uuid ?? entry.episode?.uuid ?? ""
+        Analytics.track(.searchHistoryItemTapped, properties: ["source": source, "uuid": uuid, "type": entry.type])
```
emilylaguna: We could use the `AnalyticsSearchResultItem` protocol here too:

```
extension SearchHistoryEntry: AnalyticsSearchResultItem {
    var uuid: String {
        podcast?.uuid ?? episode?.uuid ?? ""
    }

    var analyticsDescription: String {
       ...
    }
}
```


```suggestion
        Analytics.track(.searchHistoryItemTapped, properties: ["source": source, "uuid": entry.uuid, "type": entry])
```
```

**Review by emilylaguna — CHANGES_REQUESTED** [review]
I didn't get a chance to test the events yet, but I left a few comments below. 

Let me know what you think.

File: podcasts/New Search/Model/SearchResultsModel.swift (hunk 1 above)
Comment (leandroalonso) [reply podcasts/New Search/Model/SearchResultsModel.swift]: This is to avoid having to add the source to every preview. However, I changed it to be just a default constructor param in 9d4cc45b1cfab1f6c723f17e6ad4763d6f47f33a

**Comment by leandroalonso** [comment]
@emilylaguna thanks for your suggestions! Applied all of them, way better. :)

File: podcasts/New Search/Helpers/SearchAnalyticsHelper.swift [hunk 4]
```diff
@@ -0,0 +1,80 @@
...
+    func trackShown() {
+        Analytics.track(.searchShown, properties: ["source": source])
+    }
+
+    func trackSearchPerformed() {
+        Analytics.track(.searchPerformed, properties: ["source": source])
+    }
+
+    func trackFailed() {
+        Analytics.track(.searchFailed, properties: ["source": source])
```
emilylaguna: What do you think about also tracking the reason the search failed?

In other areas we pass an error_code into the event.

podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
```diff
@@ -0,0 +1,80 @@
...
+        self.source = source
+    }
+
+    // MARK: - Search
+
+    func trackShown() {
+        Analytics.track(.searchShown, properties: ["source": source])
+    }
+
+    func trackSearchPerformed() {
```
emilylaguna, line 17: I feel like the timing of this event is a bit off when searching locally. I expected that when I typed in my search I would get a search_performed event after I was done typing, but instead it's not firing until the remote search is performed a few seconds later. 

Thoughts?

podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
```diff
@@ -0,0 +1,80 @@
...
+        Analytics.track(.searchPerformed, properties: ["source": source])
+    }
+
+    func trackFailed() {
```
emilylaguna: When I disabled my network and perform a search I'm getting 2 logs for search_failed

emilylaguna (changes requested):
All the events are being tracked as noted, I left a few comments about some timing and double tracking. Let me know what you think.

podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
leandroalonso: I agree.

However, this is how it works with the current search. So I decided to keep the behavior. I know we can improve that but maybe it would be a "nice to have"?

podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
leandroalonso: Ha! Yes.

One is the podcast's failure, the other is the episode's failure. I wonder whether would be correct to trigger just once, given maybe one might fail, and the other not.

Thoughts?

podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
leandroalonso: Done in c592e930df1988635ba97aedbcf3ee55b9a498dd

podcasts/New Search/Helpers/SearchAnalyticsHelper.swift
emilylaguna: hmm. I have no opinions on this, lets leave it as it is and see how it goes.

emilylaguna (approved):

PR #782, opened 2023-03-18
Ratings: Pulls ratings in from the API and displays them if they're available

emilylaguna:
## Description
This integrates with the ratings API to display ratings for podcasts if they're available. 

I also added a view helper items:
1. A `ThemedHostingController` that lets you wrap Swift UI views with the theme environment setup
2. A `ModifedHostingController` which lets you modify the content of a SwiftUI view and pass it easily to a hosting controller. This lets you set custom environments and do other modifications
3. A View extension that maps a SwiftUI view to a UIView
4. A generic JSONDecodableURLTask that lets you convert basic JSON responses without needing to implement all the network code again

## Screenshots
| Example | Example | Example |
|:---:|:---:|:---:|
||||

## To test

1. Launch the app
5. Navigate to a podcast
6. ✅ Verify nothing displays and the app doesn't crash
7. Enable the showRatings feature flag
8. Navigate to a podcast again, ✅ Verify nothing displays and the app doesn't crash
9. Since the API isn't available yet you can do the following to test:
10. Open RetrievePodcastRatingTask
11. Replace the `endpoint` contents with `"https://podcast.emily.tools/ratings/test.php?total=1234&average=4.32"`
12. This will generate the same response but without loading from the db
13. You can change the total and average numbers to get different values
14. ✅ Verify the star rating appears

## Checklist


pocketcasts:
### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
PCHostingController.swift | 6 | Lines should not have trailing whitespace. (trailing_whitespace)
JSONDecodableURLTask.swift | 15 | Lines should not have trailing whitespace. (trailing_whitespace)


</p>

podcasts/PCHostingController.swift
```diff
@@ -47,8 +90,15 @@ class PCHostingController<Content>: UIHostingController<Content> where Content:
...
+    func body(content: Content) -> some View {
+        content.environmentObject(theme)
+    }
+}
+
 extension View {
-    func setupDefaultEnvironment(theme: Theme = Theme.sharedTheme) -> some View {
-        self.environmentObject(theme)
+     func setupDefaultEnvironment(theme: Theme = Theme.sharedTheme) -> some View {
+         self.modifier(ThemedEnvironment(theme: theme))
```
emilylaguna: I updated this to keep it consistent

leandroalonso (commented):
Tested and it works as described! However, I faced an issue: the starts don't display if I'm loading the podcast for the first time. To reproduce:

1. Go to Discover
2. Tap on any podcast you haven't loaded previously
3. 💥 Check that starts don't load

Also, wdyt about showing a faded version of the stars while loading to avoid the view "jumping"?

Also, maybe is already on your to-do list but we might want to cache the result. Of course, we can do it in another PR.

Edit: Curiously, I'm not being able to reproduce the stars not appearing issue anymore. 🤔

leandroalonso:
@emilylaguna wdyt about tracking taps on the stars? I think this will likely give us insights on users expecting to see reviews, so we can use this data to drive decisions in the future.

emilylaguna:
> However, I faced an issue: the starts don't display if I'm loading the podcast for the first time

Thanks for catching that! It should be fixed now. 

> Also, wdyt about showing a faded version of the stars while loading to avoid the view "jumping"?

I updated how this displays now, let me know what you think. 

> Also, maybe is already on your to-do list but we might want to cache the result. Of course, we can do it in another PR.

This actually already caches the values in the network cache. The ratings API server specifies the cache duration that we respect. This lets the server control how often we should refresh and reduces the hits all in one. 

This is why after the initial load the ratings just appeared again. 

> wdyt about tracking taps on the stars

Done! 55c2fdf7f0e452afd0b98652d6640241f3855d86

emilylaguna:

I ended up restructuring this a bit more so it follows more of the MVVM pattern by adding a View Model and having the star rating view use that instead of just having it all in a view controller. 

I also moved the code into the `ratings` group.

leandroalonso (approved):
@emilylaguna :shipit: 

I have just one small comment, not a blocker though: while loading the stars, wdyt about adding some opacity to the empty stars? Or changing the colors to grey? For a brief moment, I thought some podcasts had zero stars (like no ratings), and I feel our users might feel the same.

emilylaguna:
I played around that and a few other ideas but it still felt a bit off. I'll leave it as it is for now and will explore a different idea in a future pr.

PR #785, opened 2023-03-22
Adjust multi-selection header height based on screen size

### LINKED ISSUE #46: Multi-selection: spacing issues
Sent by [a user in the Beta Slack](https://pocketcastsbeta.slack.com/archives/C0GKZ8J4A/p1653658044750189):

In "Select Episodes", the header bar ("Select All ..... Cancel") is overlapping the Search bar.

Version 7.20

mrpolanco:
Fixes #46 

When selecting multiple episodes, the multi-select header overlaps the Search bar on smaller screens (under 380 points). Sample flow for iPhone SE 2nd Gen:

Test flow - iPhone SE (2nd generation)

This PR sets constraint height for multi-select header, based on the size of the display, so that the header does not overlap the search bar. It keeps the constraint height the same as previously set for larger screens.

## To test
Using the workflow above:
1. Open a podcast using a device using a small screen (anything under 380pts width, as [indicated in this chart](https://www.ios-resolution.com/)).
2. Click on Subscribe, if necessary, to view the episode list.
3. Tap the button next to the Search bar to bring up the menu.
4. Tap "Select Episodes" to bring up the multi-select header.
5. Confirm that the header no longer overlaps the Search bar.

Tested on small screens:
- iPod Touch 7th Gen (320pts)
- iPhone SE 2nd Gen (375pts)

iPod touch - SE 2nd Gen

Confirming that the Search bar remains visible on larger screens:
- iPhone 12 Pro (390pts)
- iPhone 14 Pro Max (430pts)

iPhone 12 Pro - 14 Pro Max

## Checklist

- [✅] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [✅] I have considered adding unit tests for my changes.
- [✅] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.


podcasts/PodcastViewController.swift
```diff
@@ -149,6 +152,10 @@ class PodcastViewController: FakeNavViewController, PodcastActionsDelegate, Sync
     @IBOutlet var multiSelectHeaderView: ThemeableView!
     private let operationQueue = OperationQueue()
 
+    // Constraint to adjust multiSelectHeader based on device size
+    @IBOutlet weak var multiSelectHeaderViewConstraint: NSLayoutConstraint!
+    private let screenWidth = UIScreen.main.bounds.width
```
leandroalonso: This feels out of place here.

Instead of having this here, a better idea would be to extract the `self.screenWidth < 380.0 ? 65.0 : 90.0` logic to a method, keeping `UIScreen.main.bounds.width` inside it.

leandroalonso (changes requested):
@mrpolanco on iPad the issue is still happening:

podcasts/PodcastViewController.swift
mrpolanco: Extracted logic to a method to better cover current and future edge cases.

Tested on smaller devices:
- iPod Touch (320)
- iPhone SE (375)

320 - 375 (iPod - Iphone SE)

Tested on iPads:
- iPad Mini 4 (768)
- iPad Pro 6th Gen (1024)

iPad 768 - 1024

leandroalonso (approved):
@mrpolanco we're good to go!

Before we merge though, can you please update the `CHANGELOG.md` on this PR? No need to open another one, you can do it on this one. :)

leandroalonso:
@mrpolanco JFYI, I fixed the merge conflict here and moved your bug fixes description to `7.36` (which is the version they'll land to).

When in doubt, always check the Milestones, you'll be able to see in which version your changes will land. The version with the ❄️ icon is "frozen", which means that it only accepts critical bug fixes. Again, thanks for contributing to Pocket Casts. :)

mrpolanco:
@leandroalonso Thank you for fixing the merge conflict 🌟

PR #788, opened 2023-03-28
Fix Auto Add to Up Next reset bug

### LINKED ISSUE #712: Auto Add to Up Next: All change "To Bottom" in some cases
### Expected behavior

Your custom "To Top" setting should remain intact

### Actual behavior

They all are reset to "To Bottom"

Initially, it was a bit tricky to replicate, but here is what I did: 

- Under 'xxxx' Podcasts Chosen, selected all podcasts
- Deselected the first podcast
- Went back to the 'Auto Add Podcasts' page, and set a few of them to use "To Top"
- Switched back to the 'xxxx' Podcasts Chosen page, and selected the first podcast
- Upon going back, all of the podcasts were now using the "To Bottom" option 

Recording of how to replicate: 

https://user-images.githubusercontent.com/16253818/218732707-eedf8b69-b57d-4755-a33c-3589abd47aa4.MP4

##### Tested on [iPhone 14 Pro Max], iOS [16.3], Pocket Casts [7.32]

5945768-zd-woothemes

mrpolanco:
Fixes #712 

When using the "Auto Add to Up Next" feature, there's a bug that occurs when choosing the Auto Add behavior. By default, podcasts selected for "Auto Add to Up Next" are podcasts at the bottom of the Up Next list. Users can change this behavior and choose podcasts to be added to the top of the Up Next list.

Users can also add all their podcasts to the Up Next lists using the Select All feature. The bug occurs when a user has all their podcasts selected to be added to Up Next. If the user deselects and re-selects a podcast on the list, all podcasts revert to the default configuration, which is to be added to the bottom of the Up Next list. 

This PR addresses this issue by keeping the individual podcasts as configured by the user, even when selecting and deselecting individual podcasts.

## To test

1. Subscribe to a few podcasts.
2. Tap Profile, then the Gear icon to go to Settings.
3. Tap Auto Add to Up Next menu
4. Tap the "0 Podcasts Chosen" menu
5. Select All podcasts and return to the previous menu
6. Change individual settings for a few podcasts to Add to Top
7. Tap the "# Podcasts Chosen" menu

Flow - iPhone SE (2nd generation-side

8. Tap to deselect one podcast from the list, then tap to select again - on the current version, doing so resets all podcasts to Add to Bottom
9. Verify that this PR allows settings to remain

Flow 2 - iPhone SE (2nd generation)

Note: The podcast that was deselected and reselected will be reset to Add to Bottom. This is intended behavior when an unselected podcast is first selected.

## Checklist

- [✅] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [✅] I have considered adding unit tests for my changes.
- [✅] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

### DIFF
--- a/podcasts/AutoAddToUpNextViewController+PodcastSelect.swift
+++ b/podcasts/AutoAddToUpNextViewController+PodcastSelect.swift
@@ -9,12 +9,11 @@ extension AutoAddToUpNextViewController: PodcastSelectionDelegate {
 
         if selected {
             // Checks for existing AutoAddToUpNextSetting value before assigning a default value
-            allPodcasts.forEach {podcast in
-                if podcast.autoAddToUpNext == 0 {
-                    let status = AutoAddToUpNextSetting.addLast.rawValue
-                    DataManager.sharedManager.saveAutoAddToUpNext(podcastUuid: podcast.uuid, autoAddToUpNext: status)
-                }
+            let podcasts = allPodcasts.filter {
+                $0.isSubscribed() && $0.autoAddToUpNext == AutoAddToUpNextSetting.off.rawValue
             }
+
+            DataManager.sharedManager.updateAutoAddToUpNext(to: .addLast, for: podcasts)
         } else {
             setting = AutoAddToUpNextSetting.off.rawValue
             DataManager.sharedManager.saveAutoAddToUpNextForAllPodcasts(autoAddToUpNext: setting)
--- a/podcasts/AutoAddToUpNextViewController+PodcastSelect.swift
+++ b/podcasts/AutoAddToUpNextViewController+PodcastSelect.swift
@@ -3,9 +3,23 @@ import PocketCastsDataModel
 
 extension AutoAddToUpNextViewController: PodcastSelectionDelegate {
     func bulkSelectionChange(selected: Bool) {
-        let setting = selected ? AutoAddToUpNextSetting.addLast.rawValue : AutoAddToUpNextSetting.off.rawValue
-        DataManager.sharedManager.saveAutoAddToUpNextForAllPodcasts(autoAddToUpNext: setting)
+        var setting = Int32()
+
         let allPodcasts = DataManager.sharedManager.allPodcasts(includeUnsubscribed: false)
+
+        if selected {
+            // Checks for existing AutoAddToUpNextSetting value before assigning a default value
+            allPodcasts.forEach {podcast in
+                if podcast.autoAddToUpNext == 0 {
+                    let status = AutoAddToUpNextSetting.addLast.rawValue
+                    DataManager.sharedManager.saveAutoAddToUpNext(podcastUuid: podcast.uuid, autoAddToUpNext: status)
+                }
+            }
+        } else {
+            setting = AutoAddToUpNextSetting.off.rawValue
+            DataManager.sharedManager.saveAutoAddToUpNextForAllPodcasts(autoAddToUpNext: setting)
+        }
+
         allPodcasts.forEach { NotificationCenter.postOnMainThread(notification: Constants.Notifications.podcastUpdated, object: $0.uuid) }
 
         reloadDownloadedPodcasts()
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
@@ -409,6 +409,31 @@ class PodcastDataManager {
         setOnAllPodcasts(value: autoAddToUpNext, propertyName: "autoAddToUpNext", subscribedOnly: true, dbQueue: dbQueue)
     }
 
+
+    /// TODO: Documentation
+    /// - Parameters:
+    ///   - value: <#value description#>
+    ///   - podcasts: <#podcasts description#>
+    ///   - dbQueue: <#dbQueue description#>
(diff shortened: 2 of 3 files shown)


podcasts/AutoAddToUpNextViewController+PodcastSelect.swift
```diff
@@ -3,9 +3,23 @@ import PocketCastsDataModel
...
+            // Checks for existing AutoAddToUpNextSetting value before assigning a default value
+            allPodcasts.forEach {podcast in
+                if podcast.autoAddToUpNext == 0 {
+                    let status = AutoAddToUpNextSetting.addLast.rawValue
+                    DataManager.sharedManager.saveAutoAddToUpNext(podcastUuid: podcast.uuid, autoAddToUpNext: status)
+                }
+            }
+        } else {
+            setting = AutoAddToUpNextSetting.off.rawValue
+            DataManager.sharedManager.saveAutoAddToUpNextForAllPodcasts(autoAddToUpNext: setting)
```
emilylaguna, line 19: The setting var only looks to be used here wdyt about removing it?

```suggestion
            DataManager.sharedManager.saveAutoAddToUpNextForAllPodcasts(autoAddToUpNext: AutoAddToUpNextSetting.off.rawValue)
```

podcasts/AutoAddToUpNextViewController+PodcastSelect.swift
```diff
@@ -3,9 +3,23 @@ import PocketCastsDataModel
...
         let allPodcasts = DataManager.sharedManager.allPodcasts(includeUnsubscribed: false)
+
+        if selected {
+            // Checks for existing AutoAddToUpNextSetting value before assigning a default value
+            allPodcasts.forEach {podcast in
+                if podcast.autoAddToUpNext == 0 {
+                    let status = AutoAddToUpNextSetting.addLast.rawValue
+                    DataManager.sharedManager.saveAutoAddToUpNext(podcastUuid: podcast.uuid, autoAddToUpNext: status)
+                }
+            }
```
emilylaguna: Sorry for the long comment. 

In my testing I found this causes a considerable lag when running on databases with lots of podcasts. I currently have 200 that I'm subscribed to and my phone hangs for about a second when running through this. 

Did some digging on why and since we're looping through the podcasts and calling `DataManager.sharedManager.saveAutoAddToUpNext` for each of them results in:

- Making a [DB call to update](https://github.com/Automattic/pocket-casts-ios/blob/cef84ace4cb45ad3d9fd39fc27f87c50ca3af8e9/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift#L472) the saved value
- Making another [DB call to refresh](https://github.com/Automattic/pocket-casts-ios/blob/cef84ace4cb45ad3d9fd39fc27f87c50ca3af8e9/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift#L479-L500) our internal podcast cache which includes ALL the podcasts stored in the database, not just subscribed ones
- The update cache call also needs to update the internal cache array using a [synchronous call](https://github.com/Automattic/pocket-casts-ios/blob/cef84ace4cb45ad3d9fd39fc27f87c50ca3af8e9/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift#L493) which adds extra time the rest of the queries 
- The update cache call needs to convert all the DB values to Podcast objects again

If we run the code through the Time Profiler we can see the breakdown of the process, and see the majority of the time spent is just in the rebuilding of the cache and converting the DB values to Podcast objects again. 


Screen Shot 2023-03-29 at 2 14 56 PM

emilylaguna: I think based on what we know from the above we can reduce this down to 2 total queries by moving the updating of all the podcasts from the client to the database by:

1. Adding a new method to the PodcastDataManager similar to `saveAutoAddToUpNextForAllPodcasts` but instead takes a podcasts array and passes their `UUID`'s in the WHERE query
2. Updating this to filter the podcasts similar to what we have now but adding an `.isSubscribed()` check
3. Passing the podcasts to the new method

I made a quick idea branch demonstrating this here: https://github.com/Automattic/pocket-casts-ios/compare/issue/712_Auto_Add_to_Up_Next_bug...idea/712-auto-add-reset-bug

Let me know what you think.

emilylaguna (changes requested):
Hey @mrpolanco, nice work on identifying and fixing this bug! 

I left some comments below on how we can improve the performance of the update process a bit. 

I think it would also be good to add a unit test for this process, what do you think?

podcasts/AutoAddToUpNextViewController+PodcastSelect.swift
mrpolanco: Thought about readability, but you're right, it can be done away with :)

podcasts/AutoAddToUpNextViewController+PodcastSelect.swift
mrpolanco: Tested the proposed method on this branch by adding 250 podcasts and using the Time Profiler to test db queries. It worked splendidly, reducing cache rebuild time from 494.0 ms to 2.0 ms 🙌🏼

**Initial reset bug fix:**

Initial reset bug fix

**Bug fix refactor:**

Bug fix refactor

PR #795, opened 2023-03-30
Search Improvements: fix folders updating, history saving and duplicated podcasts

leandroalonso:
| 📘 Project: #709 |
|:---:|

This PR fixes 3 issues:

1. Search history saving each letter of the search term
2. Folders updating and/or subscription expiring
3. Duplicated podcasts appearing when searching locally

### Before everything

1. Run the app

### Duplicated podcasts

1. Search for something like "hello"
1. Subscribe to the first podcast
1. Go to "Podcasts"
1. Search for "hello" again
1. ✅  The podcast you subscribed should not appear twice

### Search history

To behave like apps such as Instagram/Twitter/Facebook I changed how a search term is saved: it only happens if the user taps the "Search" button.

1. Search for something and do not tap "Search" (or, if in the simulator, do not type "return")
1. Dismiss the results
1. ✅ The term should not be added to the list
1. Repeat the search but this time type "Search" on the keyboard (or, if in the simulator, type "return" on your keyboard)
1. ✅ The term should be added to the history

### Folders

1. Make sure you're logged in to a Plus account
1. Create two folders
1. Search for those folders and tap on them, so they get added to the search history
1. Go to "Podcasts" and change the name of one folder
1. Open the search again
1. ✅ The folder should have the updated name
1. Go back to "Podcasts"
1. Delete this folder
1. Open the search
1. ✅ The folder should not be on the history
1. Go to the users web console and remove your Plus status (gift yourself `0.5` day)
1. Go back to the app and refresh it
1. Open the search history
1. ✅ Folders should not be on the history

## Checklist

### DIFF
--- a/podcasts/New Search/Model/SearchHistoryModel.swift
+++ b/podcasts/New Search/Model/SearchHistoryModel.swift
@@ -1,6 +1,7 @@
 import SwiftUI
 import PocketCastsServer
 import PocketCastsDataModel
+import Combine
 
 struct SearchHistoryEntry: Codable, Hashable {
     var searchTerm: String?
@@ -16,6 +17,8 @@ class SearchHistoryModel: ObservableObject {
     private let defaults: UserDefaults
     private let maxNumberOfEntries = 20
 
+    private var notifications = Set<AnyCancellable>()
+
     init(userDefaults: UserDefaults = UserDefaults.standard) {
         self.defaults = userDefaults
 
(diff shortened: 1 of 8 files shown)


podcasts/New Search/Model/SearchHistoryModel.swift
```diff
@@ -21,6 +22,48 @@ class SearchHistoryModel: ObservableObject {
...
+        NotificationCenter.default.addObserver(self, selector: #selector(updateFolders), name: Constants.Notifications.folderChanged, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(folderDeleted), name: Constants.Notifications.folderDeleted, object: nil)
+
+        updateFolders()
+    }
+
+    @objc private func updateFolders() {
+        guard SubscriptionHelper.hasActiveSubscription() else {
+            // User is not subscribed anymore, remove all folders from search history
+            DispatchQueue.main.async { [weak self] in
```
leandroalonso: When refreshing and losing the Plus status, this part is triggering this warning:

> Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.

I then changed this class to use `@MainActor` but the warning still persist. When dispatching on the main queue it goes away. 🤷

emilylaguna (approved):
🎉 Nice work!!

PR #803, opened 2023-04-04
Address some existing warnings and prevent future Xcode 14.3 ones 

mokagio:
This PR preemptively addresses code that will generated warnings in Xcode 14.3. While at it, I found a few more warnings which I don't think are new in Xcode 14.3 (but I didn't check by building in an older Xcode) and tackled them, too.

## To test
If CI is green, we should be good.

One thing that might be worth checking is image downloading. I had to update Kingfisher to the latest version to avoid a build failure in Xcode 14.3.

## Checklist

### DIFF
--- a/Pocket Casts Watch App Extension/EpisodeView.swift
+++ b/Pocket Casts Watch App Extension/EpisodeView.swift
@@ -94,8 +94,8 @@ struct EpisodeView: View {
                 }
             }
 
-            if viewModel.supportsPodcastNavigation, let podcast = viewModel.parentPodcast, let viewModel = PodcastEpisodeListViewModel(podcast: podcast) {
-                NavigationLink(destination: PodcastEpisodeListView(viewModel: viewModel)) {
+            if viewModel.supportsPodcastNavigation, let podcast = viewModel.parentPodcast  {
+                NavigationLink(destination: PodcastEpisodeListView(viewModel: .init(podcast: podcast))) {
                     EpisodeActionView(iconName: "episode_goto", title: L10n.goToPodcast)
                 }
             }
--- a/Pocket Casts Watch App Extension/EpisodeView.swift
+++ b/Pocket Casts Watch App Extension/EpisodeView.swift
@@ -94,7 +94,7 @@ struct EpisodeView: View {
                 }
             }
 
-            if viewModel.supportsPodcastNavigation, let podcast = viewModel.parentPodcast  {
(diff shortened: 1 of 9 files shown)


Pocket Casts Watch App Extension/EpisodeView.swift
```diff
@@ -94,8 +94,8 @@ struct EpisodeView: View {
                 }
             }
 
-            if viewModel.supportsPodcastNavigation, let podcast = viewModel.parentPodcast, let viewModel = PodcastEpisodeListViewModel(podcast: podcast) {
-                NavigationLink(destination: PodcastEpisodeListView(viewModel: viewModel)) {
+            if viewModel.supportsPodcastNavigation, let podcast = viewModel.parentPodcast {
+                NavigationLink(destination: PodcastEpisodeListView(viewModel: .init(podcast: podcast))) {
```
mokagio, line 98: `let viewModel = PodcastEpisodeListViewModel(podcast: podcast)` did not produce an `Optional`. Xcode 14.3 doesn't allow this inconsistent syntax to build.

WidgetExtension/Up Next/UpNextMediumWidgetView.swift
```diff
@@ -71,13 +71,11 @@ struct MediumFilterView: View {
...
-                            .font(.callout)
-                            .fontWeight(.regular)
-                            .foregroundColor(Color.secondary)
-                            .frame(height: 18)
-                    }
+                    Text(filterName)
+                        .font(.callout)
+                        .fontWeight(.regular)
+                        .foregroundColor(Color.secondary)
+                        .frame(height: 18)
```
mokagio, line 78: Similar to the view model above, `filterName` is not `Optional` so we don't need the optional binding.

podcasts/DownloadManager.swift
```diff
@@ -168,7 +168,6 @@ class DownloadManager: NSObject, FilePathProtocol {
         if episode.bufferedForStreaming(), autoDownloadStatus != AutoDownloadStatus.playerDownloadedForStreaming {
             let sourceUrl = URL(fileURLWithPath: streamingBufferPathForEpisode(episode))
             let destinationUrl = URL(fileURLWithPath: pathForEpisode(episode))
-            let fileManager = FileManager.default
```
mokagio, line 171: Unused.

leandroalonso (approved):
Tested Kingfisher by installing this version over an existing one and it all works just fine. Thanks @mokagio

PR #805, opened 2023-04-04
Search Improvements: design tweaks

leandroalonso:
Fixes some of the issues mentioned here: https://github.com/Automattic/pocket-casts-ios/issues/709#issuecomment-1495723140

* Updated icon
* Updated shadows
* Updated leading/trailing space
* Increased podcast covers

https://user-images.githubusercontent.com/7040243/229880523-9f9b26fa-86e8-49d1-a6db-06a2f99fb4fc.mp4

## To test

1. Run the app
2. Search for podcasts, folders, or episodes
3. ✅ Check the big covers and the small ones
4. Tap on "Show all" in both podcasts and episodes
5. ✅ Check that the covers appear just fine with a lighter shadow

## Checklist


podcasts/New Search/Views/History/SearchHistoryCell.swift
```diff
@@ -53,15 +53,18 @@ struct SearchHistoryCell: View {
...
                         } else {
-                            PodcastCover(podcastUuid: uuid)
-                                .frame(width: 48, height: 48)
+                            PodcastImage(uuid: uuid)
+                                .cornerRadius(4)
+                                .frame(width: 56, height: 56)
                                 .allowsHitTesting(false)
                                 .padding(.trailing, 12)
+                                .shadow(radius: 3, x: 0, y: 1)
                         }
```
emilylaguna: Since this code is duplicated in a few places I wonder if we could make a common component so we don't need to change the values in 4 places?

```
struct SearchEntryImage: View {
    let uuid: String
    let kind: PodcastFolderSearchResult.Kind?
    
    var body: some View {
        image
            .frame(width: 56, height: 56)
            .cornerRadius(4)
            .shadow(radius: 3, x: 0, y: 1)
            .allowsHitTesting(false)
    }
    
    @ViewBuilder
    private var image: some View {
        if kind == .folder {
            SearchFolderPreviewWrapper(uuid: uuid)
        } else {
            PodcastImage(uuid: uuid)
        }
    }
}
```

Then here we just need to do:

```
                        SearchEntryImage(uuid: uuid, kind: entry.podcast?.kind)
                            .padding(.trailing, 12)
```

podcasts/New Search/Views/Results/SearchResultCell.swift
```diff
@@ -33,13 +33,16 @@ struct SearchResultCell: View {
...
+                            .cornerRadius(4)
+                            .shadow(radius: 3, x: 0, y: 1)
                             .allowsHitTesting(false)
                     } else if let result {
                         SearchFolderPreviewWrapper(uuid: result.uuid)
-                            .frame(width: 48, height: 48)
+                            .frame(width: 56, height: 56)
                             .cornerRadius(4)
+                            .shadow(radius: 3, x: 0, y: 1)
                             .allowsHitTesting(false)
```
emilylaguna: If we decide to go with the common view we can replace the code here with something like: 

```
                    (episode?.podcastUuid ?? result?.uuid).map {
                        SearchEntryImage(uuid: $0, kind: result?.kind)
                    }
```

podcasts/PodcastImage.swift
```diff
@@ -0,0 +1,20 @@
+import SwiftUI
+import Kingfisher
+import PocketCastsServer
+
+struct PodcastImage: View {
+    let uuid: String
+    let size: Int
+
+    init(uuid: String, size: Int = 280) {
```
emilylaguna: Should we have this take a `PodcastThumbnailSize` instead of manually specifying the size?

```suggestion
    init(uuid: String, size: Int = 280) {
```

podcasts/PodcastImage.swift
```diff
@@ -0,0 +1,20 @@
...
+        self.uuid = uuid
+        self.size = size
+    }
+
+    var body: some View {
+        KFImage(ServerHelper.imageUrl(podcastUuid: uuid, size: size))
```
emilylaguna: If we use the podcast image size enum we can use the ImageManager to get the url:
```suggestion
        KFImage(ImageManager.sharedManager.podcastUrl(imageSize: size, uuid: uuid))
```

podcasts/New Search/Views/History/SearchHistoryCell.swift
```diff
@@ -76,7 +79,7 @@ struct SearchHistoryCell: View {
...
                         .allowsHitTesting(false)
                     } else if let searchTerm = entry.searchTerm {
-                        Image("custom_search")
+                        Image("search")
                             .frame(width: 48, height: 48)
```
emilylaguna: Should this be set to `.frame(width: 56, height: 56)`  as well?

emilylaguna (changes requested):
Looking good! I left some comments about reducing some duplicated code.

podcasts/New Search/Views/History/SearchHistoryCell.swift
leandroalonso: Done in f6d977fcdc03ed9200c138469a2fec491959e3fd

podcasts/New Search/Views/Results/SearchResultCell.swift
leandroalonso: Done in f6d977fcdc03ed9200c138469a2fec491959e3fd

podcasts/PodcastImage.swift
leandroalonso: Done in c3adc37cfc97454a4815fa92f1404c52678a2f5a

podcasts/PodcastImage.swift
leandroalonso: Done in c3adc37cfc97454a4815fa92f1404c52678a2f5a

podcasts/New Search/Views/History/SearchHistoryCell.swift
leandroalonso: Done in 0386e3ba2e429abc5e34040b18e082b78652e58b

emilylaguna (approved):
🚢

PR #810, opened 2023-04-07
Make author and title optional to avoid decoding issues

leandroalonso:
Fixes search not returning results in case author is not available.

## To test

1. Run the app
2. Search for "Knifecast"
3. ✅ Podcasts results should appear

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
@@ -23,7 +23,7 @@ public struct PodcastFolderSearchResult: Codable, Hashable {
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         self.uuid = try container.decode(String.self, forKey: .uuid)
-        self.title = try container.decode(String.self, forKey: .title)
+        self.title = try? container.decode(String.self, forKey: .title)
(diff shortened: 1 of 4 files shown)


Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
```diff
@@ -71,7 +71,7 @@ public class PodcastSearchTask {
         let (data, _) = try await session.data(for: request!)
         let decoder = JSONDecoder()
         decoder.keyDecodingStrategy = .convertFromSnakeCase
-        let envelope = try decoder.decode(PodcastsSearchEnvelope.self, from: data)
+        let envelope = try! decoder.decode(PodcastsSearchEnvelope.self, from: data)
```
emilylaguna:
```suggestion
        let envelope = try decoder.decode(PodcastsSearchEnvelope.self, from: data)
```

emilylaguna (changes requested):
This fixes the results not appearing, however this got me curious so I did a check and it is possible for title to be nil as well. 

Can we add that check as well?

leandroalonso:
@emilylaguna changed `title` to be optional too.

emilylaguna (approved):
🚢

geekygecko:
Can we help on the server by always returning an empty string in the author field for now?

leandroalonso:
@geekygecko This only affects the new search, so there is no need to do that (the previous search works just fine!)

PR #811, opened 2023-04-07
Updates the podcast results scrollview on the new search to use the new HorizontalCarousel

emilylaguna:
This updates the search carousel with the new carousel and also makes some improvements to the HorizontalCarousel:
- Adds lazy loading of the items to prevent performance issues
- Allows swiping to a full next page instead of just to the next item

This also makes some other changes:
- Changes the dismiss keyboard gesture to a `simultaneousGesture` to prevent it from eating the carousels gesture
- Adds iPad portrait and landscape dynamic number of carousel items

I also made a helper extension to convert a `NSNotification.Name` to a publisher without the extra boilerplate

Also, I'll need to cherry pick this into `release/7.36` but am targeting trunk since the carousel was already there.

## Demo Video

https://user-images.githubusercontent.com/793774/230679351-c67a4941-76de-4015-a9e4-1a5edd3bc46f.mov

## To test

1. Launch the app
2. Search for some podcasts
3. ✅ Verify the search appears and looks good
4. Swipe between pages
5. ✅ Verify it goes to the next page
6. Swipe back, ✅ Verify it works
7. Swipe to the end
8. ✅ Verify that works
9. Search for something that will return a single result
   - I used local search for a single podcast item
11. ✅ Verify this also works
12. Tap on a podcast and ✅ verify it opens
13. With the keyboard open swipe and ✅ verify the board dismisses
14. Test both on the Podcast tab and Discover tab
15. Test on iPad Portrait and Landscape, iPhone SE and  iOS 15.5

## Checklist

### DIFF
--- a/podcasts/New Search/Views/Results/PodcastCarouselView.swift
+++ b/podcasts/New Search/Views/Results/PodcastCarouselView.swift
@@ -30,6 +30,11 @@ struct PodcastsCarouselView: View {
         UIScreen.main.bounds.width / Double(carouselItemsToDisplay)
     }
 
+    init() {
+        // Get the initial landscape value from the scene since UIDevice may not have the value yet
+        _isLandscape = State(initialValue: SceneHelper.foregroundActiveAppScene()?.interfaceOrientation.isLandscape ?? false)
+    }
+
     var body: some View {
         Group {
             if shouldShowLoadingActivity {
--- a/podcasts/New Search/Views/Results/PodcastCarouselView.swift
+++ b/podcasts/New Search/Views/Results/PodcastCarouselView.swift
@@ -53,10 +53,14 @@ struct PodcastsCarouselView: View {
                     PodcastResultCell(result: podcast)
                 }
                 .carouselPeekAmount(.constant(20))
-                .carouselItemSpacing(10)
-                .carouselItemsToDisplay(UIDevice.current.isiPad() ? 4 : 2)
-                .aspectRatio(UIDevice.current.isiPad() ? 4 : 1.75, contentMode: .fit)
-                .padding(.bottom, 10)
+                .carouselItemSpacing(16)
+                .carouselItemsToDisplay(carouselItemsToDisplay)
+
+                // Apply an aspect ratio to the carousel to auto adjust the height
+                // while maintaining the correct ratios for the items inside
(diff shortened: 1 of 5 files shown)


podcasts/New Search/Views/Results/PodcastCarouselView.swift
```diff
@@ -60,8 +105,28 @@ struct PodcastsCarouselView: View {
             ThemedDivider()
                 .padding(.leading, 8)
         }
-        .tabViewStyle(PageTabViewStyle(indexDisplayMode: .never))
         .background(AppTheme.color(for: .primaryUi02, theme: theme))
+        .onReceive(UIDevice.orientationDidChangeNotification.publisher(), perform: { _ in
+            isLandscape = UIDevice.current.orientation.isLandscape
```
leandroalonso, line 115: You might want to do this check when the view appears.

If you run the app already on landscape it will show 4 items because  `isLandscape` will be false.

leandroalonso (changes requested):
@emilylaguna I had 3 issues while testing, one is the already mentioned launching iPad in landscape.

The other is the space between the end of podcasts and the separator:



Finally, I noticed that when swiping gently, only one podcast is "moved" — instead of the whole page. I wonder if that's on purpose or not, check on the video below (second and third swipe):


https://user-images.githubusercontent.com/7040243/230907040-5f819040-c233-4c54-97d9-8e04706010b8.MP4

podcasts/New Search/Views/Results/PodcastCarouselView.swift
emilylaguna: Fixed here: a72829341589f6713684f29960f8a7a654a37535

emilylaguna:
> The other is the space between the end of podcasts and the separator:

Added more padding here: b20ea000c632fb3d69b7d17a9826a3fbce855b32




> Finally, I noticed that when swiping gently, only one podcast is "moved" — instead of the whole page. I wonder if that's on purpose or not, check on the video below (second and third swipe):

This is actually intentional, the carousel allows you to just click to the next item in the list without needing to go to a full next page. 

When I was testing I was checking out the podcasts in the results and wanted to just see the next item but when the carousel would go to the "next" page I found myself losing track of where the next item was. Allowing smaller swipes felt the most natural to me.

leandroalonso (approved):

PR #814, opened 2023-04-10
Search Improvements: cancel ongoing searches

leandroalonso:
This PR attempts to fix the following crash:

> Fatal error: Duplicate keys of type 'Canonical' were found in a Dictionary.
This usually means either that the type violates Hashable's requirements, or
that members of such a dictionary were mutated after insertion.

Based on my investigation, it looks like this crash only occurs when:

1. Searching from Podcasts
2. Searching for a word that a local podcast would be returned
3. There are some concurrency involved

## To test

The easiest way to reproduce the crash is to:

1. Change `podcastSearchDebounceMsDefault` to `0`
2. On `SearchResultsViewController.swift` add this after `searchResults.search(term: searchTerm)`:

```swift
        DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
            self.searchResults.search(term: searchTerm)
        }
```

You can [comment on this line](https://github.com/Automattic/pocket-casts-ios/pull/814/files#diff-064ba4ecb03923e49ba8a00ea9395af07e59bf230749fe8da8f3141099115086R40) to test the crash. Sometimes it will happen, sometimes not.

✅ [Uncomment this line](https://github.com/Automattic/pocket-casts-ios/pull/814/files#diff-064ba4ecb03923e49ba8a00ea9395af07e59bf230749fe8da8f3141099115086R40) and test again, the crash should not occur anymore.

## Checklist

### DIFF
--- a/podcasts/New Search/Model/SearchResultsModel.swift
+++ b/podcasts/New Search/Model/SearchResultsModel.swift
@@ -19,6 +19,8 @@ class SearchResultsModel: ObservableObject {
 
     @Published var hideEpisodes = false
 
+    private var tasksInProgress: [Task<(), Never>] = []
+
     init(analyticsHelper: SearchAnalyticsHelper = SearchAnalyticsHelper(source: .unknown)) {
         self.analyticsHelper = analyticsHelper
     }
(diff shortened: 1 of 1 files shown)


emilylaguna:
The crash was still occurring with these changes but we found that the changes in https://github.com/Automattic/pocket-casts-ios/pull/811 fixed the crash since it started using Identifiable on the PodcastFolderSearchResult. 

We think this issue was because using the `id: \.self` when rendering the previous carousel was causing some random hashing conflicts in SwiftUI when rendering the new results.

PR #819, opened 2023-04-11
Carousel: Fix an issue where the next page animation could happen too fast

emilylaguna:
This adds the animation when changing the visibleIndex to prevent it from messing with the gesture offset animation and causing it to jump. 

This also:
- Allows the swipe animation to be configured
- Adds a damping effect for the first and last pages to reduce the effect of the gesture offset

## To test

1. Launch the app
2. Open the search
3. Swipe to the left
4. ✅ Verify it feels harder to swipe
5. Swipe to the last page, and swipe to the right
6. ✅ Verify it feels harder to swipe
7. ✅ Verify the swipe animation doesn't jump

## Checklist


leandroalonso (approved):
:shipit:

PR #829, opened 2023-04-19
Load embedded artwork when streaming

leandroalonso:
While we currently offer the "Use Embedded Artwork" option, it only works for downloaded episodes. This PR adds the capability to display artwork when streaming an episode (if there's any).

**This PR doesn't add the feature to display episode images contained in the feed.**

## To test

1. Run the app
2. Go to Settings > Appearance > enable "Use Embedded Artwork"
3. Search for a podcast that has embedded images in episode ("Tecnocast" is a good one)
4. Play the episode
5. ✅ After a bit, the image should change to show the embedded artwork
6. ✅ Check that this artwork is shown on the now playing widget
7. Run CarPlay simulator
8. ✅ Check that the image is reflected on CarPlay too

## Checklist

### DIFF
--- a/podcasts/Constants.swift
+++ b/podcasts/Constants.swift
@@ -24,7 +24,6 @@ struct Constants {
         static let dimmingViewTapped = NSNotification.Name(rawValue: "SJDimViewTapped")
         static let downloadProgress = NSNotification.Name(rawValue: "SJDwnProg")
         static let podcastImageReCacheRequired = NSNotification.Name(rawValue: "PCPodcastImageReCacheRequired")
-        static let episodeEmbeddedArtworkLoaded = NSNotification.Name(rawValue: "PCEpisodeEmbeddedArtworkLoaded")
 
         static let skipTimesChanged = NSNotification.Name(rawValue: "SJSkipTimesChanged")
         static let subscribeRequestedFromCell = NSNotification.Name(rawValue: "SJSubscribeRequestFromCell")
--- a/podcasts/Constants.swift
+++ b/podcasts/Constants.swift
@@ -24,6 +24,7 @@ struct Constants {
         static let dimmingViewTapped = NSNotification.Name(rawValue: "SJDimViewTapped")
         static let downloadProgress = NSNotification.Name(rawValue: "SJDwnProg")
         static let podcastImageReCacheRequired = NSNotification.Name(rawValue: "PCPodcastImageReCacheRequired")
+        static let episodeEmbeddedArtworkLoaded = NSNotification.Name(rawValue: "PCEpisodeEmbeddedArtworkLoaded")
 
         static let skipTimesChanged = NSNotification.Name(rawValue: "SJSkipTimesChanged")
         static let subscribeRequestedFromCell = NSNotification.Name(rawValue: "SJSubscribeRequestFromCell")
--- a/podcasts/DefaultPlayer.swift
+++ b/podcasts/DefaultPlayer.swift
@@ -671,9 +671,10 @@ class DefaultPlayer: PlaybackProtocol, Hashable {
     }
 
     func loadEmbeddedImage() {
-        let playerItem = player?.currentItem
-        if let asset = playerItem?.asset, let episodeUuid {
-            StreamingEpisodeArtwork.shared.loadEmbeddedImage(asset: asset, episodeUuid: episodeUuid)
+        guard let asset = player?.currentItem?.asset, let episodeUuid else {
+            return
         }
+
+        StreamingEpisodeArtwork.shared.loadEmbeddedImage(asset: asset, episodeUuid: episodeUuid)
     }
 }
--- a/podcasts/DefaultPlayer.swift
+++ b/podcasts/DefaultPlayer.swift
@@ -36,6 +36,10 @@ class DefaultPlayer: PlaybackProtocol, Hashable {
     private var episodeUuid: String?
 
     #if !os(watchOS)
+        private lazy var streamingArtwork: StreamingEpisodeArtwork = {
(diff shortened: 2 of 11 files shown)


podcasts/StreamingEpisodeArtwork.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+import Kingfisher
+
+/// Extracts artwork from a streaming episode (if there's any)
+class StreamingEpisodeArtwork {
+    static let shared = StreamingEpisodeArtwork()
+
+    private lazy var subscribedPodcastsCache: ImageCache = {
+        let path = (NSHomeDirectory() as NSString).appendingPathComponent("Documents/artworkv3")
```
leandroalonso: This is the same cache used in `ImageManager`. It would be good to extract this to have this logic in a single place.

I haven't done it yet because I'd like to hear what other thinks about this approach.

podcasts/DefaultPlayer.swift
```diff
@@ -663,4 +669,11 @@ class DefaultPlayer: PlaybackProtocol, Hashable {
     func hash(into hasher: inout Hasher) {
         hasher.combine(ObjectIdentifier(self))
     }
+
+    func loadEmbeddedImage() {
+        let playerItem = player?.currentItem
+        if let asset = playerItem?.asset, let episodeUuid {
```
emilylaguna: Wdyt about using a guard here instead?

podcasts/Constants.swift
```diff
@@ -24,6 +24,7 @@ struct Constants {
         static let dimmingViewTapped = NSNotification.Name(rawValue: "SJDimViewTapped")
         static let downloadProgress = NSNotification.Name(rawValue: "SJDwnProg")
         static let podcastImageReCacheRequired = NSNotification.Name(rawValue: "PCPodcastImageReCacheRequired")
+        static let episodeEmbeddedArtworkLoaded = NSNotification.Name(rawValue: "PCEpisodeEmbeddedArtworkLoaded")
```
emilylaguna: Wdyt about moving this to the extension in [Utilities/Notifications](https://github.com/Automattic/pocket-casts-ios/blob/4d580ed449b313dc25aba508234af8a52b7037be/podcasts/Notifications.swift)?

podcasts/StreamingEpisodeArtwork.swift
```diff
@@ -0,0 +1,57 @@
...
+                    if let image {
+                        set(for: episodeUuid, image: image) {
+                            NotificationCenter.postOnMainThread(notification: Constants.Notifications.episodeEmbeddedArtworkLoaded)
+                        }
+                    }
+
+                }
+            }
+        }
+    }
```
emilylaguna: Looking at the [AVFoundation docs](https://developer.apple.com/documentation/avfoundation/media_assets/retrieving_media_metadata) they recommend using [AVMetadataItem.metadataItems(from:filteredByIdentifier:)](https://developer.apple.com/documentation/avfoundation/avmetadataitem/1385843-metadataitems), which means we can reduce this down to:

```suggestion
        let artworkItems = AVMetadataItem.metadataItems(from: asset.commonMetadata, filteredByIdentifier: .commonIdentifierArtwork)
        let image = artworkItems.compactMap { $0.dataValue.flatMap { UIImage(data: $0) } }.first

        guard let image else { return }

        set(for: episodeUuid, image: image) {
            NotificationCenter.postOnMainThread(notification: Constants.Notifications.episodeEmbeddedArtworkLoaded)
        }
```

podcasts/StreamingEpisodeArtwork.swift
emilylaguna: I like this approach since it's a bit more semantic and task oriented than the ImageManager.

I wonder if we can just change the cache property in the ImageManager to `private (set) var subscribedPodcastsCache: ImageCache`, that way we can still access it elsewhere, and don't need to refactor the ImageManager. 

If we do that this class will no longer need to be a singleton since it'll be using the singleton cache property from the ImageManager already. 

Thoughts?

emilylaguna (changes requested):
nice work, works well in my testing. 

Left a couple comments below.

podcasts/DefaultPlayer.swift
leandroalonso: Done in 1d4fc161486821fdda55b1e0c5bdcfa7f3d6d8fe

podcasts/Constants.swift
leandroalonso: Done in 231ea04bd81de0fabef4cadfc0a359dafc413c92

podcasts/StreamingEpisodeArtwork.swift
leandroalonso: Done in 8e0ee5b63e2f69351c50fcc9dc4ea973c02f0ddf

podcasts/StreamingEpisodeArtwork.swift
leandroalonso: I made some changes.

I wasn't able to fully make the `StreamingEpisodeArtwork` not a singleton so open to any ideas on how to improve it.

I could move all the logic to `ImageManager` but, oh well, it's a Singleton too. :|

leandroalonso:
@emilylaguna ready for another review!

podcasts/StreamingEpisodeArtwork.swift
emilylaguna: If we remove the `shared`, we can move it to a lazy property on DefaultPlayer since it's only being used in that 1 place right now.

emilylaguna (approved):
LGTM! 🚢

PR #833, opened 2023-04-25
Patron: update upgrade screen

leandroalonso:
Updates the new upgrade screen to display prices right away and allow the purchase without the need to show the modal.

### Pre-testing setup:
1. Open the Xcode Project
3. Hit <kbd>Command</kbd>+<kbd>Shift</kbd>+<kbd>,</kbd> to open the Scheme editor
4. Click the options tab
5. Select the `Pocket Casts Configuration.storekit` file in the StoreKit configuration dropdown

### Test the current flow

1. Enable the `patron` flag in `FeatureFlag.swift`
2. Do a clean install of the app
3. Go through the whole onboarding process and buy any plan
4. ✅ Make sure you see the confetti screen at the end
5. Go to Settings > Developer > tap "Set to No Plus"
6. Go to Podcasts > tap the Folder button
7. Buy any plan
8. ✅ Make sure you see the confetti screen
9. On Xcode, go to Debug > StoreKit > Manage Transactions
10. Delete Everything
11. Display the `.sorekit` file on Xcode
12. Go to Editor > Fail Transactions > select any failure except the "unknown"
13. Re-run the app
13. Go to Settings > Developer > tap "Set to No Plus"
6. Go to Podcasts > tap the Folder button
7. Buy any plan
1. ✅ An error should be displayed

### Test messages

1. Open the `.storekit` file
2. Add free trial in all 4 available tiers, choose anything you want but make sure they're different
13. Re-run the app
13. Go to Settings > Developer > tap "Set to No Plus"
1. ✅ Make sure the correct prices and trial period are shown

### Different devices

Please test the screen on smaller devices and big devices.

## Checklist

### DIFF
--- a/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
+++ b/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
@@ -135,8 +135,9 @@ extension PlusLandingViewModel {
     @ViewBuilder
     private static func view(with viewModel: PlusLandingViewModel, purchaseModel: PlusPurchaseModel? = nil) -> some View {
         if FeatureFlag.patron.enabled, let purchaseModel {
-            UpgradeLandingView(purchaseModel: purchaseModel)
+            UpgradeLandingView()
                 .environmentObject(viewModel)
+                .environmentObject(purchaseModel)
                 .setupDefaultEnvironment()
         } else {
             PlusLandingView(viewModel: viewModel)
--- a/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
+++ b/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
@@ -116,8 +116,9 @@ private extension PlusLandingViewModel {
 extension PlusLandingViewModel {
     static func make(in navigationController: UINavigationController? = nil, from source: Source, continueUpgrade: Bool = false) -> UIViewController {
         let viewModel = PlusLandingViewModel(source: source, continueUpgrade: continueUpgrade)
+        let purchaseModel = FeatureFlag.patron.enabled ? PlusPurchaseModel() : nil
 
-        let view = Self.view(with: viewModel)
+        let view = Self.view(with: viewModel, purchaseModel: purchaseModel)
         let controller = PlusHostingViewController(rootView: view)
 
         controller.viewModel = viewModel
@@ -126,14 +127,15 @@ extension PlusLandingViewModel {
         // Create our own nav controller if we're not already going in one
         let navController = navigationController ?? UINavigationController(rootViewController: controller)
         viewModel.navigationController = navController
+        purchaseModel?.parentController = navController
 
         return (navigationController == nil) ? navController : controller
     }
 
     @ViewBuilder
-    private static func view(with viewModel: PlusLandingViewModel) -> some View {
-        if FeatureFlag.patron.enabled {
-            UpgradeLandingView()
+    private static func view(with viewModel: PlusLandingViewModel, purchaseModel: PlusPurchaseModel? = nil) -> some View {
+        if FeatureFlag.patron.enabled, let purchaseModel {
+            UpgradeLandingView(purchaseModel: purchaseModel)
                 .environmentObject(viewModel)
                 .setupDefaultEnvironment()
         } else {
--- a/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
+++ b/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
@@ -27,6 +27,18 @@ class PlusLandingViewModel: PlusPricingInfoModel, OnboardingModel {
         loadPricesAndContinue(plan: plan, selectedPrice: selectedPrice)
     }
 
+    func unlockTapped(ifLoggedIn: () -> Void) {
+        OnboardingFlow.shared.track(.plusPromotionUpgradeButtonTapped)
+
+        guard SyncManager.isUserLoggedIn() else {
+            let controller = LoginCoordinator.make(in: navigationController, fromUpgrade: true)
+            navigationController?.pushViewController(controller, animated: true)
+            return
+        }
+
+        ifLoggedIn()
+    }
+
     func didAppear() {
         OnboardingFlow.shared.track(.plusPromotionShown)
 
--- a/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
+++ b/podcasts/Onboarding/Plus/PlusLandingViewModel.swift
@@ -60,6 +60,30 @@ class PlusLandingViewModel: PlusPricingInfoModel, OnboardingModel {
         pricingInfo.products.first(where: { $0.identifier == (frequency == .yearly ? tier.plan.yearly : tier.plan.monthly) })?.rawPrice ?? "?"
     }
 
+    func purchaseTitle(for tier: UpgradeTier, frequency: PlusPricingInfoModel.DisplayPrice) -> String {
+        guard let product = pricingInfo.products.first(where: { $0.identifier == (frequency == .yearly ? tier.plan.yearly : tier.plan.monthly) }) else {
+            return ""
+        }
+
+        if let _ = product.freeTrialDuration {
+            return L10n.plusStartMyFreeTrial
+        } else {
+            return tier.buttonLabel
+        }
+    }
+
+    func purchaseSubtitle(for tier: UpgradeTier, frequency: PlusPricingInfoModel.DisplayPrice) -> String {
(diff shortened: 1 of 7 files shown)


podcasts/Onboarding/Plus/UpgradeLandingView.swift
```diff
@@ -296,10 +341,16 @@ struct UpgradeCard: View {
         .shadow(color: .black.opacity(0.1), radius: 3, x: 0, y: 2)
         .shadow(color: .black.opacity(0.1), radius: 0, x: 0, y: 0)
     }
+
+    @ViewBuilder
+    var termsAndConditions: some View {
+        let purchaseTerms = L10n.purchaseTerms("$", "$", "$", "$").components(separatedBy: "$")
+        Text(purchaseTerms[safe: 0] ?? "") + Text("[\(purchaseTerms[safe: 1] ?? "")](https://support.pocketcasts.com/article/privacy-policy/)").underline() + Text(purchaseTerms[safe: 2] ?? "") + Text("[\(purchaseTerms[safe: 3] ?? "")](https://support.pocketcasts.com/article/terms-of-use/)").underline()
```
leandroalonso: I cannot use a constant for the URL here, unfortunately. If I do that, the link doesn't work.

podcasts/Onboarding/Plus/PlusLandingViewModel.swift
```diff
@@ -56,8 +56,28 @@ class PlusLandingViewModel: PlusPricingInfoModel, OnboardingModel {
...
     }
 
-    func price(for tier: UpgradeTier, frequency: PlusPricingInfoModel.DisplayPrice) -> String {
-        pricingInfo.products.first(where: { $0.identifier == (frequency == .yearly ? tier.plan.yearly : tier.plan.monthly) })?.rawPrice ?? "?"
+    func purchaseTitle(for tier: UpgradeTier, frequency: PlusPricingInfoModel.DisplayPrice) -> String {
+        guard let product = pricingInfo.products.first(where: { $0.identifier == (frequency == .yearly ? tier.plan.yearly : tier.plan.monthly) }) else {
+            return L10n.loading
+        }
+
+        if let _ = product.freeTrialDuration {
```
emilylaguna:
```suggestion
        if product.freeTrialDuration != nil {
```

podcasts/Onboarding/Plus/PlusLandingViewModel.swift
```diff
@@ -106,14 +127,15 @@ extension PlusLandingViewModel {
...
         return (navigationController == nil) ? navController : controller
     }
 
     @ViewBuilder
-    private static func view(with viewModel: PlusLandingViewModel) -> some View {
-        if FeatureFlag.patron.enabled {
-            UpgradeLandingView()
+    private static func view(with viewModel: PlusLandingViewModel, purchaseModel: PlusPurchaseModel? = nil) -> some View {
+        if FeatureFlag.patron.enabled, let purchaseModel {
+            UpgradeLandingView(purchaseModel: purchaseModel)
```
emilylaguna: Since we're passing the other view model via an environment object, should we do the same thing for the purchase model?

podcasts/Onboarding/Plus/UpgradeLandingView.swift
```diff
@@ -13,41 +16,100 @@ struct UpgradeLandingView: View {
 
     @State private var displayPrice: PlusPricingInfoModel.DisplayPrice = .yearly
 
+    @State private var loadingPrices = true
```
emilylaguna: This doesn't seem to be used?

podcasts/Onboarding/Plus/UpgradeLandingView.swift
emilylaguna: You can wrap the markdown in `.init` to fix that:

```
        let privacyPolicy = ServerConstants.Urls.privacyPolicy
        let termsOfUse = ServerConstants.Urls.termsOfUse

        Text(purchaseTerms[safe: 0] ?? "") +
        Text(.init("[\(purchaseTerms[safe: 1] ?? "")](\(privacyPolicy))")).underline() +
        Text(purchaseTerms[safe: 2] ?? "") +
        Text(.init("[\(purchaseTerms[safe: 3] ?? "")](\(termsOfUse))")).underline()
```

podcasts/Onboarding/Plus/UpgradeLandingView.swift
```diff
@@ -296,10 +341,16 @@ struct UpgradeCard: View {
         .shadow(color: .black.opacity(0.1), radius: 3, x: 0, y: 2)
         .shadow(color: .black.opacity(0.1), radius: 0, x: 0, y: 0)
     }
+
+    @ViewBuilder
+    var termsAndConditions: some View {
```
emilylaguna, line 395: When I'm in dark mode the text disappears:

podcasts/Onboarding/Plus/UpgradeLandingView.swift
```diff
@@ -277,13 +320,15 @@ struct UpgradeCard: View {
...
-                .padding(.bottom, 24)
 
-                Button(tier.buttonLabel) {
-                    viewModel.unlockTapped(plan: tier.plan, selectedPrice: currentPrice.wrappedValue)
+                    Group {
+                        termsAndConditions
+                    }
+                    .font(style: .footnote).fixedSize(horizontal: false, vertical: true)
+                    .tint(.black)
+                    .opacity(0.64)
```
emilylaguna: No need for the Group here. 

```suggestion
                    termsAndConditions
                    .font(style: .footnote).fixedSize(horizontal: false, vertical: true)
                    .tint(.black)
                    .opacity(0.64)
```

emilylaguna (changes requested):
Overall this is working really well, nice job!

While testing I noticed a few issues which I noted below. 

For the layout issues I think we can fix those in another PR since I they're kind of outside the scope of this one.
---

There's a bug that I saw a few times during testing where the model loads and is waiting for the prices but will display blank values.

I can repro easily by disabling the `IapHelper.shared.requestProductInfo` call in the AppDelegate.




It looks like this is happening because the PlusPricingInfoModel.pricingInfo is only being set on init, so if the pricing is not available at init then it won't be updated. 

I think we can fix that by reloading when setting the `priceAvailability` to available.

---

On the iPhone 14 Pro the content is aligned to the bottom and leaves large spacing at the top of the view.
 


---

On the smallest device that we support (iPod Touch 7th Gen) the view layout breaks:

podcasts/Onboarding/Plus/UpgradeLandingView.swift
leandroalonso: Removed in 8c3ced2b7e93e198d18b3e29beb7f91e6605f8bc

podcasts/Onboarding/Plus/UpgradeLandingView.swift
leandroalonso: Done in 4b3834831c223c2f07d4c617892356291ac99613

podcasts/Onboarding/Plus/UpgradeLandingView.swift
leandroalonso: Ops, left it by mistake, done in a1c6d072c374a2a84c66fa733c843db5cabd8b69

podcasts/Onboarding/Plus/PlusLandingViewModel.swift
leandroalonso: I don't think `PlusPurchaseModel` needs to be an environment object but it's better to be consistent, changed in 3525856a458b28992e0bba01b533d3a85df5b3d0

podcasts/Onboarding/Plus/UpgradeLandingView.swift
leandroalonso: Awesome, thanks! Done in d8bbfc8548940e7b2306ad9833b54cd63c7b152c

leandroalonso:
@emilylaguna ready for another look!

I changed the layout a bit so the examples you mentioned should be fixed. We have another possibility on fixing that by adding a scroll view on the whole screen but I'll wait for some feedback before doing this change (and we can do in another PR).

emilylaguna (changes requested):
This is working on smaller devices much better now. 

On the iPhone SE the paging indicator is appearing on top of the button which looks a bit odd. I wonder if on smaller devices we should hide the indicator to prevent this? 

Screen Shot 2023-04-26 at 1 07 22 PM

Also, while testing the logged out flow I found that if I try to purchase from an upsell dialog I am not able to continue and instead get an error.

leandroalonso:
@emilylaguna fixed!

I also changed the page indicator to not appear in case of scrollable content.

emilylaguna (approved):
Looks good, the logged in issue is still there but we can address that in another PR. 

🚢

PR #846, opened 2023-05-08
Fix playback speed label collapsed in German language

### LINKED ISSUE #836: Unable to read the configured custom playback speed in german app
### Expected behavior
Playback speed is readable

### Actual behavior
Unable to read the configured custom playback speed.
image

Open custom settings of a podcast with language set to german

##### Tested on [device], iOS [version], Pocket Casts [version]
iPhone 11 Pro, iOS 16.4.1, Pocket Casts 7.36 (7.36.0.3)

gibachan:
Fixes #836

Based on ticket #836, the `Speed` label is not appearing correctly in the playback settings when using the German language setting.
And I have also noticed that similar issue is present in the playback sheet, so I have fixed both of these problems.

### English

| Before | After |
| --- | --- |
|  |  |
|  |  |

### German

| Before | After |
| --- | --- |
|  |  |
|  |  |

## To test

- Change language settings to German in iOS settings
- Launch the app
- Open custom settings of a podcast and check the `Speed` label is not collapsed
- Play an episode
- Open playback effects modal sheet and checke the `Play Speed` label is not collapsed

## Checklist

### DIFF
--- a/podcasts/EffectsViewController.xib
+++ b/podcasts/EffectsViewController.xib
@@ -1,8 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="18122" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="21701" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
     <device id="retina6_1" orientation="portrait" appearance="light"/>
     <dependencies>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="18093"/>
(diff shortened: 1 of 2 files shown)


frosty (dismissed):
Thanks for the fix, @gibachan! However, I think wrapping the text like this when it often only overflows by a couple of characters doesn't look ideal. I wonder if we can instead allow the font size to shrink slightly (up to a minimum) if the text is too long? We could also potentially, in the Now Playing settings, _slightly_ shrink the width of the box showing the current speed.

Would you be able to have a go at this alternative, and see how it looks?

gibachan:
Thank you for your feedback!
I would love to make the changes you suggested. Let me see how it goes.

gibachan:
I set a minimum font size for autoshrink on those labels. [d5aed14](https://github.com/Automattic/pocket-casts-ios/pull/846/commits/d5aed14ec725d78f0a5223e8c21cd0aca5d753ad)
I think looks better than before. What do you think?

| Device | Screen 1 | Screen 2 |
| --- | --- | --- |
|  iPhone SE (3rd) |  |  |
|  iPhone 14 |  |  |
|  iPhone 14 Pro Max |  |  |

emilylaguna (approved):
Hey @gibachan!! Excellent work on this, and thank you for providing all the screenshots!!

I gave this a test and it works great for me in multiple different languages. 

Approved!

PR #853, opened 2023-05-10
Remove starred as a default filter

geekygecko:
On a fresh install, one of the default filters is 'starred'. This change removes this filter so it matches Android. Having a starred filter in two places, on the 'Filters' and 'Profile' tab is confusing as they behave differently. As the Profile tab starred list includes episodes from unsubscribed podcasts, it's more useful there.

The reason for fixing this is one of our Automotive partners was confused by the two starred episode lists.

| iOS | Android |
| --- | --- |
| Simulator Screenshot - iPhone 14 Pro - 2023-05-10 at 14 06 45 | Screenshot_20230510_141433 |

iOS After

## To test

1. Clean install the app
2. Go to the 'Filters' tab
3. ✅ Verify the 'Starred' filter is not in the list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 7.39
 -----
-
+- Removed the starred filter from the list of default filters
 
 7.38
 -----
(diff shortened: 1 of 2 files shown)


frosty (approved):
Working as described!

PR #876, opened 2023-05-26
Fix: Tracks Seeking Crash

emilylaguna:
This adds an additional check to prevent the following crash from happening:

`Fatal error: Double value cannot be converted to Int because it is either infinite or NaN`

## To test

I'm not sure exactly how to reproduce this but I added some Unit Tests for it so if the CI is 🟢  then we should be good. 

If you want to manually test you can do the following:

1. Open the PlaybackManager.swift in Xcode
2. Go to line 394
3. Replace it with `analyticsPlaybackHelper.seek(from: .infinity, to: .infinity, duration: .infinity)`
4. Run the app
5. Open the player
6. Seek to a different location
7. ✅ Verify it doesn't crash
8. Go back to line 394
9. Change it to `analyticsPlaybackHelper.seek(from: .nan, to: .nan, duration: .nan)`
10. Follow the steps again
11. ✅ Verify it doesn't crash

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Extensions/Double+Extensions.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Extensions/Double+Extensions.swift
@@ -1,0 +1,9 @@
+import Foundation
+
+extension Double {
+
+    /// Checks whether the value is not NaN or Infinite
+    public var isValid: Bool {
+        !self.isNaN && !self.isInfinite
+    }
+}
--- a/Modules/Utils/Sources/PocketCastsUtils/Extensions/Double+Extensions.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Extensions/Double+Extensions.swift
@@ -3,7 +3,7 @@ import Foundation
 extension Double {
 
     /// Checks whether the value is not NaN or Infinite
-    public var isValid: Bool {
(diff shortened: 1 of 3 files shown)


Modules/Utils/Sources/PocketCastsUtils/Extensions/Double+Extensions.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+extension Double {
+
+    /// Checks whether the value is not NaN or Infinite
+    public var isValid: Bool {
```
leandroalonso: While I think `isNaN` configures a "not valid" value, I'm not so sure about infinite. Maybe we should call this `isNumeric`? Or `isANumber`?

leandroalonso (approved):
@emilylaguna left a small suggestion to improve the number of the new property in the `Double` extension.

Modules/Utils/Sources/PocketCastsUtils/Extensions/Double+Extensions.swift
emilylaguna: Ah! `isNumeric` is better, I wasn't sure what to call it. 🥳 

Done here: e9d735c18f3576ca2a5a6beb2a2635088d56a965

PR #877, opened 2023-05-26
Tracks: Only track the application installed / updated events if protected data is available

emilylaguna:
This adds a `UIApplication.shared.isProtectedDataAvailable` check before trying record the app installed/updated events. 

If the app somehow is launched in the background and `didFinishLaunchingWithOptions` is called but protected data isn't available then we'll save the latest app version in the user defaults but the analytics won't be setup so we don't record it. 

If that occurs, then we'll only record the events once the user relaunches the app.

## To test

Unfortunately there is not a way to test this since it only happens on app launch with protected data disabled. 

## Checklist

### DIFF
--- a/podcasts/Analytics/AppLifecyleAnalytics.swift
+++ b/podcasts/Analytics/AppLifecyleAnalytics.swift
@@ -58,8 +58,8 @@ extension AppLifecycleAnalytics {
         //
         // If the app is launched "in the background" and protected data is enabled then the analytics won't
         // be enabled and we may miss some events.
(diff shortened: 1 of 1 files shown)


pocketcasts:
### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
AppLifecyleAnalytics.swift | 62 | Lines should not have trailing whitespace. (trailing_whitespace)


</p>

leandroalonso (approved):

PR #879, opened 2023-05-26
Add Network Connection Testing

### LINKED ISSUE #854: Add Network Connection Testing
Let's add a network connection test feature similar to what we already have on Android: 
image

[Here is the code used for this in the Android app](https://github.com/Automattic/pocket-casts-android/blob/046c81ac46eeefd6a9ec55fb75da541deaa512f2/modules/features/settings/src/main/java/au/com/shiftyjelly/pocketcasts/settings/status/StatusViewModel.kt#L35-L76)

This will be handy when troubleshooting network connection issues and will be a step we will be adding to our support guides

leandroalonso:
Fixes #854 

Adds a status page so users can debug their own network connection. This is mainly a copy of what was done on Android to keep consistency.

| Success | Failure |
| -------- | ------- |
| Simulator Screen Shot - iPhone 14 - 2023-05-26 at 17 30 59 | Simulator Screen Shot - iPhone 14 - 2023-05-26 at 17 31 15 |

Running:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/7f35bbe8-fafa-4fb2-98cb-76f201db78d7

## To test

1. Run the app
2. Go to Profile > Help & Feedback > tap the top right icon
3. Tap "Run now"
4. ✅ All the checks should complete
5. Disable your wi-fi
6. ✅ All tests should fail

## Checklist


podcasts/StatusPageViewModel.swift
```diff
@@ -0,0 +1,121 @@
...
+                if let httpResponse = response as? HTTPURLResponse, error == nil {
+                    continuation.resume(returning: httpResponse.statusCode)
+                } else {
+                    continuation.resume(returning: nil)
+                }
+            }
+            task.resume()
+        }
+    }
+}
```
emilylaguna: Since this is an async function we can use the URLSession async methods and reduce this down to:

```suggestion
    func requestHTTPStatus() async -> Int? {
        let response = try? await URLSession.shared.data(from: self).1 as? HTTPURLResponse
        return response?.statusCode
    }
}
```

podcasts/OnlineSupportController.swift
```diff
@@ -61,6 +64,11 @@ class OnlineSupportController: PCViewController, WKNavigationDelegate {
         dismiss(animated: true, completion: nil)
     }
 
+    @objc private func showStatusPage() {
+        let hostingController = UIHostingController(rootView: StatusPageView().setupDefaultEnvironment())
```
emilylaguna: We can use the `ThemedHostingController` here.

```suggestion
        let hostingController = ThemedHostingController(rootView: StatusPageView())
```

emilylaguna (approved):
Nice work! Left some comments on using the async URLSession methods, and the ThemedHostingController.

Good to go after that!

PR #882, opened 2023-05-31
Support: improve Watch logs

### LINKED ISSUE #39: Apple Watch: no Wearable Logs Sent in Ticket
User having trouble logging in to the Pocket Casts Watch app. They sent a ticket along with logs and no wearable logs were included.

### Cases

* 4629344-zen
* 4851559-zen
* 4858784-zen
* 4528819-zen
* 4528819-zen

leandroalonso:
Fixes #39 

This change includes a few improvements in how the Watch logs are retrieved and how we communicate stuff to the user:

* Users are now able to copy their logs. So in case a user can see their Watch logs but they're not attached, they can manually share them with us — although with the subsequent changes, this shouldn't be necessary
* Cache: whenever a request to retrieve Watch logs succeeds, we cache the log. If any subsequent call fails, we use what we have in the cache. This will fix the issue of users being able to see their logs but the logs not coming through;
* In the "Attached Logs" screen the message that says "No wearable logs were available" was changed to include more context telling the user they should open the Watch app;
* Last, but not least, we try to identify if the user is mentioning "watch" in their issue description. If they are and watch logs are not available, we show an alert explaining that they should open the Watch app so the logs can be send. The user can then confirm that they opened the Watch app or ignore this recommendation and send the ticket as it is.

| Attached logs screen | Warning when mentioning Watch issue |
| --------------------- | --------------------------------------- |
| Simulator Screen Shot - iPhone 14 - 2023-05-31 at 10 50 39 | Simulator Screen Shot - iPhone 14 - 2023-05-31 at 10 50 50 |

### On simulator

1. Run the app
2. Go to Profile > Help & Feedback > Get in touch > Get support > Attached logs
3. ✅ Check that the no wearable message now has context about opening the Watch app
4. Tap and hold any of the fields
5. ✅ You should see a Copy option
6. Tap on it
7. ✅ Paste the content in another place and make sure the whole content was pasted

### On a real device with a connected Watch app

These steps need to run on a real device with an Apple Watch paired.

1. Run the app
2. Go to Profile > Help & Feedback > Get in touch > Get support
3. Open the Watch app
4. Tap Attached Logs
5. ✅ You can see the Watch logs
6. Close the Watch app (or even remove the Watch from your wrist)
7. Go back to the support filling screen
8. On description, type anything that contains the word "watch" and tap "Submit"
9. ✅ Your support request should be sent, you should be able to see it in Sentry and it should contain your Watch logs
10. Close the app and run it again
11. Go to Profile > Help & Feedback > Get in touch > Get support
12. On the description field, type any message containing the word "watch" and tap "Submit"
13. ✅ You should see an alert about sending your Watch logs
14. Don't open the Watch app and tap "I just opened the Watch app"
15. ✅ You should see the warning again, given the logs were not retrieved
16. Open Pocket Casts on the watch
17. Tap "I just opened the Watch app"
18. ✅ Your support request should be sent, you should be able to see it in Sentry and it should contain your Watch logs
19. Close and re-run the app on your iPhone
20. Go to Profile > Help & Feedback > Get in touch > Get support
21. On the description field, type any message NOT containing the word "watch" and tap "Submit"
22. ✅ Your support request should be sent and you should be able to see it in Sentry 

## Checklist


podcasts/Zendesk/MessageSupportView.swift
```diff
@@ -72,10 +72,15 @@ struct MessageSupportView: View {
                     viewModel.completion = nil
                     dismiss?()
                 }))
-            case .failure:
-                return Alert(title: Text(L10n.supportErrorTitle), message: Text(L10n.supportErrorMsg), dismissButton: .default(Text(L10n.supportOK), action: {
-                    viewModel.completion = nil
-                }))
+            case .failure(let error):
+                switch error {
+                case PlaybackError.errorDuringPlayback:
```
emilylaguna: Should we add a custom error for this?

podcasts/Zendesk/MessageSupportViewModel.swift
```diff
@@ -121,18 +122,27 @@ class MessageSupportViewModel: ObservableObject {
...
-                                                     email: self.requesterEmail,
-                                                     comment: self.comment,
-                                                     customFields: customFields,
-                                                     tags: self.config.tags)
-
-                return self.supportService.submitSupportRequest(requestObject)
+
+                // Check if the user mentioned watch on their issue description and if there
+                // are any Apple Watch logs available.
+                if self.comment.lowercased().contains("watch") && customFields.first(where: { $0.value.contains(FileLog.noWearableLogsAvailable) }) != nil && !ignoreUnavailableWatchLogs {
```
emilylaguna: Wdyt about also checking for the localized version of the word watch?

```suggestion
                let containsWatch = self.comment.localizedCaseInsensitiveContains(L10n.watch) || self.comment.lowercased().contains("watch")
                if containsWatch && customFields.first(where: { $0.value.contains(FileLog.noWearableLogsAvailable) }) != nil && !ignoreUnavailableWatchLogs {


```

podcasts/en.lproj/Localizable.strings
```diff
@@ -3721,3 +3721,18 @@
...
+"copy" = "Copy";
+
+/* Title of an alert shown to the user */
+"support_watch_help_title" = "Looking for Watch app help?";
+
+/* Description of an alert shown to the user */
+"support_watch_help_message" = "If you're having issues with the Pocket Casts Watch app we can send your wearable logs to better assist you. In order to do so, please open Pocket Casts on your Watch.";
+
+/* Button that the user can press if they opened the Watch app */
+"support_watch_help_opened_app" = "I just opened the Watch app";
```
emilylaguna: Wdyt about: 

```suggestion
"support_watch_help_opened_app" = "I've opened the Watch app";
```

podcasts/Zendesk/MessageSupportView.swift
```diff
@@ -72,10 +72,15 @@ struct MessageSupportView: View {
...
+                    return Alert(title: Text(L10n.supportWatchHelpTitle), message: Text(L10n.supportWatchHelpMessage), primaryButton: .default(Text(L10n.supportWatchHelpOpenedApp)) { viewModel.submitRequest() }, secondaryButton: .default(Text(L10n.supportWatchHelpSendWithoutLog)) { viewModel.submitRequest(ignoreUnavailableWatchLogs: true) })
```
emilylaguna, line 78: I wonder, should we use `viewModel.submitRequest(ignoreUnavailableWatchLogs: true)` for both actions to handle for a case where the user has opened the watch app but we still weren't able to get logs for it. 

I think currently from the users perspective they may feel like they're stuck in a loop and may keep tapping the "I opened the watch app" button but continually end up being prompted.

emilylaguna (changes requested):
Nice work! I tested and it works great. I left some comments below. 

Let me know what you think.

podcasts/Zendesk/MessageSupportView.swift
leandroalonso: Boooooo shame on me: dc12fb0669164053115e633549d035e700f173e1

podcasts/Zendesk/MessageSupportView.swift
leandroalonso: I did that intentionally, actually.

If we send it, the user will perceive it as if it worked, while it didn't. They have the option in the alert to intentionally send the ticket by acknowledging that Apple Watch logs are not attached.

While I think the perception of being stuck in a loop is bad I think is something that it's easier for users to share with us: "the app can't retrieve Watch logs" instead of not sending it, then support realizing, then informing us, etc.

Of course, the flow could be better but I think this is better than what we had previously anyway.

emilylaguna: Makes sense to me.

emilylaguna (approved):
Works for me! :shipit:

PR #899, opened 2023-06-12
Fix KingFisher from infinitely requesting the profile image when it returns 404

emilylaguna:
Screenshot 2023-06-12 at 4 42 35 PM

I noticed when debugging another issue a bunch of requests to Gravatar when the user doesn't have a profile image. 

After some debugging it turns out that for some reason KingFisher will infinitely request the profile image when it returns 404 and the `startLoadingBeforeViewAppear` flag is set on the `KFImage`.

I've removed the `startLoadingBeforeViewAppear` flag which fixes this issue longer continually requests the image. I've also removed the `cancelOnDisappear` flag since I don't think it's needed.

## To test

1. Before applying the PR
2. Use Charles to inspect the traffic of either the simulator or a real device
3. Enable SSL Proxying on gravatar.com
4. Login to an account without a Gravatar profile
5. 💥 You should see a lot of requests now
6. Apply the PR
7. Run the app
8. ✅ Verify you no longer see all the requests

## Checklist

### DIFF
--- a/podcasts/Profile - SwiftUI/Common Views/ProfileImage.swift
+++ b/podcasts/Profile - SwiftUI/Common Views/ProfileImage.swift
@@ -10,8 +10,6 @@ struct ProfileImage: View {
         if let url {
             KFImage
                 .url(url, cacheKey: email)
-                .startLoadingBeforeViewAppear()
-                .cancelOnDisappear(true)
                 .placeholder { _ in defaultProfileView }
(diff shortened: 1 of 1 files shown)


leandroalonso (approved):
Tested before and now with an account without Gravatar and with Gravatar.

It works as always and I don't see a lot of requests with this fix.

Issue #900, opened 2023-06-12
An episode can't be played when downloaded, but will play when streamed.


emilylaguna:
Related: https://github.com/Automattic/pocket-casts-ios/issues/898

As of right now I have only been able to reproduce this with this single episode: https://pca.st/22iph0yt

The episode https://pca.st/22iph0yt is able to be streamed, but when after downloading it the player will not be able to play it and will instead result in the episode ceasing to play.

### Expected behavior

The episode will either play, or an error message should be shown.

### Actual behavior

The episode just stops playing or is archived if the setting is enabled.

### Steps to reproduce the behavior

1. Go to episode https://pca.st/22iph0yt
2. Play it to start streaming
3. ✅ Verify the episode plays
4. Download the episode
6. Play it
7. 💥 The episode will stop playing

emilylaguna:
I did some investigating to figure out what is causing this to happen, and how we might be able to fix it. 

### What's going on?

Ultimately what is happening is there is some corruption in the downloaded episode that is causing the EffectsPlayer to not be able to play it due to an issue internally with the `AVAudioFile`. The error that is being shown is:

```
Packet with multiple raw data blocks was found: unsupported
```

### Why does it play if I play it on the web player?

It turns out there are dynamic ads in the episode and when you download it on the web or elsewhere you end up with a version of the episode with no issues at all, but when downloading on mobile you end up with a file with some minor corruption. 

You can see the file size difference here:
<img width="456" src="https://github.com/Automattic/pocket-casts-ios/assets/793774/27ced267-e9a0-4fb7-bd41-3932d66f02b2">

### Why are we able to stream the episode without any issues?

When we stream a file we use the `DefaultPlayer` which uses the simple `AVPlayer` under the hood which for some reason doesn't have this bug, so the file will play correctly. 

The AVPlayer is also what Apple Podcasts, and other apps use for their standard players which is why this appears not to happen to them.

I have also confirmed this issue exists outside of Pocket Casts by using a [sample project](https://developer.apple.com/documentation/avfaudio/audio_engine/playing_custom_audio_with_your_own_player) from Apple.

### What's the fix?

Unfortunately since this is an issue with the file and an internal issue with AVFoundation there isn't anything we can do directly to fix this. However we can improve how we handle this and similar issues:

#### 1. When playing a file using the EffectsPlayer, verify the file was loaded correctly.

[...]

PR #901, opened 2023-06-12
Fixes a crash that can happen when playing a corrupted file

### LINKED ISSUE #898: Crash: App crashes during playback for specific episode
It was reported in 6382056-zd-woothemes, that the app crashes when playing this episode: https://pca.st/22iph0yt.

The app archived the episode when I tried to play the downloaded episode, and crashed when I wanted to play it via streaming. The issue doesn't happen to other newer episodes, but I didn't test them all.

Crash logs:

[podcasts-2023-06-09-135150.ips.zip](https://github.com/Automattic/pocket-casts-ios/files/11702805/podcasts-2023-06-09-135150.ips.zip)

1. Open this podcast in the app: https://pca.st/22iph0yt.
2. Download the "Billy Madison Show Podcast - June 5, 2023" episode.
3. Play the downloaded episode and notice it will be marked as archived almost immediately.
4. Remove the downloaded file by going to the podcast page again > long press on the episode > ... > Remove Download.
5. Play the same episode again via streaming.
6. The app crashes after a few seconds of playback.

##### Tested on iPhone 11, iOS 16.5, Pocket Casts 7.39

emilylaguna:
Fixes #898

This fixes a crash that can happen if the EffectsPlayer tries to play a corrupted episode and adds error handling to the EffectsPlayer.

## What's the crash
The specific crash we get is:
```
Fatal error: Double value cannot be converted to Int64 because it is either infinite or NaN
```

### This can happen if:
- The file is being played with the EffectsPlayer (either by being downloaded or otherwise)
- The file is unable to be read by the player but doesn't throw an error and [results in a 0 length file](https://github.com/Automattic/pocket-casts-ios/blob/455842de94f1905e920b47e1ef5afc52f1a7cd25/podcasts/EffectsPlayer.swift#L99)
- The player [tries to skip forward](https://github.com/Automattic/pocket-casts-ios/blob/455842de94f1905e920b47e1ef5afc52f1a7cd25/podcasts/AudioReadTask.swift#L48) when its being created
   - some reasons could be:
      - Because there's playback progress
      - the podcast is set to skip forward
- The `framePositionForTime` calculation results in `123 / 0 = +Inf`

## What's the fix?
This PR applies an [`isNumeric`](https://github.com/Automattic/pocket-casts-ios/blob/455842de94f1905e920b47e1ef5afc52f1a7cd25/Modules/Utils/Sources/PocketCastsUtils/Extensions/Double%2BExtensions.swift#L6) check on the percent seeking value which checks for inf/nan and returns false for either of them. 

This results in the Inf value being ignored, and a default value being returned. 

This also adds a check to verify the frameCount value is not 0 after loading it, and will throw an error if that happens.

### ⚠️ This only fixes the crash, not the underlaying issue of a corrupt file not being handled correctly by the EffectsPlayer

I want to note that this only fixes this crash from happening and not the reason why a file would get to this state. We'll need to investigate this further, I have made an issue for it here: https://github.com/Automattic/pocket-casts-ios/issues/900

## To test

1. Locate the following episode: https://pca.st/22iph0yt
2. Start streaming it
3. While it's streaming, download it
4. Wait until the download is complete
5. ✅ Verify the episode stops playing and the app doesn't crash

## Checklist

### DIFF
--- a/podcasts/EffectsPlayer.swift
+++ b/podcasts/EffectsPlayer.swift
@@ -99,7 +99,11 @@ class EffectsPlayer: PlaybackProtocol, Hashable {
                     strongSelf.cachedFrameCount = strongSelf.audioFile!.length
                     DataManager.sharedManager.saveFrameCount(episode: episode, frameCount: strongSelf.cachedFrameCount)
 
-                    // If the count is still 0 then way may not be able to read the file correctly, so throw an error
+                    // Throw an error if the frame count is 0
+                    // `cachedFrameCount` is used to calculate the duration of the episode and a 0 value
+                    // will result in the episode immediately being marked as played or can cause Inf/NaN issues.
+                    //
+                    // For more info, see: https://github.com/Automattic/pocket-casts-ios/issues/900
                     if strongSelf.cachedFrameCount == 0 {
                         throw PlaybackError.unableToOpenFile
                     }
--- a/podcasts/EffectsPlayer.swift
+++ b/podcasts/EffectsPlayer.swift
@@ -105,7 +105,7 @@ class EffectsPlayer: PlaybackProtocol, Hashable {
                     //
                     // For more info, see: https://github.com/Automattic/pocket-casts-ios/issues/900
                     if strongSelf.cachedFrameCount == 0 {
-                        throw PlaybackError.unableToOpenFile
+                        throw PlaybackError.effectsPlayerFrameCountZero
                     }
                 }
             } catch {
--- a/podcasts/EffectsPlayer.swift
+++ b/podcasts/EffectsPlayer.swift
@@ -98,6 +98,11 @@ class EffectsPlayer: PlaybackProtocol, Hashable {
                     // we haven't cached a frame count for this episode, do that now
                     strongSelf.cachedFrameCount = strongSelf.audioFile!.length
                     DataManager.sharedManager.saveFrameCount(episode: episode, frameCount: strongSelf.cachedFrameCount)
+
+                    // If the count is still 0 then way may not be able to read the file correctly, so throw an error
+                    if strongSelf.cachedFrameCount == 0 {
+                        throw PlaybackError.unableToOpenFile
+                    }
                 }
             } catch {
                 objc_sync_exit(strongSelf.playerLock)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 7.41
 -----
 - Improve support section to help the user to send Apple Watch logs
+- Fixed a crash that could occur when playing some downloaded episodes (#901)
 
 7.40
 -----
--- a/PocketCastsTests/Tests/Analytics/AnalyticsCoordinatorTests.swift
+++ b/PocketCastsTests/Tests/Analytics/AnalyticsCoordinatorTests.swift
@@ -1,0 +1,46 @@
+import XCTest
+
+@testable import podcasts
+
+final class AnalyticsCoordinatorTests: XCTestCase {
+    func testFallsBackToPreviousSourceOnNilCurrentSource() {
+        let coordinator = AnalyticsCoordinator()
+        coordinator.currentSource = .carPlay
+
+        // Resets the current source and sets the previous source
+        let previousSource = coordinator.currentAnalyticsSource
+
+        // The current source is expected to be nil, so this should reset to the original source value
+        coordinator.fallbackToPreviousSourceIfNeeded()
+
+        XCTAssertEqual(previousSource, coordinator.currentSource)
+    }
+
+    func testFallsBackToPreviousSourceOnUnknownCurrentSource() {
+        let coordinator = AnalyticsCoordinator()
+        coordinator.currentSource = .carPlay
+
+        // Resets the current source and sets the previous source
+        let _ = coordinator.currentAnalyticsSource
+        coordinator.currentSource = .unknown
+
+        // The current source is expected to be nil, so this should reset to the original source value
+        coordinator.fallbackToPreviousSourceIfNeeded()
+
+        XCTAssertEqual(coordinator.currentSource, .carPlay)
+    }
+
+    func testFallbackDoesntHappenIfCurrentSourceIsSet() {
+        let coordinator = AnalyticsCoordinator()
+        coordinator.currentSource = .carPlay
+
+        // Resets the current source and sets the previous source
+        let _ = coordinator.currentAnalyticsSource
+        coordinator.currentSource = .chooseFolder
+
+        // The current source is set, this should do nothing
+        coordinator.fallbackToPreviousSourceIfNeeded()
+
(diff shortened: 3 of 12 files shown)


yaelirub:
@emilylaguna, I tested and I don't get the crash anymore, but I have a few of questions: 
1. The episode does stop playing when download is complete, but it also gets archived. Is that expected? 
2. Is the episode stopping the desired behavior or is it just a side affect of a temporary fix? its hard to know from the description.
3.  I understand we want to fix the crash first, but if the behaviors are not expected and just temporary. should we give the [issue](https://github.com/Automattic/pocket-casts-ios/issues/900) you opened high priority? If so, can we label it as such?

emilylaguna:
> The episode does stop playing when download is complete, but it also gets archived. Is that expected?

That is expected when the Auto Archive when played setting is enabled. This is because the player treats it as a completed episode instead of an error. 

> Is the episode stopping the desired behavior or is it just a side affect of a temporary fix? its hard to know from the description.

The episode stopping or being marked as played is not the desired behavior, ideally the episode should play correctly as it does when its being streamed. 

However once the episode is downloaded the EffectsPlayer will not play it, nor will forcing the player to use the DefaultPlayer (which is what is used when the episode is being streamed) will play it.

The default player does throw an error and says the episode is damaged:
```
Error Domain=AVFoundationErrorDomain Code=-11829 "Cannot Open" UserInfo={NSLocalizedFailureReason=This media may be damaged., NSLocalizedDescription=Cannot Open, NSUnderlyingError=0x28357e610 {Error Domain=NSOSStatusErrorDomain Code=-12848 "(null)"}}
```

> I understand we want to fix the crash first, but if the behaviors are not expected and just temporary. should we give the https://github.com/Automattic/pocket-casts-ios/issues/900 you opened high priority? If so, can we label it as such?

Looking at the issue on [Sentry](https://a8c.sentry.io/issues/4091778776/events/b42eca0b5e7c4d40b7f50bb70195baed/?project=4503921846583296), this crash has only occurred for 7 users and only 1 or 2 episodes. Because of the low counts I don't think its necessary to mark this as high priority.

When looking into this I did try an idea that would check the [`cachedFrameCount`](https://github.com/Automattic/pocket-casts-ios/blob/455842de94f1905e920b47e1ef5afc52f1a7cd25/podcasts/EffectsPlayer.swift#L99) value after reading the length and if it was still 0 throwing an error. 

I didn't use that idea because adding that check changes how the player works for everyone and could cause more issues if for some reason there's a state where having a 0 length is okay. While I don't think that's likely based on my current knowledge, I am not confident enough to commit to

emilylaguna:
> The episode does stop playing when download is complete, but it also gets archived. Is that expected?

That is expected when the Auto Archive when played setting is enabled. This is because the player treats it as a completed episode instead of an error.


While looking into this I did [try an idea](https://gist.github.com/emilylaguna/72286e3326b6678750608aba28d6632c#file-effectsplayerlengthcheck-swift-L13) that would check the [`cachedFrameCount`](https://github.com/Automattic/pocket-casts-ios/blob/455842de94f1905e920b47e1ef5afc52f1a7cd25/podcasts/EffectsPlayer.swift#L99) value after reading the length and if it was still 0 it would throwing an error and result in showing a "this episode might be corrupted" error.

I didn't use this because adding that check changes how the player works for everyone and could cause more issues if for some reason there's a state where having a 0 length is okay. While I don't think that's likely based on my current knowledge, I am not confident enough to commit to it without more investigation and testing.

> Is the episode stopping the desired behavior or is it just a side affect of a temporary fix? its hard to know from the description.

The episode stopping or being marked as played is not the desired behavior, ideally the episode should play correctly as it does when its being streamed. 

However once the episode is downloaded the EffectsPlayer will not play it, nor will forcing the player to use the DefaultPlayer, which is what is used when the episode is being streamed.

The default player does throw an error and says the episode is damaged:
```
Error Domain=AVFoundationErrorDomain Code=-11829 "Cannot Open" UserInfo={NSLocalizedFailureReason=This media may be damaged., NSLocalizedDescription=Cannot Open, NSUnderlyingError=0x28357e610 {Error Domain=NSOSStatusErrorDomain Code=-12848 "(null)"}}
```

> I understand we want to fix the crash first, but if the behaviors are not expected and just temporary. should we give the https://github.com/Automattic/pocket-casts-ios/issues/900 you opened high priority? If so, can we label it as such?

Looking at the issue on [Sentry](https://a8c.sentry.io/issues/4091778776/events/b42eca0b5e7c4d40b7f50bb70195baed/?project=4503921846583296), this crash has only occurred for 7 users and only from 1 or 2 episodes (I was only able to see 1 episode from the breadcrumbs, but since this first appeared before that episode was even published I'm assuming there is at least 1 other episode with this issue). Because of the low counts I don't think its necessary to mark this as high priority.

emilylaguna:
I pushed a change that will throw an error if the EffectsPlayer fails due to a 0 frameCount. 

I also pushed 2 concept branches that convert the AAC m4a files to AAC ones, and adds fallback handling if a player fails to play:

- https://github.com/Automattic/pocket-casts-ios/compare/fix/player-crash...fix/aac-to-m4a-files?expand=1
- https://github.com/Automattic/pocket-casts-ios/compare/fix/aac-to-m4a-files...task/add-player-fallback?expand=1

podcasts/EffectsPlayer.swift
```diff
@@ -98,6 +98,11 @@ class EffectsPlayer: PlaybackProtocol, Hashable {
                     // we haven't cached a frame count for this episode, do that now
                     strongSelf.cachedFrameCount = strongSelf.audioFile!.length
                     DataManager.sharedManager.saveFrameCount(episode: episode, frameCount: strongSelf.cachedFrameCount)
+
+                    // If the count is still 0 then way may not be able to read the file correctly, so throw an error
```
yaelirub: should this be we instead of way?

podcasts/EffectsPlayer.swift
```diff
@@ -98,6 +98,11 @@ class EffectsPlayer: PlaybackProtocol, Hashable {
...
+                    if strongSelf.cachedFrameCount == 0 {
+                        throw PlaybackError.unableToOpenFile
```
yaelirub: should we add tracking here to know how often this happens?

podcasts/EffectsPlayer.swift
emilylaguna: 🫣 oops, fixed here:  cf19c0e84b99036a80fb56b4a298acb198068ffe

podcasts/EffectsPlayer.swift
emilylaguna: Great idea! I did this in a separate PR since there were some changes that were a bit more involved: https://github.com/Automattic/pocket-casts-ios/pull/918

mokagio:
@claude give me a **short** review of this PR. Asking to verify the new API key works in GitHub Actions.

PR #908, opened 2023-06-15
Use the new show notes endpoint and show embedded episodes images

leandroalonso:
I should have probably broken this PR into two different parts. However, as I made some changes it grew to the point that most of the initial changes were changed so it's probably easier to review this one.

This PR adds two changes:

1. It replaces the current show notes endpoint that is being used;
2. By using the new endpoint we're able to support episode-embedded artworks (🥳🥳🥳)

### Notes

Before testing it, open Charles or any other proxy debug application so we can test that the cache is working just fine.

1. Run the app
2. Go to Discover > tap on any podcast > tap on an episode
4. ✅ You should see the show notes appear
5. Go to Charles
6. Check requests made to the `shownotes.pocketcasts.com` subdomain
7. ✅ A request should have been made that returned all show notes
8. On the app, dismiss the episode screen
9. Open it again
10. ✅ You should see the notes again
11. On Charles, inspect the last request made to `shownotes.pocketcasts.com`
12. ✅ The server should have responded with a 304 and the headers must contain the `if-none-match` and `if-modified-since` parameters
13. Play the episode
14. Open the player and tap "Notes"
15. ✅ You should see the notes
16. Again, check the last request made to `shownotes.pocketcasts.com` on Charles
17. ✅ The server should have responded with a 304 and the headers must contain the `if-none-match` and `if-modified-since` parameters
18. Stop the app, turn off your Wi-fi (or go airplane mode)
19. Open the app and then tap on the miniplayer
20. ✅ The notes should shown

### Embedded artwork images

1. Go to Profile > Settings (cog icon) > Appearance > enable "Use Embedded Artwork"
2. Search for "The Bulwark Podcast" or "Clublife" (both have embedded XML artwork)
3. Play any episode
4. ✅ After a while the image should change to the embedded episode artwork
5. Download any episode from "The Bulwark Podcast" or "Clublife"
6. Close the app
7. Go offline
8. Reopen the app and play any of those episodes
9. ✅ You should see the episode artwork
10. ✅ Also check that you see the episode artwork on CarPlay and Now Playing screen

## Where to show episode artwork?

We show the episode artwork on the player, Now Playing screen on iPhone and CarPlay. In the listing inside the app the podcast image is still being shown. We can discuss whether this is a behavior we would like to keep or change, but I'd wait for users' feedback on it.

Also, once we test this feature, we can consider having "Use Embedded Artwork" enabled by default.

cc @geekygecko

## Checklist


Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
```diff
@@ -0,0 +1,119 @@
...
+
+        let decoder = JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+
+        let url = ServerHelper.asUrl(ServerConstants.Urls.cache() + "mobile/show_notes/full/\(podcastUuid)")
+        var request = URLRequest(url: url)
+        request.cachePolicy = .reloadRevalidatingCacheData
+
+        // Check for any cached version and if there's any call the cached completion blocks
+        if let cachedResponse = showNotesUrlCache.cachedResponse(for: request),
```
leandroalonso, line 92: This might look weird at first (it looked to me) but the `dataTask` below don't return what is cached in case the user is offline.

So we need this in order to support offline devices.

Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
emilylaguna, line 1: While this works using the testing steps while I noticed some issues.

emilylaguna: The code is assuming only a single podcast will be loaded multiple times which will result in some show notes not being loaded when they should.

If you have Auto Download Up Next enabled, the [`ShowNotesUpdater`](https://github.com/Automattic/pocket-casts-ios/pull/908/files#diff-2bb6e01efa45508c1c4def787390c9eb8b2ff448b7764c397709ffbe2b447bdeR8) is called [for each episode](https://github.com/Automattic/pocket-casts-ios/pull/908/files#diff-3487020f91a46ff169fb61f5aad0c5decde6e68d7d3ad436e92baa540c56ac2dR171) that is being downloaded, which means if I bulk add episodes from my filters some different podcasts will not be loaded because the [flag is true](https://github.com/Automattic/pocket-casts-ios/pull/908/files#diff-dec6fb6f2dae8f8038347163a886401d352dab7f79dcd87de54b6a4037f8bcdcR66)

Here's some steps to reproduce:

1. Put a breakpoint on the return statement in [`guard !requestingNotes else { return }`](https://github.com/Automattic/pocket-casts-ios/pull/908/files#diff-dec6fb6f2dae8f8038347163a886401d352dab7f79dcd87de54b6a4037f8bcdcR66)
2. Enable Auto Download Up Next
3. Go to a filter with multiple episodes from different podcasts
4. Tap and hold to add select them, and tap and hold to add them all to your up next queue
5. After the first item is added the flag is set to true and the breakpoint is hit
6. Each different podcast UUID from the first will not be loaded until the first request finishes

emilylaguna: If multiple podcasts/episodes show notes are being loaded with callbacks, the callbacks will be prematurely called and reset after the first network request finishes. And the callbacks will be called with the wrong podcast show notes, but the episode filter will prevent wrong data from being returned to the caller.

I don't have steps for this one but you can look at the logic here:

- Caller 1, 2, and 3 all call `loadShowNotes` around the same time for different podcasts and episodes with both a cache and a completion callback
- `requestShowNotes` is called
- In the first `requestShowNotes` call:
   - the completion handlers are added to the array
   - requestingNotes is set to true
   - the rest of the cache and network code starts loading
- Then 2 and 3 start going through `requestShowNotes`:
   - the completion handlers are added to the array
   - They stop at the `requestingNotes` check
- The first `requestShowNotes` call starts finishing:
  - It finishes loading from the network the show notes from the first podcast
  - It sees 3 items in `showNotesCompletionBlocks` 
  - It then calls each of the 3 blocks with the first podcasts data
  - The data is reset

emilylaguna: The loading code can be called from multiple different threads and will crash when trying to mutate the completion block arrays



- Enable auto download up next and only on WiFi
- Disable your WifI
- Go to a filter with lots of episodes 
- Add all of them to your up next
- They will be queued
- Enable your WiFi
- Perform a Refresh
- 💥 The app will crash



<details>
  <summary>🎥 <strong>Demo Video</strong></summary>

https://github.com/Automattic/pocket-casts-ios/assets/793774/4db11e09-9b87-49e4-996a-e01019740153
</details>

Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
```diff
@@ -0,0 +1,119 @@
...
+        guard !requestingNotes else { return }
+
+        requestingNotes = true
+
+        let decoder = JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+
+        let url = ServerHelper.asUrl(ServerConstants.Urls.cache() + "mobile/show_notes/full/\(podcastUuid)")
+        var request = URLRequest(url: url)
+        request.cachePolicy = .reloadRevalidatingCacheData
```
emilylaguna:
```suggestion
        let request = URLRequest(url: url, cachePolicy: .reloadRevalidatingCacheData)
```

Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
```diff
@@ -0,0 +1,119 @@
+import Foundation
+
+/// Request information about an episode using the show notes endpoint
+public class EpisodeInfoHandler {
+    public static var shared = EpisodeInfoHandler()
```
emilylaguna: Since this is being only being used by the `CacheServerHandler` which is a singleton, do we need to also make this a singleton?

Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
```diff
@@ -0,0 +1,119 @@
...
+        if let cachedResponse = showNotesUrlCache.cachedResponse(for: request),
+           let showNotes = try? decoder.decode(ShowNotes.self, from: cachedResponse.data) {
+
+            showNotesCachedCompletionBlocks.forEach { $0(showNotes.podcast) }
+            showNotesCachedCompletionBlocks = []
+            requestingNotes = false
+        }
+
+        // Call the endpoint to request for a more up-to-date notes information
+        URLSession.shared.dataTask(with: request) { [weak self] data, response, _ in
```
emilylaguna, line 102: Can we switch this to the async version since it's much cleaner?

Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
```diff
@@ -0,0 +1,119 @@
...
+        let showNotes: String
+        let image: String?
+    }
+
+    public func loadShowNotes(podcastUuid: String, episodeUuid: String, cached: ((String) -> Void)? = nil, completion: ((String?) -> Void)?) {
+        var cachedNotes = ""
+        var didSendCachedNotes = false
+
+        requestShowNotes(for: podcastUuid, cached: { showNotes in
+            if let notes = showNotes?.episodes.first(where: { $0.uuid == episodeUuid })?.showNotes {
```
emilylaguna: Wdyt about reducing the duplicate code by adding a method on the podcast object?

```
    struct ShowNotesPodcast: Decodable {
        let episodes: [ShowNotesEpisode]

        func episode(with uuid: String) -> ShowNotesEpisode? {
            episodes.first(where: { $0.uuid == uuid })
        }
    }
```

podcasts/EpisodeArtwork.swift
```diff
@@ -0,0 +1,74 @@
...
+        return artworkItems.compactMap { $0.dataValue.flatMap { UIImage(data: $0) } }.first
+    }
+
+    private func loadEpisodeArtworkFromUrl(podcastUuid: String, episodeUuid: String) {
+        CacheServerHandler.shared.loadEpisodeArtworkUrl(podcastUuid: podcastUuid, episodeUuid: episodeUuid) { [weak self] imageUrl in
+            guard let self else {
+                return
+            }
+
+            if let imageUrl, let url = URL(string: imageUrl) {
```
emilylaguna: Can we swap this to a guard to reduce the indenting of the containing code block?

emilylaguna (changes requested):
Heya, this is working well. I hit some issues during testing that I outlined below. 

Also, I think in the future it would be great if we could:
- Update the areas where we show the podcast artwork for episodes (episode details, filters, etc) to show the episode artwork if its available
- Change how we're caching the show notes so we don't need to load and parse the entire show notes to find the information for a single episode. 
   - In my testing the show notes I was using were 700kb and when downloading a lot of episodes from filters this spiked the memory and CPU usage up a lot.

Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
leandroalonso: Do you mean the async `URLSession.shared.data`? Or `JSONDecodableURLTask`? I give a try with the async `URLSession.shared.data` and I thought that the end result was much more confusing:

```
Task {
            // Call the endpoint to request for a more up-to-date notes information
            do {
                let (data, response) = try await URLSession.shared.data(for: request)

                let responseToCache = CachedURLResponse(response: response, data: data)
                showNotesUrlCache.storeCachedResponse(responseToCache, for: request)

                guard let showNotes = try? decoder.decode(ShowNotes.self, from: data) else {
                    showNotesCompletionBlocks.forEach { $0(nil) }
                    showNotesCompletionBlocks = []
                    return
                }

                showNotesCompletionBlocks.forEach { $0(showNotes.podcast) }
                showNotesCompletionBlocks = []
            } catch {
                showNotesCompletionBlocks.forEach { $0(nil) }
                showNotesCompletionBlocks = []
            }

            requestingNotes = false
        }
```

Alternatively, I can add a new method to `JSONDecodableURLTask` to return the entity **and** the request result, but I wonder if this clashes with the single responsibility of `JSONDecodableURLTask`.

Maybe I'm missing something tho! Thoughts?

podcasts/EpisodeArtwork.swift
leandroalonso: Done in 35ce42dcd990c9815dc01e76b3a55972d0cb32e2

Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
leandroalonso: Done in 78ad1ba29859b9493a9c4829ddf5ab64e4a1b0b1

Modules/Server/Sources/PocketCastsServer/Public/Cache/EpisodeInfoHandler.swift
leandroalonso: Done in 2a293eab80f2c00f3b1cb82c9a39f18d15137019

leandroalonso:
@emilylaguna I changed the code in 55a0cfdc1e4722ccb6be89450dfa523265f28555 to make use of `NSLock`. After this change, I'm unable to make the app crash and everything seems to work as intended.

The final code needs refactoring so it's more readable. Before doing that I'd like to hear your thoughts and see if there's any other issue you might find.

leandroalonso:
Moved to #1033

PR #909, opened 2023-06-15
Fix SmartLess not appearing in search results

leandroalonso:
The SmartLess podcast is not showing the podcast carousel.

## To test

1. Run the app
2. Search for "smartless"
3. ✅ Check that the SmartLess podcast appears as the first result
4. Search for a feed (eg.: `https://feeds.megaphone.fm/WWO9257218156`)
5. ✅ A single podcast should appear correctly on the left side

## Checklist

### DIFF
--- a/podcasts/New Search/Views/Results/PodcastCarouselView.swift
+++ b/podcasts/New Search/Views/Results/PodcastCarouselView.swift
@@ -130,7 +130,7 @@ struct PodcastsCarouselView: View {
         let podcast = Podcast()
         podcast.title = ""
         podcast.author = ""
-        podcast.uuid = ""
+        podcast.uuid = UUID().uuidString
 
         return podcast
     }
--- a/podcasts/New Search/Views/Results/PodcastCarouselView.swift
+++ b/podcasts/New Search/Views/Results/PodcastCarouselView.swift
@@ -54,7 +54,7 @@ struct PodcastsCarouselView: View {
                     }
                 }
             } else if searchResults.podcasts.count > 0 {
-                let fillerPodcast = Podcast.previewPodcast()
+                let fillerPodcast = fillerPodcast()
 
                 // If needed, fill the results with filler podcasts to make sure the sizing and positioning is consistent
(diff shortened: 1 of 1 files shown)


podcasts/New Search/Views/Results/PodcastCarouselView.swift
```diff
@@ -126,6 +126,15 @@ struct PodcastsCarouselView: View {
         }
     }
 
+    private func fillerPodcast() -> Podcast {
+        let podcast = Podcast()
+        podcast.title = ""
+        podcast.author = ""
+        podcast.uuid = ""
```
emilylaguna: Since the uuid is used to identify the view, should we return a random one to prevent any possible collisions?

```suggestion
        podcast.uuid = UUID().uuidString
```

podcasts/New Search/Views/Results/PodcastCarouselView.swift
```diff
@@ -126,6 +126,15 @@ struct PodcastsCarouselView: View {
         }
     }
 
+    private func fillerPodcast() -> Podcast {
```
emilylaguna, line 129: Thoughts about renaming to placeholder? I can also see this being useful elsewhere, should we move this to a static var? Then we can just use `Podcast.placeholder` or `\.placeholder`?

emilylaguna (changes requested):
Nice, confirmed this fixes the issue of the SmartLess podcast not appearing. I left some suggestions, let me know what you think.

podcasts/New Search/Views/Results/PodcastCarouselView.swift
leandroalonso: I'd rather not use this outside. 😅

I think a better solution would be to just make `Podcast` optional in the view or create a "`FillerView`".

Anyway, I think we can extract when the opportunity comes. Wdyt?

emilylaguna: Sounds good.

emilylaguna (approved):

PR #914, opened 2023-06-20
Episodes Autoplay: extract SQL queries to its own class

leandroalonso:
This PR extracts the SQL queries that power the following screens:

* Podcast
* Filters
* Downloads
* Files
* Starred
* Listening History

This step is necessary for the autoplay, given we need a way to retrieve the same list of episodes so we can accurately play the next one.

No new features are added to this PR.

## To test

Ps.: you can use `trunk` or the main app running in another simulator/device so you can compare the list of episodes.

1. Run the app (I recommend running into on a logged account, with Filters, Downloaded files, Starred, etc)
2. Go to Discover and check any random podcast
3. ✅ The episodes should appear correctly
4. Do the same in your "Podcasts" tab
5. ✅ Check that episodes appear correctly
6. Go to "Profile" and tap "Downloads"
7. ✅ Check that episodes appear correctly
8. Go back to "Profile" and tap "Files"
9. ✅ Check that your files appear correctly
10. Go back to "Profile" and tap "Starred"
11. ✅ Check that your episodes appear correctly
12. Go back to "Profile" and tap "Listening History"
11. ✅ Check that your played episodes appear correctly

## Checklist

### DIFF
--- a/podcasts/DownloadsViewController.swift
+++ b/podcasts/DownloadsViewController.swift
@@ -177,7 +177,7 @@ class DownloadsViewController: PCViewController {
         operationQueue.addOperation { [weak self] in
             guard let strongSelf = self else { return }
 
-            let newData: [ArraySection<String, ListEpisode>] = strongSelf.episodesDataManager.get(.downloads)
+            let newData = strongSelf.episodesDataManager.downloadedEpisodes()
 
             DispatchQueue.main.sync {
                 strongSelf.downloadsTable.isHidden = (newData.count == 0)
--- a/podcasts/DownloadsViewController.swift
+++ b/podcasts/DownloadsViewController.swift
@@ -175,14 +175,14 @@ class DownloadsViewController: PCViewController {
 
     func reloadEpisodes() {
         operationQueue.addOperation { [weak self] in
-            guard let strongSelf = self else { return }
+            guard let self else { return }
 
-            let newData = strongSelf.episodesDataManager.downloadedEpisodes()
+            let newData = self.episodesDataManager.downloadedEpisodes()
 
             DispatchQueue.main.sync {
-                strongSelf.downloadsTable.isHidden = (newData.count == 0)
-                strongSelf.episodes = newData
-                strongSelf.downloadsTable.reloadData()
+                self.downloadsTable.isHidden = (newData.count == 0)
+                self.episodes = newData
+                self.downloadsTable.reloadData()
             }
         }
     }
--- a/podcasts/DownloadsViewController.swift
+++ b/podcasts/DownloadsViewController.swift
@@ -8,6 +8,8 @@ class DownloadsViewController: PCViewController {
     var episodes = [ArraySection<String, ListEpisode>]()
     var cellHeights: [IndexPath: CGFloat] = [:]
 
+    private let episodesDataManager = EpisodesDataManager()
+
     private lazy var operationQueue: OperationQueue = {
         let queue = OperationQueue()
         queue.maxConcurrentOperationCount = 1
@@ -172,17 +174,12 @@ class DownloadsViewController: PCViewController {
     }
 
     func reloadEpisodes() {
-        operationQueue.addOperation {
-            let query = "( (downloadTaskId IS NOT NULL OR episodeStatus = \(DownloadStatus.downloaded.rawValue) OR episodeStatus = \(DownloadStatus.waitingForWifi.rawValue)) OR (episodeStatus = \(DownloadStatus.downloadFailed.rawValue) AND lastDownloadAttemptDate > ?) ) ORDER BY lastDownloadAttemptDate DESC LIMIT 1000"
-            let arguments = [Date().weeksAgo(1)] as [Any]
+        operationQueue.addOperation { [weak self] in
+            guard let strongSelf = self else { return }
 
-            let newData = EpisodeTableHelper.loadSectionedEpisodes(tintColor: AppTheme.appTintColor(), query: query, arguments: arguments, episodeShortKey: { episode -> String in
-                episode.shortLastDownloadAttemptDate()
-            })
+            let newData: [ArraySection<String, ListEpisode>] = strongSelf.episodesDataManager.get(.downloads)
 
-            DispatchQueue.main.sync { [weak self] in
-                guard let strongSelf = self else { return }
-
+            DispatchQueue.main.sync {
                 strongSelf.downloadsTable.isHidden = (newData.count == 0)
                 strongSelf.episodes = newData
                 strongSelf.downloadsTable.reloadData()
--- a/podcasts/EpisodesDataManager.swift
+++ b/podcasts/EpisodesDataManager.swift
@@ -2,11 +2,11 @@ import PocketCastsDataModel
 import PocketCastsServer
 import DifferenceKit
 
-/// Returns a list of episodes for an specific section
-/// Depending on the section, it returns a DifferenceKit ArraySection
 class EpisodesDataManager {
     // MARK: - Podcast episodes list
 
+    /// Returns a podcasts episodes that are grouped by `PodcastGrouping`
+    /// Use `uuidsToFilter` to filter the episode UUIDs to only those in the array
     func episodes(for podcast: Podcast, uuidsToFilter: [String]? = nil) -> [ArraySection<String, ListItem>] {
         // the podcast page has a header, for simplicity in table animations, we add it here
         let searchHeader = ListHeader(headerTitle: L10n.search, isSectionHeader: true)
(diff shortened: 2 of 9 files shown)


podcasts/EpisodesDataManager.swift
```diff
@@ -0,0 +1,127 @@
+import PocketCastsDataModel
+import PocketCastsServer
+import DifferenceKit
+
+/// Returns a list of episodes for an specific section
+/// Depending on the section, it returns a DifferenceKit ArraySection
+class EpisodesDataManager {
```
leandroalonso, line 5: I'm not 100% sold on the name of this class. Any suggestions are highly appreciated. Maybe `EpisodesListManager` is a better choice?

rviljoen: Scanning through this briefly, it looks more like a Helper class (as opposed to a Manager class)? I see it also calls out to several other Helpers.

podcasts/DownloadsViewController.swift
```diff
@@ -172,17 +174,12 @@ class DownloadsViewController: PCViewController {
...
     func reloadEpisodes() {
-        operationQueue.addOperation {
-            let query = "( (downloadTaskId IS NOT NULL OR episodeStatus = \(DownloadStatus.downloaded.rawValue) OR episodeStatus = \(DownloadStatus.waitingForWifi.rawValue)) OR (episodeStatus = \(DownloadStatus.downloadFailed.rawValue) AND lastDownloadAttemptDate > ?) ) ORDER BY lastDownloadAttemptDate DESC LIMIT 1000"
-            let arguments = [Date().weeksAgo(1)] as [Any]
-
-            let newData = EpisodeTableHelper.loadSectionedEpisodes(tintColor: AppTheme.appTintColor(), query: query, arguments: arguments, episodeShortKey: { episode -> String in
-                episode.shortLastDownloadAttemptDate()
-            })
+        operationQueue.addOperation { [weak self] in
+            guard let strongSelf = self else { return }
```
emilylaguna:
```suggestion
            guard let self else { return }
```

podcasts/UploadedViewController.swift
```diff
@@ -225,13 +227,7 @@ class UploadedViewController: PCViewController, UserEpisodeDetailProtocol {
...
 
     func reloadLocalFiles() {
-        let sortBy = UploadedSort(rawValue: Settings.userEpisodeSortBy()) ?? UploadedSort.newestToOldest
-
-        if SubscriptionHelper.hasActiveSubscription() {
-            uploadedEpisodes = DataManager.sharedManager.allUserEpisodes(sortedBy: sortBy)
-        } else {
-            uploadedEpisodes = DataManager.sharedManager.allUserEpisodesDownloaded(sortedBy: sortBy)
-        }
+        let uploadedEpisodes = episodesDataManager.uploadedEpisodes()
```
emilylaguna: This results in the episodes no longer being displayed because it's no longer setting the uploadedEpisodes property.

podcasts/EpisodesDataManager.swift
```diff
@@ -0,0 +1,127 @@
+import PocketCastsDataModel
+import PocketCastsServer
+import DifferenceKit
+
+/// Returns a list of episodes for an specific section
+/// Depending on the section, it returns a DifferenceKit ArraySection
```
emilylaguna: This seems like it should be above the episodes for podcast. 

```suggestion
/// Returns a podcasts episodes that are grouped by `PodcastGrouping` 
/// Use `uuidsToFilter` to filter the episode UUIDs to only those in the array
```
Thoughts?

podcasts/EpisodesDataManager.swift
```diff
@@ -0,0 +1,127 @@
...
+class EpisodesDataManager {
+    // MARK: - Podcast episodes list
+
+    func episodes(for podcast: Podcast, uuidsToFilter: [String]? = nil) -> [ArraySection<String, ListItem>] {
+        // the podcast page has a header, for simplicity in table animations, we add it here
+        let searchHeader = ListHeader(headerTitle: L10n.search, isSectionHeader: true)
+        var newData = [ArraySection<String, ListItem>(model: searchHeader.headerTitle, elements: [searchHeader])]
+
+        let tintColor = AppTheme.appTintColor()
```
emilylaguna: This is only used once, can we also update the rest of the AppTheme.appTintColor() references to use this? 

Alternatively, could we just make this the default value since most places this is used right now uses this color as the default?

podcasts/EpisodesDataManager.swift
```diff
@@ -0,0 +1,127 @@
...
+            }, episodeShortKey: { episode -> String in
+                episode.keepEpisode ? L10n.statusStarred : L10n.statusNotStarred
+            })
+            newData.append(contentsOf: groupedEpisodes)
+        }
+
+        return newData
+    }
+
+    func createEpisodesQuery(_ podcast: Podcast, uuidsToFilter: [String]? = nil) -> String {
```
emilylaguna, line 53: Wdyt about making this private to enforce some limitations on episode querying in other places in the app?

podcasts/PlaylistRefreshOperation.swift
```diff
@@ -2,11 +2,13 @@ import Foundation
 import PocketCastsDataModel
 
 class PlaylistRefreshOperation: Operation {
+    private let episodesDataManager: EpisodesDataManager
     private let tableView: UITableView
     private let filter: EpisodeFilter
     private let completion: ([ListEpisode]) -> Void
 
-    init(tableView: UITableView, filter: EpisodeFilter, completion: @escaping (([ListEpisode]) -> Void)) {
+    init(episodesDataManager: EpisodesDataManager = EpisodesDataManager(), tableView: UITableView, filter: EpisodeFilter, completion: @escaping (([ListEpisode]) -> Void)) {
```
emilylaguna: [nit]

```suggestion
    init(episodesDataManager: EpisodesDataManager = .init(), tableView: UITableView, filter: EpisodeFilter, completion: @escaping (([ListEpisode]) -> Void)) {
```

podcasts/EpisodesDataManager.swift
emilylaguna: I agree, the naming of this isn't great since we also have the `EpisodeDataManager`. I think maybe using something like 'EpisodesHelper'?

podcasts/EpisodesDataManager.swift
emilylaguna, line 1: I know the code itself doesn't allow for much freedom here but I feel like this is still VERY view specific since it includes strings, colors, etc. 

Is the goal with this to be able to use the result of this to later query the same episodes? If so, I'm wondering how effective this will be since from my cursory glance it looks like we'll still need to further process the result to be used more programmatically. 

I wonder if it may be more worthwhile to do a further refactor to remove the view elements, and make it more testable. Then only adding the view related items back in when needed?

This isn't something we have to tackle here since it will be a bit more involved, but I wanted to get your thoughts.

emilylaguna (changes requested):
Working well with the exception of one item, left some comments below.

podcasts/EpisodesDataManager.swift
leandroalonso: Ha! I tried that. Turns out CarPlay uses this method, so I thought that changing it there was out of scope for this task.

podcasts/UploadedViewController.swift
leandroalonso: 🤦

Fixed!

podcasts/EpisodesDataManager.swift
leandroalonso: My intention was for it to be the description of the class itself, section being a "part of the app". But now I see how confusing that is, moved to the podcast method.

podcasts/EpisodesDataManager.swift
leandroalonso: Done in 4900b58354ac8072dd46623a489f27360a4f3222

podcasts/EpisodesDataManager.swift
leandroalonso: You are 100% right. To be honest, this is not only very view specific but also very general and has a billion different responsibilities.

> we'll still need to further process the result to be used more programmatically.

We'll definitely have to do this, although it should be simple.

> I wonder if it may be more worthwhile to do a further refactor to remove the view elements, and make it more testable. Then only adding the view related items back in when needed?

I agree. I think we can tackle this on a screen basis. As you mentioned though this is a more extensive task but I'll add it to our board so I can get to it once I finish the groundwork for this feature.

podcasts/EpisodesDataManager.swift
leandroalonso: I agree! It's ok if I change this in a follow-up PR? Just want to avoid a few merging conflicts.

emilylaguna: Sounds good.

emilylaguna (approved):
Tested all the areas this effects:
- ✅ Downloads
- ✅ Files
- ✅ Starred
- ✅ Listening History
- ✅ Podcast Details
- ✅ Filters

All return the correct values. 

🚢

PR #917, opened 2023-06-21
Episodes Autoplay: add a new option in Settings

leandroalonso:
This PR adds a feature flag and an option so the user can toggle on/off the Episodes Autoplay feature.

## To test

1. Run the app
2. Go to Profile > Settings > General and scroll all the way down
3. ✅ You should not see a continuous playback option
4. Go to Profile > Beta Features > enable `autoplay`
5. Then, go to Profile > Settings > General and scroll all the way down
6. ✅ You should see a "Continuous Playback" option toggle **on**
7. Disable the toggle, close the app, and reopen it again
8. Then, go to Profile > Settings > General and scroll all the way down
9. ✅ You should see a "Continuous Playback" option toggled **off**
10. Enable the toggle, close the app, and reopen it again
11. Then, go to Profile > Settings > General and scroll all the way down
12. ✅ You should see a "Continuous Playback" option toggled **on**

Ps.: check all the "General" footers. I did a small refactor on how they're displayed to allow us to move options easily.

## Checklist

### DIFF
--- a/podcasts/GeneralSettingsViewController.swift
+++ b/podcasts/GeneralSettingsViewController.swift
@@ -383,6 +383,8 @@ class GeneralSettingsViewController: UIViewController, UITableViewDelegate, UITa
             return L10n.settingsGeneralMultiSelectGestureSubtitle
         case [.publishChapterTitles]:
             return L10n.settingsGeneralPublishChapterTitlesSubtitle
+        case [.autoplay]:
+            return "Continue playing after an episode ends if Up Next is empty."
         default:
             return nil
         }
--- a/podcasts/GeneralSettingsViewController.swift
+++ b/podcasts/GeneralSettingsViewController.swift
@@ -366,9 +366,9 @@ class GeneralSettingsViewController: UIViewController, UITableViewDelegate, UITa
     }
 
     func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
-        guard let sectionItems = tableData[safe: section]?.last else { return nil }
+        let lastSectionItem = tableData[safe: section]?.last
 
-        switch sectionItems {
+        switch lastSectionItem {
         case .intelligentPlaybackResumption:
             return L10n.settingsGeneralSmartPlaybackSubtitle
         case .playUpNextOnTap:
--- a/podcasts/GeneralSettingsViewController.swift
+++ b/podcasts/GeneralSettingsViewController.swift
@@ -9,8 +9,8 @@ class GeneralSettingsViewController: UIViewController, UITableViewDelegate, UITa
 
     let debounce = Debounce(delay: Constants.defaultDebounceTime)
 
-    private enum TableRow { case skipForward, skipBack, remoteSkipChapters, keepScreenAwake, openPlayer, intelligentPlaybackResumption, defaultRowAction, extraMediaActions, defaultAddToUpNextSwipe, defaultGrouping, defaultArchive, playUpNextOnTap, legacyBluetooth, multiSelectGesture, openLinksInBrowser, publishChapterTitles }
-    private let tableData: [[TableRow]] = [[.defaultRowAction, .defaultGrouping, .defaultArchive, .defaultAddToUpNextSwipe, .openLinksInBrowser], [.skipForward, .skipBack, .keepScreenAwake, .openPlayer, .intelligentPlaybackResumption], [.playUpNextOnTap], [.remoteSkipChapters], [.extraMediaActions], [.legacyBluetooth], [.multiSelectGesture], [.publishChapterTitles]]
+    private enum TableRow { case skipForward, skipBack, remoteSkipChapters, keepScreenAwake, openPlayer, intelligentPlaybackResumption, defaultRowAction, extraMediaActions, defaultAddToUpNextSwipe, defaultGrouping, defaultArchive, playUpNextOnTap, legacyBluetooth, multiSelectGesture, openLinksInBrowser, publishChapterTitles, autoplay }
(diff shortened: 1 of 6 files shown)


podcasts/en.lproj/Localizable.strings
```diff
@@ -2473,12 +2473,18 @@
...
+"settings_general_autoplay" = "Continuous Playback";
+
 /* Subtitle explaining the toggle to modify how a tap is handled in the up next queue. This is used when the toggle is off. */
 "settings_general_up_next_tap_off_subtitle" = "Tapping an episode in Up Next shows the actions page. Long press plays the episode. Turn on to switch these around.";
 
 /* Subtitle explaining the toggle to modify how a tap is handled in the up next queue. This is used when the toggle is on. */
 "settings_general_up_next_tap_on_subtitle" = "Tapping an episode in Up Next will play it. Long press shows episode options. Turn off to switch these around.";
 
+/* Subtitle explaining the toggle to modify if a new episode will be reproduced after the current one ends. */
+"settings_general_autoplay_subtitle" = "Continue playing after an episode ends if Up Next is empty.";
```
yaelirub: It took me a moment to understand this. How about something like: "When on, if your Up Next queue is empty, we'll automatically play episodes from the same podcast or list you are currently listening to." or "If your Up Next queue is empty, we'll play episodes from the same podcast or list you're currently listening to."

podcasts/GeneralSettingsViewController.swift
```diff
@@ -226,6 +230,17 @@ class GeneralSettingsViewController: UIViewController, UITableViewDelegate, UITa
...
 
+            return cell
+
+        case .autoplay:
+            let cell = tableView.dequeueReusableCell(withIdentifier: switchCellId, for: indexPath) as! SwitchCell
+
+            cell.cellLabel.text = L10n.settingsGeneralAutoplay
+            cell.cellSwitch.isOn = Settings.autoplay
+
+            cell.cellSwitch.removeTarget(self, action: nil, for: UIControl.Event.valueChanged)
```
yaelirub: Do you intentionally state the full `UIControl.Event.valueChanged` instead of `.valueChanged`?

podcasts/GeneralSettingsViewController.swift
```diff
@@ -351,23 +366,28 @@ class GeneralSettingsViewController: UIViewController, UITableViewDelegate, UITa
     }
 
     func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
-        if section == 1 {
+        guard let sectionItems = tableData[safe: section] else { return nil }
+
+        switch sectionItems {
```
yaelirub: Since its for footers, can we switch on `sectionItems.last` instead and then we wouldn't have to do the special treatment for `.intelligentPlaybackResumption` and we could remove the array from the cases?

podcasts/Settings.swift
```diff
@@ -784,6 +784,21 @@ class Settings: NSObject {
...
 
+    // MARK: - Autoplay
+
+    static var autoplay: Bool {
+        set {
+            UserDefaults.standard.set(newValue, forKey: Constants.UserDefaults.autoplay)
+        }
+        get {
+            guard let isEnabled = UserDefaults.standard.object(forKey: Constants.UserDefaults.autoplay) as? Bool else {
+                return true
```
yaelirub, line 795: This means its going to be enabled by default, right? All good if so, but not sure I saw that in the project thread

yaelirub (changes requested):

yaelirub:
Your changes to FeatureFlag made me create this issue https://github.com/Automattic/pocket-casts-ios/issues/920 which I hope to work on.

podcasts/en.lproj/Localizable.strings
leandroalonso: I forgot to add a comment asking for input on the option's title and description. I liked the latter, changed in 5461202300d7921c654adb72fa6ecb67b2c11bc7

podcasts/GeneralSettingsViewController.swift
leandroalonso: I just copied/pasted this from the previous one. I changed everything in 898aa3cd05ef8c5c034a6a4c3721397660cad9db to keep it consistent

podcasts/Settings.swift
leandroalonso: Yes, it is! I'll update this info there

podcasts/GeneralSettingsViewController.swift
leandroalonso: Changed in 893a499dfcdb104ad9f9c70acfb23d6e351729e9

podcasts/GeneralSettingsViewController.swift
```diff
@@ -351,23 +366,28 @@ class GeneralSettingsViewController: UIViewController, UITableViewDelegate, UITa
     }
 
     func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
-        if section == 1 {
+        guard let sectionItems = tableData[safe: section]?.last else { return nil }
```
yaelirub: Thats even better then switching on sectionItems.last. Nitpick, but this could be just sectionItem now.

yaelirub (approved):
Looking good! I had one small nitpick comment

podcasts/GeneralSettingsViewController.swift
leandroalonso: I improved it in 5deebe969906b873e8a1cd95111f075a79b5ed6b

I also removed the `guard` since it's not needed

PR #923, opened 2023-06-23
Episodes Autoplay: persist from where the episode started playing

leandroalonso:
Adds logic so the app can know from where an episode started playing and persist it.

## To test

This PR doesn't have any user visible changes, so in order to test add this line to `AutoplayHelper.swift:33`:

```swift
print("🎧 played from: \(playlist)")
```

And this line to `AppDelegate.swift:86`:

```swift
print("🎧 latest played from: \(AutoplayHelper.shared.lastPlaylist)")
```

Then:

1. Run the app
2. Go to Podcasts tab, select any podcast and play an episode by tapping the play button at the right of the episode name
3. ✅ You should see `🎧 played from: podcast(uuid: UUID)` in the console (`UUID` being the podcast UUID from the episode you are playing)
3. Tap any other episode so the episode detail opens and tap play
3. ✅ You should see `🎧 played from: podcast(uuid: UUID)` in the console (`UUID` being the podcast UUID from the episode you are playing)
4. Stop the app and run it again
5. ✅ `🎧 latest played from: Optional(podcasts.EpisodesDataManager.Playlist.podcast(uuid: UUID))` in the console (`UUID` being the podcast UUID from the episode you are playing)
6. Repeat the steps above for: filters (filter tab), Downloads, Files, and Starred

You can also try Siri shortcuts,  Now Playing widget and Listening History: those should not be tracked.

## Checklist


podcasts/PlaybackActionHelper.swift
```diff
@@ -7,6 +7,8 @@ class PlaybackActionHelper {
     class func play(episode: BaseEpisode, filterUuid: String? = nil, podcastUuid: String? = nil) {
         HapticsHelper.triggerPlayPauseHaptic()
 
+        AutoplayHelper.shared.savePlaylist()
```
leandroalonso: I don't like that this is a singleton. I pretend to change that once I make `PlaybackActionHelper` a singleton (it's a collection of static methods right now).

Refactoring this would skyrocket the size of this PR.

podcasts/EpisodeDetailViewController.swift
```diff
@@ -493,3 +493,11 @@ enum EpisodeDetailViewSource: String, AnalyticsDescribable {
 
     var analyticsDescription: String { rawValue }
 }
+
+// MARK: - Autoplay
+
+extension EpisodeDetailViewController: PlaylistAutoplay {
+    var playlist: EpisodesDataManager.Playlist {
+        .podcast(uuid: podcast.uuid)
```
leandroalonso: This logic is wrong. `EpisodeDetailViewController` can be opened from different places, I'm working on a fix.

podcasts/AutoplayHelper.swift
```diff
@@ -0,0 +1,58 @@
...
+    var lastPlaylist: EpisodesDataManager.Playlist? {
+        if let rawDictionary = userDefaults.dictionary(forKey: userDefaultsKey),
+           let dictData = try? JSONSerialization.data(withJSONObject: rawDictionary) {
+            return try? JSONDecoder().decode(EpisodesDataManager.Playlist.self, from: dictData)
+        }
+
+        return nil
+    }
+
+    init(userDefaults: UserDefaults = UserDefaults.standard,
```
emilylaguna:
```suggestion
    init(userDefaults: UserDefaults =.standard,
```

podcasts/Analytics/Helpers/AnalyticsCoordinator.swift
```diff
@@ -59,7 +59,7 @@ class AnalyticsCoordinator {
                 return currentSource
             }
 
-            return (getTopViewController() as? AnalyticsSourceProvider)?.analyticsSource ?? .unknown
+            return (UIApplication.shared.getTopViewController() as? AnalyticsSourceProvider)?.analyticsSource ?? .unknown
```
emilylaguna: Since this isn't being used anymore, do we still need to make this change?

podcasts/AutoplayHelper.swift
```diff
@@ -0,0 +1,53 @@
...
+    func getTopViewController(base: UIViewController?) -> UIViewController?
+}
+
+extension TopViewControllerGetter {
+    func getTopViewController(base: UIViewController? = SceneHelper.rootViewController()) -> UIViewController? {
+        getTopViewController(base: base)
+    }
+}
+
+extension UIApplication: TopViewControllerGetter { }
```
emilylaguna, line 69: Since this isn't being used anymore, do we still need to make this change?

podcasts/AutoplayHelper.swift
```diff
@@ -0,0 +1,53 @@
...
+    private func save(selectedPlaylist playlist: EpisodesDataManager.Playlist?) {
+        guard let playlist else {
+            userDefaults.removeObject(forKey: userDefaultsKey)
+            return
+        }
+
+        if let data = try? JSONEncoder().encode(playlist),
+           let dict = try? JSONSerialization.jsonObject(with: data, options: .allowFragments) {
+            userDefaults.set(dict, forKey: userDefaultsKey)
+        }
```
emilylaguna: Is there a reason we're converting the encoded data into a dictionary before saving it when we could just save the data directly?


```suggestion
        guard let data = try? JSONEncoder().encode(playlist) else {
            return
        }

        userDefaults.set(data, forKey: userDefaultsKey)
```

podcasts/AutoplayHelper.swift
```diff
@@ -0,0 +1,53 @@
...
+    private let userDefaults: UserDefaults
+    private let userDefaultsKey = "playlist"
+
+    var lastPlaylist: EpisodesDataManager.Playlist? {
+        if let rawDictionary = userDefaults.dictionary(forKey: userDefaultsKey),
+           let dictData = try? JSONSerialization.data(withJSONObject: rawDictionary) {
+            return try? JSONDecoder().decode(EpisodesDataManager.Playlist.self, from: dictData)
+        }
+
+        return nil
```
emilylaguna: If we switch to saving the data directly (mentioned below) this can be refactored down to:

```suggestion
        userDefaults.data(forKey: userDefaultsKey).flatMap {
            try? JSONDecoder().decode(EpisodesDataManager.Playlist.self, from: $0)
        }
```

podcasts/EpisodesDataManager.swift
```diff
@@ -3,6 +3,15 @@ import PocketCastsServer
 import DifferenceKit
 
 class EpisodesDataManager {
+    enum Playlist: Codable {
```
emilylaguna: It feels like this struct should be defined in the AutoplayHelper since the autoplay helper is the one actively using it and the episode data manager is just the one providing the data to the helper. 

Thoughts?

podcasts/EpisodesDataManager.swift
```diff
@@ -3,6 +3,15 @@ import PocketCastsServer
...
+        case podcast(uuid: String)
+        case filter(uuid: String)
+        case downloads
+        case files
+        case starred
+        case unknown
```
emilylaguna: Do we have special handling for the unknown case? If not, does it make more sense to remove it and use `nil` as the default like we're doing in some other places?

podcasts/UIApplication+topViewController.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+#if !os(watchOS)
```
emilylaguna: Similar to above, do we still need this since we're not using it anymore?

podcasts/AutoplayHelper.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    init(userDefaults: UserDefaults = UserDefaults.standard) {
```
emilylaguna: [nit]

```suggestion
    init(userDefaults: UserDefaults = .standard) {
```

podcasts/AutoplayHelper.swift
emilylaguna, line 1: Can we also add documentation to the class and methods?

podcasts/AutoplayHelper.swift
```diff
@@ -0,0 +1,53 @@
...
+        self.userDefaults = userDefaults
+    }
+
+    func playedAt(playlist: EpisodesDataManager.Playlist?) {
```
emilylaguna: Wdyt about using `playedFrom` or `func played(from playlist: EpisodesDataManager.Playlist?) {` instead of at? 

At first glance the word `at` had my mind thinking this was referring to playback position and not a playlist source. I think switching to `from` could help clarify it.

podcasts/AutoplayHelper.swift
emilylaguna, line 50: Should we add some debug FileLog's if we're not able to save or restore the playlist? It could help debugging issues in the future.

emilylaguna (changes requested):
Nice work! This is working nicely so far. 

**I tested the following flows:**

- Podcast Details saves: .podcast(uuid:)
    - ✅ When tapping the play button on the episode cell
    - ✅ Tapping an episode and playing from the episode details screen
- Filters saves: .filter(uuid:)
    - ✅ When tapping the play button on the episode cell
    - ✅ Tapping an episode and playing from the episode details screen
- Downloads saves: .downloads
    - ✅ When tapping the play button on the episode cell
    - ✅ Tapping an episode and playing from the episode details screen
- Starred saves: .starred
    - ✅ When tapping the play button on the episode cell
    - ✅ Tapping an episode and playing from the episode details screen
- Files saves: .files
    - ✅ When tapping the play button on the episode cell
    - ✅ Tapping an episode and playing from the episode details screen
- ✅ Playing an episode from Discover resets the saved playlist


**While testing I noticed:**

- If I open my up next → tap on one of a podcast episodes to open the episode details view → tap play
   - The source that is saved is `unknown`
   - Should the playlist source be reset in this case?

- If I add a user file to my up next → open the up next queue → tap on the episode to open the episode details view → tap play
    - The source is always `files`
    - Same as above, should this reset?

podcasts/AutoplayHelper.swift
leandroalonso: Done in 9c24b5dad5e3510d3636a918f931288ba1031626

podcasts/EpisodesDataManager.swift
leandroalonso: That makes a lot more sense, done in c1a128afcd429f99745700f31a2dce68b349201f

podcasts/EpisodesDataManager.swift
leandroalonso: Done in ccd12f2e06190c976bc74444267cc1d570bedd60

podcasts/AutoplayHelper.swift
leandroalonso: I'm not so sure about that. I think the class name is already descriptive (Autoplay). Maybe the only method that might be confusing is `playedFrom`. I added a few comments in 2547a9945a1091bd06b47b642646991ae0d82f2c but I feel they are redundant.

podcasts/AutoplayHelper.swift
leandroalonso: Done in 3e95383aea438434d04a1c29a3747cdfa4f38e84

leandroalonso:
@emilylaguna 

> Should the playlist source be reset in this case?

It is now reset.

> Same as above, should this reset?

This was a bug, it now resets it.

CookieyedCodes:
Will this wipe the current upnext que or dose this logic only apply when hitting play all-filters-/on an empty upnext que, also with this logic, if the last episode is played will it then play the next oldest episode? & Will it skip over archived episodes -I would prefer that tbc- Just wondering 

Also what happens if an upnext que finishes (pre 7.0 apparently it auto played the in progress filter, might be a good default fallback 😉)

(Also just so your aware if people change the play icon to download, long pressing the icon gives the option to stream this logic should also apply their)

Can you also confirm this logic is coming to android?

emilylaguna (approved):
Nice work!

leandroalonso:
@CookieyedCodes autoplay will only trigger if you enable it AND the Up Next queue is empty. And yes, this will come to Android too.

CookieyedCodes:
Great stuff, I hope it skips archived episodes:)

leandroalonso:
@CookieyedCodes is up to you! It will reflect what is appearing on your episode list. :)

CookieyedCodes:
I suppose that makes sence if the show archive enabled it will if not it won't 😉 -good to QC & doubble check this tho 😅 (especially with grouping unplayed) -
(Might be good to get it to stop if it reaches the end of an episode grouping to prevent loops 😅

I probably won't use this feature for a few years but when I will I'll definitely use it when I finally have caught up on stuff/enough stuff❤️ 😉
(Maby sooner the more I think on it tho 🤔🤔)

Some questions 

Is the intention that it will keep automatically occupying the now playing (ie keep checking every time the que is empty) or will the upnext get occupied with all the intended episodes 🤔


& Some more thoughts

if it's the former it might be good to write a note in the empty upnext que that you can play any episode/filter/list with some signage of what is comeing next just to help with user clarity 🤔 -eg a notice like "the upnext is empty but automatically upcomeing is..." (maby a link to the setting too if it's disabled as well because most won't check settings unless informed) and will this empty que thing also be picked up on the webapp/watch app (which is one plus for the latter method as you wouldn't have to worry about that but you lose the good hopping functionality & non destructiveness *deep thoughts*)

(The only thing I would add tho is if an episode is upcomeing make sure it is being auto downloaded at that point not when it's in now playing 🤔 😉 (just to avoid the skip back & D-Ad breakages), seemless is what makes it so elegant 😉 ❤️-)

One nice thing about this elegant solution (either) is it priortises your add to upnext function while being non destructive too what's been added too it. tho I wonder will it resume from the list it was doing or will it go off of the last thing in now playing -this former is more logical & my preference but I can image this being hard to retain that but correct me if I'm wrong 😅-

Sorry for the many edits, this is a real power user potential feature & makes my head think fast 😉 😂

Issue #191, opened 2022-08-29
Apple Watch :: Increase the Episode Limit


joashrajin:
### Expected behavior

The filter episodes list, the downloads list, and the podcast page show allow you to view more than 50 episodes, or add a load more option. 

### Actual behavior

The filter episodes list, the downloads list, and the podcast page are limited to just 50 episodes on the Apple Watch.

### Steps to reproduce the behavior

View a Podcast page which has more than 50 episodes. More context can be found at https://a8c.slack.com/archives/C02A333D8LQ/p1661768420458879

##### Tested on [iPhone 13], iOS [15.6], Pocket Casts [7.20.3]

#5513945-zen

jinnypark:
8722061-zd-a8c
> Can you add the ability to show more than 50 episodes at once on a podcast on the Apple Watch app?

PR #927, opened 2023-06-26
Improve UI responsiveness of the Up Next view on the Apple Watch for large collections

DanReyLop:
Hi folks! I'm learning Swift, and my most used app happens to be open-source, so it was a no-brainer to try this.

I have a large amount of episodes on my Up Next list (~800, I just put my whole library in there), and my Apple Watch S5 struggles when doing any operation with it. Opening the Up Next view takes about 3 seconds, in which the UI is completely frozen. So I gave Instruments a try and got to hacking. I can only test it in the Simulator (my watch doesn't pair with XCode for some reason), but with all these changes, opening Up Next should be instant. If these changes are implemented for the other lists (Podcast Episodes, Filters, Files), maybe the [limit of rows](https://github.com/Automattic/pocket-casts-ios/blob/297a81580766c13005a56bdf3cba758fb5971349/podcasts/Constants.swift#L192) on the Apple Watch can be removed altogether (fixing #191).

I'll explain what every commit does, in order:
- https://github.com/Automattic/pocket-casts-ios/commit/867b24e01e354c6c5d025da8bf40b04d3c0e8b1c: For every episode in the list, [this code](https://github.com/Automattic/pocket-casts-ios/blob/297a81580766c13005a56bdf3cba758fb5971349/Pocket%20Casts%20Watch%20App%20Extension/EpisodeViewModel.swift#L15) checks if it's in UpNext. That performs a linear search on an array of `PlaylistEpisode`s. This commit adds an auxiliary data structure (a `Set`) to speed up that check. The memory impact should be minimal because it only stores the UUIDs. This commit doesn't have a big impact if the rest of the fixes are implemented, but it doesn't hurt.
- https://github.com/Automattic/pocket-casts-ios/commit/aeae11443174bdfa7517886ae45416c2702eb2e9: This is a simple optimization. We want to merge 2 arrays based on the order of a 3rd array. Because both arrays are already sorted, there's no need to do a full array scan for every insertion into the new list. The code is certainly not as elegant, but it's instant (as opposed to maybe 500ms on a real device). I'm open to rewriting it in a more "swift" way if you give me some pointers though.
- https://github.com/Automattic/pocket-casts-ios/commit/2f93b761aa4449371b60ddefd68c3cff4fe7a8fa: The big one. Most of the time when rendering the Up Next view is spend initializing all of the `EpisodeRowViewModel`s. The initialization (I called it `hydration` but I'm not married to that, naming is hard) is now postponed until the last possible moment, when the row view is going to be rendered.
- https://github.com/Automattic/pocket-casts-ios/commit/488400c17bf1f4afbce381621b32878b98162983: This extends the "hydration" concept to Episode/UserEpisode objects too. Instead of building 800+ Episode objects from database records, the Episode objects are created in a minimal way (just with the `uuid`), and they will be "filled" just in time as required. Strictly speaking, 800 calls to the database will be slower than a single big one, but those 800 calls will be spread through a large amount of time if the user decides to scroll through the whole list.

As an alternative to the "hydration" changes, I have [another branch](https://github.com/DanReyLop/pocket-casts-ios/compare/update/upnext-performance...DanReyLop:pocket-casts-ios:try/upnext-performance-lazy-rowmodel-collection) with a different approach.  It involves having a custom collection of `EpisodeRowViewModel`s that will manage fetching the full Episode objects and building the view models. The result is the same, but I feel it's a bit more clunky. I don't like that `EpisodeStub` is a whole different object than `Episode`, for example. Let me know if you see more potential in that solution and I'll give it another think.

Let me know how do you prefer to proceed. Maybe it's worth it to add some comments here and there, and I haven't even checked if tests pass, but I wanted to put this in front of y'all in case you disagree fundamentally with those changes. The first 2 commits can be merged independently, for sure, so let me know if I should open separate PRs.

## To test

- On `trunk`, any of those operations would result in a ~3s freeze (maybe a bit less on newer Watches). With this PR, it should be near-instant.

## Checklist


leandroalonso:
Hey @DanReyLop! Thanks for your contribution.

This PR seems to have a lot of changes, so it will take some time for us to go through and review it properly. Just want to give you a heads-up so you don't think we're ignoring it. :)

DanReyLop:
Yeah, I understand it touched a lot of files. I tried to keep the changes to a minimum, but to completely fix the issue I had to change more than I expected.

As I pointed out, the first 2 commits are small, independent from each other and the rest of the PR, and should be non-controversial. Would it help if I opened separate PRs for those?

Also, the 3rd commit ( https://github.com/Automattic/pocket-casts-ios/commit/2f93b761aa4449371b60ddefd68c3cff4fe7a8fa ) could be merged independently of the 4th ( https://github.com/Automattic/pocket-casts-ios/commit/488400c17bf1f4afbce381621b32878b98162983 ). I personally like having a "big-picture" PR with possible discussion, all the changes, and a clear `Before -> After` improvements, but I understand wanting to review/merge each commit as a separate PR.

CookieyedCodes:
Hmmmm my upnext & podcast screen on apple watch aren't limited since folders were introduced(rusty rewrote the app then) at that I do only have 81 in my upnext tho (but like 1856+ subscriptions & its pretty instant)
I do get lag with filters but I do have an all filter which is probably makeing it chug (& I recommend that limit be kept in place at least on that page, files tho it would be nice not to have the limit)
Like fair enough you have more pods then me in upnext but have you reinstalled the app in the past year?+ (At least since folders, do be aware their is a bug with filters not displaying currently tho) at that oof 800 😅 I'd expect a bit of lag maby 🤔

Regardless any enhancements to my s5 would also be appreciated (hopefully I'll have a new backup phone once a family member upgrades so I can upgrade to watch os 10 I really dig widgits 😅)

DanReyLop:
> Hmmmm my upnext & podcast screen on apple watch aren't limited since folders were introduced

The UpNext isn't limited since about a year ago, but the rest of the lists are. At least filters and podcast episodes, I don't know about Downloads and Files because I have less than 50 on each at the moment. That's with an app directly compiled from `trunk` and freshly installed. But I don't use folders so I don't know if that changes things.

> at that I do only have 81 in my upnext tho (but like 1856+ subscriptions & its pretty instant)

Well, yeah, the lag when opening UpNext only depends on the number of episodes there are in UpNext. I assume with that many subscriptions you'll get more lag in other places, but not in UpNext.

> Like fair enough you have more pods then me in upnext but have you reinstalled the app in the past year?

The app has kept auto-updating as normal. I freshly installed it in the Simulator and I confirmed that it's doing a lot of work when opening a large list of episodes (specifically UpNext). Obviously not as long in the Simulator as in the real Watch (a Core i9 is faster than the Watch CPU), but the bottleneck is there.

>  at that oof 800 😅 I'd expect a bit of lag maby 🤔

There's no reason to expect lag anywhere. Displaying a big list is not rocket science, you just need to be mindful of performance when writing the code. And there's a setting in iOS that allows you to keep auto-adding episodes to UpNext until 1000, so this is using the app within the limits that are expected.

> Regardless any enhancements to my s5 would also be appreciated (hopefully I'll have a new backup phone once a family member upgrades so I can upgrade to watch os 10 I really dig widgits 😅)

Meh, I'm not a fan of the widgets so far, you aren't missing much IMHO. Although they may be cool if you like having a "clean" watchface and still have quick access to complications. To each their own :)

CookieyedCodes:
Ah yes your correct filters & episodes are limited, I would like if the episodes limit was at least lifted 🤔

Downloads & files are limited to 50 each too, files can be at least flipped so you can get 100+, 50 downloads is a limitation of wear os tho maby less so now 🤔🤔

I do find it questionably great when my upnext on watch has updated faster on my wrist then my app XD(android is my main device), well a bit if the bottle neck is they have to rebuild their episode back end, once that's done I find the app quite zippy 🤔 but ya an i9 would be faster 

Ya agreed that below 1000 it should be good but I also expect it a bit due to watch cpu limitations 😅

While fair on widgits I want the option with PC to display the currently playing artwork, have the option to play pause the current episode, mark it as played & the option to star the current episode, doing this outside the app would be lovely 😍😍

leandroalonso:
@DanReyLop sorry for taking too long in getting back.

> Would it help if I opened separate PRs for those?

Could you please do that? I think it will make our lives easier if we separate this into multiple PRs. :)

DanReyLop:
@leandroalonso Done. While I was doing some git gardening, this PR got closed in the process. The new one with the main "hydration" changes is https://github.com/Automattic/pocket-casts-ios/pull/950 . The 2 small, independent improvements are #951 and #952

leandroalonso:
Thank you @DanReyLop!

PR #932, opened 2023-06-28
Episodes Autoplay: play the next episode

leandroalonso:
Adds the autoplay feature to the app.

### Flag disabled

1. Run the app and make sure the `autoplay` feature flag is **disabled**
2. Clear your Up Next queue
3. Play any episode and move the scrubber to the end
4. ✅ Nothing should autoplay after the episode finishes
5. **Enable** the `autoplay` feature flag, go to Profile > Settings > disable "Continuos Playback"
3. Play any episode and move the scrubber to the end
4. ✅ Nothing should autoplay after the episode finishes

#### Empty Up Next

1. Run the app and make sure the `autoplay` feature flag is **enabled** and Profile > Settings > enable "Continuos Playback" 
2. Make sure your Up Next queue is **empty**

Then, use this table as a guide for testing:

| Any podcast screen | The next episode should play (and the app will respect the order you defined) |
| A filter | The next episode in the filter should play |
| Starred screen | The next starred episode should play |
| Listening History | Nothing should be played after the current episode finishes |
| Files | The next episode should play |
| Downloads | The next downloaded episode should play |

#### Not Empty Up Next

1. Add two or three episodes to your Up Next
2. Play them til the end (moving the scrubber to the very end)
3. ✅ After the last one finishes playing nothing should happen

## Checklist


podcasts/PodcastViewController+TableData.swift
```diff
@@ -71,7 +71,11 @@ extension PodcastViewController: UITableViewDataSource, UITableViewDelegate {
         if let listEpisode = itemAtRow as? ListEpisode {
             let cell = tableView.dequeueReusableCell(withIdentifier: PodcastViewController.episodeCellId, for: indexPath) as! EpisodeCell
             cell.hidesArtwork = true
-            cell.playlist = .podcast(uuid: listEpisode.episode.uuid)
+
+            if let podcast {
+                cell.playlist = .podcast(uuid: podcast.uuid)
```
leandroalonso, line 76: ~Curiously, `listEpisode.episode.uuid` returns a wrong UUID 🫠~

I'm stupid, episode.uuid returns the episode UUID, not the PODCAST UUID, which is what we want. Thanks @emilylaguna.

podcasts/EpisodesDataManager.swift
```diff
@@ -3,6 +3,30 @@ import PocketCastsServer
...
 
 class EpisodesDataManager {
+    // MARK: - Playlist episodes
+
+    /// Return the list of episodes for a given playlist
+    func episodes(for playlist: AutoplayHelper.Playlist) -> [BaseEpisode] {
+        switch playlist {
+        case .podcast(uuid: let uuid):
+            if let podcast = DataManager.sharedManager.findPodcast(uuid: uuid, includeUnsubscribed: true) {
+                return Array(episodes(for: podcast).compactMap { $0.elements as? [ListEpisode] }.joined()).map { $0.episode }
```
emilylaguna: I ran into an issue where the autoplay wouldn't work when played from the podcast details when my podcast is grouped by season and sorted newest to oldest. 

After some digging it looks like the data structure returned is:

```swift
[
    ArraySection("Search", [ListHeader()]),
    ArraySection("episodes", [
        ListHeader("No Season"),
        ListEpisode()
        ListEpisode(),
        .....
        ]
    )
]
```

This means that the first compactMap will return nil for the episodes section because not all items are ListEpisodes. 

We can fix this and refactor the code to be a bit faster by using a flatMap to flatten the array, and validating each element of the array.

```suggestion
                return episodes(for: podcast).flatMap { $0.elements.compactMap { ($0 as? ListEpisode)?.episode } }
```

podcasts/EpisodesDataManager.swift
```diff
@@ -3,6 +3,30 @@ import PocketCastsServer
...
+            }
+        case .filter(uuid: let uuid):
+            if let filter = DataManager.sharedManager.findFilter(uuid: uuid) {
+                return episodes(for: filter).compactMap { $0.episode }
+            }
+        case .downloads:
+            return Array(downloadedEpisodes().compactMap { $0.elements }.joined()).map { $0.episode }
```
emilylaguna: We can skip the joining/Array casting step by using `flatMap`:

```suggestion
            return downloadedEpisodes().flatMap { $0.elements.map { $0.episode } }
```

podcasts/EpisodesDataManager.swift
```diff
@@ -3,6 +3,30 @@ import PocketCastsServer
...
+    /// Return the list of episodes for a given playlist
+    func episodes(for playlist: AutoplayHelper.Playlist) -> [BaseEpisode] {
+        switch playlist {
+        case .podcast(uuid: let uuid):
+            if let podcast = DataManager.sharedManager.findPodcast(uuid: uuid, includeUnsubscribed: true) {
+                return Array(episodes(for: podcast).compactMap { $0.elements as? [ListEpisode] }.joined()).map { $0.episode }
+            }
+        case .filter(uuid: let uuid):
+            if let filter = DataManager.sharedManager.findFilter(uuid: uuid) {
+                return episodes(for: filter).compactMap { $0.episode }
```
emilylaguna: [nit] since `episodes(for: filter)` is returning an array without any nil items we can just use `map` here. 


```suggestion
                return episodes(for: filter).map { $0.episode }
```

podcasts/PlaybackManager.swift
```diff
@@ -886,6 +886,19 @@ class PlaybackManager: ServerPlaybackDelegate {
...
+            // If Autoplay is enabled we check if there's another episode to play
+            if Settings.autoplay,
+               queue.upNextCount() == 0,
+               let nextEpisode = AutoplayHelper.shared.nextEpisode(currentEpisodeUuid: episode.uuid) {
+                FileLog.shared.addMessage("Autoplaying next episode: \(nextEpisode.displayableTitle())")
+                queue.add(episode: nextEpisode, fireNotification: false)
+            } else {
+                // Reset the latest played from
+                AutoplayHelper.shared.playedFrom(playlist: nil)
+            }
```
emilylaguna: Wdyt about moving this code block into its own method to increase its visibility

podcasts/AutoplayHelper.swift
```diff
@@ -38,6 +42,23 @@ class AutoplayHelper {
         #endif
     }
 
+    /// Given the current episode UUID, checks if there's any
+    /// episode to play.
+    /// This is done by checking the list from the last place the user
+    /// started playing it, locating the current episode there and
+    /// returning the subsequent one.
+    func nextEpisode(currentEpisodeUuid: String) -> BaseEpisode? {
+        if let lastPlaylist = lastPlaylist {
```
emilylaguna:
```suggestion
        if let lastPlaylist {
```

emilylaguna (changes requested):
So far this is working well, I hit one issue when playing podcasts but elsewhere it seems to be working. 

However I am still testing this fully but I figured I would submit some changes.

podcasts/EpisodesDataManager.swift
emilylaguna: .map is also slightly faster when processing lots of items than .compactMap

Screenshot 2023-06-29 at 4 33 42 PM

podcasts/PlaybackManager.swift
leandroalonso: Done in 1d75527be9381ac105497eb8e8cbcf81ea3dbab6

emilylaguna (approved):
I've been testing this and its been working well! Nice job.

PR #935, opened 2023-06-28
Bookmarks: Add the Headphone Controls item to the settings list

emilylaguna:
| 🎨 **Designs:** [Figma](https://www.figma.com/file/Io7PCz1H1hmOgEE9eljoYW/Bookmarks?type=design&node-id=2098-48475&mode=design&t=fj8Bq5qHTIpZ7BPL-4) |
|:---:|

This adds the basic Headphone Controls setting to the settings list. I also did some refactoring the SettingsViewController to make it easier to hide and show items from the list without having to add another if/else check. 

Items can now hidden by returning `false` in the `var visible: Bool` property on the `TableData` enum. 

I spent some time trying to get some tests working for the refactor but I would have to refactor a lot more to get it to work. 

### Making sure the settings appear correctly
1. Launch the release version of the app
2. Go to Profile > Cog
3. Make note of the items and their order
4. Launch the PR
5. ✅ Verify the order and items is the same

### Headphone Controls item
1. With the Bookmarks Feature Flag disabled
7. Go to Profile > Cog
8. ✅ Verify you do not see the Headphone Controls item under Siri Shortcuts
9. Tap the Beta Features item and Enable the Bookmarks item
10. Go Back
11. ✅ Verify you see the option
12. Tap it
13. ✅ Verify you see `🎧 Coming Soon` in the logs

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #936, opened 2023-06-28
Bookmarks: Adds the Settings properties

emilylaguna:
| 🚨  Depends on: https://github.com/Automattic/pocket-casts-ios/pull/935 |
|:---:|

This adds the underlaying setting values for the headphone/bookmark settings. In normal fashion I also did some refactoring to hopefully make this a bit easier in the future. 

<details>
  <summary><h3>JSONCodable</h3></summary>
  
When storing enum values to the UserDefaults we have [previously](https://github.com/Automattic/pocket-casts-ios/blob/3601468f8c4d1af6a4f6ca20e4eec6815d2f9ded/podcasts/Settings.swift#L14) on using `Int` and storing the `rawValue` but this can end up causing issues if we want to change the order of enum values. The fix would be to manually define each value which, to me, defeats the purpose of magical enums. 

Instead of storing a raw value we can instead encode and decode the value using `JSONEncoder`/`JSONDecoder` but it also results in some extra boilerplate;

```swift
enum HelloWorld: Codable {
    case one, two, three
}

let value: HelloWorld = .two

// Save
let data = try? JSONEncoder().encode(value)
UserDefaults.standard.set(data, forKey: "hello_world")

// Get
if let data = UserDefaults.standard.data(forKey: "hello_world") {
    let value = try? JSONDecoder().decode(HelloWorld.self, from: data) 
    print(value)
}

print(value)

```

Instead we can now simplify it down to:

```swift
enum HelloWorld: JSONCodable {
    case one, two, three
}

let value: HelloWorld = .two
UserDefaults.standard.set(value, forKey: "hello_world")

let value = UserDefaults.standard.jsonObject(HelloWorld.self, forKey: "hello_world)
print(value)

```

If needed objects can override the default encoder/decoder to fit their needs. 

</details>

<details>
  <summary><h3>SettingValue</h3></summary>

Continuing with the trend of storing things in the UserDefaults, I've created a new struct named `SettingValue` that intends to reduce the boilerplate of adding a new setting and make it easier to understand the code. 

Right now, the items we store UserDefaults are defined in:
- The Key is stored in [`Constants.UserDefaults`](https://github.com/Automattic/pocket-casts-ios/blob/3601468f8c4d1af6a4f6ca20e4eec6815d2f9ded/podcasts/Constants.swift#L100)
- The default value is sometimes stored in:
    - [`AppDelegate+Defaults`](https://github.com/Automattic/pocket-casts-ios/blob/3601468f8c4d1af6a4f6ca20e4eec6815d2f9ded/podcasts/AppDelegate%2BDefaults.swift#L17)
    - As a fallback in the `Settings`
        - [Example 1](https://github.com/Automattic/pocket-casts-ios/blob/3601468f8c4d1af6a4f6ca20e4eec6815d2f9ded/podcasts/Settings.swift#L149)
        - [Example 2](https://github.com/Automattic/pocket-casts-ios/blob/3601468f8c4d1af6a4f6ca20e4eec6815d2f9ded/podcasts/Settings.swift#L170)
    - Other places 🤷‍♀️

This aims to unify this by moving the configuration, default value, setting and retrieving to one place. 

Using it is simple, and can be dropped into our current Constants setup. And it works with JSONCodable.

```swift
let setting = SettingValue("hello", defaultValue: "world")

print(setting.value()) -> defaults to "world"
setting.save("hello world")
print(setting.value()) -> prints "hello world"
```

I wrote more about the details and usage in the [documentation comment](https://github.com/Automattic/pocket-casts-ios/blob/8f1d7bd50a03d782f886e008f747e8b037ff475c/podcasts/Constants.swift#L314)
</details>

## To test
There's only code change in this so 👀 on the code and a 🟢 CI build should be enough. 

## Checklist


Modules/Utils/Sources/PocketCastsUtils/Extensions/JSONCodable.swift
```diff
@@ -0,0 +1,86 @@
...
+ This also provides a UserDefaults extension that allows easy reading and writing of
+ encoded values using `set(encodedValue:forKey)` and `jsonObject(type:forKey)`
+
+ The methods behave similar to the default `set(value:forKey)` and `object(forKey:)` methods.
+ */
+
+// MARK: - JSONEncodable
+
+/// A type that can encode itself as JSON data using a JSONEncoder
+public protocol JSONEncodable: Encodable {
```
leandroalonso, line 42: There's any reason why we need an abstraction on top of `Encodable` and `Decodable`?

I think we can highly simplify the logic if we keep using `Codable` in general and I think it increases the readability, we can reduce everything to:

```swift
public extension UserDefaults {
    /// Saves a JSONEncodable object to the UserDefaults for the given key.
    /// Passing nil to this will delete the value.
    func set<Value: Codable>(_ encodedValue: Value?, forKey key: String) {
        setValue(try? JSONEncoder().encode(encodedValue), forKey: key)
    }

    /// Retrieves the JSONDecodable object from the UserDefaults for the given key.
    /// THe value and cast it to the given `type`
    func jsonObject<Value: Decodable>(_ type: Value.Type, forKey: String) -> Value? {
        data(forKey: forKey).flatMap {
            try? JSONDecoder().decode(type, from: $0)
        }
    }
}
```

Then we can make `HeadphoneControlAction` conform to `Codable`.

podcasts/Constants.swift
```diff
@@ -298,6 +310,56 @@ struct Constants {
...
+        let key: String
+        let defaultValue: Value
+
+        init(_ key: String, defaultValue: Value) {
+            self.key = key
+            self.defaultValue = defaultValue
+        }
+
+        /// Retrieve the stored value from the UserDefaults or the `defaultValue` if there isn't a stored value
+        func value(in defaults: Foundation.UserDefaults = .standard) -> Value {
```
leandroalonso: Having the `defaults` as a function argument might open this class to use `save` with one UserDefaults and `value()` with another.

What do you think about having the `UserDefaults` in `init` and changing `value` to be a property?

I think we will eliminate the chance of different defaults happening plus it will be a bit more Swift.

leandroalonso (commented):
Nice changes, left a few ideas so we can remove the abstractions on top of `Codable` and remove some code, let me know what do you think!

Modules/Utils/Sources/PocketCastsUtils/Extensions/JSONCodable.swift
emilylaguna: I originally wrote it like this but this also means any object that conforms to Codable will default to being saved as JSON which includes Bool, String, etc...

I felt like doing that was too much overhead for storing simple values since even a bool will be saved as:

```swift
// true
<74727565>
```

Using the abstraction ensures we're only saving the values we need to save as JSON.

podcasts/Constants.swift
emilylaguna: Makes sense, done here: 91c8d5e77934db58ccf57c9ab3c8a40eb53196e3

leandroalonso (approved):

PR #937, opened 2023-06-28
Bookmarks: Adds PCTableViewController

emilylaguna:
| 🚨  Depends on: https://github.com/Automattic/pocket-casts-ios/pull/936|
|:---:|

This adds a new helper class `PCTableViewController` and the `ReusableTableCell` protocol to make adding UIKit tables easier. 

### PCTableViewController
When adding the new settings view I noticed there is a lot of duplicated code and unnecessary xibs to create a themed table view. Currently the process looks like:
- Create a new subclass of `PCViewController`
- Create a xib
- Add a `ThemeableTable` to the xib
- Setup the rest of the view controller

To reduce this, this creates a `PCTableViewController` class will automatically handle most of that and functions very similar to `UITableViewController`, so now the process looks like:

- Create a new subclass of `PCTableViewController`
- Setup the rest of the view

### ReusableTableCell

Part of the duplicated/boilerplate code that I noticed too was manually defining the `reuseIdentifier` in [so many places](https://github.com/search?q=repo%3AAutomattic%2Fpocket-casts-ios%20disclosureCellId&type=code), and the `ReusableTableCell` hopes to reduce that. 

The `ReusableTableCell` defines 2 properties:
- `static var reuseIdentifier: String { get }`
- `static var nib: UINib? { get }`

The default implementation of the `reuseIdentifier` uses the class name of the conforming object, and that is also used to create the UINib.

The reason for this is because our setup is to have the class name the same as the nib and we are currently just defining the reuseIdentifier as a string literal of the class name anyways. However it can be overridden if needed.

### ReusableTableCell + Dequeuing 

One benefit of the protocol is we can simplify the process of dequeuing the cell and can do away with [having to `as!`](https://github.com/search?q=repo%3AAutomattic%2Fpocket-casts-ios+%22for%3A+indexPath%29+as%21+%22&type=code) to get the correct type. 

We can now use [`dequeueReusableCell(type:for)`](https://github.com/Automattic/pocket-casts-ios/blob/e83b7cbb283bf252734169c70434f5456c01512d/podcasts/PCTableViewController.swift#L96) and specify our ReusableTableCell object and get it as the output. 

```swift

// Before
static let reuseIdentifier = "something"
let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier, for: indexPath) as! Something

// After
let cell = tableView.dequeueReusableCell(Something.self, for: indexPath)
```

### ReusableTableCell + PCTableViewController 

The `PCTableViewController` also makes registering the custom cells easier by providing the `var customCellTypes: [ReusableTableCell.Type]` property that subclasses can override to have the cells automagically registered when the table is created. 

## To test

No visible changes, just 👀 on the code and 🟢 CI

## Checklist


podcasts/PCTableViewController.swift
```diff
@@ -0,0 +1,121 @@
+import UIKit
+
+/// Allows for dynamic creation of reusable table cells.
+/// By default the `reuseIdentifier` is equal to the name of the cell's class
+/// and the nib is loaded from a file with the same name
+///
+protocol ReusableTableCell {
```
leandroalonso, line 7: [nit] I think in the end this is a `ReusableNib`. Anyway, for the context of this task I think it's ok to leave it as it is.

podcasts/PCTableViewController.swift
```diff
@@ -0,0 +1,121 @@
...
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        0
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        UITableViewCell()
+    }
+
+
+    func tableView(_ tableView: UITableView, willDisplayFooterView view: UIView, forSection section: Int) {
```
leandroalonso, line 113: If the class that inherits from `PCTableViewController` wants to implement, let's say, `tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath)` are they able to?

leandroalonso (approved):

podcasts/PCTableViewController.swift
emilylaguna: Yup, all they have to do is implement it as they normally would.

PR #940, opened 2023-06-29
Episodes Autoplay: add CarPlay support

leandroalonso:
Adds CarPlay support so that episodes played in CarPlay are autoplayed depending on whether the user started playing it.

## To test

1. Run this branch on a real phone and use the CarPlay Simulator
2. Make sure the `autoplay` Feature Flag is enabled and "Continuous Playback" is on (Profile > Settings > General > scroll all the way down)
3. make sure your Up Next is empty
3. Use the table below as a guide for testing:

| A podcast list | The next episode should play |
| A filter | The next episode in the filter should play |
| Downloads | The next downloaded episode should play |
| Listening History (under More) | Nothing should be played after the current episode finishes |
| Files (under More) | The next episode should play |

#### Not Empty Up Next

If your Up Next queue has items the next episode in Up Next queue should play after the current one ends.

## Checklist

### DIFF
--- a/podcasts/CarPlaySceneDelegate+Interaction.swift
+++ b/podcasts/CarPlaySceneDelegate+Interaction.swift
@@ -141,7 +141,7 @@ extension CarPlaySceneDelegate {
             guard let self else { return nil }
 
             let episodes = episodeLoader()
-            let episodeItems = self.convertToListItems(episodes: episodes, showArtwork: showArtwork)
+            let episodeItems = self.convertToListItems(episodes: episodes, showArtwork: showArtwork, playlist: playlist)
             return [CPListSection(items: episodeItems)]
         }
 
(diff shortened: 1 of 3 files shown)


podcasts/CarPlaySceneDelegate+Interaction.swift
```diff
@@ -51,17 +51,20 @@ extension CarPlaySceneDelegate {
 
         // Anything else, load the episode and start playing it
         PlaybackManager.shared.load(episode: episode, autoPlay: true, overrideUpNext: false)
+
+        // Store the playlist
+        AutoplayHelper.shared.playedFrom(playlist: playlist)
```
leandroalonso, line 56: Ideally, this will be in a single place — that place being `load`. However, `load` is used in multiple places so I think we can live with this for now.

emilylaguna (approved):
This works in my testing!

PR #942, opened 2023-06-29
Episodes Autoplay: add Watch (Phone source) support

leandroalonso:
Adds Watch (Phone source) support so that episodes played using the Watch are autoplayed depending on whether the user started playing it.

## To test

1. Run this branch on a real phone paired with an Apple Watch
2. Make sure the `autoplay` Feature Flag is enabled and "Continuous Playback" is on (Profile > Settings > General > scroll all the way down)
3. Make sure your Up Next is empty
3. Use the table below as a guide for testing:

| Phone > Filters | The next episode in the filter should play |
| Phone > Downloads | The next downloaded episode should play |
| Phone > Files | The next episode should play |

#### Not Empty Up Next

If your Up Next queue has items the next episode in the Up Next queue should play after the current one finishes.

## Checklist

### DIFF
--- a/Pocket Casts Watch App Extension/SessionManager+Send.swift
+++ b/Pocket Casts Watch App Extension/SessionManager+Send.swift
@@ -16,7 +16,7 @@ extension SessionManager {
         if !WCSession.default.isReachable { return }
 
         let playEpisodeRequest = [WatchConstants.Messages.messageType: WatchConstants.Messages.PlayEpisodeRequest.type, WatchConstants.Messages.PlayEpisodeRequest.episodeUuid: episode.uuid,
-            WatchConstants.Messages.PlayEpisodeRequest.playlist: playlist as Any] as [String: Any]
+            WatchConstants.Messages.PlayEpisodeRequest.playlist: (try? JSONEncoder().encode(playlist)) as Any] as [String: Any]
         WCSession.default.sendMessage(playEpisodeRequest, replyHandler: nil)
     }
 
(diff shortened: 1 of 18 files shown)


Pocket Casts Watch App Extension/SessionManager+Send.swift
```diff
@@ -12,10 +12,11 @@ extension SessionManager {
         sendResponseless(messageType: WatchConstants.Messages.MinorSyncableUpdate.type)
     }
 
-    func play(episode: BaseEpisode) {
+    func play(episode: BaseEpisode, playlist: AutoplayHelper.Playlist?) {
         if !WCSession.default.isReachable { return }
 
-        let playEpisodeRequest = [WatchConstants.Messages.messageType: WatchConstants.Messages.PlayEpisodeRequest.type, WatchConstants.Messages.PlayEpisodeRequest.episodeUuid: episode.uuid] as [String: Any]
+        let playEpisodeRequest = [WatchConstants.Messages.messageType: WatchConstants.Messages.PlayEpisodeRequest.type, WatchConstants.Messages.PlayEpisodeRequest.episodeUuid: episode.uuid,
+            WatchConstants.Messages.PlayEpisodeRequest.playlist: (try? JSONEncoder().encode(playlist)) as Any] as [String: Any]
```
leandroalonso, line 19: We can't send the enum directly, so it needs to be encoded/decoded.

emilylaguna (approved):
This is working for me on a real device!

PR #948, opened 2023-07-03
What's New: add logic to show it

leandroalonso:
Adds logic to show a new "What's New". We currently have code for the current What's New but this one differs in design and I'd like to have i18n as part of it. So reusing a few pieces of information we already have on the app I came up with this solution.

Designs are not part of this task and it will be covered in future tasks.

## To test

1. Review the code
2. Make sure CI is happy green

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
+++ b/PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
@@ -37,4 +37,15 @@ class WhatsNewtests: XCTestCase {
 
         XCTAssertTrue(whatsNew.showIfNeeded())
     }
+
+    /// When opening the app for the first time, show nothing
+    func testDontShowWhenFirstOpening() {
+        let whatsNew = WhatsNew(
+            announcements: [.init(version: 7.41, image: "", title: "", message: "")],
+            previousOpenedVersion: nil,
+            currentVersion: 7.41
+        )
+
+        XCTAssertFalse(whatsNew.showIfNeeded())
+    }
 }
--- a/PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
+++ b/PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
@@ -48,4 +48,15 @@ class WhatsNewtests: XCTestCase {
 
         XCTAssertNil(whatsNew.viewControllerToShow())
     }
+
+    // If the announcement is for a futue version, don't show
+    func testDontShowWhenFutureVersion() {
+        let whatsNew = WhatsNew(
+            announcements: [.init(version: 7.50, image: "", title: "", message: "")],
+            previousOpenedVersion: 7.41,
+            currentVersion: 7.41
+        )
+
+        XCTAssertNil(whatsNew.viewControllerToShow())
+    }
 }
--- a/PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
+++ b/PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
@@ -83,6 +83,18 @@ class WhatsNewtests: XCTestCase {
 
         XCTAssertNotNil(whatsNew.viewControllerToShow())
     }
+
+    // If there's an announcement for the current hotfix and the user
+    // has opened this version yet, show what's new
+    func testDontShowWhatsNewForHotfix() {
+        let whatsNew = WhatsNew(
+            announcements: [.init(version: 7.42, image: "", title: "", message: "")],
+            previousOpenedVersion: 7.42,
+            currentVersion: "7.42.1".toDouble()
+        )
+
+        XCTAssertNil(whatsNew.viewControllerToShow())
+    }
 }
 
 extension String {
--- a/PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
+++ b/PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
@@ -49,7 +49,7 @@ class WhatsNewtests: XCTestCase {
         XCTAssertNil(whatsNew.viewControllerToShow())
     }
 
-    // If the announcement is for a futue version, don't show
+    // If the announcement is for a future version, don't show
     func testDontShowWhenFutureVersion() {
         let whatsNew = WhatsNew(
             announcements: [.init(version: 7.50, image: "", title: "", message: "")],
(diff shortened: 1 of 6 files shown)


podcasts/AppDelegate.swift
```diff
@@ -26,13 +26,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    var whatsNew: WhatsNew?
+
     // MARK: - App Lifecycle
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
         configureFirebase()
         TraceManager.shared.setup(handler: traceHandler)
         FileLog.shared.setup()
 
+        setupWhatsNew()
```
leandroalonso, line 38: This is here mainly because I need the number of the last opened version before the analytics logic overrides it.

But now that I'm look at it, maybe I just need to store something like `lastOpenedVersion`?

This way, the reason why this has to be called before `appLifecycleAnalytics.checkApplicationInstalledOrUpgraded` is not implicit. Any feedback is appreciated.

podcasts/Whats New/WhatsNew.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+struct WhatsNewAnnouncement {
```
emilylaguna: Instead of prefixing this struct wdyt about namespacing it under the WhatsNew class? It achieves the same thing, but imo is a bit cleaner:

```
WhatsNew.Announcement
```

podcasts/Whats New/WhatsNew.swift
```diff
@@ -0,0 +1,29 @@
...
+    let version: Double
```
emilylaguna: Do you think we should handle for a case where an announcement could target a hotfix/patch version number? IE: 7.40.1

podcasts/Whats New/WhatsNew.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    init(announcements: [WhatsNewAnnouncement] = Announcements().announcements, previousOpenedVersion: Double? = UserDefaults.standard.string(forKey: Constants.UserDefaults.lastRunVersion)?.toDouble(), currentVersion: Double = Settings.appVersion().toDouble()) {
+        self.announcements = announcements
+        self.previousOpenedVersion = previousOpenedVersion
+        self.currentVersion = currentVersion
+    }
+
+    func viewControllerToShow() -> UIViewController? {
+        guard let previousOpenedVersion,
+              let announcement = announcements.filter { $0.version >= previousOpenedVersion && $0.version <= currentVersion }.last else {
```
emilylaguna: [nit]
```suggestion
              let announcement = announcements.last(where: { $0.version >= previousOpenedVersion && $0.version <= currentVersion }) else {
```

emilylaguna: Based on this logic a user will always see the announcement if it targets the currently running version, even after they have updated (second launch on the new version)

```suggestion
              let announcement = announcements.filter { $0.version > previousOpenedVersion && $0.version <= currentVersion }.last else {
```

PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
```diff
@@ -0,0 +1,62 @@
...
+        let whatsNew = WhatsNew(
+            announcements: [.init(version: 7.40, image: "", title: "", message: "")],
+            previousOpenedVersion: 7.39,
+            currentVersion: 7.40
+        )
+
+        XCTAssertNotNil(whatsNew.viewControllerToShow())
+    }
+
+    /// When just opening the same version, do nothing
```
emilylaguna, line 18: I mentioned this below, but if we add a test where all 3 versions are the same, `viewControllerToShow` will always return a view controller.

```
    func testDontShowWhatsNewAgain() {
        let whatsNew = WhatsNew(
            announcements: [.init(version: 7.40, image: "", title: "", message: "")],
            previousOpenedVersion: 7.40,
            currentVersion: 7.40
        )

        XCTAssertNil(whatsNew.viewControllerToShow())
    }
```

podcasts/AppDelegate.swift
emilylaguna: Couldn't we just set the WhatsNew via the property which will be init'd before anything is called?

```
let whatsNew = WhatsNew()
```

emilylaguna (changes requested):
Nice work, the tests pass!

I did some additional testing and maybe found one issue where the whats new will repeatedly show even after upgrading. 

Let me know your thoughts

podcasts/Whats New/WhatsNew.swift
leandroalonso: Done in 1b9878fc9c083e1361b5b53a67de274ae088fc28

PocketCastsTests/Tests/Whats New/WhatsNewTests.swift
leandroalonso: Oh! Thanks for spotting that, fixed in c545923d4c104ccb24863db901aa6427e18b5f41

podcasts/Whats New/WhatsNew.swift
leandroalonso: I thought about that but couldn't find any argument to justify it. Hotfixes are meant for critical bug fixes and I wondered if an internationalized message should be deployed on a hotfix.

However, your question makes me wonder how that would work for hotfixes and while implementing tests for it I discovered a bug and it is now fixed. You can see the updated tests in `WhatsNewTests`.

podcasts/AppDelegate.swift
leandroalonso: Not sure I follow, this is what `setupWhatsNew()` already does.

podcasts/Whats New/WhatsNew.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+class WhatsNew {
+    struct Announcement {
+        let version: Double
+        let image: String
+        let title: String
+        let message: String
```
emilylaguna, line 8: We should also specify the button title here as well

leandroalonso: I'm leaving that for a subsequent PR. There's also the button behavior.

podcasts/AppDelegate.swift
emilylaguna: Ah I think I misunderstood. I don't think we need to track another version number in the defaults. But I'm also not super keen on creating a long living object that isn't used that regularly.

Right now I don't have any better ideas and since this is lightweight we can leave it as it is.

emilylaguna (approved):
:shipit:

PR #950, opened 2023-07-05
Improve UI responsiveness of the Up Next view on the Apple Watch for large collections

DanReyLop:
Hi folks! I'm learning Swift, and my most used app happens to be open-source, so it was a no-brainer to try this.

I have a large amount of episodes on my Up Next list (~800, I just put my whole library in there), and my Apple Watch S5 struggles when doing any operation with it. Opening the Up Next view takes about 3 seconds, in which the UI is completely frozen. So I gave Instruments a try and got to hacking. I can only test it in the Simulator (my watch doesn't pair with XCode for some reason), but with all these changes, opening Up Next should be instant. If these changes are implemented for the other lists (Podcast Episodes, Filters, Files), maybe the [limit of rows](https://github.com/Automattic/pocket-casts-ios/blob/297a81580766c13005a56bdf3cba758fb5971349/podcasts/Constants.swift#L192) on the Apple Watch can be removed altogether (fixing #191).

I'll explain what every commit does, in order:
- https://github.com/Automattic/pocket-casts-ios/pull/950/commits/39d81989178c930ea9c34239b2baab946dd46fa7: A quick refactor so other commits are smaller and more focused. Completely initializes the `EpisodeRowViewModel` object properties in the constructor instead of waiting for the main thread. Without this change, with the rest of the commits applied, you would see the rows "fill up" only when you stop scrolling on the list.
- https://github.com/Automattic/pocket-casts-ios/pull/950/commits/916b5ca36088df9d45758b6f11539b97c4eda135: Most of the time when rendering the Up Next view is spend initializing all of the `EpisodeRowViewModel`s. The initialization (I called it `hydration` but I'm not married to that, naming is hard) is now postponed until the last possible moment, when the row view is going to be rendered.
- ~https://github.com/Automattic/pocket-casts-ios/commit/5fb97b312c3c7c8481e250c57da03c666f566094: This extends the "hydration" concept to Episode/UserEpisode objects too. Instead of building 800+ Episode objects from database records, the Episode objects are created in a minimal way (just with the `uuid`), and they will be "filled" just in time as required. Strictly speaking, 800 calls to the database will be slower than a single big one, but those 800 calls will be spread through a large amount of time if the user decides to scroll through the whole list.~ _[Deferred to a future PR]_
- ~https://github.com/Automattic/pocket-casts-ios/commit/12153a12a89ded36bea2338b1d8ce7273fa7aec0: For all the code I could find that only reads the `uuid` of the Episode objects, I just changed the call to ask for the "bare-bones" dehydrated objects. I doubt there will be a dramatic speed difference, but why not.~ _[Deferred to a future PR]_

As an alternative, I have [another branch](https://github.com/DanReyLop/pocket-casts-ios/compare/update/upnext-performance...DanReyLop:pocket-casts-ios:try/upnext-performance-lazy-rowmodel-collection) with a different approach.  It involves having a custom collection of `EpisodeRowViewModel`s that will manage fetching the full Episode objects and building the view models. The result is the same, but I feel it's a bit more clunky. I don't like that `EpisodeStub` is a whole different object than `Episode`, for example. Let me know if you see more potential in that solution and I'll give it another think.

Let me know how do you prefer to proceed. Maybe it's worth it to add some comments here and there, and I haven't even checked if tests pass, but I wanted to put this in front of y'all in case you disagree fundamentally with those changes.

## To test

- On `trunk`, any of those operations would result in a ~3s freeze (maybe a bit less on newer Watches). With this PR, it should be near-instant.

## Checklist

### DIFF
--- a/Pocket Casts Watch App Extension/EpisodeRowViewModel.swift
+++ b/Pocket Casts Watch App Extension/EpisodeRowViewModel.swift
@@ -45,16 +45,16 @@ class EpisodeRowViewModel: EpisodeViewModel, Identifiable {
 
         let info: String
         let statusText: String?
-        switch episode.episodeStatus {
-        case DownloadStatus.downloaded.rawValue:
+        switch DownloadStatus(rawValue: episode.episodeStatus) {
+        case .downloaded:
             self.downloadStatusIconName = "episodedownloaded"
             info = episode.displayableTimeLeft()
             statusText = L10n.statusDownloaded
-        case DownloadStatus.downloading.rawValue:
+        case .downloading:
             self.downloadStatusIconName = nil
             info = episode.displayableInfo(includeSize: false)
             statusText = L10n.statusDownloading
-        case DownloadStatus.downloadFailed.rawValue:
+        case .downloadFailed:
             self.downloadStatusIconName = "downloadfailed"
             informationLabel = []
             accessibilityLabel = [L10n.downloadFailed]
--- a/Pocket Casts Watch App Extension/EpisodeRowViewModel.swift
+++ b/Pocket Casts Watch App Extension/EpisodeRowViewModel.swift
@@ -14,53 +14,59 @@ class EpisodeRowViewModel: EpisodeViewModel, Identifiable {
     override init(episode: BaseEpisode) {
         super.init(episode: episode)
 
+        updateProperties(episode: episode, downloadProgress: downloadProgress)
         Publishers.CombineLatest($episode, $downloadProgress)
             .receive(on: RunLoop.main)
+            .dropFirst()
             .sink(receiveValue: { [unowned self] episode, downloadProgress in
-                var informationLabel = [String]()
-                var accessibilityLabel = [episode.title ?? ""]
-
-                if let publishedDate = episode.publishedDate {
-                    let info = DateFormatHelper.sharedHelper.shortLocalizedFormat(publishedDate)
-                    informationLabel.append(info)
-                    accessibilityLabel.append(info)
-                }
-
-                isDownloading = episode.downloading() || downloadProgress != nil
-
-                let info: String
-                let statusText: String?
-                switch episode.episodeStatus {
-                case DownloadStatus.downloaded.rawValue:
-                    self.downloadStatusIconName = "episodedownloaded"
-                    info = episode.displayableTimeLeft()
-                    statusText = L10n.statusDownloaded
-                case DownloadStatus.downloading.rawValue:
-                    self.downloadStatusIconName = nil
-                    info = episode.displayableInfo(includeSize: false)
-                    statusText = L10n.statusDownloading
-                case DownloadStatus.downloadFailed.rawValue:
-                    self.downloadStatusIconName = "downloadfailed"
-                    informationLabel = []
-                    accessibilityLabel = [L10n.downloadFailed]
-                    info = episode.displayableInfo(includeSize: false)
-                    statusText = nil
-                default:
-                    self.downloadStatusIconName = nil
-                    info = episode.displayableInfo(includeSize: false)
-                    statusText = L10n.statusNotDownloaded
-                }
-
-                informationLabel.append(info)
-                accessibilityLabel.append(info)
-
-                if let statusText = statusText {
-                    accessibilityLabel.append(statusText)
-                }
-
-                self.displayInfo = informationLabel.joined(separator: " • ")
-                self.accessibilityInfo = accessibilityLabel.joined(separator: " , ")
+                updateProperties(episode: episode, downloadProgress: downloadProgress)
             })
             .store(in: &cancellables)
+    }
+
+    private func updateProperties(episode: BaseEpisode, downloadProgress: DownloadProgress?) {
+        var informationLabel = [String]()
+        var accessibilityLabel = [episode.title ?? ""]
+
+        if let publishedDate = episode.publishedDate {
+            let info = DateFormatHelper.sharedHelper.shortLocalizedFormat(publishedDate)
+            informationLabel.append(info)
+            accessibilityLabel.append(info)
+        }
+
+        isDownloading = episode.downloading() || downloadProgress != nil
+
+        let info: String
+        let statusText: String?
+        switch episode.episodeStatus {
+        case DownloadStatus.downloaded.rawValue:
+            self.downloadStatusIconName = "episodedownloaded"
+            info = episode.displayableTimeLeft()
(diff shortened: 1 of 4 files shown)


leandroalonso:
@DanReyLop this one is a big one. :)

First of all, thank you so much for tackling that and we're super happy that you're contributing to Pocket Casts!

While this might seem a straightforward change, it touches a lot of critical parts of the code used by different parts of the app.

I would like to propose a change that might fix the issue while keeping the areas of change as small as possible — so we have less stuff changed and testing is more focused on the main area of the problem.

What do you think about:

1. Create a new method (on the data layer) to return only the UUIDs of episodes in the Up Next (instead of changing the already existing methods);
2. Use this new method to populate the Watch's Up Next;
3. As you did, load (or hydrate) the episode just when it's needed

This way we keep the changes focused on Apple Watch, reduce the scope of tests, make sure we're solving the original issue, and then we can think about what other parts we can improve next — once at a time.

One challenge I foresee by doing that is the inheritance from `EpisodeViewModel` and the fact that this expects a full `BaseEpisode` to be initialized. We can still make this work by returning an anemic `BaseEpisode` (which might be the responsibility of the method I mentioned in number 1).

All in all, my proposal is to try to change less code that is used by a lot of different areas of the app. While the changes seem reasonable, we don't have unit tests covering these areas so there are a lot of risks that things might go wrong.

DanReyLop:
> While this might seem a straightforward change,

Oh, I'm not under any delusion to think this is a simple, straightforward change. I see the number of files I had to touch :/

> All in all, my proposal is to try to change less code that is used by a lot of different areas of the app.

Yeah, that's a perfectly reasonable course of action. I'm fairly confident that my changes work without problems, but I understand wanting to minimize risk. Swift's type checking is doing a lot of heavy lifting here, I come from the Javascript world and it's night and day, I can be fairly sure that a refactor of this size will just work if it compiles successfully. I could get used to this :)

> [...] anemic BaseEpisode

LOL, I like that name :)

> One challenge I foresee by doing that is the inheritance from EpisodeViewModel and the fact that this expects a full BaseEpisode to be initialized.

If we keep the changes on `EpisodeViewModel`/`EpisodeRow`/`EpisodeRowViewModel`, then it doesn't matter if it's initialized with a full `BaseEpisode` or an "anemic" one. I think that's a reasonable compromise because those are relatively simple changes, easy to test (if something doesn't get initialized correctly it would be very noticeable), and contained to the Watch app.

I'll make the changes you suggest:
- Create new methods in the data layer instead of relying on optional parameters of existing methods. This may cause some redundant code, but that's a good trade-off for now.
- Leave all the changes unrelated to the Apple Watch UpNext view out of this PR.

It will take me a bit because I'll need to basically rewrite the whole PR :)

Each of the commits of this PR leaves the app in a good state, what do you think about splitting this PR further? Commits 1 and 2 could be their own PR, commit 3 its own, and commit 4 its own too. After I do the changes you requested, they won't exactly be these commits, obviously. But "hydrating the RowModel", "hydrating the Episode objects" and "use the hydration in other parts of the app" are 3 distinct improvements that could be merged one after the other.

leandroalonso:
> It will take me a bit because I'll need to basically rewrite the whole PR :)

No worries! Again, thank you so much for your patience and for your contributions. :)

> what do you think about splitting this PR further?

If it makes reviewing easier, I'm all for it!

DanReyLop:
In order to make this PR easier to review, I've kept only the first 2 commits. That means the full list of `BaseEpisode` objects are fetched at once from the database, bu at least the `EpisodeListRowModel` objects are lazily initialized. I'll open another PR with the rest of the changes once this one lands. I've also rebased the changes with `trunk`.

@leandroalonso If I understood your feedback correctly, the part where you disagreed about my approach was the changes related with the database and the Episode objects (it's used by many places on the app), so this PR should have no controversial changes :)

I profiled opening the Up Next list on the Watch source, with my current library (around 800 episodes, of which around 50 are Files). I used the watchOS 10 simulator in a 2019 MBP 16" i9. On Instruments, I chose the "Time Profiler" preset, and in the Recording Options I selected "High Frequency" and "Deferred" recording mode.

Profiling for `trunk` (the interesting part is in the Main Thread between seconds 3.8 and 4.5):
trunk

Profiling for this PR (the interesting part is in the Main Thread between seconds 3.4 and 3.8):
pr1

I've attached the full profiling files so you can check for yourselves:
[trunk.trace.zip](https://github.com/Automattic/pocket-casts-ios/files/12497663/trunk.trace.zip)
[pr1.trace.zip](https://github.com/Automattic/pocket-casts-ios/files/12497667/pr1.trace.zip)

leandroalonso:
Hey @DanReyLop! Just to let you know that I have this on my list to review soon. :)

Pocket Casts Watch App Extension/EpisodeRowViewModel.swift
```diff
@@ -11,56 +11,69 @@ class EpisodeRowViewModel: EpisodeViewModel, Identifiable {
...
-                    statusText = L10n.statusNotDownloaded
-                }
+        isDownloading = episode.downloading() || downloadProgress != nil
 
-                informationLabel.append(info)
-                accessibilityLabel.append(info)
+        let info: String
+        let statusText: String?
+        switch episode.episodeStatus {
+        case DownloadStatus.downloaded.rawValue:
```
leandroalonso: I know this is just old code in a new place but we can improve readability:

```suggestion
        case DownloadStatus(rawValue: episode.episodeStatus):
```

This way you can change every `case`. So instead of:

```switch
case DownloadStatus.downloading.rawValue:
```

Just:

```swift
case .downloading:
```

leandroalonso (approved):
@DanReyLop I've been testing this on my watch and it has been working just fine! Thanks for your patience and contribution.

I've just left one minor suggestion for improvement.

Additionally, could you please update the `CHANGELOG.md`?

Pocket Casts Watch App Extension/EpisodeRowViewModel.swift
DanReyLop: Oh, I didn't even *look* at that block of code, I didn't want to mess around with more than strictly needed :)

But that's a good change, done in https://github.com/Automattic/pocket-casts-ios/pull/950/commits/70f4cb57e6f6098bfb2a2b80bb294b4884367edf

I've also rebased with `trunk` and added a changelog entry. I'm still not sure what is changelog-worthy, but this may be noticeable for power users :)

I'll open a follow-up PR with the rest of the changes once this is merged.

PR #951, opened 2023-07-05
Speed up checking if a podcast is in UpNext by caching a Set of all the UUIDs in UpNext

DanReyLop:
For every episode in the UpNext list, [this code](https://github.com/Automattic/pocket-casts-ios/blob/297a81580766c13005a56bdf3cba758fb5971349/Pocket%20Casts%20Watch%20App%20Extension/EpisodeViewModel.swift#L15) checks if it's in UpNext. That performs a linear search on an array of `PlaylistEpisode`s. This PR adds an auxiliary data structure (a `Set`) to speed up that check. The memory impact should be minimal because it only stores the UUIDs. This PR doesn't have a big impact if the rest of the proposed speed improvements are implemented, but it doesn't hurt.

## To test

- On this PR, those operations should freeze the watch UI for a bit less time than in `trunk`. It's more noticeable in a real watch than in the simulator.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 7.44
 -----
 - Add a share action from the episode row [#966]
+- Speed up checking if a podcast is in UpNext, slightly improving app responsiveness [#951]
 
 7.43
 -----
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/UpNextDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/UpNextDataManager.swift
@@ -14,6 +14,7 @@ class UpNextDataManager {
     ]
 
     private var cachedItems = [PlaylistEpisode]()
+    private var allUuids = Set<String>()
     private lazy var cachedItemsQueue: DispatchQueue = {
         let queue = DispatchQueue(label: "au.com.pocketcasts.UpNextItemsQueue")
 
@@ -65,6 +66,12 @@ class UpNextDataManager {
     func playlistEpisodeCount(dbQueue: FMDatabaseQueue) -> Int {
         cachedItemsQueue.sync {
             cachedItems.count
+        }
+    }
+
+    func isEpisodePresent(uuid: String, dbQueue: FMDatabaseQueue) -> Bool {
+        cachedItemsQueue.sync {
+            return allUuids.contains(uuid)
         }
     }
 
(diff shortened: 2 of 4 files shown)


leandroalonso (approved):
@DanReyLop thanks for improving that!

I'm approving the changes but I'd like to provide some feedback:

It would be good to provide some data in future PRs regarding how much of a change improves the code in terms of performance. Just looking at the UI differences might not be enough: there are different hardware, simulators, computers, etc.

Ideally, using `measure` on unit tests would be the best way to retrieve this data: https://developer.apple.com/documentation/xctest/xctestcase/1496290-measure

What I did in my tests though was to run the code from`trunk` and from your change and calculate the average — while scrolling through a podcast where none of the episodes were in the Up Next queue —, comparing both. This is still not ideal but gives a number. :)

With your changes it took `0.00027751922607421875` and in `trunk` it took `0.007229208946228027`.

Last, but not least, could you please edit `CHANGELOG.md` adding this change before we merge? Thanks!

DanReyLop:
> It would be good to provide some data in future PRs regarding how much of a change improves the code in terms of performance. Just looking at the UI differences might not be enough: there are different hardware, simulators, computers, etc.

Yeah, that's an unfortunate side-effect of splitting the big PR into multiple ones. These are incremental changes that eventually take loading UpNext from "a lot" to "instant". I don't remember exactly the improvement that can be attributed to each change, I just remember they were noticeable.

> Ideally, using measure on unit tests would be the best way to retrieve this data:

What do you normally do? Do you create an ad-hoc test that won't be committed? I'll admit that I haven't yet looked at the test setup y'all have.

If it's reasonably easy to create a large UpNext queue as part of a test setup, I'll do that for the big PR. If not, I'll just attach my Instruments runs for `trunk` and the PR, that should paint a pretty clear picture and be easy to look at (I just need to run them again because I didn't have the ~foresight~ *common sense* to save them).

> Last, but not least, could you please edit CHANGELOG.md adding this change before we merge?

I didn't think these kind of changes would merit a changelog entry (they aren't user-facing), but sure! Done.

leandroalonso:
> I didn't think these kind of changes would merit a changelog entry (they aren't user-facing), but sure! Done.

Oh, we need to change this part of the PR checklist! Thank you so much!

leandroalonso (approved):

leandroalonso:
@DanReyLop thank you so much for your patience and this contribution!

PR #952, opened 2023-07-05
Speed up merging the Episodes and UserEpisodes in UpNext

DanReyLop:
On the `allUpNextEpisodes()` function, the list of `Episode`s and `UserEpisode`s need to be merged, based on the order of a 3rd list (of `PlaylistEpisode`s). Both arrays are already sorted, so there's no need to do a full array scan for every insertion into the new list. The code is certainly not as elegant, but it's instant (as opposed to maybe 500ms on a real device). I'm open to rewriting it in a more "swift" way if you give me some pointers though.

## To test

- On this PR, those operations should freeze the watch UI for a bit less time than in `trunk`. It's more noticeable in a real watch than in the simulator.

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
@@ -93,22 +93,20 @@ public class DataManager {
         if userEpisodes.isEmpty {
             return episodes
         }
+
         var convertedEpisodes = [BaseEpisode]()
         var episodeIndex = 0
         var userEpisodeIndex = 0
         for upNextEpisode in allUpNextEpisodes {
-            if let episode = episodes[safe: episodeIndex] {
-                if episode.uuid == upNextEpisode.episodeUuid {
-                    convertedEpisodes.append(episode)
-                    episodeIndex += 1
-                    continue
-                }
+            if let episode = episodes[safe: episodeIndex],
+               episode.uuid == upNextEpisode.episodeUuid {
+                convertedEpisodes.append(episode)
+                episodeIndex += 1
+                continue
             }
-            if let userEpisode = userEpisodes[safe: userEpisodeIndex] {
-                if userEpisode.uuid == upNextEpisode.episodeUuid {
-                    convertedEpisodes.append(userEpisode)
-                    userEpisodeIndex += 1
-                }
+            if let userEpisode = userEpisodes[safe: userEpisodeIndex], userEpisode.uuid == upNextEpisode.episodeUuid {
+                convertedEpisodes.append(userEpisode)
+                userEpisodeIndex += 1
             }
         }
 
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
@@ -90,6 +90,9 @@ public class DataManager {
         let userEpisodes = userEpisodeManager.allUpNextEpisodes(dbQueue: dbQueue)
 
         // this extra step is to make sure we return the episodes in the order they are in the up next list, which they won't be if there's both Episodes and UserEpisodes in Up Next
+        if userEpisodes.isEmpty {
+            return episodes
+        }
         var convertedEpisodes = [BaseEpisode]()
         var episodeIndex = 0
         var userEpisodeIndex = 0
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
@@ -91,10 +91,22 @@ public class DataManager {
 
         // this extra step is to make sure we return the episodes in the order they are in the up next list, which they won't be if there's both Episodes and UserEpisodes in Up Next
         var convertedEpisodes = [BaseEpisode]()
+        var episodeIndex = 0
+        var userEpisodeIndex = 0
         for upNextEpisode in allUpNextEpisodes {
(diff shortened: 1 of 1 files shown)


leandroalonso (changes requested):
@DanReyLop s[imilar to this comment](https://github.com/Automattic/pocket-casts-ios/pull/951#pullrequestreview-1530694629), would you be able to provide some numbers?

I also wonder: do we really need to go through this logic if there are no `userEpisodes` at all?

DanReyLop:
> would you be able to provide some numbers?

I'm on it.

> I also wonder: do we really need to go through this logic if there are no userEpisodes at all?

Technically, no. For people without UserEpisodes in UpNext, it would marginally speed this up even further. But we're talking microseconds here, the big win is to speed up a `O(n^2)` algorithm into an `O(n)` algorithm.

I've added a check for that special case, it doesn't hurt :)

DanReyLop:
Well, this has been an eventful summer, I wasn't unoccupied enough to give these PRs a final push until now :)

I've rebased this PR against `trunk`.

> would you be able to provide some numbers?

The whole DataManager and related classes aren't exactly built to be tested, so I did what I could.

I've put the following test on `PocketCastTests/Tests`, it simulates an UpNext queue similar to mine (800 episodes, of which 50 are files):

```swift
import Foundation
@testable import podcasts
@testable import PocketCastsDataModel
import XCTest
import GameplayKit

class UpNextPerfTest: XCTestCase {
    override func setUp() -> Void {
        var allEpisodes = [BaseEpisode]()
        for _ in 0..<750 {
            let episode = Episode()
            episode.uuid = UUID().uuidString
            episode.podcastUuid = UUID().uuidString
            allEpisodes.append(episode)
        }
        for _ in 0..<50 {
            let episode = UserEpisode()
            episode.uuid = UUID().uuidString
            allEpisodes.append(episode)
        }
        let randomGen = GKMersenneTwisterRandomSource(seed: 1) // Consistent shuffling so tests are repeatable
        allEpisodes = randomGen.arrayByShufflingObjects(in: allEpisodes) as! [BaseEpisode]

        allEpisodes.forEach {
            DataManager.sharedManager.save(episode: $0)
        }
        PlaybackManager.shared.queue.bulkAdd(allEpisodes)
    }

    func testMergeEpisodesAndUserEpisodesForUpNext() {
        measure {
            _ = DataManager.sharedManager.allUpNextEpisodes()
        }
    }
}
```

Average runtime in this PR: 0.0514s
Average runtime in `trunk`: 1.08s

DanReyLop:
Hey @leandroalonso, I was finishing the follow-up to #950 and noticed this PR hasn't been merged yet. Could you give it a final thumbs-up?

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
```diff
@@ -90,11 +90,26 @@ public class DataManager {
         let userEpisodes = userEpisodeManager.allUpNextEpisodes(dbQueue: dbQueue)
 
         // this extra step is to make sure we return the episodes in the order they are in the up next list, which they won't be if there's both Episodes and UserEpisodes in Up Next
+        if userEpisodes.isEmpty {
+            return episodes
+        }
```
leandroalonso, line 95:
```suggestion
        }

```

Nitpick: to slightly improve readibility

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
```diff
@@ -90,11 +90,26 @@ public class DataManager {
...
-            guard let episode: BaseEpisode = episodes.first(where: { $0.uuid == upNextEpisode.episodeUuid }) ?? userEpisodes.first(where: { $0.uuid == upNextEpisode.episodeUuid }) else { continue }
-
-            convertedEpisodes.append(episode)
+            if let episode = episodes[safe: episodeIndex] {
+                if episode.uuid == upNextEpisode.episodeUuid {
+                    convertedEpisodes.append(episode)
+                    episodeIndex += 1
+                    continue
+                }
+            }
```
leandroalonso: You can have the two conditions in a single `if`:

```suggestion
            if let episode = episodes[safe: episodeIndex],
               episode.uuid == upNextEpisode.episodeUuid {
                convertedEpisodes.append(episode)
                episodeIndex += 1
                continue
            }
```

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
```diff
@@ -90,11 +90,26 @@ public class DataManager {
...
+            if let userEpisode = userEpisodes[safe: userEpisodeIndex] {
+                if userEpisode.uuid == upNextEpisode.episodeUuid {
+                    convertedEpisodes.append(userEpisode)
+                    userEpisodeIndex += 1
+                }
+            }
```
leandroalonso: Same as before!

```suggestion
            if let userEpisode = userEpisodes[safe: userEpisodeIndex], userEpisode.uuid == upNextEpisode.episodeUuid {
                convertedEpisodes.append(userEpisode)
                userEpisodeIndex += 1
            }
```

leandroalonso (changes requested):
@DanReyLop sorry that this fell under the cracks! Left some minor suggestions for improving!

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
DanReyLop: WOW! Swift's syntax can still surprise me haha, thanks for this unexpected lesson!

DanReyLop:
@leandroalonso No worries! "Fun" fact, I've just opened https://github.com/Automattic/pocket-casts-ios/pull/1139 as a follow-up and, since I added a new method just for fetching the episodes for the UpNext view, this PR doesn't optimize that use case any more. So, this PR will make *something* faster, but I don't know what. Probably syncing? In any case, it doesn't hurt to merge now that it's reviewed :)

leandroalonso (approved):

PR #956, opened 2023-07-06
What's new: Autoplay

leandroalonso:
This PR adds the "What's New" view for the Autoplay feature.

https://github.com/Automattic/pocket-casts-ios/assets/7040243/d194c679-2a2d-4764-bf7a-38837f1aae01

## To test

1. Run the app
2. Enable the `autoplay` feature flag (Settings > Beta Features)
3. Change `WhatsNew.viewControllerToShow()` to:

```swift
func viewControllerToShow() -> UIViewController? {
        let announcement = announcements.last!

        let whatsNewViewController = ThemedHostingController(rootView: WhatsNewView(announcement: announcement))
        whatsNewViewController.modalPresentationStyle = .overCurrentContext
        whatsNewViewController.modalTransitionStyle = .crossDissolve
        whatsNewViewController.view.backgroundColor = .init(red: 0, green: 0, blue: 0, alpha: 0.5)

        return whatsNewViewController
    }
```

4. Run the app again
5. You should see the popup
6. ✅ Tap "Maybe later"
7. The popup should dismiss
8. Re-run the app
9. Tap "Enable it"
10. ✅ Profile should show, then the general settings and then the screen should scroll to the Autoplay option

Test the "Enable it" flow from the popup appearing above Podcasts/Filters/Discover and above Profile, the flows are slightly different and they all should work.

## Checklist


podcasts/Whats New/Announcements.swift
```diff
@@ -1,15 +1,30 @@
...
     // there were two announcements, the last one will be picked.
     var announcements: [WhatsNew.Announcement] = [
         .init(
-            version: 7.80,
-            image: "",
-            title: "",
-            message: ""
+            version: 7.43,
+            header: {
+                AnyView(AutoplayWhatsNewHeader())
```
leandroalonso, line 12: I know using `AnyView` is bad, but I couldn't think of any other solution. :(

Any input is appreciated. :)

podcasts/en.lproj/Localizable.strings
```diff
@@ -3752,5 +3752,17 @@
...
 "add_bookmark" = "Add Bookmark";
 
+/* Autoplay feature announcement title */
+"announcement_autoplay_title" = "Autoplay is here!";
+
+/* Autoplay feature announcement description */
+"announcement_autoplay_description" = "Enable it and Pocket Casts will autoplay episodes from your current list, if your Up Next queue is empty.";
+
+/* Button label for a feature that the user can enable */
+"enable_it" = "Enable it";
```
emilylaguna: Should this be 'Enable it now' based on [the designs](https://www.figma.com/file/G60UqxPhTynbalo0Vfr1NC/Playground?type=design&node-id=442-6923&mode=design&t=GmD9ZoU0s75yJuw3-4)?

```suggestion
"enable_it" = "Enable it";
```

podcasts/Whats New/WhatsNewView.swift
```diff
@@ -5,14 +5,17 @@ struct WhatsNewView: View {
...
             ZStack(alignment: .topTrailing) {
-                Rectangle()
-                    .frame(height: 195)
+                ZStack {
+                    announcement.header()
+                }
 
                 Button {
-                    NavigationManager.sharedManager.dismissPresentedViewController()
+                    dismiss()
```
emilylaguna: Do we still need the X icon if we have a dismiss button at the bottom now?

podcasts/Whats New/Announcements.swift
emilylaguna: I gave this a shot but couldn't come up with something better than AnyView for now.

podcasts/Whats New/WhatsNew.swift
```diff
@@ -4,9 +4,11 @@ import SwiftUI
 class WhatsNew {
     struct Announcement {
         let version: Double
-        let image: String
+        let header: () -> AnyView
```
emilylaguna, line 7: Does this need to be a closure?

```suggestion
        let header: AnyView
```

podcasts/Whats New/WhatsNewView.swift
```diff
@@ -5,14 +5,17 @@ struct WhatsNewView: View {
 
     let announcement: WhatsNew.Announcement
 
+    @State private var scale = 1.0
```
emilylaguna:
```suggestion
```

emilylaguna (changes requested):
Looking good! I left some comments below.

podcasts/Whats New/WhatsNew.swift
leandroalonso: I thought about being a closure because then the view will only be returned if the closure is called. My concern is that if the property is a view itself, in the future all views might be loaded to the memory — even if we didn't need them.

emilylaguna: Makes sense. 👍

emilylaguna (approved):
:shipit:

podcasts/Whats New/AutoplayWhatsNewHeader.swift
```diff
@@ -0,0 +1,35 @@
+import SwiftUI
+
+struct AutoplayWhatsNewHeader: View {
+    @State private var animate = false
+
+    var body: some View {
+        ZStack {
+            LinearGradient(colors: [.init(hex: "03A9F4"), .init(hex: "50D0F1")], startPoint: .top, endPoint: .bottom)
```
mchowning, line 8: 👋 @leandroalonso ! I think we need to update these gradient colors to be themed so they work with all the themes. For example, this is what radioactive looks like on iOS currently.



I'm using `primaryInteractive01` and `primaryInteractive01Hover` on Android, and @david-gonzalez-a8c said that worked.

PR #968, opened 2023-07-11
Bookmarks: Add the bookmark list to the player tab

emilylaguna:
This adds the base bookmarks list design to the player tab and adds pulling of real data from the database,

There's still a few items to add which will be done in a future PR:
- Multi select
- Empty state
- Navigational routing
- Editing/Deleting

## Screenshots

| iPod Touch | iPhone 14 Pro | iPad |
|:---:|:---:|:---:|
||||

## To test
3. Swipe over to Bookmarks
4. ✅ Verify you see an empty list that says 0 bookmarks
5. Swipe over to the Now Playing
6. Tap the ... button in the player shelf
7. Tap Edit
8. Move the Add Bookmark item to the top of the list
9. Tap done
10. Tap the Add Bookmark item
11. Swipe to the Bookmarks tab
12. ✅ Verify you see the bookmark and the title is the correct timestamp
13. Go back to Now Playing, move to a new position in the episode. Tap the Add Bookmark action
14. Go back to Bookmarks, ✅ Verify it appears
15. Dismiss the player
16. Reopen the player
17. Swipe to the Bookmarks tab
18. ✅ Verify it appears
19. Pause the playing episode
20. Tap the Play button on the bookmarks list
21. ✅ Verify it plays at the correct time. 
22. Switch themes, and ✅ Verify it looks good on each theme.

## Checklist


podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
```diff
@@ -1,44 +1,91 @@
...
+
     var body: some View {
-        ZStack {
-            Color.blue
+        VStack(spacing: 0) {
+            headerView
+            divider
+            scrollView
+        }
+        .environmentObject(viewModel)
```
emilylaguna, line 17: Why pass the viewModel around when I can just inject it into an environment object. 🤔

podcasts/PlayerTabsView.swift
```diff
@@ -36,7 +36,7 @@ class PlayerTabsView: UIScrollView {
         didSet {
             animateTabChange(fromIndex: oldValue, toIndex: currentTab)
 
-            guard oldValue != currentTab, let tab = PlayerTabs(rawValue: currentTab) else {
+            guard oldValue != currentTab, let tab = tabs[safe: currentTab] else {
```
emilylaguna, line 39: I noticed a bug where the wrong tab was being returned here, so I fixed it.

podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
```diff
@@ -1,44 +1,91 @@
...
     }
-}
 
-// MARK: - PlayerItemViewController Wrapper
+    private var scrollView: some View {
+        ZStack(alignment: .top) {
+            ScrollViewWithContentOffset {
+                LazyVStack(spacing: 0) {
+                    //   enumerated to get both the index and the bookmark to hide the last divider
+                    ForEach(Array(viewModel.bookmarks.enumerated()), id: \.element.id) { index, bookmark in
```
emilylaguna, line 44: This is ugly, but it's the only way if I want to get the index and object.

podcasts/Bookmarks/List/BookmarkRow.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    private let bookmark: BookmarkManager.Bookmark
+
+    private let title: String
+    private let subtitle: String
+    private let playButton: String
+
+    @State private var higlighted = false
+
+    init(bookmark: BookmarkManager.Bookmark) {
```
emilylaguna: I'm not super sold on passing the bookmark here cause it feels like it should be a view model, but I'm not 100% sure what would be right.

podcasts/BookmarkManager.swift
```diff
@@ -29,9 +33,12 @@ class BookmarkManager {
...
         let podcastUuid: String? = (episode as? Episode)?.podcastUuid
 
-        dataManager.add(episodeUuid: episode.uuid, podcastUuid: podcastUuid, time: time)
+        let uuid = dataManager.add(episodeUuid: episode.uuid, podcastUuid: podcastUuid, time: time)
 
-        playTone()
+        // Inform the subscribers a bookmark was added
+        uuid.flatMap { dataManager.bookmark(for: $0) }.map {
+            onBookmarkCreated.send((episode, $0))
+        }
```
leandroalonso, line 39: No `playTone()` anymore?

podcasts/Bookmarks/List/BookmarkListViewModel.swift
```diff
@@ -0,0 +1,49 @@
+import Combine
+import PocketCastsDataModel
+
+class BookmarkListViewModel: ObservableObject {
+    let bookmarkManager: BookmarkManager
```
leandroalonso:
```suggestion
    private let bookmarkManager: BookmarkManager
```

podcasts/Bookmarks/List/BookmarkListViewModel.swift
```diff
@@ -0,0 +1,49 @@
...
+        bookmarkCount = bookmarks.count
+        self.bookmarks = bookmarks
+    }
+
+    private func listenForAddedBookmarks() {
+        bookmarkManager.onBookmarkCreated
+            .filter { [weak self] (episode, _) in
+                self?.episode?.uuid == episode.uuid
+            }
+            .sink { [weak self] (_, _) in
```
leandroalonso: [nit]

```suggestion
        bookmarkManager.onBookmarkCreated
            .filter { [weak self] episode, _ in
                self?.episode?.uuid == episode.uuid
            }
            .sink { [weak self] _, _ in
```

podcasts/Bookmarks/List/BookmarkListViewModel.swift
```diff
@@ -0,0 +1,49 @@
...
+            }
+            .store(in: &cancellables)
+    }
+
+    func bookmarkTapped(_ bookmark: BookmarkManager.Bookmark) {
+        print("Tapped")
+    }
+
+    func bookmarkPlayTapped(_ bookmark: BookmarkManager.Bookmark) {
+        PlaybackManager.shared.seekTo(time: bookmark.time, startPlaybackAfterSeek: true)
```
leandroalonso, line 46: What do you think about injecting `PlaybackManager.shared` on `init`? It's ok to leave it as it is if you prefer, but this class is brand new and readable which makes this part even uglier. :P

podcasts/Bookmarks/List/BookmarkRow.swift
```diff
@@ -0,0 +1,104 @@
+import SwiftUI
+import PocketCastsUtils
+import PocketCastsDataModel
+
+struct BookmarkRow: View {
+    @EnvironmentObject var theme: Theme
+    @EnvironmentObject var viewModel: BookmarkListViewModel
+
+    private let bookmark: BookmarkManager.Bookmark
```
leandroalonso: I believe `Bookmark` is a model that can live outside any class — and is not necessarily tied to `BookmarkManager`.

What do you think about extracting it?

podcasts/Bookmarks/List/BookmarkRow.swift
```diff
@@ -0,0 +1,104 @@
...
+    private let title: String
+    private let subtitle: String
+    private let playButton: String
+
+    @State private var higlighted = false
+
+    init(bookmark: BookmarkManager.Bookmark) {
+        self.bookmark = bookmark
+
+        let title = bookmark.title?.isEmpty == false ? bookmark.title : nil
```
leandroalonso, line 20: I wonder if this logic can be on `Bookmark`?

podcasts/Bookmarks/List/BookmarkRow.swift
leandroalonso: I think it's fine. If this `row` grows then we can have a VM for it.

podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
leandroalonso: This is somewhat common! We did that on search too, to avoid manually injecting dependencies over and over again.

podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
leandroalonso, line 44: You can achieve the same result and get rid of this `enumerated` and the `divider` logic if you add the divider on the top (before `BookmarkRow`) and remove the `body` `divider`.

podcasts/Bookmarks/Player/BookmarksPlayerTabController.swift
```diff
@@ -0,0 +1,54 @@
...
+            self?.updateCurrentEpisode()
+        }.store(in: &cancellables)
+    }
+
+    override func willBeRemovedFromPlayer() {
+        viewModel.episode = nil
+    }
+
+    private func updateCurrentEpisode() {
+        viewModel.episode = PlaybackManager.shared.currentEpisode()
```
leandroalonso: Same comment as before, maybe we can inject `PlaybackManager`?

leandroalonso (changes requested):
Worked as described! Left a few comments to improve code

podcasts/BookmarkManager.swift
emilylaguna: Yeah I removed it because I think it playing the tone should be closer to the view model than in a manager.

podcasts/Bookmarks/List/BookmarkListViewModel.swift
emilylaguna: I agree, In a future PR I've replaced this with a Router to allow areas of the app to customize the functionality.

podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
emilylaguna: I actually ended up changing how this works in a future PR.

podcasts/Bookmarks/List/BookmarkRow.swift
emilylaguna: Makes sense, I can move it there once I add the title to the database.

podcasts/Bookmarks/Player/BookmarksPlayerTabController.swift
emilylaguna: Done here: cdd6bd2c823b7f8311678b08c8d5424503183dfe

podcasts/Bookmarks/List/BookmarkRow.swift
emilylaguna: I agree, I was having the same feeling. I've gone ahead and done that here:

- f6ca347cca14263a3a57b50672ce54bd48dc4d81
- 77f82458dddcda40ff234c2693aeb991b7e43a5a
- ba87649556dbc68b902111adbf72ace75fbc352c

I also took the liberty to refactor the code a bit to remove some unnecessary items from the Bookmark struct.

podcasts/Bookmarks/List/BookmarkListViewModel.swift
emilylaguna: Done here: 84da6b376c6b1ba7e367589b106a654d81e48732

podcasts/Bookmarks/List/BookmarkListViewModel.swift
emilylaguna: Done here: c2b95167a26db2d77b8ef37186e44073ff5b2f40

emilylaguna:
@leandroalonso back to you.

podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
leandroalonso: Oh ok!

leandroalonso (approved):

PR #969, opened 2023-07-11
Multi selection: add share action

leandroalonso:
Finishes https://github.com/Automattic/pocket-casts-ios/discussions/962

This PR adds a new "Share" action to the multi episodes selection.

While the change seems simple, this section has a few things that I had to account for:

1. It's customizable
2. The first two items are displayed as the two buttons on the multi-selection UI;

A little demo:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/c7cbcb55-57d0-4850-892a-25e84aa07442

### Showing the new option for previous users

1. Checkout and run `trunk`
2. Go to any podcast and long press any episode
3. Tap "..."
4. Tap "Edit"
5. Reorder and tap "Done"
6. Run this branch
7. Go to any podcast and long press an episode
8. Tap "..."
9. ✅ "Share" should appear

### Testing share action

1. Run this branch
2. Go to any podcast and long press a single episode
3. Tap "..."
4. Tap "Share"
5. ✅ The share sheet should appear
6. Dismiss it
7. Select more episodes
8. ✅ "Share" should not be available (only sharing a single episode is available)

### Reordering

1. Run this branch
2. Go to any podcast and long press a single episode
3. Tap "..."
4. Tap "Edit"
5. Move "Share" to be one of the first two actions
6. Tap "Done"
7. ✅ Share should be displayed on the action bar
8. Tap it
9. ✅ The share sheet should appear
10. Dismiss it
11. Select more episodes
12. ✅ The share button will be replaced by the next available action
13. Tap "..." and then "Edit"
14. Move "Share" to below
15. Tap "Done"
16. ✅ "Share" should not be available as you have multiple selected episodes

### Events

1. Run this branch
2. Go to Profile > Settings > Beta Features > enable `tracksLogging`
3. Go to any podcast and long press any episode
4. Tap Share
5. ✅ When showing the share sheet you should see a `🔵 Tracked: podcast_shared ["type": "episode", "source": "multi_select"]` log

### iPad

Please test the flows on iPad, given the way the share sheet appears is slightly different.

## Checklist

### DIFF
--- a/podcasts/MultiSelectActionController.swift
+++ b/podcasts/MultiSelectActionController.swift
@@ -95,7 +95,7 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
         updateColors()
         setPreferredSize(animated: false)
 
-        removeShareIfMultipleEpisodesAreSelected()
+        filterUnavailableActions()
 
         Analytics.track(.multiSelectViewOverflowMenuShown, properties: ["source": actionDelegate.multiSelectViewSource])
     }
@@ -237,7 +237,7 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
     }
 
     @IBAction func doneTapped(_ sender: UIButton) {
-        removeShareIfMultipleEpisodesAreSelected()
+        filterUnavailableActions()
         setActionsFunc(orderedActions)
         delegate.actionOrderChanged()
         dismiss(animated: true, completion: nil)
@@ -272,10 +272,9 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
         editButton.setTitleColor(AppTheme.colorForStyle(.primaryInteractive01, themeOverride: themeOverride), for: .normal)
     }
 
-    private func removeShareIfMultipleEpisodesAreSelected() {
-        if numSelectedEpisodes > 1 {
-            orderedActions = orderedActions.filter { $0 != .share }
-        }
+    private func filterUnavailableActions() {
+        let episodes = actionDelegate.multiSelectedBaseEpisodes()
+        orderedActions = orderedActions.filter { $0.isVisible(with: episodes) }
     }
 
     private func showAllActions() {
--- a/podcasts/MultiSelectActionController.swift
+++ b/podcasts/MultiSelectActionController.swift
@@ -279,11 +279,15 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
 
     private func showAllActions() {
         orderedActions = originalActions
+        let shareableEpisodesCount = actionDelegate
+            .multiSelectedBaseEpisodes()
+            .filter { $0 is Episode }
+            .count
 
         // If a user already saved the actions, share will be
         // missing since was added later to the app.
-        // This ensures it's displayed.
-        if !orderedActions.contains(.share) {
+        // This ensures it's displayed except for user's files.
+        if !orderedActions.contains(.share) && shareableEpisodesCount > 0 {
             orderedActions.append(.share)
         }
     }
--- a/podcasts/MultiSelectActionController.swift
+++ b/podcasts/MultiSelectActionController.swift
@@ -79,6 +79,7 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
         self.setActionsFunc = setActionsFunc
         self.themeOverride = themeOverride
         super.init(nibName: "MultiSelectActionController", bundle: nil)
+        filterUnavailableActions()
     }
 
     @available(*, unavailable)
@@ -94,8 +95,6 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
         doneButton.isHidden = true
         updateColors()
         setPreferredSize(animated: false)
-
-        filterUnavailableActions()
 
         Analytics.track(.multiSelectViewOverflowMenuShown, properties: ["source": actionDelegate.multiSelectViewSource])
     }
--- a/podcasts/MultiSelectActionController.swift
+++ b/podcasts/MultiSelectActionController.swift
@@ -278,15 +278,14 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
 
     private func showAllActions() {
         orderedActions = originalActions
-        let shareableEpisodesCount = actionDelegate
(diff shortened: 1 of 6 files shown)


podcasts/MultiSelectActionController.swift
```diff
@@ -264,4 +271,21 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
         doneButton.setTitleColor(AppTheme.colorForStyle(.primaryInteractive01, themeOverride: themeOverride), for: .normal)
         editButton.setTitleColor(AppTheme.colorForStyle(.primaryInteractive01, themeOverride: themeOverride), for: .normal)
     }
+
+    private func removeShareIfMultipleEpisodesAreSelected() {
+        if numSelectedEpisodes > 1 {
+            orderedActions = orderedActions.filter { $0 != .share }
```
emilylaguna: Since we're using this logic in a few places what do you think about moving this logic to the enum itself?

We could have:

```
    func isVisible(with episodes: [BaseEpisode]) -> Bool {
        switch self {
        case .share:
            return episodes.count > 1 && episodes.allSatisfy({ $0 is Episode })

        default:
            return true
        }
    }
```

Then this gets replaced with: 

```
let episodes = actionDelegate.multiSelectedBaseEpisodes()
orderedActions = orderedActions.filter { $0.isVisible(with: episodes) }
```

podcasts/MultiSelectFooterView.swift
```diff
@@ -64,13 +64,16 @@ class MultiSelectFooterView: UIView, MultiSelectActionOrderDelegate {
...
+        numberOfEpisodes = 0
         super.init(frame: frame)
         commonInit()
     }
 
     required init?(coder aDecoder: NSCoder) {
+        numberOfEpisodes = 0
         super.init(coder: aDecoder)
         commonInit()
     }
```
emilylaguna: Couldn't we set the default value on the var?

```suggestion
    private var numberOfEpisodes = 0

    override init(frame: CGRect) {
        super.init(frame: frame)
        commonInit()
    }

    required init?(coder aDecoder: NSCoder) {
        super.init(coder: aDecoder)
        commonInit()
    }
```

podcasts/MultiSelectHelper.swift
```diff
@@ -318,6 +321,23 @@ class MultiSelectHelper {
         actionDelegate.multiSelectActionCompleted()
     }
 
+    private class func share(actionDelegate: MultiSelectActionDelegate, view: UIView?) {
+        guard let episode = actionDelegate.multiSelectedBaseEpisodes().first as? Episode else {
```
emilylaguna, line 325: I wonder if we should add an extension to the BaseEpisode that adds a "shareable" property that we can use instead of hardcoding as? Episode. 

Something like: 
```
extension BaseEpisode {
    var shareable: Bool {
        self is Episode
    }
}
```

podcasts/MultiSelectHelper.swift
```diff
@@ -318,6 +321,23 @@ class MultiSelectHelper {
...
+        Analytics.track(.podcastShared, properties: ["type": "episode", "source": "multi_select"])
+
+        if let view {
+            SharingHelper.shared.shareLinkTo(episode: episode, shareTime: 0, fromController: actionDelegate.multiSelectPresentingViewController(), sourceRect: view.bounds, sourceView: view)
+        }
+
+        // If no view is provided, the app is sharing from the bottom sheet
+        if let view = actionDelegate.multiSelectPresentingViewController().view {
+            SharingHelper.shared.shareLinkTo(episode: episode, shareTime: 0, fromController: actionDelegate.multiSelectPresentingViewController(), sourceRect: view.bounds, sourceView: view, showArrow: false)
+        }
```
emilylaguna: Both of the `SharingHelper` methods are called when I share from the overlay button. 

Wdyt about refactoring this to:

```suggestion
        guard let sourceView = view ?? actionDelegate.multiSelectPresentingViewController().view else {
            return
        }

        SharingHelper.shared.shareLinkTo(episode: episode,
                                         shareTime: 0,
                                         fromController: actionDelegate.multiSelectPresentingViewController(),
                                         sourceRect: sourceView.bounds,
                                         sourceView: sourceView,
                                         showArrow: view != nil)
```

emilylaguna (changes requested):
Nice, this works in my testing. 

I noticed that the Share option is available in the Edit menu when I open the multi select from the User's files.

leandroalonso:
@emilylaguna ready for another 👀

podcasts/MultiSelectActionController.swift
```diff
@@ -279,11 +279,15 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
...
+            .multiSelectedBaseEpisodes()
+            .filter { $0 is Episode }
+            .count
 
         // If a user already saved the actions, share will be
         // missing since was added later to the app.
-        // This ensures it's displayed.
-        if !orderedActions.contains(.share) {
+        // This ensures it's displayed except for user's files.
+        if !orderedActions.contains(.share) && shareableEpisodesCount > 0 {
```
leandroalonso: I quite didn't like this solution for user's files :(

Open to any better idea.

podcasts/MultiSelectActionController.swift
```diff
@@ -272,19 +271,22 @@ class MultiSelectActionController: UIViewController, UITableViewDelegate, UITabl
...
+        let episodes = actionDelegate.multiSelectedBaseEpisodes()
+        orderedActions = orderedActions.filter { $0.isVisible(with: episodes) }
     }
 
     private func showAllActions() {
         orderedActions = originalActions
+        let shareableEpisodesCount = actionDelegate
+            .multiSelectedBaseEpisodes()
+            .filter { $0 is Episode }
+            .count
```
emilylaguna: We could just use allSatisfy here instead of filtering and checking the count. 

```suggestion
        let hasShareableEpisodes = actionDelegate.multiSelectedBaseEpisodes().allSatisfy({ $0 is Episode })
```

emilylaguna (approved):
This is working for me now, thanks for fixing the issues. 

I left a small non-blocking suggestion.

podcasts/MultiSelectActionController.swift
leandroalonso: Done in 5aa0a7a83eb3149bf3ce079e53a38790fbc1b4a6

PR #973, opened 2023-07-13
Adds the generic ListViewModel and MutliSelectListViewModel view model parent classes

emilylaguna:
This introduces the `ListViewModel` and `MutliSelectListViewModel` parent classes that represent and handle showing of lists in SwiftUI and providing mutliple selection capabilities without needing to duplicate logic.

The `ListViewModel` is a simple model that defines an items array and a calculates the `numberOfItems` when it changes. It also provides a common way to reload the data using `reload`. 

`MutliSelectListViewModel` adds quite a bit more functionality but encompasses everything that is needed to provide the multiple selection capability to any list view. Views can call `toggleMultiSelection` to enter or exit the selection mode, and use the variety of methods to interact with the selection. 

The `MutliSelectListViewModel` also provides the default long press functionality that will enter multi select from a long press, and select the long pressed item, and show the Select All Above/Below options picker. 

## To test

No functionality was added so just a code review.

## Checklist


yaelirub:
@emilylaguna , such a beautiful work! 👏 
I went over the code, and it looks good, but it would have been great if it included tests. Wdyt?

podcasts/Common SwiftUI/List View/ListViewModel.swift
```diff
@@ -0,0 +1,31 @@
...
+    @Published private(set) var numberOfItems = 0
+
+    // MARK: - Data
+
+    func reload() { /* Subclasses should override this */ }
+
+    // MARK: - Helpers
+
+    /// Whether the given item is the last in the list
+    func itemIsLast(_ item: Model) -> Bool {
```
leandroalonso: Do you think `isLast(item: Model)` is more Swift-y?

This also applies to all other `MutliSelectListViewModel` methods.

podcasts/Common SwiftUI/List View/MutliSelectListViewModel.swift
```diff
@@ -0,0 +1,121 @@
+import Foundation
+
+/// A generic list view model that allows the user to enter a mode that allows them to select multiple items from the list and perform actions on those items
+///
+/// Usage:
+///
+///     class MyListViewModel: MutliSelectListViewModel<MyCustomModel> {
+///         ...
+///     }
+class MutliSelectListViewModel<Model: Hashable>: ListViewModel<Model> {
```
leandroalonso: Typo (on the file name and on the model): `MutliSelectListViewModel` ("mutli")

podcasts/Common SwiftUI/List View/ListViewModel.swift
emilylaguna: Done here: bf6eddb8dd9407e9663f80661f1577dc51325df7

podcasts/Common SwiftUI/List View/MutliSelectListViewModel.swift
emilylaguna: Fixed 😁 816b97a2416dee52c53fea67f3c542a2603b7366

emilylaguna:
@leandroalonso @yaelirub Tests have been added! I also made some more improvements:
-  ListViewModel is now MainActor only to make sure the events are published on the main thread
- I added an convenience init to the ListViewModel to init with an items array
- I removed `reload()` from the ListViewModel since it didn't really do anything outside of providing a common way to reload the data. However this could be deceiving since subclasses can mutate the `items` array directly which could result in changes being ignored until reload is called. Instead I'll leave it up the the subclasses to determine how they want to load their data, and I've made the `MultiSelectListViewModel` reactive to the items array changing. 
- I added a `tapped(item:)` method to the `MultiSelectListViewModel` to provide default functionality of selecting the tapped item.
- I made the `showOptionsPicker` overridable so I can test it, and also allow subclasses the ability to customize it if they'd like. 
- `MultiSelectListViewModel.selectedItems` getter is now exposed to subclasses. 
- I added a change to the `MultiSelectListViewModel` that updates the selectedItems to remove any objects that are no longer available in the `items` array. This will be used when deletion is added.

leandroalonso (approved):

PR #974, opened 2023-07-13
Adds the MultiSelectRow and ActionBarOverlayView views

emilylaguna:
| 🛫 Depends on: https://github.com/Automattic/pocket-casts-ios/pull/973 |
|:---:|

This adds the common MultiSelectRow and ActionBarOverlayView views, along with the `ScaledMetricWithMaxSize` property wrapper. 

`MultiSelectRow` is a list row wrapper that will display the selection checkbox, and inform the parent view when the selection changes. 

`ActionBarOverlayView` displays a simple overlay view with text and action buttons. 

`ScaledMetricWithMaxSize` provides the same functionality as the [@ScaledMetric](https://developer.apple.com/documentation/swiftui/scaledmetric) property wrapper, however it also adds the ability to limit the maximum size of the value.

## To test

Check the SwiftUI previews, and code review. 

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #975, opened 2023-07-13
Bookmarks: Adds the multi select mode and action bar

emilylaguna:
| 🛫 Depends on: https://github.com/Automattic/pocket-casts-ios/pull/974 | 🎨 Designs: [Figma](https://www.figma.com/file/Io7PCz1H1hmOgEE9eljoYW/Bookmarks?type=design&node-id=2032-54831) |
|:---:|:---:|

## Description
This adds the multi select ability to the bookmark list, showing of the action bar in multi select mode, and showing of the Select All/Cancel header. 

The designs have the select all/cancel buttons at the very top of the view, but unfortunately adding that to the player view would require a lot more work since it isn't contained in a navigation controller and would require forcing the buttons there somehow.

For now, I've added those options in replace of the normal header that appears. 

Also, when the action bar is visible and more than 1 item is selected the Edit option is hidden. This is something that I've confirmed with @david-gonzalez-a8c. 

## Demo Video

https://github.com/Automattic/pocket-casts-ios/assets/793774/0e371aa1-1ca8-46ab-b73f-7717e0f2b80f

## To test

3. Tap the ... button in the player shelf
4. Tap Edit
5. Move the Add Bookmark item to the top of the list
6. Tap done
7. Tap the Add Bookmark action a few times to add multiple bookmarks
8. Swipe over to Bookmarks
9. Long press on a row
10. ✅ Verify you enter multi selection mode and the action bar appears
11. Tap on a row, ✅ verify the item is now selected
12. ✅ Verify the action bar selection count updated
13. ✅ Verify the edit option is now hidden
14. Tap the row again to deselect it, ✅ Verify the edit option appears again
15. Tap the Select All button
16. ✅ Verify all the items are selected, and the button now reads Deselect All
17. ✅ Tap it again to, and verify all the items are deselected
18. Tap and hold on a row again
19. ✅ Verify you see the 'Select All Above' and 'Select All Below' options
20. Tap each option and ✅ verify the correct items are selected
21. Tap the Cancel button
22. ✅ Verify the selection buttons are removed

## Checklist

### DIFF
--- a/podcasts/Bookmarks/List/BookmarkListViewModel.swift
+++ b/podcasts/Bookmarks/List/BookmarkListViewModel.swift
@@ -6,6 +6,7 @@ protocol BookmarkListRouter {
     func bookmarkPlay(_ bookmark: Bookmark)
 }
 
+class BookmarkListViewModel: MutliSelectListViewModel<Bookmark> {
     var router: BookmarkListRouter?
 
     private let bookmarkManager: BookmarkManager
@@ -19,15 +20,15 @@ protocol BookmarkListRouter {
 
     init(bookmarkManager: BookmarkManager) {
         self.bookmarkManager = bookmarkManager
+        super.init()
 
         listenForAddedBookmarks()
     }
 
-    func reload() {
-        let bookmarks = episode.map { bookmarkManager.bookmarks(for: $0) } ?? []
+    override func reload() {
+        items = episode.map { bookmarkManager.bookmarks(for: $0) } ?? []
 
-        bookmarkCount = bookmarks.count
-        self.bookmarks = bookmarks
+        super.reload()
     }
 
     private func listenForAddedBookmarks() {
--- a/podcasts/Bookmarks/List/BookmarkListViewModel.swift
+++ b/podcasts/Bookmarks/List/BookmarkListViewModel.swift
@@ -46,6 +46,6 @@ protocol BookmarkListRouter {
     }
 
     func bookmarkPlayTapped(_ bookmark: Bookmark) {
-        PlaybackManager.shared.seekTo(time: bookmark.time, startPlaybackAfterSeek: true)
+        router?.bookmarkPlay(bookmark)
     }
 }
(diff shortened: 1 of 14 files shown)


pocketcasts:
### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
UIFont+FontStyle.swift | 217 | Files should have a single trailing newline. (trailing_newline)


</p>

podcasts/Bookmarks/List/BookmarkListViewModel.swift
```diff
@@ -16,15 +20,15 @@ class BookmarkListViewModel: ObservableObject {
...
     init(bookmarkManager: BookmarkManager) {
         self.bookmarkManager = bookmarkManager
+        super.init()
 
         listenForAddedBookmarks()
     }
 
-    func reload() {
-        let bookmarks = episode.map { bookmarkManager.bookmarks(for: $0) } ?? []
+    override func reload() {
```
leandroalonso: Why this overrides a method?

It seems that `MultiSelectVM` and `ListVM` don't make any use of `reload` (and the method does nothing). Maybe we can just remove this?

leandroalonso (changes requested):
@emilylaguna worked as described!

One [nit] thing: shall we disable the multi-selection if the user swipes to another tab?

podcasts/Bookmarks/List/BookmarkListViewModel.swift
emilylaguna: Haha, :spiderman-point: I removed it from the model. The main reason it overrode before was because when adding deletion I needed to also reflect changes to the multi select model. This has just been replaced with a didSet.

emilylaguna:
@leandroalonso 
> One [nit] thing: shall we disable the multi-selection if the user swipes to another tab?

Annoyingly the view doesn't when the tabs change. I can look into adding a handler for that but I think this PR is large enough so I'll tackle it sometime in the future.

leandroalonso:
> I can look into adding a handler for that but I think this PR is large enough so I'll tackle it sometime in the future.

No worries, ok!

leandroalonso (approved):

PR #979, opened 2023-07-15
Bookmarks: Update the BookmarkManager

emilylaguna:
| 🛫 Depends on: https://github.com/Automattic/pocket-casts-ios/pull/978 |
|:---:|

This updates the BookmarkManager to include the ability to update the title, and provide more specific context to the notifications.

## To test

Code review and 🟢 CI

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
@@ -21,3 +21,13 @@ public struct Bookmark: Hashable {
 extension Bookmark: Identifiable {
     public var id: String { uuid }
 }
+
+extension Bookmark {
+    public static let preview = Bookmark(uuid: "bookmark",
+                                         title: "Interesting Part",
+                                         time: 3600,
+                                         created: Date(),
+                                         modified: Date(),
+                                         episodeUuid: "episode",
+                                         podcastUuid: "podcast")
+}
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
@@ -1,4 +1,5 @@
 import Foundation
+import SwiftUI
 
 /// A bookmark that represents a position in time within an episode
 public struct Bookmark: Hashable {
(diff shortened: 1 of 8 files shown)


Modules/DataModel/Tests/PocketCastsDataModelTests/BookmarkDataManagerTests.swift
```diff
@@ -43,10 +43,10 @@ final class BookmarkDataManagerTests: XCTestCase {
     }
 
     func testAddingExistingBookmarkIsNotAdded() {
-        let firstUuid = addBookmark()
-        let secondUuid = addBookmark()
+        dataManager.add(episodeUuid: "episode-uuid", podcastUuid: "podcast-uuid", title: "Title", time: 1)
 
-        XCTAssertEqual(firstUuid, secondUuid)
+        let second = dataManager.add(episodeUuid: "episode-uuid", podcastUuid: "podcast-uuid", title: "Title", time: 1)
+        XCTAssertNil(second)
```
leandroalonso, line 54: I think it would be good to add a few additional checks that the current `episode-uuid` bookmark didn't have its title, time, etc. changed.

Modules/Utils/Sources/PocketCastsUtils/Extensions/Equatable+When.swift
```diff
@@ -0,0 +1,36 @@
...
+    ///         let success = database.addTheThing()
+    ///
+    ///         if success { sendSomething() }
+    ///
+    ///         return success
+    ///     }
+    ///
+    ///     // After:
+    ///     func doSomething() -> Bool {
+    ///         database.addTheThing().when(true) {
```
leandroalonso, line 22: [nit] While reading it for the first time my brain went: "when true? true what?"

I wonder if we could just do a `database.addTheThing().succeeds { }` or `succeeded { }`, this way it's pretty clear that the block will be executed when `addTheThing()` succeeds and there's no margin to interpret from where the `true` will come from.

Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
```diff
@@ -11,9 +11,23 @@ public struct Bookmark: Hashable {
...
+    }
 }
 
 // MARK: - Identifiable
 extension Bookmark: Identifiable {
     public var id: String { uuid }
 }
+
+extension Bookmark {
+    public static let preview = Bookmark(uuid: "bookmark",
```
leandroalonso: Is this for SwiftUI previews? I wonder if we should extend `PreviewProvider` instead, given this is removed from the final build.

Modules/Utils/Sources/PocketCastsUtils/Extensions/Equatable+When.swift
emilylaguna: I initially went the direction of adding a private extension for Bool that adds an `onSuccess` but I feel like I can see wanting to solve this again in the future outside of just Bool, especially as we move towards more async functional type programming. 

This is why I went with the `.when` extension on Equatable. However I do see your point about the lack of context when first looking at the code. 

Would renaming it from `when` to something more event based like `on` would make more sense? To me seeing `.on(true) { ... }` solves that. 

Ideally in the future it wouldn't be as "generic" as `.on(true)` and would be something more like `something.on(.success){ ... }`.

Modules/DataModel/Tests/PocketCastsDataModelTests/BookmarkDataManagerTests.swift
emilylaguna: Done here: 80e613ead30858da2dbe68e68f5edbb09356eb59

Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
emilylaguna: Done here: 0bd77ac107d84e1eb5e22f91d6ec7471c3c6e6c5

leandroalonso (approved):

PR #989, opened 2023-07-18
Share Extension: add it back and fix behavior

leandroalonso:
1. Adds the share extension back
2. Fix the activation rule (thus not crashing the OS share sheet)
3. Handle the OPML import process

## To test

**📱 Note:** You should test this on a real device.

1. Build and run the app on your device
2. Reboot your device (this is important to re-run all the share extensions code)
3. Go to Photos and share any photo
4. ✅ The share sheet should contain your usual non-Apple apps
5. Open any other podcast app
6. Export the OPML file
7. ✅ When the share sheet appears you should see Pocket Casts
8. Tap on it
9. ✅ The app should open and import the podcasts

## Checklist

### DIFF
--- a/Share Extension/ShareViewController.swift
+++ b/Share Extension/ShareViewController.swift
@@ -1,22 +1,30 @@
 import UIKit
 import Social
 
-class ShareViewController: SLComposeServiceViewController {
+class ShareViewController: UIViewController {
 
-    override func isContentValid() -> Bool {
-        // Do validation of contentText and/or NSExtensionContext attachments here
-        return true
-    }
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
 
-    override func didSelectPost() {
-        // This is called after the user selects Post. Do the upload of contentText and/or NSExtensionContext attachments.
-        // Inform the host that we're done, so it un-blocks its UI. Note: Alternatively you could call super's -didSelectPost, which will similarly complete the extension context.
         self.extensionContext!.completeRequest(returningItems: [], completionHandler: nil)
+
+        redirectToHostApp()
     }
 
-    override func configurationItems() -> [Any]! {
-        // To add configuration options via table cells at the bottom of the sheet, return an array of SLComposeSheetConfigurationItem here.
-        return []
+    func redirectToHostApp() {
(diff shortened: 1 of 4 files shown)


Share Extension/ShareViewController.swift
```diff
@@ -1,22 +1,59 @@
...
+                }
+
+                guard let opmlData = try? Data(contentsOf: url) else {
+                    self?.close()
+                    return
+                }
+
+                self?.close()
+
+                self?.redirectToHostApp(opmlData.base64EncodedString())
```
leandroalonso, line 38: My original approach was to copy the file to the shared directory but our current provisioning profile is not ready for the app group.

I decided to go ahead with the Base 64 approach given it works and [it's somewhat safe](https://stackoverflow.com/a/58869478).

Once the provisioning profile is done I can revert to this commit: https://github.com/Automattic/pocket-casts-ios/pull/989/commits/660f69a17208b63cd755ae757addead291b9fafa

yaelirub: Trying to understand what this code does and why we need it and I think I kinda understand. I think some in code documentation can be helpful. 
`if attachment.hasItemConformingToTypeIdentifier(UTType.data.identifier)` - when will it not conform?

yaelirub:
I followed the testing steps and it works as expected. I'm not sure I understand everything we're doing and why needed special handling to open the an attachment. Would love to go over it

Share Extension/ShareViewController.swift
leandroalonso: > I think some in code documentation can be helpful.

I'll add!

> when will it not conform?

This should be hit only, and if only, the shared file is an `OPML` file (which has the `.data` type). We can be sure of that because the share extension is activated only for `OPML` as defined here: https://github.com/Automattic/pocket-casts-ios/pull/989/files#diff-014aaa0834a60bcc7c5f4da354f564222d6bdcc7ac2da030d5c1e9189bccb00cR11-R22

I'll add documentation to make that clear. :)

leandroalonso: @yaelirub added documentation in 35f0086952a10eae7064cde5b02a1dcfcf388f62

yaelirub (approved):

PR #991, opened 2023-07-18
Bookmarks: Adds a way to display Toast messages

emilylaguna:
| 🎨 Designs: [Figma](https://www.figma.com/file/Io7PCz1H1hmOgEE9eljoYW/Bookmarks?type=design&node-id=2032-52595&mode=dev) |
|:---:|

This adds a way to display a temporary notification overlay message with actions. The view can be dismissed either automatically after a timeout period, or by flicking the message up or down.

The Toast class is designed to be global so it uses a singleton to keep track of the visible toast window and dismiss it when needed. 

Simple notifications can be displayed by using 

```swift
Toast.show("Hello World")
```

## Demo Video

https://github.com/Automattic/pocket-casts-ios/assets/793774/fc492f9c-1aa8-45ac-ac5a-f295311d48d1

## To test

3. Tap on the Add Bookmark player shelf item
   - This has been replaced with a debug action
4. ✅ Verify you see the toast message
5. Wait 5 seconds
6. ✅ Verify the view is automatically hidden
7. Tap the add bookmark action again
8. Swipe up or down to dismiss the action
9. ✅ Verify it fades out and dismisses in the direction you flicked
10. Tap the add bookmark action again
11. Tap the 'View' button in the action
12. ✅ Verify you see 'View Action` printed in console and the toast is hidden automatically

## Checklist


podcasts/Common SwiftUI/Toast/ToastViewModel.swift
```diff
@@ -0,0 +1,75 @@
+import Foundation
+
+protocol ToastCoordinator: AnyObject {
```
leandroalonso: Do you think we can name this `ToastDelegate`? Coordinator always reminds me of the [coordinator pattern](https://www.hackingwithswift.com/articles/71/how-to-use-the-coordinator-pattern-in-ios-apps).

leandroalonso (approved):
Superb job on that @emilylaguna!

I tested on iPad with multiple screens and while rotating the device and everything works as expected! 💯

podcasts/Common SwiftUI/Toast/ToastViewModel.swift
emilylaguna: That makes more sense that coordinator, done here: 4c0e5e522a6ab00bd66faccab60f6d78ab783b15

PR #993, opened 2023-07-19
Bookmarks: Adds Empty State

emilylaguna:
This adds the empty state view when there are no bookmarks to display. This doesn't include the upgrade state yet, but I've written the code to be able to support that in the future. 

## Demo Video

https://github.com/Automattic/pocket-casts-ios/assets/793774/eda40f22-ea63-4783-b285-a073f0ef960c

## To test

2. Play an episode that has no bookmarks
3. Open the Full Screen Player
4. Swipe to the Bookmarks tab
5. ✅ Verify you see the empty state view
6. Swipe to the now playing
7. Add a bookmark
8. Swipe back to the bookmarks tab
9. ✅ Verify the empty state is gone
10. Delete the bookmarks
11. ✅ Verify the empty state view shows

## Checklist

### DIFF
--- a/podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
+++ b/podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
@@ -14,21 +14,29 @@ struct BookmarksPlayerTab: View {
 
     var body: some View {
         VStack(spacing: 0) {
-            if viewModel.items.isEmpty {
-                BookmarksEmptyStateView(style: .playerStyle)
-                Spacer()
-            } else {
-                headerView
-                divider
-
-                actionBarView {
-                    scrollView
-                }
-            }
+            viewModel.items.isEmpty ? emptyView : listView
         }
         .environmentObject(viewModel)
         .padding(.bottom)
         .background(style.background.ignoresSafeArea())
+    }
+
+    /// An empty state view that displays instructions
+    @ViewBuilder
+    private var emptyView: some View {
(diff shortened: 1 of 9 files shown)


podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
```diff
@@ -14,11 +14,16 @@ struct BookmarksPlayerTab: View {
...
+                Spacer()
+            } else {
+                headerView
+                divider
 
-            actionBarView {
-                scrollView
+                actionBarView {
+                    scrollView
+                }
```
leandroalonso: Maybe extract these views to their own properties? Then we can do something like:

```
viewModel.items.isEmpty ? emptyView : listView
```

leandroalonso (approved):

podcasts/Bookmarks/Player/BookmarksPlayerTab.swift
emilylaguna: Nice, done here: 9e5df15fae53130dda03c4a3c1fade25c7eaa8db

leandroalonso (approved):

leandroalonso:
@emilylaguna it looks like CI somehow is not happy with the latest view change, I think we can revert to the prev one. 😬

emilylaguna:
Weird, builds fine for me locally 🤔 

@leandroalonso Fixed here: bbd77b2b80cb3b92b6b57e6ad84d0fd57d59d269

Issue #791, opened 2023-03-30
Podcasts being removed from folders


reginabally:
Internal discussions: p1679634149115029-slack-C02A333D8LQ

### Expected behavior

Podcasts added to the folders are not removed without the user taking any action.

### Actual behavior

Podcasts were removed from folders. Two reports from users and one from @joashrajin. Some of them noticed the issue after updating the app.

User reports:
* 6063249-zd-woothemes
* 6071012-zd-woothemes

##### Both issues were reported from Pocket Casts 7.34

reginabally:
Another report in 6180668-zd-woothemes.

App Version: 7.34
Device: iPhone15,2
OS: 16.3.1

reginabally:
Another report in 6391970-zd-woothemes. I've turned on server loggings for this user and asked them to contact us again when the issue happens.

App Version: 7.39
Device: iPhone15,2
OS: 16.5

joashrajin:
This happened to me while travelling. I just have one or two podcasts in my folders now and the rest are out. I’ve been on some unreliable internet at times so it’s possible that the sync wasn’t successful and that is what stuck. 

Device info:

iPhone 14 Pro Max
Pocket Casts 7.40
iOS 16.5

![IMG_5229](https://github.com/Automattic/pocket-casts-ios/assets/16253818/af439e1b-5dd6-499f-b15a-4ece502b27b7)
![IMG_5228](https://github.com/Automattic/pocket-casts-ios/assets/16253818/2e9d7cea-9f77-4dab-ba7c-fae4b6981656)

igotdes:
+1 in #6378898-zen:

> Out of the blue about 10 podcasts seem to forget which folders they belonged to.  The folders are there, they are still in order, some are still in those folders.but about 10 of them just popped out.
>
> I checked the online app and it seemed to forget ALL the folders and a series of the podcasts I had subscribed to.

reginabally:
> My folders are empty since yesterday
Today I sorted my feeds back to the folders but now they are empty again.

I turned on server logging for this user and we requested them to contact us when the issue happens again.

6391970-zd-woothemes

reginabally:
> Ever since updating to iOS 16.5.1, I’ve noticed podcasts randomly popping out of folders and back in the main list. I’ll put them back and many others will have popped out the next time I open the app.

I turned on server logging for this user and we requested them to contact us when the issue happens again.

6460680-zd-woothemes

[...]

PR #996, opened 2023-07-20
Podcasts out of folders issue: add additional logs

leandroalonso:
Related to #791

While I've been unable to reproduce the podcasts out of folders issue I think that adding additional logs might help us to understand better the problem.

This basically adds additional logs in different parts of the code where a folder from a podcast is changed.

## To test

1. Run the app
2. Create a folder and add podcasts to it
3. Unsubscribe from podcasts
4. ✅ Go to Profile > Settings (top-right cog) > Help & Feedback > Get in touch > Get support > check your "Debug Logs" for the recently added logs

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
@@ -316,6 +316,9 @@ class PodcastDataManager {
     // MARK: - Updates
 
     func save(podcast: Podcast, dbQueue: FMDatabaseQueue) {
+        // Get the existing podcast to compare if folder is being changed
+        let existentPodcast = DataManager.sharedManager.findPodcast(uuid: podcast.uuid)
+
         dbQueue.inDatabase { db in
             do {
                 if podcast.id == 0 {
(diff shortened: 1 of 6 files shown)


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
```diff
@@ -316,6 +316,9 @@ class PodcastDataManager {
     // MARK: - Updates
 
     func save(podcast: Podcast, dbQueue: FMDatabaseQueue) {
+        // Get the existing podcast to compare if folder is being changed
+        let existentPodcast = DataManager.sharedManager.findPodcast(uuid: podcast.uuid)
```
leandroalonso: My initial plan was to release these logs in `7.43`, however, due to this change I think some time out there in beta would be good, a time that unfortunately we don't have right now, so I'm targeting 7.44.

emilylaguna: [nit] 

```suggestion
        let existingPodcast = DataManager.sharedManager.findPodcast(uuid: podcast.uuid)
```

emilylaguna (approved):
I was able to see the new logs when updating folders and syncing. 

I wonder should we also log when a podcast is added to a folder and when a folder is added or deleted from a sync?

leandroalonso:
> I wonder should we also log when a podcast is added to a folder and when a folder is added or deleted from a sync?

When a podcast is added/removed from a folder is already added: https://github.com/Automattic/pocket-casts-ios/pull/996/files#diff-9c2e713f2ea02338ed4b82898f4c80b0e6125e8772bd3a042eb594f4842bfc0dR121

Deleting a folder from a sync was missing and I added here: 6a59e23a866b740de1d740c51888a215168da077

You can test both flows by by logging into another device (or web) and removing/adding a podcast from a folder and removing a folder.

PR #999, opened 2023-07-21
Bookmarks: Add a way to access the bookmarks from the podcast details

emilylaguna:
| 🎨 Designs: [Figma](https://www.figma.com/file/Io7PCz1H1hmOgEE9eljoYW/Bookmarks?type=design&node-id=2033-33461&mode=dev) |
|:---:|

This adds the bookmarks tab button to the podcast details view. 

## Demo Video

https://github.com/Automattic/pocket-casts-ios/assets/793774/4b797571-f227-4e80-8163-a50d1262f463

## To test

1. Launch the app with the bookmarks feature flag enabled
2. Go to the podcasts list and select a podcast
3. ✅ Verify you see the Episodes / Bookmarks "Tab" buttons
4. Tap on the Bookmarks button
5. ✅ Verify the bookmarks placeholder screen is presented
6. Swipe to dismiss it
7. Expand the podcasts details
8. ✅ Verify the tabs appear correctly
9. Go to Profile > Cog > Beta Features
10. Disable Bookmarks
11. Reopen a podcast details view
12. ✅ Verify the tabs don't appear

## Checklist

### DIFF
--- a/podcasts/Bookmarks/Podcast/BookmarksPodcastListView.swift
+++ b/podcasts/Bookmarks/Podcast/BookmarksPodcastListView.swift
@@ -1,0 +1,11 @@
+import SwiftUI
+
+struct BookmarksPodcastListView: View {
+    var body: some View {
+        ZStack {
+            Color.blue
+            Text("COMING SOON")
+        }
+        .ignoresSafeArea()
+    }
+}
--- a/podcasts/EpisodeBookmarksTabsView.swift
+++ b/podcasts/EpisodeBookmarksTabsView.swift
@@ -3,13 +3,14 @@ import SwiftUI
 /// Displays a fake set of tabs that allows the user to open the bookmarks view from the podcast list
 struct EpisodeBookmarksTabsView: View {
     @EnvironmentObject var theme: Theme
-    
+
     weak var delegate: PodcastActionsDelegate?
-    
+
     var body: some View {
         HStack(spacing: 12) {
-            Text(L10n.episodes).applyStyle(theme: theme)
-            
+            Text(L10n.episodes)
+                .applyStyle(theme: theme, highlighted: true)
+
             Text(L10n.bookmarks)
                 .buttonize {
                     delegate?.showBookmarks()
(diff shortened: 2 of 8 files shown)


podcasts/Bookmarks/Podcast/BookmarksPodcastListView.swift
```diff
@@ -0,0 +1,11 @@
+import SwiftUI
+
+struct BookmarksPodcastListView: View {
+    var body: some View {
+        ZStack {
+            Color.blue
+            Text("COMING SOON")
```
leandroalonso, line 7: OK DON'T YELL AT ME

podcasts/EpisodeBookmarksTabsView.swift
```diff
@@ -0,0 +1,50 @@
+import SwiftUI
+
+/// Displays a fake set of tabs that allows the user to open the bookmarks view from the podcast list
+struct EpisodeBookmarksTabsView: View {
```
leandroalonso, line 4: Can't we just call this `RoundedTabsView`? Maybe it's a change we don't necessarily need to do now (I believe this is also gonna be used on the Episode Detail) so we can leave it for later, but the name is a little bit confusing.

podcasts/EpisodeBookmarksTabsView.swift
```diff
@@ -0,0 +1,50 @@
...
+    @EnvironmentObject var theme: Theme
+
+    weak var delegate: PodcastActionsDelegate?
```
leandroalonso, line 7: I also believe we can use a `TabsDelegate` or something like that. But again, we can leave for later.

leandroalonso (approved):
Tested on iOS 15/16, iPad, and iPhone, all works as expected. Left a few comments, none of them are blockers.

podcasts/EpisodeBookmarksTabsView.swift
emilylaguna: I didn't make this generic because of the way this "tab view" works where the episode "button" isn't real and tapping the bookmarks button doesn't change state or anything and just triggers a single action. 

For the episode detail I'll make something better.

emilylaguna: That's also why I am using the `PodcastActionsDelegate` directly.

PR #1000, opened 2023-07-21
Zendesk: prepare for the URL change

leandroalonso:
We really on Zendesk API and our domain is gonna change soon. This PR adds a logic to:

1. In case the request fails to the current domain;
2. It attempts to make the request to the new one

This should be enough to provide a smooth transition once the domain is actually changed.

## To test

1. Run `bundle exec fastlane run configure_apply`
2. Go to `SupportConfig` and invert `baseURL` and `newBaseURL` values (so you can force an error):

```swift
let baseURL = ApiCredentials.zendeskNewUrl
    let newBaseURL = ApiCredentials.zendeskUrl
```

3. Run the app
4. Go to Profile > Settings > Help & Feedback > scroll all the way down > Get in touch > Get Support
5. Write any message and submit
6. ✅ You should see the confirmation message
7. Go to Zendesk and check that your ticket is there

## Checklist


emilylaguna (approved):
This works well, I tested:
- ✅ Fallback work correctly
- ✅ Failure state is still handled
- ✅ Current domain still works correctly

PR #1002, opened 2023-07-22
Bookmarks: Add SearchableListViewModel

emilylaguna:
| 🛫 Depends on: https://github.com/Automattic/pocket-casts-ios/pull/1001 |
|:---:|

This creates a view model that allows searching of the items array. This extends off of the `MultiSelectListViewModel`. 

## To test
Just a code review and 🟢 CI

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1003, opened 2023-07-22
Bookmarks: Add showing of podcast bookmarks

emilylaguna:
| 🎨 Design: [Figma](https://www.figma.com/file/Io7PCz1H1hmOgEE9eljoYW/Bookmarks?type=design&node-id=2033-33810&mode=dev) | 🛫 Depends on: https://github.com/Automattic/pocket-casts-ios/pull/1002 |
|:---:|:---:|

This adds showing of the podcast bookmarks in an overlay view. Since the designs don't quite handle for showing the list in an overlay I combined the look of the original designs, the episode list, and the profile bookmark designs. 

I also didn't see a "No results" view designed so I repurposed the empty state view to show this. 

The designs show, showing of the season and episode number but not all podcasts have this information so I've defaulted to showing the episode title for now. 

## Screenshots
| Default View | Searching | No Results |
|:---:|:---:|:---:|
||||

## Demo Video

https://github.com/Automattic/pocket-casts-ios/assets/793774/f3e7682c-15a9-4ae7-aae9-29a3b59a3116

## To test

2. Open a podcast's details view for a podcast that has some bookmarks
3. Tap the Bookmarks item
4. ✅ Verify you see the bookmarks list
5. Tap the search bar
6. ✅ Verify the keyboard opens and you see the cancel item
7. Tap the cancel item
8. ✅ Verify the keyboard dismisses
9. Tap the search bar again
10. Enter some text that does not appear in your bookmarks
11. ✅ Verify you see the 'No Results' view
12. Tap the 'Clear Search' item in the no results view
13. ✅ Verify the search resets
14. Enter some text that only appears in some of the bookmarks
15. ✅ Verify the items are filtered
16. Press and hold on one of those items
17. Edit the title of the bookmark to something else
18. ✅ Verify that bookmark is then filtered out automatically
19. Tap the Cancel button
20. ✅ Verify all the bookmarks are displayed
21. Tap and hold to enter multi select
22. ✅ Verify the Select All and Cancel buttons appear in the top header
23. ✅ Verify they work
24. Tap Cancel
25. Tap the dismiss icon
26. ✅ Verify the view dismisses
27. Open the bookmarks list again
28. Tap the play button for an episode that is not currently playing
29. ✅ Verify the view dismisses, and the episode starts playing at the bookmark time
30. Open the bookmarks list again
31. Tap on the play button for an episode that is currently playing but at a different time
32. ✅ Verify the view dismisses and the episode jumps to that time

## Checklist

### DIFF
--- a/podcasts/BookmarkManager.swift
+++ b/podcasts/BookmarkManager.swift
@@ -5,6 +5,8 @@ import Combine
 
 class BookmarkManager {
     private let dataManager: BookmarkDataManager
+    private let generalManager: DataManager
+    private let playbackManager: PlaybackManager
 
     /// Called when a bookmark is created
     let onBookmarkCreated = PassthroughSubject<Event.Created, Never>()
@@ -15,8 +17,12 @@ class BookmarkManager {
     /// Called when a value of the bookmark changes
     let onBookmarkChanged = PassthroughSubject<Event.Changed, Never>()
 
-    init(dataManager: BookmarkDataManager = DataManager.sharedManager.bookmarks) {
+    init(dataManager: BookmarkDataManager = DataManager.sharedManager.bookmarks,
+         generalManager: DataManager = .sharedManager,
+         playbackManager: PlaybackManager = .shared) {
         self.dataManager = dataManager
+        self.generalManager = generalManager
+        self.playbackManager = playbackManager
     }
 
     /// Plays the "bookmark created" tone
@@ -76,6 +82,30 @@ class BookmarkManager {
         await dataManager.update(title: title, for: bookmark).when(true) {
             onBookmarkChanged.send(.init(uuid: bookmark.uuid, change: .title(title)))
         }
+    }
+
+    // MARK: - Playback
+
+    /// Play the bookmark
+    func play(_ bookmark: Bookmark) {
+        // If we're already the now playing episode, then just seek to the bookmark time
+        if playbackManager.isNowPlayingEpisode(episodeUuid: bookmark.episodeUuid) {
+            playbackManager.seekTo(time: bookmark.time, startPlaybackAfterSeek: true)
+            return
+        }
+
+        // Get the bookmark's BaseEpisode so we can load it
+        guard let episode = bookmark.episode ?? generalManager.findBaseEpisode(uuid: bookmark.episodeUuid) else {
+            return
+        }
+
+        #if !os(watchOS)
+        // Save the playback time before we start playing so the player will jump to the correct starting time when it does load
+        generalManager.saveEpisode(playedUpTo: bookmark.time, episode: episode, updateSyncFlag: false)
+
+        // Start the play process
+        PlaybackActionHelper.play(episode: episode, podcastUuid: bookmark.podcastUuid)
+        #endif
     }
 
     // MARK: - Named Events
--- a/podcasts/BookmarkManager.swift
+++ b/podcasts/BookmarkManager.swift
@@ -155,6 +155,8 @@ extension BookmarkManager {
     }
 }
 
+// MARK: - BookmarkSortOption
+
 private extension BookmarkSortOption {
     var dataSortOption: BookmarkDataManager.SortOption {
         switch self {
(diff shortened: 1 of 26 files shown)


podcasts/BookmarkManager.swift
```diff
@@ -78,6 +84,30 @@ class BookmarkManager {
         }
     }
 
+    // MARK: - Playback
+
+    /// Play the bookmark
+    func play(_ bookmark: Bookmark) {
```
emilylaguna: I thought about where to put this, but this made the most sense since it's central and specific to bookmarks.

leandroalonso: Hmmm, that's indeed a tricky one. However, I wonder if `PlaybackManager` wouldn't be the correct place to have this.

Perhaps, a new method like `PlaybackManager.play(episodeUuid: String, time: TimeInterval)`. This method knows a lot about `PlaybackManager`, which I think is a signal maybe this might belong there.

Also, by doing this you can get rid of the `DataManager` from here.

Thoughts?

podcasts/Bookmarks/List/BookmarkListViewModel.swift
```diff
@@ -4,9 +4,18 @@ import PocketCastsDataModel
 protocol BookmarkListRouter: AnyObject {
     func bookmarkPlay(_ bookmark: Bookmark)
     func bookmarkEdit(_ bookmark: Bookmark)
+    func dismissBookmarksList()
+
+    var alertController: UIViewController? { get }
```
leandroalonso: This isn't really an alert controller, but more of a `UIViewController` in which an actual alert controller will be presented, right?

I think a more generic name + a comment can help clarify that.

podcasts/Bookmarks/List/BookmarkListViewModel.swift
```diff
@@ -160,3 +182,9 @@ private extension BookmarkSortOption {
         }
     }
 }
+
+extension Bookmark: SearchableDataModel {
+    var searchField: String {
+        [title, episode?.title].compactMap { $0 }.joined()
```
leandroalonso: This will join both strings without a separator between them. While in most cases this shouldn't be a problem if the user searches for the bookmark title and episode title with a space between them the bookmark might not return for them.

podcasts/Bookmarks/List/BookmarkPodcastListViewModel.swift
```diff
@@ -0,0 +1,46 @@
+import Combine
+import PocketCastsDataModel
+
+class BookmarkPodcastListViewModel: BookmarkListViewModel {
+    weak var podcast: Podcast?
```
leandroalonso: Why this needs to be `weak`?

podcasts/Bookmarks/List/BookmarkRowViewModel.swift
```diff
@@ -0,0 +1,45 @@
...
+        _episodeImage = .init(initialValue: imageManager.placeHolderImage(.list))
+
+        self.heading = (episode as? Episode).flatMap {
+            $0.title
+        }
+
+        loadImage(for: episode)
+    }
+
+    private func loadImage(for episode: BaseEpisode) {
```
leandroalonso: [nit] While I believe this logic belongs here, I think we can use episode images in a lot of other places, so I wonder if a component specific for that — or an `Image` extension might be a good approach?

leandroalonso (commented):
@emilylaguna tested and it works as described!

One small issue with iPad though: as opposed to iPhone, it seems that the "visible area" don't resize when the keyboard appears:

Screen Shot 2023-07-24 at 13 45 35

emilylaguna:
@leandroalonso 

> One small issue with iPad though: as opposed to iPhone, it seems that the "visible area" don't resize when the keyboard appears:

I tested this on iOS 16 and iOS 15 and the list resizes correctly when the keyboard opens and I can access all the items still. 
iPad Mini 15.5


https://github.com/Automattic/pocket-casts-ios/assets/793774/fb1c671c-55d9-407c-8394-938c9b16c63a

podcasts/Bookmarks/List/BookmarkRowViewModel.swift
emilylaguna: That makes sense, I moved this into a separate component called `EpisodeImage`: dcded9238f3c94d7f0af386d6a84221dc66c461c

podcasts/Bookmarks/List/BookmarkPodcastListViewModel.swift
emilylaguna: hmm. I don't think I had a reason. Fixed here: 2e65ef51894587d9eaee6f68900e4d1cc904c3b5

podcasts/Bookmarks/List/BookmarkListViewModel.swift
emilylaguna: Ah good catch, for some reason I thought the default separator was a space already. Done here: cf9707b9c31c31c14aae0411df2329cddab239d3

podcasts/Bookmarks/List/BookmarkListViewModel.swift
emilylaguna: That is correct, I originally tried `presentationController` but sadly that's already taken :(

I shifted to moving the presentation logic to the router and using a method instead of a property. 

Done here: 4fe251c947f2964387ee5006b84ba4516e93085c

leandroalonso (approved):
:shipit:

PR #1013, opened 2023-08-01
Bookmarks: Adds the episode bookmark list page

emilylaguna:
| 🛫 Depends on: https://github.com/Automattic/pocket-casts-ios/pull/1012 |
|:---:|

This adds the ability to swipe to the bookmarks page and displaying of the bookmarks for the specific episode. This doesn't quite match the designs yet, I'll be updating this later.

## Demo Video

https://github.com/Automattic/pocket-casts-ios/assets/793774/bf794d09-858b-473b-b4e9-23796dfa232f

## To test

1. Launch the app
2. Enable the Bookmarks feature flag
3. Go to the Podcasts tab
4. Open a podcast that has bookmarks
5. Tap on an episode that has bookmarks
6. ✅ Verify the episode details load, and you can swipe to see all the show notes
7. ✅ Verify there are no auto layout warnings
8. Swipe to the left
9. ✅ Verify the bookmarks view appears and displays the episodes bookmarks
10. Swipe back to the episode details
11. Swipe down until the shadow appears under the nav bar
12. Swipe left to show the bookmarks
13. ✅ Verify the shadow is not visible
14. Swipe right to show the episode details again
15. ✅ Verify the shadow appears again

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1014, opened 2023-08-01
Bookmarks: Adds the tabs to the episode detail view

emilylaguna:
| 🛫 Depends on: https://github.com/Automattic/pocket-casts-ios/pull/1013 | 🎨 Design: [Figma](https://www.figma.com/file/Io7PCz1H1hmOgEE9eljoYW/Bookmarks?type=design&node-id=2036-47334&mode=dev) |
|:---:|:---:|

This adds the Details / Bookmarks tabs to the episode detail view. 

Note: There are some clipping issues on smaller screens that I will address in a future PR.

## Demo Video

https://github.com/Automattic/pocket-casts-ios/assets/793774/1b9c99b1-94b5-47c3-a15c-951d5a50ccfe

## To test

1. Launch the app
2. Enable the Bookmarks feature flag
3. Go to the Podcasts tab
4. Open a podcast that has bookmarks
5. Tap on an episode that has bookmarks
6. ✅ Verify you see the Details / Bookmarks tab
7. Swipe to the bookmarks tab
8. ✅ Verify the tab updates with the current page
9. ✅ Verify the right buttons update to show the more button ( ... )
10. Tap the More button
11. ✅ Verify you see the options picker, and the options work
12. Tap the Details tab
13. ✅ Verify you are brought to the details view
14. ✅ Verify the right buttons update to the toggle star / share items
15. ✅ Verify they work
16. Tap the Bookmarks tab
17. ✅ Verify you are brought to the bookmarks view

## Checklist


podcasts/EpisodeDetailViewController.swift
```diff
@@ -552,6 +580,125 @@ private extension EpisodeDetailViewController {
...
+    private func updateRightButtons() {
+        removeAllButtons()
+
+        switch currentTab {
+        case .details:
+            addRightAction(image: UIImage(named: "podcast-share"), accessibilityLabel: L10n.share, action: #selector(shareTapped(_:)))
+            starButton = addRightAction(image: UIImage(named: "star_empty"), accessibilityLabel: L10n.starEpisode, action: #selector(starTapped(_:)))
+            updateStar()
+        case .bookmarks:
+            if bookmarksController?.viewModel.numberOfItems != 0 {
```
leandroalonso, line 673: Do we really need to handle this special case?

I would show the "..." anyways — we do this in Filters, for example. It doesn't matter if there are episode or not, it will be shown there.

I know the design doesn't show that but it looks like a lot of code for something minor. 🤷

leandroalonso (approved):
Seems you had a lot of fun with fake nav!

Nice work, tested on iPad/iPhone with different themes and it all works as expected.

podcasts/EpisodeDetailViewController.swift
emilylaguna: I don't think it's too bad, it's really just a couple of lines to listen for the change and reload the items.

PR #1015, opened 2023-08-01
Bookmarks: Fixes an iOS 15 issue when adding a hosting controller as a child

emilylaguna:
There's a bug on iOS 15 that causes the nav bar to appear when adding a hosting controller as a child controller. 

This fixes it by adding a flag to hide the nav bar in will appear. 

## Screenshots

| Before | After |
|:---:|:---:|
|||

| iPad Before | iPad After |
|:---:|:---:|
|||

## To test

1. Launch the app on iOS 15
2. Enable the Bookmarks Feature Flag
3. Go to the Discover tab
4. Tap on any podcast
5. ✅ Verify the nav bar isn't shown at the top
6. Tap back
7. ✅ Verify the Discover nav bar is still shown
8. Do the same from the Podcasts tab
6. Relaunch the app on iOS 16
7. ✅ Repeat the steps, and verify everything works as intended

## Checklist

### DIFF
--- a/podcasts/FakeNavViewController.swift
+++ b/podcasts/FakeNavViewController.swift
@@ -85,6 +85,15 @@ class FakeNavViewController: PCViewController, UIScrollViewDelegate {
         if !navigationTitleSetOnScroll { fakeNavTitle.text = navTitle }
     }
 
+    override func addChild(_ childController: UIViewController) {
+        super.addChild(childController)
+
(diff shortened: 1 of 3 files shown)


leandroalonso (approved):

emilylaguna:
@leandroalonso I actually realized my solution didn't work as intended since it will hide the nav bar and never show it again. 

I pushed a new solution that is working, and updated the testing steps.

leandroalonso (approved):

PR #1019, opened 2023-08-02
PlaybackManager: remove lock

leandroalonso:
Our top crash seems to be related to a deadlock with the players. In order to easily reproduce that:

1. Go to `trunk`
2. Run the app
3. Clean your Up Next queue
4. Download and add an episode to your Up Next queue
5. Add a non-downloaded to your Up Next queue (after the downloaded episode)
6. Go offline
7. Play the download episode
8. Move the scrubber til the end
9. 💥 The app should freeze due to a deadlock

This happens while changing from `EffectsPlayer` to `DefaultPlayer`. Given there's no connectivity the lock is never released somehow and the whole app freezes.

## Additional context

With the help of @emilylaguna we were able to check why this check was added in the first place:

1. It seems that there was a crash because `loadEpisode` was called twice
2. A test of calling `loadEpisode` 100 times was added — which crashes the app right away
3. The locks were then added to the code preventing the crash from happening

Although I don't have access to the original test, I did the same and removed the locks. To my surprise, there wasn't any crash.

However, a few times I was able to see a `EXC BAD_ACCESS` happening on iOS 15. I then reduced the number of calls to 10 (given that 100 is an exaggeration of the issue) and wasn't unable to see the crash.

I believe we can remove the locks (as did in this PR) and monitor this during the beta period — I wonder whether any other possible crash that might happen due to this change will have fewer occurrences compared to what we have today.

## To test

1. Make sure you have at least one episode in your Up Next queue
1. In the `AppDelegate.swift` `didFinishLaunchingWithOptions` method add this just before the `return true`:

```swift
        Array(0...10).forEach { index in
            DispatchQueue.init(label: "\(index)-thread").asyncAfter(deadline: .now() + 1) {
                print("$$ Executing at \(CFAbsoluteTimeGetCurrent()) \(Thread.current)\n")
                if let firstEpisodeInUpNext = DataManager.sharedManager.episodeInUpNextAt(index: 0) {
                    PlaybackManager.shared.load(episode: firstEpisodeInUpNext, autoPlay: true, overrideUpNext: false)
                }
            }
        }
```

This will create 10 threads loading the first episode of your Up Next queue.

2. Run the app
3. ✅ The app should not crash
4. Download the episode and re-run the app
5. ✅ The app should not crash ¶

¶ This is in order to test both `DefaultPlayer` and also `EffectsPlayer`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 7.45
 -----
-
+- Attempt to fix a crash when switching episodes [1019]
 
 7.44
 -----
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -52,8 +52,6 @@ class PlaybackManager: ServerPlaybackDelegate {
     private var playersToCleanUp = [AnyHashable]()
     private let playerCleanupQueue: DispatchQueue
 
-    private let playerCreateDestroyLock = NSObject()
-
     private let catchUpHelper = PlaybackCatchUpHelper()
 
     private let analyticsPlaybackHelper = AnalyticsPlaybackHelper.shared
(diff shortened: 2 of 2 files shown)


yaelirub (approved):
Approving as the crash seems to be fixed with this change. 
There are more places in the code where we use the lock and note that the code is not thread safe. I wonder if we need to find another solution to those in the future

leandroalonso:
> There are more places in the code where we use the lock and note that the code is not thread safe. I wonder if we need to find another solution to those in the future

I also looked at these and compared them to the crash reports we've been seeing and most crashes seem to come from this part in specific, but I totally agree that if the changes here come to fruition we can tweak those too!

PR #1023, opened 2023-08-04
Bookmarks: Adds importing of bookmarks from the server

emilylaguna:
This adds the foundation to import bookmarks that are sent from the server. Along with making the SyncTask testable to verify the bookmarks are being imported as expected. 

🚨 **Heads Up:** I realized too late that I was copying the folderIssue file log. I fixed this in a later PR.

## To test

There aren't any visible changes to test, so a 🟢 CI is great. However this does change the data manager and sync task a bit so in addition to that and a code review we should also verify:

1. There are no errors being displayed in console about the database
2. The data remains intact after running
3. Syncing works as expected

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncTask+ServerChanges.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncTask+ServerChanges.swift
@@ -8,6 +8,7 @@ extension SyncTask {
         var episodesToImport = [Api_SyncUserEpisode]()
         var filtersToImport = [Api_SyncUserPlaylist]()
         var foldersToImport = [Api_SyncUserFolder]()
+        var bookmarksToImport = [Api_SyncUserBookmark]()
 
         for item in response.records {
             guard let oneOf = item.record else { continue }
@@ -22,8 +23,7 @@ extension SyncTask {
             case .folder:
                 foldersToImport.append(item.folder)
             case .bookmark:
-                #warning("TODO")
-                continue
+                bookmarksToImport.append(item.bookmark)
             case .device:
                 continue // we aren't expecting the server to send us devices
             }
@@ -65,6 +65,20 @@ extension SyncTask {
                 guard let strongSelf = self else { return }
 
                 strongSelf.importFilter(filterItem)
+            }
+        }
+
+        for bookmark in bookmarksToImport {
+            importQueue.addOperation { [weak self] in
+                guard let strongSelf = self else { return }
+                let semaphore = DispatchSemaphore(value: 0)
+
+                Task {
+                    await strongSelf.importBookmark(bookmark)
+                    semaphore.signal()
+                }
+
+                semaphore.wait()
             }
         }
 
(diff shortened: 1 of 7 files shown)


Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncTask+ServerChanges.swift
```diff
@@ -315,4 +329,77 @@ extension SyncTask {
     func isPlayerPlaying(episode: Episode) -> Bool {
         ServerConfig.shared.playbackDelegate?.isActivelyPlaying(episodeUuid: episode.uuid) ?? false
     }
+
+    func importBookmark(_ apiBookmark: Api_SyncUserBookmark) async {
+        let bookmarkManager = dataManager.bookmarks
```
leandroalonso, line 334: Not for this PR, but for the future: I wonder if we can decouple the model/database layer from the syncing logic.

I believe we can make changes in a way that the `SyncTask` doesn't need to know anything about the returned model and our stored model. What we can try to do is to move all the "BookmarkManager` logic to the `Bookmark` itself.

A few ideas we could potentially explore:

* A Bookmark can be created from a `Api_SyncUserBookmark`. It would also be able to check if it already exists or not;
* The bookmark itself has `.save()` and `.delete()` methods
* `save()` would be smart enough to know if the Bookmark needs to be added or updated on the DB layer

This way, all this function would do could be something like:

```swift
let bookmark = Bookmark(from: apiBookmark)
apiBookmark.shouldDelete ? bookmark.permanentlyDelete() : bookmark.save()
```

Modules/Server/Tests/PocketCastsServerTests/SyncTaskTests+BookmarkImport.swift
```diff
@@ -0,0 +1,181 @@
...
+
+    func testNonexistingBookmarkIsAdded() async {
+        let uuid = UUID().uuidString
+        let episode = UUID().uuidString
+        let podcast = UUID().uuidString
+        let title = "Hello World"
+        let time = 86400.0
+        let created = Date(timeIntervalSince1970: 456)
+
+        let apiBookmark = Api_SyncUserBookmark.forTesting(uuid: uuid,
```
leandroalonso: Given we're already on the tests, I believe we can just initialize `Api_SyncUserBoomark` right away with the given parameters. We can just add this to your extension:

```swift
init(uuid: String,
                     episode: String = "episode",
                     podcast: String? = nil,
                     title: String = "Title",
                     time: TimeInterval = 1234,
                     created: Date = Date(),
                     isDeleted: Bool? = nil) {
        self.init()
        bookmarkUuid = uuid
        episodeUuid = episode

        if let podcast {
            podcastUuid = podcast
        }

        self.title.value = title
        self.time.value = Int32(time)
        createdAt = .init(date: created)

        if let isDeleted {
            self.isDeleted.value = isDeleted
        }
    }
```

Then we can remove the `.forTesting`

leandroalonso (approved):

Modules/Server/Tests/PocketCastsServerTests/SyncTaskTests+BookmarkImport.swift
emilylaguna: Nice, that's cleaner. Done here: f0f5f7eaf6784bae05e07ec16206a8e6a7409227

Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncTask+ServerChanges.swift
emilylaguna: I think that would be a nice change from the way the app currently works. I think if we do this we'll need to make some bigger changes to allow our future code to also use the same patterns. Jumping off from your idea, I'm thinking:

- Create a parent `DataModel` class that contains all the base functionality for save, deleted, create, etc. 
    - It would also define the base initializer to take a DataManager and set its internal property
- Bookmarks, and other models would subclass off of the parent and add override the methods to add their own functionality.

PR #1025, opened 2023-08-04
Bookmark: Adds incremental syncing of bookmarks

emilylaguna:
This adds sending and importing of bookmarks during an incremental sync. 

I also added an internal feature flag to use to prevent sending/importing bookmarks in production. This isn't ideal, but it should work for now. 

I also replaced `foldersIssue` with `addMessage` 🤦‍♀️ 

> **🚨 Heads Up:** Syncing of bookmarks on a user file doesn't work on the server yet. 

## To test

1. Switch your environment to staging
2. Enable the Bookmarks feature flag
   - Restart the app if you did
3. Add some bookmarks to podcasts
4. Go to the Profile tab
5. Pull to refresh to sync
6. ✅ Verify you see `SyncTask: Number of changed bookmarks: {X}` where X is the number of bookmarks you added that are unsync'd
7. ✅ Verify you also see `SyncTask: Found X bookmarks to import` where X is the number of bookmarks you sent to be sync'd
    - The other processes also do this, I'm not entirely sure why we need to reimport the data that we just sync'd. 
8. Open the podcast/episode you added bookmarks to
9. ✅ Verify all your bookmarks are still there and their titles are correct
10. In a simulator or on a different device run this branch, enable the flag, etc
11. If you just signed in, perform another pull to refresh to perform the incremental sync
12. ✅ Verify you see the bookmarks from your other device 
13. Edit one of the titles
14. Delete one of the bookmarks
15. On your second device perform another sync
16. Check your bookmarks again
17. ✅ Verify the title is updated
18. ✅ Verify the deleted item is removed

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
@@ -15,8 +15,10 @@ public struct Bookmark: Hashable {
     public var episode: BaseEpisode? = nil
     public var podcast: Podcast? = nil
 
+    // For syncing
     public var titleModified: Date? = nil
     public var deletedModified: Date? = nil
+    public var deleted: Bool = false
 
     // `BaseEpisode` and `Podcast` don't conform to Hashable, so instead we implement it manually to ignore those properties
     public func hash(into hasher: inout Hasher) {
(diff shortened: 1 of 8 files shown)


leandroalonso (approved):
When testing the first time, I ran into an issue:

1. I created 5 bookmarks on one device
2. Synced and saw the message "SyncTask: Found 5 bookmarks to import"
3. Logged in to a new device
4. Performed sync
5. No bookmarks were returned 🤔 

I then edited the bookmarks and deleted them on the other device and then they started appearing. I'm not sure what might have happened. Sharing here just so this is documented in case we run into it again in the future.

Also, I had to run the app first in `DEBUG` mode because of the `#if DEBUG` condition. Otherwise, the Bookmarks database is not created.

leandroalonso:
More information about the issue I experienced: with the full sync I was able to see all Bookmarks.

PR #1027, opened 2023-08-05
Bookmarks: Migrates the Remote Skips Chapters setting to Headphone Controls

emilylaguna:
This removes the Remote Skips Chapters setting, and migrates it over to the Headphone Controls. 

When the app launches for the first time after updating we'll check to see if the remote skips chapters setting is on, if its we'll update the headphone controls settings to reflect their original option. 

This also updates the next/prev track commands to use the new headphone actions. 

## To test

1. Before updating
2. Enable the Remote Skips Chapters setting in General
3. Launch this PR
4. Disable the bookmarks feature flag if its enable
Go to Profile > Cog > Headphone Controls
5. ✅ Verify the next action shows next chapter
6. ✅ Verify the previous action shows previous chapter
7. Play a podcast with chapters
8. Using headphones or a remote use the next track action 
9. ✅ Verify you are brought to the next chapter
10. Perform the previous track action
11. ✅ Verify you are brought to the previous chapter
12. Open the Headphone Controls
13. Tap on each action
14. ✅ Verify you do not see the add bookmark option
15. Change to the Skip Back and Skip Forward options
16. Use the previous and next track actions
17. ✅ Verify you are skipped forward in the correct direction
18. Enable the Bookmarks feature flag
19. Go to the Headphone Controls
20. Change one of the actions to add bookmark
21. Perform that action
22. ✅ Verify a bookmark was created

## Checklist

### DIFF
--- a/podcasts/AppDelegate+Defaults.swift
+++ b/podcasts/AppDelegate+Defaults.swift
@@ -109,6 +109,21 @@ extension AppDelegate {
             }
         }
 
+        // With the addition of bookmarks we have added a new headphone controls setting that this is being migrated to
+        // This will check if the user has the old Remote Skips Chapters preference enabled, and will move that setting
+        // by setting both the previous and next track actions to the change chapter action.
+        performUpdateIfRequired(updateKey: "MigrateRemoteSkipsChaptersToHeadphoneControls") {
+            let key = "RemoteChapterSkip"
+
+            UserDefaults.standard.bool(forKey: key).when(true) {
+                Settings.headphonesNextAction = .nextChapter
+                Settings.headphonesPreviousAction = .previousChapter
+
+                // Remove the setting
+                UserDefaults.standard.removeObject(forKey: key)
+            }
+        }
+
         defaults.synchronize()
     }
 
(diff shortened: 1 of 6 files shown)


podcasts/AppDelegate+Defaults.swift
```diff
@@ -109,6 +109,21 @@ extension AppDelegate {
             }
         }
 
+        // With the addition of bookmarks we have added a new headphone controls setting that this is being migrated to
+        // This will check if the user has the old Remote Skips Chapters preference enabled, and will move that setting
+        // by setting both the previous and next track actions to the change chapter action.
+        performUpdateIfRequired(updateKey: "MigrateRemoteSkipsChaptersToHeadphoneControls") {
+            let key = "RemoteChapterSkip"
+
+            UserDefaults.standard.bool(forKey: key).when(true) {
```
leandroalonso: [nit] It's ok to keep the `.when(true)` here but I gotta say that this adds a cognitive load when reading it. 😬 An `if UserDefaults.standard.bool(forKey: key)` wouldn't cause this (and it wins by saving 8 characters 🤭)

leandroalonso (approved):

podcasts/AppDelegate+Defaults.swift
emilylaguna: Oops, that the correct way to do this. Fixed here: 6509518881da746c147349449de017cff70b27f7

PR #1028, opened 2023-08-05
Bookmarks: Improve some interactions when creating a bookmark from headphones

emilylaguna:
This fixes some odd UI issues when the edit title action appears while in the background by disabling showing of update bookmark title view if the app is not active, or if connected to CarPlay. 

I've disabled this while connected to CarPlay for a couple reasons:
- It could be distracting for someone who is driving
- In my testing being connected to CarPlay has made it show the app always appears active no matter what. 

This also adds some visual feedback in the form of a toast when the user makes a bookmark and the full screen is not open. This toast gives the user the option to update the title of the bookmark, and then opens the edit title view, if they update the title we then show another toast confirming their change with the option to view it in the fullscreen player. 

## Demo Video

https://github.com/Automattic/pocket-casts-ios/assets/793774/a94d83a4-43dd-4146-8960-4f10d552d4eb

## To test

1. Connect some headphones to your device
2. Enable the bookmarks feature flag
3. Enable the headphone controls so one of the actions is add bookmark
4. Open the full screen player
5. Move the app into the background
6. Use the headphones to add a bookmark
7. ✅ Verify you see a log indicating a bookmark was created
8. Open the app
9. ✅ Verify the edit title screen is not open
10. Dismiss the full screen player
11. Use the add bookmark action again
12. ✅ Verify you see a toast that the bookmark was created
13. Tap the Change title action
14. ✅ Verify the edit controller opens
15. Update the title
16. ✅ Verify you see a 'Bookmark X updated' toast where X is the new title of the bookmark
17. Repeat the steps while connected to CarPlay.

## Checklist

### DIFF
--- a/podcasts/MainTabBarController.swift
+++ b/podcasts/MainTabBarController.swift
@@ -593,6 +593,7 @@ private extension MainTabBarController {
     // Shows a toast notification when a bookmark is created and we're not in the full screen player
     func addBookmarkCreatedToastHandler() {
         let bookmarkManager = PlaybackManager.shared.bookmarkManager
+
         bookmarkManager.onBookmarkCreated
             .receive(on: RunLoop.main)
             .filter { event in
@@ -601,7 +602,7 @@ private extension MainTabBarController {
                 && NavigationManager.sharedManager.miniPlayer?.playerOpenState == .closed
             }
             .compactMap { event in
-                PlaybackManager.shared.bookmarkManager.bookmark(for: event.uuid)
+                bookmarkManager.bookmark(for: event.uuid)
             }
             .sink { [weak self] bookmark in
                 self?.showToast(for: bookmark)
@@ -622,7 +623,7 @@ private extension MainTabBarController {
                 self?.handleBookmarkTitleUpdated(updatedTitle: updatedTitle)
             })
 
-            self?.present(controller, animated: true)
+            self?.presentFromRootController(controller)
         }
 
         Toast.show(message, actions: [action], theme: .playerTheme)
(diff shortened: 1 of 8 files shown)


podcasts/MainTabBarController.swift
```diff
@@ -583,6 +587,62 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
...
+            .compactMap { event in
+                PlaybackManager.shared.bookmarkManager.bookmark(for: event.uuid)
+            }
+            .sink { [weak self] bookmark in
+                self?.showToast(for: bookmark)
+            }
+            .store(in: &cancellables)
+    }
+
+    func showToast(for bookmark: Bookmark) {
```
leandroalonso, line 613: One edge case is that if there's any ViewController already being present any interaction with the popup will fail:

```
[Presentation] Attempt to present <podcasts.BookmarkEditTitleViewController: 0x157647170> on <podcasts.MainTabBarController: 0x15602aa00> (from <podcasts.MainTabBarController: 0x15602aa00>) which is already presenting <podcasts.EpisodeDetailViewController: 0x159069600>.
```

To repro this:

1. Go to Discover
2. Open a podcast
3. Open an episode
4. Bookmark
5. Try to interact with the toast and nothing will happen

I don't think that's a big problem but I wonder if we want to handle this case somehow.

leandroalonso (approved):
For CarPlay, I tested by being connected to it while performing the actions on my headphones — I think that's what was expected — although is a bit weird to use headphones while connected to CarPlay.

Anyways, it all works as described.

emilylaguna:
> although is a bit weird to use headphones while connected to CarPlay.

This is true. I should have included different steps for CarPlay because the headphone gesture is triggered when using the next/previous track buttons and you don't need to use headphones to trigger it.

podcasts/MainTabBarController.swift
emilylaguna: Good catch! I created UIViewController extension that tries to safely present the controller either on the existing presented view, or the root controller:

- 2691a2026500c30e15bc2982d4879f87eb4d34f3
- fd6e579bc9d6fceac697e276bbd5e39cc75dc05c

PR #1035, opened 2023-08-11
Share Extension: copy files through shared group container

leandroalonso:
In the original #989 we were "transmitting" the OPML file to Pocket Casts by encoding it to Base 64. This was not the correct behavior but we were missing the right permissions on the certificates to do "the correct way" at the time.

Now that this is fixed we can copy the file to the shared group container and just read it on the app, without any hack stuff.

This also fixes an issue with some OPML files not being able to import.

## To test

1. Run the app
2. [Import this file](https://a8c.slack.com/archives/C02A333D8LQ/p1691728449019899) to Pocket Casts
3. ✅ The app should open, the import process should start and be successful

## Checklist

### DIFF
--- a/Share Extension/ShareViewController.swift
+++ b/Share Extension/ShareViewController.swift
@@ -7,7 +7,6 @@ class ShareViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
-        // Get the attachment
         let content = extensionContext?.inputItems.first as? NSExtensionItem
         guard let attachment = content?.attachments?.first as? NSItemProvider else {
             close()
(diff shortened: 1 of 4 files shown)


Share Extension/ShareViewController.swift
```diff
@@ -22,26 +21,29 @@ class ShareViewController: UIViewController {
...
-                    self?.close()
                     return
                 }
 
-                // Convert the file to Data
-                guard let opmlData = try? Data(contentsOf: url) else {
-                    self?.close()
+                // Save the file to the shared group directory
+                let fileManager = FileManager.default
+                guard let container = fileManager.containerURL(forSecurityApplicationGroupIdentifier: "group.au.com.shiftyjelly.pocketcasts") else {
```
yaelirub: "group.au.com.shiftyjelly.pocketcasts" can be  `SharedConstants.GroupUserDefaults.groupContainerId`

yaelirub:
Tested and it work, but it worked for me before too. Is there a file you know couldn't be imported before and now it can?

leandroalonso:
@yaelirub You can find a file that didn't work [here](https://a8c.slack.com/archives/C02A333D8LQ/p1691728449019899).

In `trunk` I'm unable to import it

Share Extension/ShareViewController.swift
leandroalonso: Done in 65ca83342d2a5cc54485016399e75af175a0dc03

yaelirub (approved):
Tested with the problematic file and it works.

PR #1038, opened 2023-08-15
Player did not see an IO cycle: attempt to fix the crash by removing locks

leandroalonso:
Based on the crash reports we've seen so far our top crash "Player did not see an IO cycle" seems to be happening because of a deadlock. Also, on iOS 17 this crash seems way worse.

This is an attempt to fix it by removing the locks in `EffectsPlayer` and observing the results. I also added this change behind a remote feature flag so we can enable/disable the locks without doing an app release. In case anything goes wrong, we can simply revert it.

## To test

First, configure the environment:

1. Go to `AppDelegate.swift`, locate the `configureFirebase` method, and change `2.hour` to `30.seconds`.
2. Then, on `EffectsPlayer.swift` add a breakpoint on line `68`.
3. Finally, download some episodes

### Testing the flag changing

1. Run the app
2. Hit play on any downloaded episode
3. When the breakpoint is hit, type `po strongSelf.playerLock` on the console
4. ✅ The returned value should be `nil`
5. Go to Firebase Console
6. Change `lock_effects_player` to `true` and publish the changes
7. Re-run the app
8. Play any downloaded episode
3. When the breakpoint is hit, type `po strongSelf.playerLock` on the console
4. ✅ The returned value should be **NOT** `nil`
4. Go to Firebase Console
6. Change `lock_effects_player` to `false` and publish the changes

### Testing the player without locks

It's better to test that on a real device.

1. Run the app
2. ✅ Make sure `Settings.lockEffectsPlayer` is `false`
3. Play any downloaded episode
4. Connect a Bluetooth earphone
5. ✅ Episode should keep playing on Bluetooth
6. Remove the Bluetooth device
7. ✅ Episode should pause
8. Connect to CarPlay
9. Play a downloaded episode
10. Open Maps and set any route
11. ✅ When Maps announce a route or information the episode should pause and resume after the announcement ends
12. Go back to your phone
13. Play/pause the episode a few times (to trigger player creation/destruction)
14. ✅ Everything should work as expected

## Checklist

### DIFF
--- a/podcasts/Settings.swift
+++ b/podcasts/Settings.swift
@@ -871,6 +871,11 @@ class Settings: NSObject {
             return remote.boolValue
         }
 
+        static var lockEffectsPlayer: Bool {
+            let remote = RemoteConfig.remoteConfig().configValue(forKey: Constants.RemoteParams.lockEffectsPlayer)
+            return remote.boolValue
+        }
+
         private class func remoteMsToTime(key: String) -> TimeInterval {
             let remoteMs = RemoteConfig.remoteConfig().configValue(forKey: key)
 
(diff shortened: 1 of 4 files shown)


podcasts/Settings.swift
```diff
@@ -871,6 +871,11 @@ class Settings: NSObject {
             return remote.boolValue
         }
 
+        static var lockEffectsPlayer: Bool {
```
leandroalonso, line 874: We should definitely create a `RemoteFeatureFlag` or add the remote feature flags in `FeatureFlag`.

However, this change is a little bit bigger, and given this is targeting a frozen branch I decided to leave this out of the scope for this task.

emilylaguna (approved):
The remote feature flag works as expected, and connecting / disconnecting is working as well. 

I tried my best to break the player but I wasn't able to. 

:shipit:

PR #1041, opened 2023-08-17
Bookmarks: Fix swipe to dismiss on the episode detail view

emilylaguna:
This fixes an issue reported on the beta slack where the episode detail can't be swiped to be dismissed anymore. 

This is happening because the episode detail view has been wrapped in a parent UIScrollView to support the tabs which is interfering with the interactive dismiss. 

This PR fixes the issue by adding a new UIScrollView subclass called `DismissableNestedScrollView` that prevents the panning gesture from triggering if the scroll view would be scrolled past the top offset. 

This means the container scrollview's gesture is triggered instead, which correctly calls the UIPresentationController's internal methods to dismiss the view. 

### Default system swipe to dismiss

https://github.com/Automattic/pocket-casts-ios/assets/793774/bc8a013d-7b88-4e03-b99e-ffc323b8b900

### Swipe to dismiss introduce in this PR

https://github.com/Automattic/pocket-casts-ios/assets/793774/e80a4da9-4619-44b9-8413-c1689f8c9637

## To test

1. Launch the app
2. Disable the bookmarks feature flag if it's enabled
3. Open any episode detail view
4. Swipe down
5. ✅ Verify the interactive dismiss is engaged, and swiping down fully dismisses the view
6. Open the episode detail again
7. Scroll up a small amount, then scroll back down to dismiss
8. ✅ Verify the view **does not** start dismissing
   - This is the default system behavior
9. Stop dragging, then swipe to dismiss from the top again
10. ✅ Verify it dismisses
11. Enable the bookmarks feature flag
12. Open the episode details again
13. ✅ Verify you can swipe to dismiss
14. ✅ Verify you can still swipe to the next page

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Extensions/UI/DismissableNestedScrollView.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Extensions/UI/DismissableNestedScrollView.swift
@@ -1,0 +1,10 @@
+#if !os(watchOS)
+import UIKit
+
+/// A UIScrollView subclass that allows the swipe to dismiss in a presented view when its contained in a parent UIScrollView
+public class DismissableNestedScrollView: UIScrollView, UIGestureRecognizerDelegate {
+    public override func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+        return (contentOffset.y + contentInset.top) - panGestureRecognizer.translation(in: self).y > 0
+    }
+}
+#endif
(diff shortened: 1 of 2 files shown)


leandroalonso:
@emilylaguna I believe we should target `release/7.45`?

emilylaguna:
@leandroalonso 🤦 oops forgot to change base branch. Done now

leandroalonso (approved):
Tested on a real-device and it workes just fine! :shipit:

PR #1043, opened 2023-08-17
Bookmarks: Adds a way to unlock features for subscription tiers

emilylaguna:
This adds a new class called `PaidFeature` that represents represents a feature that is unlocked with a subscription tier, and is considered to be unlocked if the tier is equal to or higher than the set `tier` value. 

#### Why a new class instead of an extension on `FeatureFlag`
This was my initial idea, but in practice it ended up feeling really clunky, and felt like the FeatureFlag was trying to do too much. 

There are could be instances where we don't want the feature tied to a development flag. So I went with creating a new way to manage this specific state.

#### Class naming
I also struggled with the naming of the class and I think I landed on something that's contextually easy to understand without having to dive into the code itself. 

#### Why a class instead of an enum like `FeatureFlag`?
1. The class can self manage its unlock state which reduces the need to listen for the necessary notifications over and over. Instead we can just observe the class itself. 
2. Testability. Using the class allows for injection of a `SubscriptionHelper` class which gives the ability to test the various states. 
3. It's an `ObservableObject` so the unlock state can easily be used in SwiftUI
4. When we add new features we don't have to then add another case to multiple switches. 

## To test

Only code changes and unit tests in this PR so a code review and green CI should be good.

## Checklist


Modules/Server/Sources/PocketCastsServer/Public/SubscriptionHelper.swift
```diff
@@ -1,6 +1,6 @@
 import UIKit
 
-public class SubscriptionHelper: NSObject {
+open class SubscriptionHelper: NSObject {
```
emilylaguna, line 3: `public` -> `open` to allow it to be subclassed in tests.

podcasts/PaidFeature.swift
```diff
@@ -0,0 +1,84 @@
...
+        )
+        .receive(on: DispatchQueue.main)
+        .sink { [weak self] _ in
+            self?.objectWillChange.send()
+        }
+        .store(in: &cancellables)
+    }
+}
+
+// MARK: - Helpers
```
emilylaguna, line 66: I added these because I found myself repeating the same logic in a few places so it made sense to make it reusable.

PocketCastsTests/Tests/Utilities/PaidFeatureTests.swift
```diff
@@ -0,0 +1,80 @@
...
+    static func activePlus() {
+        _hasActiveSubscription = true
+        _subscriptionTier = .plus
+        _subscriptionType = .plus
+    }
+
+    static func activePatron() {
+        _hasActiveSubscription = true
+        _subscriptionTier = .patron
+        _subscriptionType = .plus
```
yaelirub: should this be `.supporter`?

emilylaguna: Nah, supporter is from an old feature and isn't used anymore.

emilylaguna: This is a great question, and reminds me that I can actually remove some code here because its not needed. 

The short answer is, no because the `.supporter` value is for the paid podcasts feature that isn't being used. It's kept in the enum just to maintain compatibility. 

The longer answer of why I was setting the 3 propergties is because the subscriptionType is the old way of checking for plus, and there is [fallback handling](https://github.com/Automattic/pocket-casts-ios/blob/4528d2fc405a8b49c06659c85ea3e2937e9fcc28/Modules/Server/Sources/PocketCastsServer/Public/SubscriptionHelper.swift#L17-L22) in the activeTier code that allowed us to use the new `SubscriptionTier` even when it wasn't on the production yet servers. 

And since the SubscriptionType only ever returns `none` or `plus`, even if the user has Patron (Internal Ref: `pdeCcb-2fL#comment-2074`), using `.plus` for this value is technically correct and reflects what we receive from the server for Patron (`tier = patron, type = plus`)

However for the purposes of testing there isn't a need to set the subscriptionType so I've removed it here: 39e7b25274da6e7a982d56ceac3ce4230da5a5e1

I believe the tier changes are now on production so I can look into removing the fallback code in `activeTier` since it shouldn't be needed anymore.

PocketCastsTests/Tests/Utilities/PaidFeatureTests.swift
```diff
@@ -0,0 +1,67 @@
...
+@testable import podcasts
+import PocketCastsServer
+
+final class PaidFeatureTests: XCTestCase {
+    override func setUp() {
+        TestSubscriptionHelper.noSubscription()
+    }
+
+    func testFreeFeatureIsAlwaysUnlocked() {
+        assertTierValues(tier: .none, free: true, plus: true, patron: true)
```
leandroalonso: It took me a bit to understand this test because I had to understand `assertTierValues`. Overall, in test I tend to prioritize readability. I believe you can make tests more readable in this way:

```swift
// MARK: Feature available for everyone

func testFreeFeatureIsUnclocked() {
   let feature = freeFeature()

   userHasNoSubscription()

   XCTAssertEqual(feature.isUnlocked, true)
}
```

Of course, this is a lot more lines and duplication but when it comes to tests I think it's a good trade-off. Let me know what you think.

podcasts/PaidFeature.swift
```diff
@@ -0,0 +1,84 @@
...
+/// Since this is a subclass of `ObservableObject` it can easily be used in SwiftUI. Outside SwiftUI, you can use the `objectWillChange`
+/// publisher to be notified about changes.
+///
+/// And while the class is an `ObservableObject` it doesn't use any `@Published` properties and instead manually triggers `objectWillChange`.
+/// This is done to ensure future compatibility of any property changes by not allowing listeners to directly access a published properties publisher.
+///
+class PaidFeature: ObservableObject {
+    /// Whether the feature is unlocked for the active subscription tier
+    var isUnlocked: Bool {
+        subscriptionHelper.activeTier >= tier
```
leandroalonso, line 33: Since all we use here is `activeTier` do we really need to inject `SubscriptionHelper` here? Can't we just inject the `activeTier` on `init` and then update it when the listeners are triggered?

emilylaguna: Hmm. I don't think that would work because if we injected the tier we would still need to access the SubscriptionHelper in order to update the value.

PocketCastsTests/Tests/Utilities/PaidFeatureTests.swift
emilylaguna: Done here: 245b82931c3e556d7c991aff79ea7132c76006ad

podcasts/PaidFeature.swift
leandroalonso: Got it!

Another suggestion (feel free to tackle in the future if you'd prefer and, of course, if you agree): what do you think about creating a `SubscriptionHelper` singleton and injecting that? This singleton would contain only the `activeTier` (which accesses the static `activeTier` property) but I think a singleton is better than a whole static class. Of course, you would make the change only here because changing other places that use `SubscriptionHelper` is out of scope.

leandroalonso (approved):

podcasts/PaidFeature.swift
emilylaguna: I like that idea, it makes more sense and pushes us in a better directly. I updated the code here:
- 67840024b07cd5cba2dcafd1edc21b6ab3853455
- 9c8a47344c73c2af5911a234d8a93f478793a955

PR #1050, opened 2023-08-19
Bookmarks: Locks the settings and other UI actions

emilylaguna:
| 🎨 Designs: [Figma](https://www.figma.com/file/Io7PCz1H1hmOgEE9eljoYW/Bookmarks?type=design&node-id=2099-32855&mode=dev) |
|:---:|

This locks the player shelf action and the headphone settings

- Headphone Settings
- Shelf Action
- Overflow Menu

### Things to note
- If the user no longer has access to bookmarks, but has the Add Bookmark headphone action selected then the setting will fallback to defaultValue set on the `SettingValue`. 
- If the user upgrades after tapping the Add Bookmark setting, then we will automatically change the setting to the add bookmark option. 
   - I didn't do this with the other places because it felt less contextual (ie: they tap add bookmark on the shelf, then 30 - 45 seconds later they finish the upgrade process and then we make a bookmark that isn't really near where they started the process at)

### 🔓 Unlockable
This also aims to reduce duplicated code, and make it easier to define an unlockable option by adding the `Unlockable` protocol. When an item adheres to Unlockable it defines which `PaidFeature` it can be unlocked with, and callers can check the `isUnlocked` property to determine if it's unlocked. 

Along with that, I added all of the Unlockable settings for Bookmarks in a the `Unlockable+Bookmarks` file to increase the visibility of which items are available with bookmarks.

## To test

**Note**: Enable the Bookmarks + Patron feature flags

1. Launch the app in the signed out state
10. Open the Full Screen player
11. Tap the Overflow icon (...)
12. Tap the Add Bookmark item
13. ✅ Verify the upsell opens
14. Dismiss it
15. Tap the Edit button and move the item into the shelf actions
16. Dismiss the overflow menu
17. Tap the shelf action
18. ✅ Verify you see the upsell
19. Dismiss it
2. Go to the Profile Tab > Settings Cog
3. Swipe down to the Headphone Controls option
4. Tap the previous action
5. ✅ Verify you see the Add Bookmark item with the Plus icon next to it
6. ✅ Verify tapping it opens the plus upsell
7. Dismiss it
8. Tap the next action
9. ✅ Verify the action is locked, and tapping it opens the upsell
10. Purchase the upgrade, and dismiss the app icon unlock view
11. ✅ Verify the setting action that was selected is now set to 'Add Bookmark'

## Checklist

### DIFF
--- a/podcasts/Common SwiftUI/BookmarksLockedStateView.swift
+++ b/podcasts/Common SwiftUI/BookmarksLockedStateView.swift
@@ -85,8 +85,7 @@ private class BookmarksUpgradeViewModel: PlusAccountPromptViewModel {
     override func showModal(for product: PlusPricingInfoModel.PlusProductPricingInfo? = nil) {
         guard let parentController = SceneHelper.rootViewController() else { return }
 
-        let controller = feature.upgradeController(source: "bookmarks_locked_state")
-        parentController.presentFromRootController(controller)
+        feature.presentUpgradeController(from: parentController, source: "bookmarks_locked_state")
     }
 
     private func product(for tier: SubscriptionTier) -> PlusProductPricingInfo? {
--- a/podcasts/Common SwiftUI/BookmarksLockedStateView.swift
+++ b/podcasts/Common SwiftUI/BookmarksLockedStateView.swift
@@ -65,6 +65,7 @@ struct BookmarksLockedStateView<Style: EmptyStateViewStyle>: View {
 
 private class BookmarksUpgradeViewModel: PlusAccountPromptViewModel {
     let feature: PaidFeature
+    let bookmarksSource: BookmarkAnalyticsSource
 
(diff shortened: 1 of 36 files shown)


podcasts/Common SwiftUI/BookmarksLockedStateView.swift
```diff
@@ -85,8 +85,7 @@ private class BookmarksUpgradeViewModel: PlusAccountPromptViewModel {
     override func showModal(for product: PlusPricingInfoModel.PlusProductPricingInfo? = nil) {
         guard let parentController = SceneHelper.rootViewController() else { return }
 
-        let controller = feature.upgradeController(source: "bookmarks_locked_state")
-        parentController.presentFromRootController(controller)
+        feature.presentUpgradeController(from: parentController, source: "bookmarks_locked_state")
```
leandroalonso: nice

leandroalonso (approved):

PR #1053, opened 2023-08-19
Bookmarks: Fixes an error that can happen when syncing bookmarks

emilylaguna:
This fixes some errors that I've seen while testing syncing:
- ` Unknown error calling sqlite3_step (19: UNIQUE constraint failed: Bookmark.uuid) rs`
  - This was happening while doing a full sync where some of the local bookmarks were marked as deleted, and when we tried to import the incoming one we would think the bookmark wasn't in our local database but it was. 
  - `SyncTask: Process Server Bookmarks - Could not add bookmark:`
      - This was triggered when importing existing bookmarks at the same timestamp on an episode as another in our local database. Since the `existingBookmark` check would return nil, the process thought it failed. I've removed that check to fix this .

## To test

I'm not entirely sure how to get into a state to reproduce these because it requires some data on the server to be in a specific state and your local device to also be in a state. 

I would try deleting some local bookmarks, logging out and back in to trigger full sync, and verifying you don't see any errors during syncing. 

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/BookmarkDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/BookmarkDataManager.swift
@@ -86,8 +86,8 @@ public struct BookmarkDataManager {
     // MARK: - Retrieving
 
     /// Retrieves a single Bookmark for the given UUID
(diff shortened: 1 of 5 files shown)


leandroalonso (approved):
Tested on two different devices, while creating bookmarks on the same timestamp and performing full and incremental syncs. Works as expected.

PR #1055, opened 2023-08-22
EffectsPlayer: add different setup mechanisms

leandroalonso:
This PR adds 3 different ways of setting up the `EffectsPlayer`:

1. The current one. It just calls `play` on `player` and if it crashes, it crashes
2. A retry mechanism. The player attempts to create and play itself three times, in the third time if it fails it crashes
3. A fallback mechanism. If the player creation/play fails, it falls back to `DefaultPlayer`

Which mechanism will be used is controlled by a remote feature flag, allowing us to test different scenarios without deploying a new app version.

Also part of this PR, I removed the old remote feature flag code for removing the `EffectsPlayer` lock as this didn't have any effect.

## To test

First, configure the environment:

1. Go to `AppDelegate.swift`, locate the `configureFirebase` method, and change `2.hour` to `30.seconds`.
3. Comment (or remove) `EffectsPlayer.swift` lines `143` to `157` (by doing this the engine won't start thus the playback will fail)
2. Download some episodes or add Files

### Normal play

Go to Firebase Console and change `effects_player_strategy` feature flag value to `1` and publish it.

1. Run the app
2. Play a downloaded episode or local file
3. ✅ The app should crash

### Retrying three times

Go to Firebase Console and change `effects_player_strategy` feature flag value to `2` and publish it.

1. Run the app
2. Play a downloaded episode or local file
3. ✅ The app should crash
4. Check the logs, you should see three `Loading <episode title>`, and for the first two you should see something like `EffectsPlayer: failed to start playback`

### Falling back to DefaultsPlayer

Go to Firebase Console and change `effects_player_strategy` feature flag value to `3` and publish it.

1. Run the app
2. Play a downloaded episode or local file
3. ✅ The episode should play
4. Check the logs and you should see `Playback Failed, attempting to fallback to: DefaultPlayer`

### Changing remote feature flag to 2

Go to Firebase Console and change `effects_player_strategy` feature flag value to `2` and publish it.

## Checklist

### DIFF
--- a/podcasts/AppDelegate.swift
+++ b/podcasts/AppDelegate.swift
@@ -275,7 +275,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             Constants.RemoteParams.episodeSearchDebounceMs: NSNumber(value: Constants.RemoteParams.episodeSearchDebounceMsDefault),
             Constants.RemoteParams.podcastSearchDebounceMs: NSNumber(value: Constants.RemoteParams.podcastSearchDebounceMsDefault),
             Constants.RemoteParams.customStorageLimitGB: NSNumber(value: Constants.RemoteParams.customStorageLimitGBDefault),
-            Constants.RemoteParams.endOfYearRequireAccount: NSNumber(value: Constants.RemoteParams.endOfYearRequireAccountDefault)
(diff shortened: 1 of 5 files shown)


emilylaguna (approved):
🤞 Tested using the steps and hit all the expected logs/errors.

PR #1056, opened 2023-08-23
Update Protobuf and fix logs

leandroalonso:
Update Protobuf to the latest version to ensure we're using the most up-to-date version.

I also fixed some logs so they report the correct server method used (POST/GET/DELETE) and the URL.

## To test

1. Run the app
2. Login
3. ✅ Perform actions such as subscribing to podcasts and make sure everything is synced

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Private/API Tasks/ApiBaseTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/API Tasks/ApiBaseTask.swift
@@ -54,7 +54,7 @@ class ApiBaseTask: Operation {
 
             return (responseData, httpResponse.statusCode)
         } catch {
-            FileLog.shared.addMessage("Failed to POST to server (\(url) \(error.localizedDescription)")
+            FileLog.shared.addMessage("Failed to POST to server (\(url)) \(error.localizedDescription)")
         }
 
         return (nil, ServerConstants.HttpConstants.serverError)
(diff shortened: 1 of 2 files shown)


Modules/Server/Sources/PocketCastsServer/Private/API Tasks/ApiBaseTask.swift
```diff
@@ -54,7 +54,7 @@ class ApiBaseTask: Operation {
 
             return (responseData, httpResponse.statusCode)
         } catch {
-            FileLog.shared.addMessage("Failed to post to server \(error.localizedDescription)")
+            FileLog.shared.addMessage("Failed to POST to server (\(url) \(error.localizedDescription)")
```
emilylaguna:
```suggestion
            FileLog.shared.addMessage("Failed to POST to server (\(url)) \(error.localizedDescription)")
```

emilylaguna (changes requested):
Nice, this is working well in my tests of logging in, performing sync actions, etc. 

I have a couple of suggestions:

1. Since we're using the same structure for each of the logs what do you think about moving the logging into a `logFailure` method?
2. What do you think about also including the subclass name in the log? 

Something like: `FileLog.shared.addMessage("[\(type(of: self))] Failed to POST to server (\(url)) \(error)")` which will output: 

`[UpNextSyncTask] Failed to POST to server (https://api.pocketcasts.com/up_next/sync) Error`

I know we can infer which subclass the log came from by looking at the URL but I think including it in the log would help make it explicitly clear.

Modules/Server/Sources/PocketCastsServer/Private/API Tasks/ApiBaseTask.swift
```diff
@@ -88,7 +88,7 @@ class ApiBaseTask: Operation {
 
             return (responseData, httpResponse)
         } catch {
-            FileLog.shared.addMessage("Failed to post to server \(error.localizedDescription)")
+            FileLog.shared.addMessage("Failed to GET from server (\(url) \(error.localizedDescription)")
```
emilylaguna:
```suggestion
            FileLog.shared.addMessage("Failed to GET from server (\(url)) \(error.localizedDescription)")
```

leandroalonso:
@emilylaguna good suggestion! Done in 112c0bc0ff8c7331bbd2f1dab74e9f9265c7a9c6

emilylaguna (approved):

PR #1059, opened 2023-08-24
Autoplay: add missing tracks

leandroalonso:
Adds two tracks that were added on the web:

* `autoplay_started`: whenever an autoplay session starts
* `autoplay_finished_last_episode`: when the last episode of a playlist finishes playing

## To test

1. Run the app
2. Go to Profile > Settings > Beta Features > enable `tracksLogging`
3. Clean your Up Next if there's any item (long press the miniplayer > "Close and Clear Up Next")
4. Play an episode from any place
5. ✅ You should see a log like: `🔵 Tracked: autoplay_started ["source": "SOURCE"]` (with SOURCE being `podcast`, `filter`, `downloads`, `files` or `starred`)
6. Play an episode from any playlist or podcast with less than 5 episodes (Discover has the "House of R")
7. Move the scrubber to the end until you finish all episodes
8. ✅ After the last one, you should see a log `🔵 Tracked: autoplay_finished_last_episode`

## Checklist

### DIFF
--- a/podcasts/AutoplayHelper.swift
+++ b/podcasts/AutoplayHelper.swift
@@ -4,12 +4,27 @@ import PocketCastsUtils
 
 /// Reponsible for handling the Autoplay of episodes
 class AutoplayHelper {
-    enum Playlist: Codable {
+    enum Playlist: Codable, AnalyticsDescribable {
         case podcast(uuid: String)
         case filter(uuid: String)
         case downloads
         case files
         case starred
+
+        var analyticsDescription: String {
+            switch self {
+            case .podcast(uuid: _):
+                return "podcast"
+            case .filter(uuid: _):
+                return "filter"
+            case .downloads:
+                return "downloads"
+            case .files:
+                return "files"
+            case .starred:
+                return "starred"
+            }
+        }
     }
 
     #if !os(watchOS)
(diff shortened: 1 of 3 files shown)


podcasts/AutoplayHelper.swift
```diff
@@ -71,6 +89,12 @@ class AutoplayHelper {
 
         userDefaults.set(data, forKey: userDefaultsKey)
         FileLog.shared.addMessage("Autoplay: saving the latest playlist: \(playlist)")
+
+        // We always save the playlist no matter if Up Next is empty or not
+        // However this event should be fired only if the Up Next is empty.
+        if upNextQueue.upNextCount() == 0 {
+            Analytics.track(.autoplayStarted, properties: ["source": playlist])
```
emilylaguna: If I have Autoplay disabled in settings this still gets fired.

emilylaguna (changes requested):
All works, except for I left 1 comment about the autoplay started event firing even if I have the setting disabled.

podcasts/AutoplayHelper.swift
leandroalonso: Fixed in 6c55829d32324cfcf1fcaca38e188502f5631667

emilylaguna (approved):

Issue #1061, opened 2023-08-24
Bookmarks: Final Touches


emilylaguna:
### What's New
- [x] Add What's New for Bookmarks: https://github.com/Automattic/pocket-casts-ios/pull/1172

### Bookmark Creation
- [x] When creating a bookmark that is at the same time as another bookmark, display the edit title screen: https://github.com/Automattic/pocket-casts-ios/pull/1068

### Bookmark Play
- [x] Stay on the same screen when playing a bookmark: https://github.com/Automattic/pocket-casts-ios/pull/1069

### Player
- [x] Rename the Show Notes tab to 'Details': https://github.com/Automattic/pocket-casts-ios/pull/1067
- [x] Update the tab style to [reflect the updated style in the designs](https://www.figma.com/file/Io7PCz1H1hmOgEE9eljoYW/Bookmarks?type=design&node-id=2032-52290&mode=dev): https://github.com/Automattic/pocket-casts-ios/pull/1067
- [x] Increase the tab switch transition speed: https://github.com/Automattic/pocket-casts-ios/pull/1067
- [x] When bookmarks are enabled, have the player action appear between mark as played and archived in the more menu: https://github.com/Automattic/pocket-casts-ios/pull/1069

### Upgrade Screen
- [x] Change the upgrade button title to be: https://github.com/Automattic/pocket-casts-ios/pull/1069
   - If the tier has a free trial: `Start Free Trial`
   - If not: `Upgrade to X`
      - Where X is name of the tier (Plus, or Patron)

### Bookmark Sorting
- [x] Store the sort option on a "per list" basis where each bookmark lists option is saved separately: https://github.com/Automattic/pocket-casts-ios/pull/1069

### Edit Title Screen
- [x] Button color can display incorrectly for some podcasts (Ie: The Pocket Cast): https://github.com/Automattic/pocket-casts-ios/pull/1066
- [x] When editing a bookmark change the save button title to 'Change Title': https://github.com/Automattic/pocket-casts-ios/pull/1066

### User Files
- [x] When creating a bookmark with the files list open, the bookmark indicator doesn't display unless you reopen the view: https://github.com/Automattic/pocket-casts-ios/pull/1065

### Headphone Controls
- [x] Add the confirmation tone: https://github.com/Automattic/pocket-casts-ios/pull/1076
- [x] Show all the available options for next and prev: https://github.com/Automattic/pocket-casts-ios/pull/1069

### Syncing
[...]

PR #1068, opened 2023-08-25
Bookmarks: Shows the edit screen if the bookmark that was created already existed at the current time

emilylaguna:
| 📘 Part of: #1061 |
|:---:|

## To test

1. Enable the bookmarks and patron feature flags
2. Open the full screen player
3. Pause the episode
4. Tap the Add Bookmark option
5. ✅ You should see the 'Add Bookmark' screen
6. Dismiss it
7. Tap the 'Add Bookmark' item again
8. ✅ You should see the edit bookmark screen

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/BookmarkDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/BookmarkDataManager.swift
@@ -9,6 +9,13 @@ public struct BookmarkDataManager {
         self.dbQueue = dbQueue
     }
 
+    /// Looks for any existing bookmarks in an episode that have the same start time
+    public func existingBookmark(forEpisode episodeUuid: String, time: TimeInterval) -> Bookmark? {
+        selectBookmarks(where: [.episode, .time],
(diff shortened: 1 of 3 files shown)


leandroalonso (approved):
@emilylaguna it works as described.

However, I think handling this specific case might potentially cause some (very rare) edge cases. What if a user is listening to an episode again and ends up adding a bookmark at the same time as a previously added one? But this time with additional context on the title. Then they check their bookmarks and the previous title is... gone. Is a destructive implicit action.

Not strong though but I'd leave this case unhandled. Added multiple bookmarks for the same timestamp? They all get added. Up to the user to manage them (plus less code).

emilylaguna:
I can see this case, although I'm not sure how often this will happen. 

I think for now we can keep it the duplicate handling and see how people are using bookmarks and adjust in the future if needed. 

> But this time with additional context on the title

One thing is that the user will see their old bookmark title with the wording that they are updating the title, so hopefully that gives them the context of what's happening.

leandroalonso:
> One thing is that the user will see their old bookmark title with the wording that they are updating the title, so hopefully that gives them the context of what's happening.

Oh I missed that. Nice!

PR #1075, opened 2023-08-28
Episode Artwork: enable only on TestFlight builds

leandroalonso:
Until this feature has cross-platform parity we'll have it enabled only on TestFlight.

## To test

I'm really not sure how we can test that without doing a release on TestFlight. 🤷 

## Checklist

### DIFF
--- a/podcasts/AppDelegate.swift
+++ b/podcasts/AppDelegate.swift
@@ -54,6 +54,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         GoogleCastManager.sharedManager.setup()
 
         CacheServerHandler.newShowNotesEndpoint = FeatureFlag.newShowNotesEndpoint.enabled
+        CacheServerHandler.episodeFeedArtwork = FeatureFlag.episodeFeedArtwork.enabled
 
         setupRoutes()
 
--- a/podcasts/FeatureFlag.swift
+++ b/podcasts/FeatureFlag.swift
@@ -51,10 +51,6 @@ enum FeatureFlag: String, CaseIterable {
             return overriddenValue
         }
 
-        if availableOnlyOnTestFlight && !FeatureFlag.isTestFlight {
-            return false
-        }
-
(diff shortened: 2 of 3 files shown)


podcasts/AppDelegate.swift
```diff
@@ -54,6 +54,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         GoogleCastManager.sharedManager.setup()
 
         CacheServerHandler.newShowNotesEndpoint = FeatureFlag.newShowNotesEndpoint.enabled
+        CacheServerHandler.episodeFeedArtwork = FeatureFlag.episodeFeedArtwork.enabled
```
leandroalonso, line 57: We already discussed about moving `FeatureFlag` to Utils but I decided to keep as it is given we're targeting a release branch.

podcasts/FeatureFlag.swift
```diff
@@ -77,6 +84,18 @@ enum FeatureFlag: String, CaseIterable {
             return true
         case .newShowNotesEndpoint:
             return true
+        case .episodeFeedArtwork:
+            return true
```
emilylaguna: I wonder if it would be easier to just define a feature as enabled on TestFlight using a check on the feature flag value itself, rather than defining a new private property?

```suggestion
            return FeatureFlag.isTestFlight ? true : false
```

emilylaguna (changes requested):
> I'm really not sure how we can test that without doing a release on TestFlight. 🤷

I found that if I run the app on a real device the isTestFlight check returns true because it also uses the sandboxReceipt. 

I left a comment below with a suggestion, let me know what you think.

podcasts/FeatureFlag.swift
leandroalonso: Done in afcef1b18f072a6b5a9bd275f8138ced210e8cad

emilylaguna (approved):
🚢

PR #1077, opened 2023-08-28
Bookmarks: Allow specifying different values when beta testing

emilylaguna:
This adds a way to specify different tiers for a feature during beta testing. 

This also adds the `BuildEnvironment` enum that describes the type environment the app is currently running in. 

## To test

Note: Enable the Bookmarks and Patron feature flags

1. Launch the app
2. Go to Profile Tab > Cog > Developer > Set No Plus
3. Go to full screen player > Bookmarks tab
4. ✅ Verify you see Upgrade to Patron
5. In Xcode, open `BuildEnvironment.swift`
6. Remove lines 17 - 19, and 26 (removing the if check)
7. Run the app on a real device
   - When testing on a real device the `appStoreReceiptURL` returns the `sandboxReceipt` value which is what we use to check if it's TestFlight.
9. Go to Profile Tab > Cog > Developer > Set No Plus
10. Go to the full screen player > Bookmarks tab
11. ✅ Verify you see Upgrade to Plus

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/General/BuildEnvironment.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/General/BuildEnvironment.swift
@@ -1,0 +1,28 @@
+import Foundation
+
+/// Describes the type of build environment the app is running in.
+/// Use `BuildEnvironment.current` to get the environment for the current running build.
+public enum BuildEnvironment {
+    /// From Xcode, or another DEBUG build
+    case debug
+
+    /// A release build from TestFlight
+    case testFlight
+
+    /// A release build from the AppStore
+    case appStore
+
+    /// Returns the `BuildEnvironment` for the current build
+    public static var current: BuildEnvironment {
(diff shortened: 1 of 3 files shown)


Modules/Utils/Sources/PocketCastsUtils/General/BuildEnvironment.swift
```diff
@@ -0,0 +1,34 @@
...
+    case debug
+
+    /// A release build from TestFlight
+    case testFlight
+
+    /// A release build from the AppStore
+    case appStore
+
+    /// Returns the `BuildEnvironment` for the current build
+    public static var current: BuildEnvironment = .determineCurrentEnvironment
```
emilylaguna, line 16: I'm using `.determineCurrentEnvironment` so the `.current` value is only calculated once since the environment state shouldn't need to change at all.

leandroalonso (approved):

PR #1088, opened 2023-09-05
Player: improve transition

leandroalonso:
Address part of #1037 

https://github.com/Automattic/pocket-casts-ios/assets/7040243/8280532e-3587-4d1f-b6d6-a82f170015f1

This PR refactors the player transition to:

1. Use the system transition instead of doing it manually
2. By doing the above, speed and responsiveness is increased
3. It does it by using a Feature Flag, so if any issues arise we can revert to what it was very quickly.

**This PR does not add a custom transition as this will be part of another task**

## Additional information

Turned out, that this change was more complex than it seemed. 🙃A lot of places on the app relied on presenting stuff from the `rootViewController` of the app (the `MainTabBarController`). However, with this change, the `rootViewController` was already presenting something, so stuff like Chromecast, Bookmarks, Links, etc. weren't being presented.

I fixed this by:

1. The `present` method from `MainTabBarController` now checks to see if there is already a `presentedViewController`. If there is, the presentation is done through `presentedViewController` (the player)
2. A side effect of these changes was `player_shown` and `player_dimissed` methods being triggered. I had to manually handle these cases to not cause this. However, there are a few instances in the current `trunk` where tracks are triggered while they shouldn't (eg.: open player for a video podcast ~> tap the video ~> dismiss it ~> the presented tab will trigger its event)

### Old transition

1. Run the app with the `newPlayerTransition` flag *disabled*
2. Open the player
3. Swipe up
4. ✅ Up Next screen should appear
5. Dismiss it
6. Tap Chromecast
7. ✅ The Chromecast screen should appear
8. Go to "Details", tap any link
9. ✅ The link should open
10. Enable the Bookmarks feature flag
11. ✅ Check that when you tap "Upgrade to Patreon" the view displays
12. Play a video podcast
13. Tap the video
14. ✅ The video should be displayed in a new screen
15. Go landscape
16. ✅ The video should fill the whole screen

### New transition

1. Run the app (preferably on a real device)
4. If you have something playing, tap the miniplayer
5. ✅ The full player should appear
6. Swipe down
7. ✅ The player should dismiss
8. Swipe up in the miniplayer
9. ✅ The player should appear

### New transition flows

1. Run the app with the `newPlayerTransition` flag *enabled*
2. Open the player
3. Swipe up
4. ✅ Up Next screen should appear
5. Dismiss it
6. Tap Chromecast
7. ✅ The Chromecast screen should appear
8. Go to "Details", tap any link
9. ✅ The link should open
10. Enable the Bookmarks feature flag
11. ✅ Check that when you tap "Upgrade to Patreon" the view displays
12. Play a video podcast
13. Tap the video
14. ✅ The video should be displayed on a new screen
15. Go Landscape
16. ✅ The video should fill the whole screen

Make sure to test on iPad and with different themes (dark and white mode).

## Checklist


podcasts/PlayerContainerViewController+Gestures.swift
```diff
@@ -11,14 +11,43 @@ extension PlayerContainerViewController: UIGestureRecognizerDelegate {
...
+                    UIView.animate(withDuration: 0.2, animations: {
+                        self.view.frame = CGRect(x: 0,
+                                                 y: 0,
+                                                 width: self.view.frame.size.width,
+                                                 height: self.view.frame.size.height)
+                    })
+                }
+            default:
+                break
+            }
```
emilylaguna: To improve readability what do you think about moving this into a separate function?

emilylaguna (changes requested):
Overall this works well, I did notice a weird issue when playing a video though. 

1. Play a video podcast
2. Tap on the video
3. The background disappears
4. Going landscape also causes the background to be clear

I don't see this with the old code. 

https://github.com/Automattic/pocket-casts-ios/assets/793774/543c9633-c875-40fb-ad26-86677c0e6a97

podcasts/PlayerContainerViewController+Gestures.swift
leandroalonso: Done in 323423413e8b1ad4d8b2dd636dd30d0cc9cd3907

leandroalonso:
@emilylaguna ready for another review. The testing steps have also been updated.

podcasts/PlayerContainerViewController.swift
```diff
@@ -78,6 +78,9 @@ class PlayerContainerViewController: SimpleNotificationsViewController, PlayerTa
 
     var finalScrollViewConstraint: NSLayoutConstraint?
 
+    /// Whether the player is appearing or disappearing because of an user interaction
+    private var isAppearingOrDisappearingThroughUserGesture = true
```
emilylaguna: Instead of adding a flag and having the child controllers present from the parent, could we move the shown/dismissed logic to the miniPlayer in the `openFullScreenPlayer` and `finishedWithFullScreenPlayer` methods?

podcasts/MainTabBarController.swift
```diff
@@ -585,6 +585,19 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
             controller.present(whatsNewViewController, animated: true)
         }
     }
+
+    // There are different areas of the app that relies on presenting VCs from the tab bar
+    // However, sometimes the tab bar is already displaying the player.
+    // This code simple checks if the tab bar is already presenting something and, if yes,
+    // present the VC through the presentedViewController
+    override func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
+        if FeatureFlag.newPlayerTransition.enabled, let presentedViewController {
```
emilylaguna: Note: I made some comments below about an alternative to presenting on top of the presented view.

If I enable 'Open Player Automatically' in the General settings and then:

1. Open a Podcast
2. Tap on an episode to see its details
3. Tap Play
4. I see: `Attempt to present <podcasts.PlayerContainerViewController: 0x11982c800> on <podcasts.EpisodeDetailViewController: 0x11782f600> (from <podcasts.EpisodeDetailViewController: 0x11782f600>) whose view is not in the window hierarchy.`

I did some digging and the isBeingDismissed flag is set to true and checking it fixes this for me.

```suggestion
        if FeatureFlag.newPlayerTransition.enabled, let presentedViewController, !presentedViewController.isBeingDismissed {
```

podcasts/MiniPlayerViewController+Positioning.swift
```diff
@@ -37,11 +37,43 @@ extension MiniPlayerViewController {
...
 
         if playerOpenState == .open || playerOpenState == .animating { return }
 
+        guard !FeatureFlag.newPlayerTransition.enabled else {
+            playerOpenState = .animating
+            aboutToDisplayFullScreenPlayer()
+
+            fullScreenPlayer?.modalPresentationStyle = .overCurrentContext
+
+            presentFromRootController(fullScreenPlayer!, animated: true) {
```
emilylaguna: Since we're presenting now there's an odd state that we can get into if the player gets presented when we are already presenting something like the episode details or folders:

We then get into the same issue of the presenting error with bookmarks, etc:

1. Add the Up Next widget to your home screen
2. Open the app and add a folder or open an episode details
3. Go back to your home screen
4. Tap on the playing episode in the widget
5. The app will open and the player will be presented on top of the view
6. Try presenting another view
7. 💥

emilylaguna: I wonder if instead of modifying the MainTabViewController's present method if instead we ensured that the player will always appear by dismissing the presented view if there is one?

However this could result in someone losing their folder their adding, or being kicked out of the upgrade flow. 

An alternative would be to check if there is a presented view and not doing anything. 

Thoughts?

emilylaguna (changes requested):
Nice work, the video changes work well. 

I did some more extensive testing and found some new issues 😢 

Let me know what you think.

podcasts/MiniPlayerViewController+Positioning.swift
leandroalonso: > Thoughts?

I think we should follow the system pattern: stack the views by presenting on the top `presentedViewController`. I did that on 31ad6b7cff3e8eea3f15808f28046dcdce6f0f0c

Here's a video of the flow involving purchase and Bookmark:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/cd02a53a-66f6-4456-9726-a7169cc0c21f

Ps.: I ended up changing the sign-out listener too: 2d48e018443055bcbbbdabf4a6b06c9bf5eb0dea

leandroalonso:
@emilylaguna as always, thanks for the thorough review. Ready for another look!

podcasts/SceneHelper.swift
```diff
@@ -24,6 +24,16 @@ class SceneHelper {
     }
 
     class func rootViewController() -> UIViewController? {
+        guard !FeatureFlag.newPlayerTransition.enabled else {
+            if let scene = connectedScene() {
+                for window in scene.windows {
+                    return Self.getPresentedViewController(base: window.rootViewController)
```
emilylaguna: This will only ever return the first value regardless of whether the value is nil or not.

podcasts/SceneHelper.swift
```diff
@@ -38,6 +48,19 @@ class SceneHelper {
...
+                return nil
+            }
+
+            if let nav = base as? UINavigationController {
+                return getPresentedViewController(base: nav.visibleViewController)
+            } else if let presented = base?.presentedViewController {
+                return getPresentedViewController(base: presented)
+            }
+            return base
+        }
```
emilylaguna: Could we move this into the UIViewController extension?

```

    var topMostViewController: UIViewController? {
        if let presentedViewController {
            return presentedViewController.topMostViewController
        }

        if let navController = self as? UINavigationController {
            return navigationController?.visibleViewController?.topMostViewController
        }

        return self
    }
```

podcasts/BackgroundSignOutListener.swift
```diff
@@ -5,15 +5,17 @@ import PocketCastsServer
...
 
+    private var presentingViewController: UIViewController? {
+        SceneHelper.rootViewController()
+    }
+
     private var canShowSignOut = true
 
-    init(notificationCenter: NotificationCenter = NotificationCenter.default, presentingViewController: UIViewController,
+    init(notificationCenter: NotificationCenter = NotificationCenter.default,
          navigationManager: NavigationManager = NavigationManager.sharedManager) {
```
emilylaguna: We can still inject the presenting controller and have it calculate the value on runtime if we use [`@autoclosure`](https://www.swiftbysundell.com/articles/using-autoclosure-when-designing-swift-apis/) in the init value:

```
class BackgroundSignOutListener {
    private var presentingViewController: () -> UIViewController?

    init(presentingViewController: @autoclosure @escaping () -> UIViewController?) {
        self.presentingViewController = presentingViewController
    }

    func showAlertIfPossible() {
        presentingViewController()?.present...
    }
}
```

This will auto wrap the value in the closure and allow us to keep the same initializer:

```
BackgroundSignOutListener(presentingViewController: SceneHelper.rootViewController())
```

podcasts/Bookmarks/Player/BookmarksPlayerTabController.swift
```diff
@@ -117,6 +117,13 @@ class BookmarksPlayerTabController: PlayerItemViewController {
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+    // MARK: - Present
+
+    /// Always present from the parent VC
+    override func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
+        parent?.present(viewControllerToPresent, animated: flag, completion: completion)
+    }
```
emilylaguna: Since the Bookmarks/Chapters/Now Player/Show Notes all subclass from `PlayerItemViewController` can we this logic there?

podcasts/SceneHelper.swift
emilylaguna: This logic looks to be used in `AnalyticsCoordinator.getTopViewController` can remove this to use the new logic?

podcasts/MiniPlayerViewController.swift
```diff
@@ -94,6 +102,20 @@ class MiniPlayerViewController: SimpleNotificationsViewController {
     func finishedWithFullScreenPlayer() {
         guard let rootVC = rootViewController() else { return }
 
+        guard !FeatureFlag.newPlayerTransition.enabled else {
+            rootViewController()?.setNeedsStatusBarAppearanceUpdate()
+            rootViewController()?.setNeedsUpdateOfHomeIndicatorAutoHidden()
+            fullScreenPlayer?.view.removeFromSuperview()
+            fullScreenPlayer = nil
+
+            Analytics.track(.playerDismissed)
```
emilylaguna: Small nit. If I swipe down/left/right on the mini player this is triggered.

podcasts/SceneHelper.swift
emilylaguna: I got inspired looking at this to make a new utility: [mapFirst](https://github.com/Automattic/pocket-casts-ios/pull/1126/files)

We could clean this up by doing replacing this with:

```
return connectedScene()?.windows.mapFirst { getPresentedViewController(base: $0.rootViewController) }
```

emilylaguna (changes requested):
This is working well in my tests, I left some comments to improve the code and reduce some duplication.

podcasts/SceneHelper.swift
leandroalonso: Done in 3f37e4051b2f644387d270ea753eebeb9ef20f46

podcasts/SceneHelper.swift
leandroalonso: Done in 9cf0e5e17ae961f5877590d3e490864db41430cf

There's a slight difference between them:

1. This one returns the top most presented view controller, so if there's nothing presented it will return the `MainTabBarController`
2. On the other hand, the one on `AnalyticsCoordinator` iterates over the presented tabs tor return the presented VC

podcasts/BackgroundSignOutListener.swift
leandroalonso: Nice!

Done in 393bb75d54159795f9efbe47cbc463a7c14c4299

podcasts/Bookmarks/Player/BookmarksPlayerTabController.swift
leandroalonso: Ohhhhhh, thanks! Done in c50c169d09e65cfe0a96816c9f7ee4533db9e64c

podcasts/MiniPlayerViewController.swift
leandroalonso: This was a bit annoying to fix, done in 8b607f748c48ce5cba3b6c371e9bb49d83304846

leandroalonso:
@emilylaguna ready for another another review! Fixed the tests too. 🤓

podcasts/UIViewController+Presenting.swift
```diff
@@ -1,6 +1,22 @@
 import UIKit
 
 extension UIViewController {
+    var topMostPresentedViewController: UIViewController? {
+        guard UIApplication.shared.applicationState == .active else {
+            return nil
+        }
+
+        if let presentedViewController {
```
emilylaguna: On iPad when I have an episode details popover opened and open player automatically enabled the player won't appear.

Adding the check below and applying the [fix from here](https://github.com/Automattic/pocket-casts-ios/pull/1088/files#r1332013567) resolves this for me. 

```suggestion
        if let presentedViewController, !presentedViewController.isBeingDismissed {
```

podcasts/MiniPlayerViewController+Positioning.swift
```diff
@@ -37,11 +37,44 @@ extension MiniPlayerViewController {
...
```
emilylaguna: Can we add a check instead of force unwrapping here?

There's a weird state I can put the app into that will cause a crash:
1. Add the up next widget
2. Go to Profile > Settings Cog > Private > Read privacy policy
3. Close app
4. Tap on Up next widget
5. 💥 crash

```suggestion
            presentFromRootController(fullScreenPlayer!, animated: true) {
```

emilylaguna (changes requested):
I gave this a test on iOS 15.5 and iPad running 16.6 and I hit an issue with the iPad as noted below.

podcasts/UIViewController+Presenting.swift
leandroalonso: Done in cd5a9fd6731cffb3708c1667f2d6136b81c72d71

podcasts/MiniPlayerViewController+Positioning.swift
leandroalonso: Done in cd5a9fd6731cffb3708c1667f2d6136b81c72d71

I know the nested `guard` is ugly as hell but once we remove the feature flag just one will be there so I think we're good.

leandroalonso:
@emilylaguna ready for another another another review!

podcasts/MiniPlayerViewController+Positioning.swift
```diff
@@ -37,11 +37,48 @@ extension MiniPlayerViewController {
...
 
         if playerOpenState == .open || playerOpenState == .animating { return }
 
+        guard !FeatureFlag.newPlayerTransition.enabled else {
+            guard let fullScreenPlayer else {
+                return
+            }
+
+            playerOpenState = .animating
+            aboutToDisplayFullScreenPlayer()
```
emilylaguna: When tapping the mini player the fullScreenPlayer check will always be nil since it's set in `aboutToDisplayFullScreenPlayer`. 

```suggestion
            aboutToDisplayFullScreenPlayer()

            guard let fullScreenPlayer else {
                return
            }

            playerOpenState = .animating

```

podcasts/MainTabBarController.swift
emilylaguna: I am still able to reproduce this in the current build, but applying the isBeingDismissed check fixes it.

emilylaguna (changes requested):
😢 Found a couple more issues, but I think we're good to go after that.

leandroalonso:
> 😢 Found a couple more issues, but I think we're good to go after that.

Sorry @emilylaguna, I thought I had addressed those issues but clearly I didn't. 🫠

pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/31">7.48</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/36">7.53</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

emilylaguna (approved):
I went through and tested everything I could and all is working as intended.

PR #1091, opened 2023-09-08
Subscribe route: fix logic that append "http://"

leandroalonso:
This PR fixes the logic when importing a new URL through the `pktc://subscribe/[…]` URL.

The way it worked before, it would append `http://` to the URL, it doesn't matter if it had it already or not.

## To test

1. Grab the XML content for any podcast ([feel free to use this one](https://gist.githubusercontent.com/leandroalonso/ad74c71380eb90ca466aa80ada883d6f/raw/7de9613e073fe8349c2c7a81e735c49e459f795e/teste.xml))
2. Create a GitHub gist with it and save the raw file URL
3. Run the app
4. Go to Safari and paste `pktc://subscribe/YOUR-GIST`, where `YOUR-GIST` is the URL to the raw file. Eg.: `pktc://subscribe/https://gist.githubusercontent.com/leandroalonso/ad74c71380eb90ca466aa80ada883d6f/raw/7de9613e073fe8349c2c7a81e735c49e459f795e/teste.xml`, then tap enter
5. When opened to Pocket Casts, accept it
6. ✅ The podcast should appear correctly
7. Test the same without the `https://` part (eg.: `pktc://subscribe/gist.githubusercontent.com/leandroalonso/ad74c71380eb90ca466aa80ada883d6f/raw/7de9613e073fe8349c2c7a81e735c49e459f795e/teste.xml`

## Checklist


emilylaguna (approved):
This works!

One thing, is there a reason we default to http instead of https?

leandroalonso:
> One thing, is there a reason we default to http instead of https?

I kept what was there. 🤷

PR #921, opened 2023-06-22
[poc/wip] Interactive Up Next widget

jgcaruso:
NOTE: This is a very rough experiment

Fixes #

- adds an AudioPlaybackIntent
   - an AppIntent is used for triggering actions in an interactive widget button or toggle
   - AudioPlaybackintent allows the app to be re-started if it isn't running in the background
- adds a new internal notification to request playback, called from the AudioPlaybackIntent. Not sure how good an idea this is, but it works enough to get a feel for this feature.

to do:
- nicer, more relevant animations?

## To test

Steps to test your PR.

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1095, opened 2023-09-13
iOS 17 widgets: add play/pause functionality

leandroalonso:
First of all, all kudos to @jgcaruso. The code that powers this feature comes from his explorations in #921. I just cleaned up a bit and added the Pocket Casts play/pause icons. 

## To test

1. Run the app on a device/simulator with iOS 17
2. Tap and hold on the apps screen and add the widgets: Now Playing, Up Next (medium) and Up Next (big)
3. Tap play on any episode
4. ✅ The episode should start playing and the other widgets should update accordingly
5. Pause the episode
6. ✅ The episode should pause and the other widgets should update accordingly
7. Play any other episode
8. ✅ The episode should start playing and the other widgets should update accordingly

## Checklist


rviljoen:
@leandroalonso Great work! I've tested this on a real device with iOS 17 RC and it works perfectly. Some (subjective!) comments on the design:
 - The play/pause icon on the album art does not look good for some podcasts, especially those with a lot of white. Can the button be moved to the right of the podcast title?
 - If it stays as an overlay, is it possible to only animate the button and not the whole podcast art image when clicking the button?
 - I really like the play/pause animation in the main app. Is it possible to bring that to the widget as well (I suspect not, but worth asking 😀)

UNDER THE RADAR

jleandroperez (approved):
✅ Verified that Play / Pause update the playback state
✅ Verified that other widgets react accordingly
✅ Code looks super clean 👌 

:shipit: sir!!

leandroalonso:
@rviljoen thanks for the feedback!

We'll likely update the UI for 7.48. So I'll pass along your feedback. About bringing the animation, that's a cool idea. Not sure how feasible it is but I'll take a look. :D

PR #1097, opened 2023-09-14
iOS 17 Now Playing widget: adapt it for StandBy and iPad lock screen

leandroalonso:
This PR adapts the Now Playing widget for:

1. iPad landscape lockscreen: it doesn't display the background
2. StandBy: it doesn't show background and improve how items are shown
3. StandBy not playing: instead of the huge Pocket Casts icon we show a message that opens Discover

#### StandBy

| Playing | Now Playing |
| ------- | ------------ |
| IMG_0064 | IMG_0065 |

## To test

1. Run the app (device or simulator)
2. Add the Now Playing on the StandBy screen (if you're using phone, put it on charge and move it to be in landscape, on simulator just go to Features > Show StandBy)
3. ✅  If something is playing, interact with the play or pause button and see that it works
4. Go to Pocket Casts, tap and hold on the miniplayer and tap "Close And Clear Up Next"
5. Go back to StandBy mode
6. Tap it
7. ✅ You should go to Discover
8. Run the app on iPad
9. Add the Now Playing widget to the landscape lock screen
10. Check that it looks good

## Checklist

### DIFF
--- a/WidgetExtension/Common/ArtworkViews.swift
+++ b/WidgetExtension/Common/ArtworkViews.swift
@@ -2,28 +2,38 @@ import SwiftUI
 
 struct LargeArtworkView: View {
     @State var imageData: Data?
+
+    var showShadow: Bool = true
+
     var body: some View {
         ZStack {
-            Rectangle()
-                .foregroundColor(Color.nowPlayingShadowColor)
-                .aspectRatio(1, contentMode: .fit)
-                .frame(maxHeight: 74)
-                .cornerRadius(9)
-                .secondaryShadow()
+            if showShadow {
+                Rectangle()
+                    .foregroundColor(Color.nowPlayingShadowColor)
+                    .aspectRatio(1, contentMode: .fit)
+                    .frame(maxHeight: 74)
+                    .cornerRadius(9)
+                    .secondaryShadow()
+            }
+
             if let imageData = imageData, let uiImage = UIImage(data: imageData) {
                 Image(uiImage: uiImage)
                     .resizable()
                     .aspectRatio(1, contentMode: .fit)
                     .frame(maxHeight: 74)
                     .cornerRadius(8)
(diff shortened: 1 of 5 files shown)


podcasts/WidgetHelper.swift
```diff
@@ -148,7 +148,7 @@ class WidgetHelper {
         var imageUrl = ""
 
         if let episode = episode as? Episode {
-            imageUrl = ServerHelper.image(podcastUuid: episode.parentIdentifier(), size: 130)
+            imageUrl = ServerHelper.image(podcastUuid: episode.parentIdentifier(), size: 340)
```
leandroalonso, line 151: Had to increase the artwork size so it looks good on StandBy.

WidgetExtension/Common/ArtworkViews.swift
```diff
@@ -2,28 +2,38 @@ import SwiftUI
 
 struct LargeArtworkView: View {
     @State var imageData: Data?
+
+    var showShadow: Bool = true
```
jleandroperez, line 6: Perhaps this one should be `let` ?

jleandroperez (approved):
Looks super on my end!! Verified on iPad (Landscape) and iPhone Simulators.

:shipit:

WidgetExtension/Common/ArtworkViews.swift
leandroalonso: Curious enough, if I do that the `showShadow` stops being a property in the `struct` constructor. I would have to manually create an `init`, so I think that maybe this is an ok-tradeoff?

jleandroperez: If you changed this to `let` (but left the default value initializer, `= true`), the compiler would have removed the property from the default initializer.

No big deal though, thanks for checking!!

PR #1131, opened 2023-09-27
Ratings: add screens design

### LINKED ISSUE #1880: Add basic UI and foundations

leandroalonso:
This PR adds logic for:

* Showing a message when the user can't rate
* Showing the stars for the user to select when they can rate
* Dealing with success/error (stubbed)

Tracks and VoiceOver are not part of this task.

## To test

1. Go to Profile > Settings > Beta features > enable `giveRatings`

### Unable to give ratings

1. Go to `RatePodcastViewModel:checkIfUserCanRate` and change `self?.userCanRate = .allowed` to `self?.userCanRate = .disallowed`
2. Re-run the app
3. Go to Discover and tap any podcast
4. Tap the stars
5. ✅ You should see a message asking to listen to the podcast first

### Able to give ratings

I recommend running on a real device to test the haptic feedback.

#### Success

1. Go to `RatePodcastViewModel:checkIfUserCanRate` and change `self?.userCanRate = .disallowed ` to `self?.userCanRate = .allowed`
2. Re-run the app
3. Go to Discover and tap any podcast
4. Tap the stars
5. ✅ You should see a screen containing the stars and the "Submit" button shouldn't work
6. Tap any star
7. ✅ See that the correct number of stars is selected
8. Swipe over the stars
9. ✅ You'll select the correct number of stars and can select half stars
10. Tap the Submit button
11. ✅ The screen should disappear and you should see a toast confirmation message

#### Error

Go to `RatePodcastViewModel.submit` and replace the method to:

```swift
    func submit() {
        isSubmitting = true
        DispatchQueue.main.asyncAfter(deadline: .now() + 2) { [weak self] in
            self?.isSubmitting = false
            self?.anErrorOccurred = true
        }
    }
```

1. Re-run the app
2. Go to Discover > tap any podcast
3. Give it a few stars
4. Tap Submit
5: ✅ You should see an error alert, tapping "OK" just dismisses it

## Checklist


podcasts/Ratings/RatePodcastView.swift
```diff
@@ -1,23 +1,154 @@
...
+        static let maxStars = 5
+    }
+}
+
+extension Double {
+    func round(nearest: Double) -> Double {
+        let n = 1/nearest
+        let numberToRound = self * n
+        return numberToRound.rounded() / n
+    }
```
emilylaguna: It doesn't look this is being used?

podcasts/Ratings/RatePodcastView.swift
```diff
@@ -1,23 +1,154 @@
...
+        case .disallowed:
+            cannotRate
+        }
+    }
+
+    private var cannotRate: some View {
+        Group {
+            PodcastCover(podcastUuid: viewModel.podcast.uuid, big: true)
+                .frame(width: 164, height: 164)
+                .padding(.bottom, 40)
```
emilylaguna: We repeat this a few times, wdyt about moving this into a shared place?

podcasts/Ratings/RatePodcastView.swift
```diff
@@ -1,23 +1,154 @@
...
+            Text(L10n.ratingListenToThisPodcastTitle)
+                .font(size: 20, style: .title3, weight: .bold)
+                .padding(.bottom, 16)
+            Text(L10n.ratingListenToThisPodcastMessage)
+                .font(style: .body)
+                .multilineTextAlignment(.center)
+        }
+    }
+
+    private var rate: some View {
```
emilylaguna, line 67: We can remove the Groups here and in cannotRate to reduce the indenting if we add the  `@ViewBuilder` attribute.

podcasts/Ratings/RatePodcastView.swift
```diff
@@ -1,23 +1,154 @@
...
+                        }
+                        .foregroundStyle(theme.primaryText01)
+                        .frame(width: 36, height: 36)
+                        .padding(4)
+                        .onTapGesture {
+                            viewModel.stars = Double(index) + 1
+                        }
+                    }
+                }
+                .gesture(
```
emilylaguna: When dragging there was a few times where the view started dismissing. using a `highPriorityGesture` prevents that.

```suggestion
                .highPriorityGesture(
```

podcasts/Ratings/PodcastRatingViewModel.swift
```diff
@@ -13,11 +14,15 @@ class PodcastRatingViewModel: ObservableObject {
...
     /// the uuid yet
-    private var uuid: String? = nil
+    private(set) var uuid: String? = nil
+
+    private(set) var podcast: Podcast?
 
     /// Updates the rating for the podcast.
     ///
-    func update(uuid: String) {
+    func update(uuid: String, podcast: Podcast?) {
```
emilylaguna: Should we remove the UUID here and just use the podcast since it's doing the same thing?

podcasts/Ratings/RatePodcastView.swift
```diff
@@ -1,23 +1,154 @@
...
+    private var rate: some View {
+        Group {
+            PodcastCover(podcastUuid: viewModel.podcast.uuid, big: true)
+                .frame(width: 164, height: 164)
+                .padding(.bottom, 40)
+            Text(L10n.ratingTitle(viewModel.podcast.title ?? ""))
+                .font(size: 20, style: .title3, weight: .bold)
+                .padding(.bottom, 16)
+                .multilineTextAlignment(.center)
+            ContentSizeGeometryReader { reader in
```
emilylaguna: When I first load the view there's a split second where the rating view isn't fully loaded and pops in.

I think this is because when the `ContentSizeGeometryReader` first loads its frame isn't set which means it will take up the entire VStack space. Then after it calculates the content size it resizes itself. 

A fix for this would be to add a state property that is set to true when the `contentSizeUpdated` listener is fired. Then on the rating content we change the opacity from 0 to 1. 

https://github.com/Automattic/pocket-casts-ios/assets/793774/e21dc828-d091-44a6-8886-3bfd720053d0

podcasts/Ratings/RatePodcastViewModel.swift
```diff
@@ -1,9 +1,68 @@
...
+
+    @Published var stars: Double = 0 {
+        didSet {
+            if oldValue != stars {
+                HapticsHelper.triggerStarHaptic()
+            }
+        }
+    }
+
+    var podcast: Podcast
```
emilylaguna: Does this need to be a var?

emilylaguna (changes requested):
Nice work. I left a couple comments below. 

Also I don't know how to reproduce it but during testing there were a couple instances where the rating popover displayed twice. 

I think this could happen if the podcast view controller reloads while the popover is open.

podcasts/Ratings/RatePodcastView.swift
leandroalonso: Done in 976a6c6ca422a3a554cd42f844c4a401f345096f

podcasts/Ratings/RatePodcastView.swift
leandroalonso: Done in 5039193bdee1b39981c6c03e65e84f60ee762c5b

podcasts/Ratings/RatePodcastView.swift
leandroalonso: Done in 99c58549dd40a70c7fab99037b82cf5369b663c3

podcasts/Ratings/PodcastRatingViewModel.swift
leandroalonso: Done in 33ce40155db603e2212653adbb70c75b1c7f26ca

podcasts/Ratings/RatePodcastView.swift
leandroalonso: Oh, good catch! Fixed in 9b687fbbca1fcd6aa2eed2b47a234c9ae5e87251

podcasts/Ratings/RatePodcastViewModel.swift
leandroalonso: Nope! Done in 62a55384a87c21c87a0c36b4515941ef0198f513

podcasts/Ratings/StarRatingView.swift
```diff
@@ -29,7 +29,11 @@ struct StarRatingView: View {
         }.onTapGesture {
             viewModel.didTapRating()
         }
-        .sheet(isPresented: $viewModel.presentingGiveRatings) { RatePodcastView(viewModel: RatePodcastViewModel(presented: $viewModel.presentingGiveRatings)) }
+        .sheet(isPresented: $viewModel.presentingGiveRatings) {
```
emilylaguna, line 32: I saw the "Attempt to present ... which is already presenting" error a couple times in the console while testing. I think what is happening is the tableview is reloading and another rating view is created and attempts to present again. 

wdyt about instead of showing the sheet here, we can add a method to the `PodcastActionsDelegate` and on tap have the `PodcastViewController` display the sheet?

emilylaguna (approved):
Nice work. I left a comment about moving the sheet presentation outside of the star rating view. 

I'm approving this one since, if we go that route we can do it in another PR to keep this one more focused on the design. 

Thoughts?

podcasts/Ratings/StarRatingView.swift
leandroalonso: Fixed in https://github.com/Automattic/pocket-casts-ios/pull/1185

PR #1132, opened 2023-09-28
EffectsPlayer: change retry mechanism to stop on first failure

leandroalonso:
We've been receiving reports from users that their playback is resuming automatically after they put the AirPods on the case:

```
2023-09-28 09:30:02 EffectsPlayer: failed to start playback (1/3: Error Domain=com.apple.coreaudio.avfaudio Code=0 "(null)"
2023-09-28 09:30:02 pausing playback
2023-09-28 09:30:02 saving played up to 1787.235918367347 for episode Cormack McCarthy's Spring Break
2023-09-28 09:30:03 PlaybackManager handleAudioInterrupt began
2023-09-28 09:30:03 Loading Cormack McCarthy's Spring Break with UUID c7ed4e20-185f-486f-a47a-b5d705eaa453 autoPlay false
2023-09-28 09:30:03 saving played up to 1787.235918367347 for episode Cormack McCarthy's Spring Break
2023-09-28 09:30:03 cleanupCurrentPlayer permanent? false
2023-09-28 09:30:03 Using EffectsPlayer
2023-09-28 09:30:03 activating audio session succeeded
2023-09-28 09:30:08 Remote control: pauseCommand
```

This PR changes the retry mechanism to just pause and do nothing — fixing this error while fixing the recurring crash we used to see.

## Discussion

I'm targeting `trunk` with this change. We could create a new remote feature flag and we release that right away but I'm not sure if that's the correct approach.

Open to any ideas.

## To test

Comment those lines on `EFfectsPlayer` to force a crash:

```
strongSelf.engine?.prepare()
try strongSelf.engine?.start()
```

1. Run the app
2. Play a downloaded episode or local file
3. ✅ The app should not crash
4. Check the logs, you should see three `Loading <episode title>`, and a log like `EffectsPlayer: failed to start playback`

## Checklist


podcasts/EffectsPlayer.swift
```diff
@@ -192,27 +192,15 @@ class EffectsPlayer: PlaybackProtocol, Hashable {
 
     /// Try to play. If an exception happens, try again until `maxNumberOfRetries`
```
rviljoen: @leandroalonso Comment to be updated in line with new logic

emilylaguna (approved):
I gave this a test and was able to reproduce the original issue now by taking my airpods in and out. 

I confirmed that this doesn't trigger a crash, and doesn't start playing again. 

I think that in the long run this is a bandaid over a larger issue that the player has when it comes to the async play/load process. 

The issue that is happening with the AirPods is the `AVAudioSession.routeChangeNotification` is being fired multiple times as I take one AirPod out, then the other, and then put them into the case, and each time the player is either paused, played, and recreated.

So, we end up with multiple players trying to do an async play (load file, start engines) process, which at the end of we finally trigger the play on the internal `AVAudioPlayerNode`. 

If that wasn't enough, whenever a pause comes in the PlaybackManager calls  the EffectsPlayer calls `PlaybackManager.playerDidRequestTermination`, which then calls `cleanupCurrentPlayer`, which uses another async process to then deactivate the `AVAudioSession.sharedInstance`. 

This ends up looking like this https://gist.github.com/emilylaguna/40b82de07d6fab6e120c30f195d4ccb0 (I added a bunch of logs to monitor what was happening)

In the logs there's 3 different route changes, PlaybackManager.pause is called 4 different times, PlaybackManager.play is called twice, and PlaybackManager.deactiveAudioSession is called 7 times. 

That's a lot happening just because the AirPods are disconnecting. 

Ultimately I think the reason this exception gets triggered is because while all of this is happening when we actually try call `player.play` either the AVAudioSession is not active (Seen in logs), or the engines have been stopped right before we try to play.

A larger fix would be to add "is cancelled" checks during the load process to stop loading if the player no longer needs to play, additionally we should probably try and revamp our async processes to improve situations like these.

PR #1139, opened 2023-10-03
[Watch] Only fetch the episode UUIDs upfront when rendering the UpNext list, lazy load the episodes as needed

DanReyLop:
Follow-up to #950, see that PR for more backstory.

On the Up Next table in the Apple Watch, instead of fetching all the episode objects from the database at once, only the UUIDs are fetched (which is an order of magnitude faster), and each episode object is lazy-loaded as needed. Loading a handful of episode objects is still fast enough so that scrolling through that list is smooth.

I focused on touching existing code as little as possible, and keeping overall changes to a minimum, since none of this is covered by tests.

I'm not happy with the following:
- The `isStub` property in the `BaseEpisode` class is a hack. I would rather create a `StubEpisode` class that implements the `BaseEpisode` protocol, but that would require tons of boilerplate.
- As requested by the feedback [here](https://github.com/Automattic/pocket-casts-ios/pull/950#issuecomment-1636192131), I created new methods on the storage layer instead of adding optional parameters to the existing ones. But naming is hard. I followed the naming pattern `getSomeEpisodes() -> getSomeEpisodeUuids()`, but in fact those new methods return `BaseEpisode` "dehydrated/anemic" objects, *not* UUID strings. Naming is hard. I considered returning just UUID strings, but that would just make the changes on the `EpisodeViewModel` classes more complicated.

Suggestions are welcome. There are no perfect solutions here, I'm just hoping the trade-offs I made align with what you consider acceptable.

## To test

- On trunk, any of those operations would result in a ~1s freeze (that has already improved a lot due to my previous PRs). With this PR, it should be near-instant.

I haven't included any tracing because this change will be immediately noticeable on a real Apple Watch. When I measured in Instruments with my ~800 ep library, I could barely even find a CPU spike when I opened the Up Next view.

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
@@ -104,6 +104,30 @@ class EpisodeDataManager {
 
         return loadMultiple(query: "SELECT \(episodeTableName).* FROM \(upNextTableName) JOIN \(episodeTableName) ON \(episodeTableName).uuid = \(upNextTableName).episodeUuid ORDER BY \(upNextTableName).episodePosition ASC", values: nil, dbQueue: dbQueue)
     }
+    
+    func allUpNextEpisodeUuids(dbQueue: FMDatabaseQueue) -> [Episode] {
+        let upNextTableName = DataManager.playlistEpisodeTableName
+        let episodeTableName = DataManager.episodeTableName
+
+        var episodes = [Episode]()
+        dbQueue.inDatabase { db in
+            do {
+                let resultSet = try db.executeQuery("SELECT \(episodeTableName).uuid FROM \(upNextTableName) JOIN \(episodeTableName) ON \(episodeTableName).uuid = \(upNextTableName).episodeUuid ORDER BY \(upNextTableName).episodePosition ASC", values: nil)
+                defer { resultSet.close() }
+
+                while resultSet.next() {
+                    let episode = Episode()
+                    episode.uuid = DBUtils.nonNilStringFromColumn(resultSet: resultSet, columnName: "uuid")
+                    episode.isStub = true
+                    episodes.append(episode)
+                }
+            } catch {
+                FileLog.shared.addMessage("EpisodeDataManager.allUpNextEpisodeUuids Episode error: \(error)")
+            }
+        }
+
+        return episodes
+    }
 
     private func loadSingle(query: String, values: [Any]?, dbQueue: FMDatabaseQueue) -> Episode? {
         var episode: Episode?
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
@@ -104,30 +104,6 @@ class EpisodeDataManager {
 
         return loadMultiple(query: "SELECT \(episodeTableName).* FROM \(upNextTableName) JOIN \(episodeTableName) ON \(episodeTableName).uuid = \(upNextTableName).episodeUuid ORDER BY \(upNextTableName).episodePosition ASC", values: nil, dbQueue: dbQueue)
     }
-    
-    func allUpNextEpisodeUuids(dbQueue: FMDatabaseQueue) -> [Episode] {
-        let upNextTableName = DataManager.playlistEpisodeTableName
-        let episodeTableName = DataManager.episodeTableName
-
-        var episodes = [Episode]()
-        dbQueue.inDatabase { db in
-            do {
-                let resultSet = try db.executeQuery("SELECT \(episodeTableName).uuid FROM \(upNextTableName) JOIN \(episodeTableName) ON \(episodeTableName).uuid = \(upNextTableName).episodeUuid ORDER BY \(upNextTableName).episodePosition ASC", values: nil)
-                defer { resultSet.close() }
-
-                while resultSet.next() {
-                    let episode = Episode()
-                    episode.uuid = DBUtils.nonNilStringFromColumn(resultSet: resultSet, columnName: "uuid")
-                    episode.isStub = true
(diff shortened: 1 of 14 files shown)


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
```diff
@@ -104,6 +104,30 @@ class EpisodeDataManager {
...
+        var episodes = [Episode]()
+        dbQueue.inDatabase { db in
+            do {
+                let resultSet = try db.executeQuery("SELECT \(episodeTableName).uuid FROM \(upNextTableName) JOIN \(episodeTableName) ON \(episodeTableName).uuid = \(upNextTableName).episodeUuid ORDER BY \(upNextTableName).episodePosition ASC", values: nil)
+                defer { resultSet.close() }
+
+                while resultSet.next() {
+                    let episode = Episode()
+                    episode.uuid = DBUtils.nonNilStringFromColumn(resultSet: resultSet, columnName: "uuid")
+                    episode.isStub = true
```
leandroalonso: I was thinking that maybe this could be a simple computed property on `BaseEpisode` like:

```swift
extension BaseEpisode {
    var isStub: Bool {
        title != nil
    }
}
```

However what properties need to be defined versus what is an Optional and can stay that way? So better to leave it as it is.

I'd change this name to something like: `isFilled` or `hasOnlyUUID` — which conveys better what this field is representing. Stub always makes me think about tests. :P

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
```diff
@@ -104,6 +104,30 @@ class EpisodeDataManager {
 
         return loadMultiple(query: "SELECT \(episodeTableName).* FROM \(upNextTableName) JOIN \(episodeTableName) ON \(episodeTableName).uuid = \(upNextTableName).episodeUuid ORDER BY \(upNextTableName).episodePosition ASC", values: nil, dbQueue: dbQueue)
     }
+    
+    func allUpNextEpisodeUuids(dbQueue: FMDatabaseQueue) -> [Episode] {
```
leandroalonso: Where is this used?

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/UserEpisodeDataManager.swift
```diff
@@ -108,6 +108,30 @@ class UserEpisodeDataManager {
         return loadMultiple(query: "SELECT \(userEpisodeTableName).* FROM \(upNextTableName) JOIN \(userEpisodeTableName) ON \(userEpisodeTableName).uuid = \(upNextTableName).episodeUuid ORDER BY \(upNextTableName).episodePosition ASC", values: nil, dbQueue: dbQueue)
     }
 
+    func allUpNextEpisodeUuids(dbQueue: FMDatabaseQueue) -> [UserEpisode] {
```
leandroalonso: Same as the other method, this doesn't seem to be used anywhere.

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
```diff
@@ -98,6 +98,15 @@ public class DataManager {
         return convertedEpisodes
     }
 
+    public func allUpNextEpisodeUuids() -> [BaseEpisode] {
+        upNextManager.allUpNextPlaylistEpisodes(dbQueue: dbQueue).map{
```
leandroalonso: I thought here we would use the new UUID methods you added on the DataManager?

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/BaseEpisode.swift
```diff
@@ -57,4 +57,5 @@ import Foundation
     func mayContainChapters() -> Bool
 
     var hasBookmarks: Bool { get }
+    var isStub: Bool { get set }
```
leandroalonso: It would be good to add a comment here:

```suggestion
    /// This property is true when...
    var isStub: Bool { get set }
```

leandroalonso (changes requested):
@DanReyLop I tested and it works as described however I'm a bit puzzled by why some of the newly implemented methods are not being used. 🤔 

Also, there are a few lint warnings to fix:

Screenshot 2023-10-04 at 13 24 29

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
DanReyLop: > However what properties need to be defined versus what is an Optional and can stay that way? So better to leave it as it is.

Yeah, that was my thinking too, I can't 100% say for sure if an Episode object is "filled" by just looking at its properties.

> I'd change this name to something like: isFilled or hasOnlyUUID — which conveys better what this field is representing. Stub always makes me think about tests. :P

Sounds good, I'll change it to `hasOnlyUuid`.

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
DanReyLop: My bad. Turns out, we don't need them! I realized it at the last minute and I forgot to delete the new methods. The list of UUIDs can be fetched by just calling the `upNextManager` table, no need to fetch from the `episodes` and `user_episodes` tables. The only "problem" is that all the "anemic" objects will be of type `Episode`, even if they really are representing UserEpisodes. That isn't an issue with the current implementation of `EpisodeViewModel` though.

DanReyLop:
I've fixed the lint warnings, removed the unused functions, and changed the `isStub` name. This PR is surprisingly compact now :)

leandroalonso (approved):
> This PR is surprisingly compact now :)

Amazing! Thanks for your patience on the many back and forth on this @DanReyLop and, of course, for your contribution! :shipit:

Issue #1142, opened 2023-10-04
Pocket Casts Playback: 2023


leandroalonso:
## Stories design

- [x] Intro Story #1145
- [x] Total episodes / shows #1150
- [x] Most listened show #1155
- [x] Top 5 shows #1157 
- [x] Top categories #1161 
- [x] Total time listening to podcasts #1163 
- [x] Longest episode #1164
- [x] Prologue #1166 

## Plus

- [x] Listening hours comparison story #1174
- [x] Completion rate story #1175 
- [x] Add a mechanism to skip "Plus stories" #1178
- [x] Plus Upsell story #1189
- [x] Plus purchase flow #1189
- [x] Add logic to sync 2022 data #1191

## Entry point

- [x] Update the modal entry point #1169
- [x] Update the Profile entry point #1169

## Tracks

- [x] Double check Tracks #1196
- [x] Add tracks for Plus #1196

## Refinements

- [x] Calculate top podcasts based on listening time AND number of episodes #1192
- [x] Gradient should be relative to the screen height and plus/normal gradient can be combined into one #1192
- [x] In small devices the font is too small #1192
- [x] Update the Pocket Casts watermark when exporting #1192
- [x] Fix the logo snapping issue on the IntroStory when the motion kicks in

## Animations

- [x] Add animation to the listened time story #1203
- [x] Add animation to the top five story to top 1 transition #1203
- [x] Add animation to the longest episode story #1203

## Additional

- [x] Update the share copies #1192
- [x] The old login flow is displayed, it should be replaced by the new one #1189
- [x] On the first story, add the numbers separately and add different parallax effects in each of them #1192
- [x] Prompt user to share when they take a screenshot #1193
- [x] Look at if we should show the YOY if 2022 is zero, [ref](https://github.com/Automattic/pocket-casts-ios/pull/1191#issuecomment-1801905103)

PR #1150, opened 2023-10-06
EOY 2023: listened number story

leandroalonso:
|:---:|:---:|

Adds the 2023 visual for the Listened Numbers Story and updates the order so it's the first story (after the intro).

**Animations are not part of this task and will be tackled later.**

## DM Sans line spacing

We have an issue with the line spacing of DM Sans. Is just too big and SwiftUI doesn't allow setting a negative value. I'll leave as it is for now.

## To test

1. Make sure you're logged in into an account that has a few episodes listened
3. Go back to Profile
4. Tap the EOY image
5. Skip to the second story
5. ✅ Check that the Listened Numbers story looks good
6. Tap "Share this story" > Save Image
7. Go to Photos
8. ✅ The story image asset should be the last one on Photos

Please test on a big device (iPhone 15 Pro Max, for example), normal device (iPhone 15), and a smaller device (iPhone SE).

## Checklist


podcasts/End of Year/Stories/Views/PodcastCover.swift
```diff
@@ -49,16 +49,28 @@ struct PodcastCover: View {
...
 
-            Action {
-                image = nil
-                KingfisherManager.shared.retrieveImage(with: ServerHelper.imageUrl(podcastUuid: podcastUuid, size: 280)) { result in
-                    switch result {
-                    case .success(let result):
-                        image = result.image
-                    default:
-                        break
+                .onAppear {
```
leandroalonso, line 52: We need to use `onAppear` so images are loaded when we render the view to snapshot. 🤷

podcasts/End of Year/Stories/Views/CommonStoryViews.swift
```diff
@@ -69,18 +73,37 @@ struct StoryLabel: View {
...
+            .custom("DM Sans", size: size).weight(.bold)
         case .pillarSubtitle:
-            return .system(size: 13, weight: .regular)
+            .custom("DM Sans", size: size).weight(.regular)
         }
     }
 
+    private var size: CGFloat {
+        let iPhone15DefaultHeight: CGFloat = 759
+        let screenHeight = geometry?.size.height ?? iPhone15DefaultHeight
```
emilylaguna, line 90: What's the reason for this?

emilylaguna (approved):
Nice work, I tested on iPhone 14, iPod Touch, iPad, and the iPhone 15 all look good. 

Left a comment asking a question.

podcasts/End of Year/Stories/Views/CommonStoryViews.swift
leandroalonso: In case the geometry size is `nil`, it uses the iPhone 15 as a base to calculate the values. I see how that's confusing... but after changing all the stories I can make the  `geometry` parameter not Optional and get rid of that.

PR #1160, opened 2023-10-11
Bookmarks: Adds early access what's new logic

emilylaguna:
| 📘 Part of: #1061 |
|:---:|

This adds the logic the display the what's new while the feature is in early access:
- If we're in beta we'll show a "join beta testing" message for Plus and Patron
- If we're in production we'll show the regular announcement for Patron

This also adds a flag that's stored in the user defaults to prevent showing the full announcement to people who already saw it while it was in early access.

**Note:** This doesn't add the design or real text yet. 

## To test

**Before running do the following:**
   1. Enable the Bookmarks feature flag
   2. Open Announcements.swift, and change the bookmarks announcement versions from `99.99` to `7.50`

The easiest way to test is by changing the values in the WhatsNew initializer so the WhatsNew will always be displayed:

```     
self.announcements = announcements
self.previousOpenedVersion = "7.49"
self.currentVersion = "7.50"
self.lastWhatsNewShown = "7.43"
```

### Early Access: What's New is shown for Patron when not in beta:
1. Launch the app
2. ✅ You should not see the What's New
3. Sign into an account with Patron, or use the developer menu Paid Patron option
4. Restart the app
5. ✅ Verify you see the What's New with the "Normal Title"

### Early Access: What's New is shown for Plus and above when in beta:
1. In Xcode, open BuildEnvironment.swift, change line 24 to `.testFlight`
2. Sign into free account
3. Restart the app
4. ✅ Verify you do not see the what's new
5. Sign into a Plus account
6. Restart the app
7. ✅ Verify you see the What's New with the "Beta Title"
8. Sign into a Patron account
9. Restart the app
10. ✅ Verify you see the What's New with the "Beta Title"

### Full Release: Don't show the announcement to people who saw it already
1. In Xcode, open PaidFeature, and change line 9 from `.inEarlyAccess` to `.plusFeature`
2. Restart the app
3. ✅ You should not see the what's new because the flag was set to true in the first testing step
4. In Xcode, run `UserDefaults.standard.setValue(nil, forKey: "WhatsNew.Bookmarks.EarlyAccess.Seen")`
5. Restart the app
6. ✅ You should now see the what's new with the "Release Title"

## Checklist

### DIFF
--- a/podcasts/Whats New/Announcements.swift
+++ b/podcasts/Whats New/Announcements.swift
@@ -53,7 +53,7 @@ struct Announcements {
 
         // TODO: Update with the real content
         let announcement = WhatsNew.Announcement(
-            version: "7.50",
+            version: "99.99",
             header: {
                 AnyView(AutoplayWhatsNewHeader().onAppear {
                     // Record when someone sees the full announcement while in early access so we don't show it again
--- a/podcasts/Whats New/Announcements.swift
+++ b/podcasts/Whats New/Announcements.swift
@@ -71,6 +71,34 @@ struct Announcements {
 
         announcements.append(announcement)
     }
+
+    /// When bookmarks is out of early access we'll show the announcement to everyone
+    /// but don't show it to people who already saw the full announcement when it was in early access
+    /// This is a placeholder for now.
+    /**
+    mutating func addBookmarksReleaseAnnouncement() {
+        guard
+            FeatureFlag.bookmarks.enabled, !PaidFeature.bookmarks.inEarlyAccess,
+            !UserDefaults.standard.bool(forKey: "WhatsNew.Bookmarks.EarlyAccess.Seen")
+        else {
+            return
+        }
+
+        // TODO: Update with the real content
+        let announcement = WhatsNew.Announcement(
+            version: "99.99",
+            header: {
+                AnyView(AutoplayWhatsNewHeader())
+            },
+            title: "Normal Title",
+            message: "Bookmarks Message",
+            buttonTitle: "Action Button",
+            action: { }
+        )
+
+        announcements.append(announcement)
+    }
+     */
 }
 
 class AnnouncementFlow {
--- a/podcasts/Whats New/Announcements.swift
+++ b/podcasts/Whats New/Announcements.swift
@@ -1,5 +1,7 @@
 import Foundation
 import SwiftUI
+import PocketCastsServer
+import PocketCastsUtils
 
 struct Announcements {
     // Order is important.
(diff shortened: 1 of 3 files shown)


podcasts/Whats New/Announcements.swift
```diff
@@ -21,6 +23,82 @@ struct Announcements {
...
+        let activeTier = SubscriptionHelper.activeTier
+        let environment = BuildEnvironment.current
+        let isBeta = environment == .testFlight
+
+        // Check to determine if we are showing the announcement
+        let shouldShowInBeta = isBeta && activeTier > .none
+        let shouldShowInRelease = !isBeta && activeTier == .patron
+
+        // Don't add the announcement if we're not showing it
+        guard shouldShowInBeta || shouldShowInRelease else { return }
```
leandroalonso: What do you think about changing `WhatsNew.Announcement` to have a property that is a `() -> Bool` that returns an additional condition whether the announcement should show or not?

I believe this might be super useful going forward and this way we tie the logic to the `Announcement` itself.

When checking to present, the version should match AND this block should return `true`. I also believe this could potentially help us in avoiding running these conditions if an announcement was already present.

podcasts/Whats New/Announcements.swift
```diff
@@ -21,6 +23,82 @@ struct Announcements {
...
+
+        // TODO: Update with the real content
+        let announcement = WhatsNew.Announcement(
+            version: "99.99",
+            header: {
+                AnyView(AutoplayWhatsNewHeader().onAppear {
+                    // Record when someone sees the full announcement while in early access so we don't show it again
+                    // to them when we move to full release.
+                    if shouldShowInRelease {
+                        UserDefaults.standard.setValue(true, forKey: "WhatsNew.Bookmarks.EarlyAccess.Seen")
```
leandroalonso: Do we really need this?

If we tie the release announcement to the same version as this one, it won't be displayed if it is already shown.

emilylaguna: To clarify, when you're suggesting tieing the release announcement to the same version you mean, when we release in early access we'll use (for example) 7.50 as the announcement version, and in the next version (7.51) when its moved into full release the announcement will still be tied to 7.50? 

If so, the main issue I see with that is people who saw the beta announcement will not see the full release announcement because its tied to the same version.

podcasts/Whats New/Announcements.swift
emilylaguna: I originally went this direction by adding an `isEnabled` flag to the announcement. I moved away from that because it meant that I would need to duplicate the announcement content for the different versions (one for the beta, one for the app store). 

I think this ends up being cleaner though so I'll switch back to that.

emilylaguna: @leandroalonso The code has been updated. I also did a bit of a refactor of how we get the visible announcement and added some more tests.

podcasts/Whats New/Announcements.swift
leandroalonso: > If so, the main issue I see with that is people who saw the beta announcement will not see the full release announcement because its tied to the same version.

Oh, got it! I assumed we didn't want that. Makes sense!

emilylaguna:
@leandroalonso ready for another look. I also updated the testing steps to be a bit simpler.

podcasts/Whats New/Announcements.swift
```diff
@@ -19,6 +22,45 @@ struct Announcements {
...
+        // Bookmarks Early Access: Beta
+        // Show only in TestFlight, for Plus and Patron
+        .init(
+            version: "99.99",
+            header: { AnyView(EmptyView()) },
+            title: "Early Access Beta Title",
+            message: "Message",
+            buttonTitle: "Button",
+            action: {},
+            isEnabled: PaidFeature.bookmarks.inEarlyAccess && BuildEnvironment.current == .testFlight && SubscriptionHelper.activeTier > .none
```
leandroalonso, line 35: The `isEnabled` is a `Bool` that is calculated when the `Announcement` is created. As it is right now, I think this is pretty quick and straightforward. However, I'm thinking about the future: what if we stack many announcements with special conditions? That would mean that this logic will always run when the app starts and I wonder if we want that?

If we make that an `@autoclosure` we can eliminate this problem.

No need to do in this PR right now but something to keep in mind.

podcasts/Whats New/WhatsNew.swift
```diff
@@ -38,6 +36,26 @@ class WhatsNew {
...
+            return nil
+        }
+
+        // Find the last announcement that:
+        // - is enabled
+        // - has not been shown already
+        // - the target version is not before the last opened version, and not for a future version
+        return announcements
+            .last(where: {
+                $0.isEnabled &&
```
leandroalonso: If we change the `isEnabled` to be an `@autoclosure` we can move this to be the last check. This would mean that the closure would only run if the announcement could be displayed, so we wouldn't need to check it for all announcements.

leandroalonso (approved):

podcasts/Whats New/Announcements.swift
emilylaguna: I can see that, I updated to use an autoclosure. Also since I was defining it, I also updated the header view to use it as well. 

6d017035c13f5c3214c2a80f7f59608f51e97eef

leandroalonso: > I also updated the header view to use it as well.

Nice!

leandroalonso:
This is good to go! :shipit:

PR #1174, opened 2023-10-18
Playback 2023: year over year listening time story

leandroalonso:
|:---:|:---:|

Adds the 2023 visual for the Year Over Year Listening Time Story.

**This doesn't tackle syncing 2022 listening history.**

## Different states

1. Go to `YearOverYearstory.swift`
2. Preview the view
3. ✅ Go over all the cases and check that their match the designs
4. You can also play with the numbers to see if the bars/title updates accordingly

## Story appearing

1. Make sure you're logged in to an account that has a few episodes listened
3. Go back to Profile
4. Tap the EOY image
5. Skip to the penultimate story
5. ✅ Check that the story looks good compared to the design
6. Tap "Share this story" > Save Image
8. Go to Photos
9. ✅ The story image asset should be the last one in Photos
10. Give yourself Plus (Settings > Developer > Set to Paid Plus)
11. ✅ Check the story again, it should show a Plus badge
10. Give yourself Patron (Settings > Developer > Set to Paid Patron)
11. ✅ Check the story again, it should show a Patron badge

Please test on a big device (iPhone 15 Pro Max, for example), a normal device (iPhone 15), and a smaller device (iPod touch).

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -16,6 +16,10 @@ class EndOfYearDataManager {
                                             lastPlaybackInteractionDate IS NOT NULL AND lastPlaybackInteractionDate BETWEEN strftime('%s', '\(startDate)') and strftime('%s', '\(endDate)')
                                            """
 
+    private lazy var listenedEpisodesPreviousYear = """
+                                            lastPlaybackInteractionDate IS NOT NULL AND lastPlaybackInteractionDate BETWEEN strftime('%s', '2021-01-01') and strftime('%s', '"2023-01-01"')
+                                           """
+
     /// If the user is eligible to see End of Year stats
     ///
     /// All it's needed is a single episode listened for more than 5 minutes.
@@ -318,6 +322,32 @@ class EndOfYearDataManager {
 
         return numberOfItemsInListeningHistory
     }
+
+    /// Returns the approximate listening time for the current year
+    func yearOverYearListeningTime(dbQueue: FMDatabaseQueue) -> YearOverYearListeningTime {
+        var listeningTimeThisYear: Double = 0
+        var listeningTimePreviousYear: Double = 0
+
+        dbQueue.inDatabase { db in
+            do {
+                let query = "SELECT DISTINCT \(DataManager.episodeTableName).uuid, SUM(playedUpTo) as totalPlayedTime from \(DataManager.episodeTableName) WHERE \(listenedEpisodesThisYear) UNION SELECT DISTINCT \(DataManager.episodeTableName).uuid, SUM(playedUpTo) as totalPlayedTime from \(DataManager.episodeTableName) WHERE \(listenedEpisodesPreviousYear)"
+                let resultSet = try db.executeQuery(query, values: nil)
+                defer { resultSet.close() }
+
+                if resultSet.next() {
+                    listeningTimeThisYear = resultSet.double(forColumn: "totalPlayedTime")
+                }
+
+                if resultSet.next() {
+                    listeningTimePreviousYear = resultSet.double(forColumn: "totalPlayedTime")
+                }
+            } catch {
+                FileLog.shared.addMessage("EndOfYearDataManager.listeningTime error: \(error)")
+            }
+        }
+
+        return YearOverYearListeningTime(totalPlayedTimeThisYear: listeningTimeThisYear, totalPlayedTimeLastYear: listeningTimePreviousYear)
+    }
 }
 
 public struct ListenedCategory {
@@ -367,7 +397,7 @@ public struct YearOverYearListeningTime {
     public let totalPlayedTimeLastYear: Double
 
     public var percentage: Double {
-        let nonRoundendPercentage = ((100 * totalPlayedTimeThisYear) / totalPlayedTimeLastYear)
+        let nonRoundendPercentage = ((100 * totalPlayedTimeThisYear) / totalPlayedTimeLastYear) - 100
         return (nonRoundendPercentage.rounded() * 100) / 100
     }
 
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -395,15 +395,12 @@ public struct ListenedNumbers {
 public struct YearOverYearListeningTime {
     public let totalPlayedTimeThisYear: Double
     public let totalPlayedTimeLastYear: Double
-
-    public var percentage: Double {
-        let nonRoundendPercentage = ((100 * totalPlayedTimeThisYear) / totalPlayedTimeLastYear) - 100
-        return (nonRoundendPercentage.rounded() * 100) / 100
-    }
+    public let percentage: Double
 
     public init(totalPlayedTimeThisYear: Double, totalPlayedTimeLastYear: Double) {
         self.totalPlayedTimeThisYear = totalPlayedTimeThisYear
         self.totalPlayedTimeLastYear = totalPlayedTimeLastYear
+        self.percentage = (((totalPlayedTimeThisYear - totalPlayedTimeLastYear) / totalPlayedTimeLastYear) * 100).rounded()
     }
 }
 
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
@@ -362,6 +362,21 @@ public struct ListenedNumbers {
     }
 }
 
+public struct YearOverYearListeningTime {
+    public let totalPlayedTimeThisYear: Double
+    public let totalPlayedTimeLastYear: Double
+
+    public var percentage: Double {
+        let nonRoundendPercentage = ((100 * totalPlayedTimeThisYear) / totalPlayedTimeLastYear)
+        return (nonRoundendPercentage.rounded() * 100) / 100
+    }
+
+    public init(totalPlayedTimeThisYear: Double, totalPlayedTimeLastYear: Double) {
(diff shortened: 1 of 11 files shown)


podcasts/End of Year/Stories/YearOverYearStory.swift
```diff
@@ -0,0 +1,211 @@
+import SwiftUI
+import PocketCastsServer
+import PocketCastsDataModel
+
+struct YearOverYearStory: ShareableStory {
+    var duration: TimeInterval = 5.seconds
+
+    let identifier: String = "year_over_year"
```
leandroalonso, line 8: @ashiagr I defined the identifier of this story as `year_over_year`. Open to suggestions if you have any other ideas. :)

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -16,6 +16,10 @@ class EndOfYearDataManager {
                                             lastPlaybackInteractionDate IS NOT NULL AND lastPlaybackInteractionDate BETWEEN strftime('%s', '\(startDate)') and strftime('%s', '\(endDate)')
                                            """
 
+    private lazy var listenedEpisodesPreviousYear = """
+                                            lastPlaybackInteractionDate IS NOT NULL AND lastPlaybackInteractionDate BETWEEN strftime('%s', '2021-01-01') and strftime('%s', '"2023-01-01"')
```
emilylaguna: Should this be between 2022-01-01 and 2023-01-01?

emilylaguna: My stats were showing that I didn't listen to anything. 

```suggestion
                                            lastPlaybackInteractionDate IS NOT NULL AND lastPlaybackInteractionDate BETWEEN strftime('%s', '2021-01-01') and strftime('%s', '2023-01-01')
```

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -318,6 +322,32 @@ class EndOfYearDataManager {
...
     }
+
+    /// Returns the approximate listening time for the current year
+    func yearOverYearListeningTime(dbQueue: FMDatabaseQueue) -> YearOverYearListeningTime {
+        var listeningTimeThisYear: Double = 0
+        var listeningTimePreviousYear: Double = 0
+
+        dbQueue.inDatabase { db in
+            do {
+                let query = "SELECT DISTINCT \(DataManager.episodeTableName).uuid, SUM(playedUpTo) as totalPlayedTime from \(DataManager.episodeTableName) WHERE \(listenedEpisodesThisYear) UNION SELECT DISTINCT \(DataManager.episodeTableName).uuid, SUM(playedUpTo) as totalPlayedTime from \(DataManager.episodeTableName) WHERE \(listenedEpisodesPreviousYear)"
```
emilylaguna: If the user doesn't have any listening history for the previous year, then the result will be:

Note: I modified the query to return the label to show the order better.

Screenshot 2023-10-31 at 2 46 14 PM

When using `UNION` the order of the queries is not preserved, so it's not guaranteed the first item will always be 2023, and the second 2022. 

Using to using `UNION ALL` will preserve the order regardless of the values:


Screenshot 2023-10-31 at 2 56 58 PM


---

```suggestion
                let query = "SELECT DISTINCT \(DataManager.episodeTableName).uuid, SUM(playedUpTo) as totalPlayedTime from \(DataManager.episodeTableName) WHERE \(listenedEpisodesThisYear) UNION ALL SELECT DISTINCT \(DataManager.episodeTableName).uuid, SUM(playedUpTo) as totalPlayedTime from \(DataManager.episodeTableName) WHERE \(listenedEpisodesPreviousYear)"
```

podcasts/End of Year/EndOfYearStoriesBuilder.swift
```diff
@@ -95,6 +96,14 @@ class EndOfYearStoriesBuilder {
                 stories.append(.longestEpisode)
             }
 
+            // Year over year listening time
+            let yearOverYearListeningTime = dataManager.yearOverYearListeningTime()
+            if yearOverYearListeningTime.totalPlayedTimeThisYear != 0 ||
+                yearOverYearListeningTime.totalPlayedTimeLastYear != 0 {
```
emilylaguna, line 102: We can't compare if we don't have history for both years, so we should hide the story. 

```suggestion
            if yearOverYearListeningTime.totalPlayedTimeThisYear != 0 &&
                yearOverYearListeningTime.totalPlayedTimeLastYear != 0 {
```

podcasts/End of Year/Stories/YearOverYearStory.swift
```diff
@@ -0,0 +1,211 @@
...
+
+    let yearOverYearListeningTime: YearOverYearListeningTime
```
emilylaguna: Can we rename this to just `data` since it's already relative to the story which is named year over year?


```suggestion
    let data: YearOverYearListeningTime
```

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -362,6 +392,21 @@ public struct ListenedNumbers {
     }
 }
 
+public struct YearOverYearListeningTime {
+    public let totalPlayedTimeThisYear: Double
+    public let totalPlayedTimeLastYear: Double
+
+    public var percentage: Double {
+        let nonRoundendPercentage = ((100 * totalPlayedTimeThisYear) / totalPlayedTimeLastYear) - 100
+        return (nonRoundendPercentage.rounded() * 100) / 100
```
emilylaguna: This looks like we're calculating the percent change between this and last years listening number. 

Wouldn't it be the same to do:

```suggestion
        (((totalPlayedTimeThisYear - totalPlayedTimeLastYear) / totalPlayedTimeLastYear) * 100).rounded()
```

emilylaguna: Also since we're using this in 4 different places in the view, and the values aren't going to change could this to be calculated in init and stored?

podcasts/End of Year/Stories/YearOverYearStory.swift
```diff
@@ -0,0 +1,211 @@
...
+        switch listeningPercentage {
+        case _ where listeningPercentage == .infinity:
+            return L10n.eoyYearOverYearTitleSkyrocketed
+        case _ where listeningPercentage > 10:
+            return L10n.eoyYearOverYearTitleWentUp("\(listeningPercentage.clean)%")
+        case _ where listeningPercentage < 0:
+            return L10n.eoyYearOverYearTitleWentDown
+        default:
+            return L10n.eoyYearOverYearTitleFlat
+        }
```
emilylaguna: We can achieve the same effect without the `where` by doing:

```suggestion
        switch yearOverYearListeningTime.percentage {
        case .infinity:
            L10n.eoyYearOverYearTitleSkyrocketed
        case 10...:
            L10n.eoyYearOverYearTitleWentUp("\(yearOverYearListeningTime.percentage.clean)%")
        case ...0:
            L10n.eoyYearOverYearTitleWentDown
        default:
            L10n.eoyYearOverYearTitleFlat
        }
```

podcasts/End of Year/Stories/YearOverYearStory.swift
```diff
@@ -0,0 +1,211 @@
...
+    var subtitle: String {
+        let listeningPercentage = yearOverYearListeningTime.percentage
+        switch listeningPercentage {
+        case _ where listeningPercentage > 10:
+            return L10n.eoyYearOverYearSubtitleWentUp
+        case _ where listeningPercentage < 0:
+            return L10n.eoyYearOverYearSubtitleWentDown
+        default:
+            return L10n.eoyYearOverYearSubtitleFlat
+        }
```
emilylaguna: We can achieve the same result a bit cleaner without the `where`:
 
```suggestion
        switch yearOverYearListeningTime.percentage {
        case 10...:
            L10n.eoyYearOverYearSubtitleWentUp
        case ...0:
            L10n.eoyYearOverYearSubtitleWentDown
        default:
            L10n.eoyYearOverYearSubtitleFlat
        }
```

emilylaguna (changes requested):
Nice work, left some comments to improve the code below.

podcasts/End of Year/EndOfYearStoriesBuilder.swift
leandroalonso: We can't compare but we'll show it, we'll display the previous year as "0 minutes listened". ([discussed here](https://a8c.slack.com/archives/C041BGTFQ5R/p1697573573229879))

Plus this makes things a bit easier for upcoming PRs, given for non-Plus users we can't really guarantee they had or not listening time in 2022.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
leandroalonso: I fixed that on an upcoming PR but did it here: b8bc790f66077a82855fc126ae594b6601af8768

leandroalonso: Fixed in b8bc790f66077a82855fc126ae594b6601af8768

podcasts/End of Year/Stories/YearOverYearStory.swift
leandroalonso: Done in 4b40128a13db3db5b98e0a3b35eb1f9444c7ffc6

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
leandroalonso: Done in b344100174bf440aeb2f3e6d27192b0574a44e84

emilylaguna (approved):

PR #1178, opened 2023-10-20
Playback 2023: add a mechanism to skip "Plus stories" in sequence

leandroalonso:
| 📘 Part of: #1142 |
|:---:|

This PR adds a mechanism to:

1. Set which stories are Plus or not;
2. Skip a bunch of Plus stories in sequence for a non-paid user;
3. Skip a bunch of Plus stories in sequence for a non-paid user when going to previous stories

Basically — given the "Plus stories" will have an upsell, we don't want users to see the same story multiple times. So we just display it once, whether is going to next stories or previous one.

## To test

Before testing, Add `let plusOnly = true` in `CompletionRateStory`, `LongestEpisodeStory` and `YearOverYearStory`.

Then:

1. Run the app
2. Go to Profile > Settings > Developer > Set to No Plus
3. Let the stories run (don't tap to skip to the next or to the previous one)
4. ✅ After the "Longest Episode" story is show, it should skip to the last one
5. Tap to go to the previous story
6. ✅ It should go back to the "Longest Episode" (skipping completion rate and year over year)
7. On the "Longest Episode", tap to go next
8. ✅ It should go to the last story
9. Exit the stories
10. Go to Profile > Settings > Developer > Set to Paid Plus
11. Repeat the testing steps
12. ✅ However, all stories should show (no stories will be skipped)
13. Go to Profile > Settings > Developer > Set to Paid Patron
14. ✅ All stories should show (no stories will be skipped)

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/End of Year/StoriesModelTests.swift
+++ b/PocketCastsTests/Tests/End of Year/StoriesModelTests.swift
@@ -65,6 +65,80 @@ class StoriesModelTests: XCTestCase {
 
         XCTAssertEqual(dataSource.didCallStoryForWithStoryNumber, 0)
     }
+
+    // MARK: - Plus stories
+
+    /// If a user has a non-paid account we want to show only the first Plus story
+    /// and skip all the following ones to the next non-Plus story.
+    /// This way they don't see the same upsell multiple times.
+    func testSkipToFirstNonPlusStoryIfFreeUser() {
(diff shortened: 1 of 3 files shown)


emilylaguna (approved):
We discussed some improvements we can do with using the PaidFeature that will be done in a future PR. Approving for now.

PR #1137, opened 2023-10-01
PCiOS: Update Fastfile with version model

spencertransier:
## Summary
- This PR integrates the new version and build code models from https://github.com/wordpress-mobile/release-toolkit/pull/512
- Many of the changes here are from breaking out the code within the various Release Toolkit `*_prechecks` actions into the Fastfile itself. Nearly all of the `prechecks` actions were tightly coupled with the existing version methods. To simplify things with the new version model, I took the relevant snippets from the `prechecks` actions and added them directly to the Fastfile. We'll be evaluating those and adding better-named and succinct shared actions to the Release Toolkit in the future. 
- "Release version" refers to the `VERSION_SHORT` in the xcconfig file: either the `1.2` or `1.2.3` (hotfix) version
-  "Build code" is what we're calling the incremented number for each build (in PCiOS's case, the format of `1.2.3.4`, `VERISON_LONG` in the xcconfig file.). This is because the `versionCode` in Android apps or the `build_number` used by other A8C apps are formatted differently. So "build code" is the overarching term for all of those. Just wanted to clarify that vocabulary.

PR #1179, opened 2023-10-23
Remove unneeded env vars from Fastfile

spencertransier:
## Summary of changes:

This PR removes several unneeded environment variables:
- `PROJECT_ROOT_FOLDER` - The only action used by PCiOS that uses this one is[`ios_update_release_notes`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_update_release_notes.rb) Release Toolkit action. PCiOS already sets an explicit path for that action, so the env var isn't needed.
- `PROJECT_NAME` - This env var is used by Android apps in a path like `PROJECT_ROOT_FOLDER/PROJECT_NAME/build.gradle`, but it's not currently used by any iOS actions. I don't know if that was different in the past, but it's not needed at all now. 
- `PUBLIC_CONFIG_FILE` - This is used by [`ios_get_build_version.rb`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_get_build_version.rb), and the `update_xc_configs` method in [`ios_version_helper`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_version_helper.rb) but none of those actions are used by PCiOS anymore after the recent versioning changes: https://github.com/Automattic/pocket-casts-ios/pull/1137
- `GHHELPER_REPO` - This env var isn't used by any actions in the Release Toolkit or anywhere else by PCiOS.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -121,7 +121,8 @@ before_all do
 end
 
 platform :ios do
-  # This explicit timeout helps resolve Xcode build errors
+  # This explicit timeout is necessary for the `xcodebuild -showBuildSettings` call to succeed in CI.
+  # See https://buildkite.com/automattic/pocket-casts-ios/builds/4875#018b7eea-e8d7-40a1-8ce5-515dd8feafce/666-805
   ENV['FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT'] = '120'
 
   desc 'Run the unit tests'
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -121,12 +121,7 @@ before_all do
 end
 
 platform :ios do
-  # Unfortunately, release toolkit still relies on certain settings being set in the environment
-  ENV['PROJECT_NAME'] = 'podcasts'
-  ENV['PROJECT_ROOT_FOLDER'] = "#{PROJECT_ROOT_FOLDER}/"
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -121,12 +121,7 @@ before_all do
 end
 
 platform :ios do
-  # Unfortunately, release toolkit still relies on certain settings being set in the environment
-  ENV['PROJECT_NAME'] = 'podcasts'
-  ENV['PROJECT_ROOT_FOLDER'] = "#{PROJECT_ROOT_FOLDER}/"
-  ENV['PUBLIC_CONFIG_FILE'] = VERSION_XCCONFIG_PATH
-  ENV['GHHELPER_REPO'] = GHHELPER_REPO
-
+  # This explicit timeout helps resolve Xcode build errors
```
mokagio: I was curious about this and tested it https://github.com/Automattic/pocket-casts-ios/pull/1187. Got some insights in what it's used for:

image

```suggestion
  # This explicit timeout is necessary for the `xcodebuild -showBuildSettings` call to succeed in CI.
  # See https://buildkite.com/automattic/pocket-casts-ios/builds/4875#018b7eea-e8d7-40a1-8ce5-515dd8feafce/666-805
```

mokagio (approved):

PR #1186, opened 2023-10-27
Use DefaultPlayer instead of EffectsPlayer for downloaded files when trim silence is disabled

### LINKED ISSUE #1184: EffectsPlayer always used for downloaded files, even when no effects are enabled
### Description

I have noticed in my device logs that `EffectsPlayer` is always used, even when I have no effects enabled. I have traced this down to the fact that I always download episodes. `DefaultPlayer` is capable of playing downloaded files, speed adjustments and Volume Boost, but `EffectsPlayer` is needed for trim silence. Looking at the logic, when trim silence is enabled, a download is initiated:

https://github.com/Automattic/pocket-casts-ios/blob/da0e130d8b11a77a3002ea69879fcb2ac9d7f069/podcasts/PlaybackManager.swift#L1836-L1841

When building the list of possible players to use, we always add `EffectsPlayer` when a file has been downloaded (in my case by the user and not because trim silence is enabled):

https://github.com/Automattic/pocket-casts-ios/blob/da0e130d8b11a77a3002ea69879fcb2ac9d7f069/podcasts/PlaybackManager.swift#L1086-L1091 

I've been running a test build for the last few weeks which adds a check for trim silence, in addition to the download check:

`(currEpisode.downloaded(pathFinder: DownloadManager.shared) && effects().trimSilence != .off)`

`DefaultPlayer` is then used for all scenarios, except where trim silence is enabled. Toggling trim silence also correctly switches between the correct player.

In my limited testing, `DefaultPlayer` uses about 50% less resources than `EffectsPlayer`, so this could be a big help with battery consumption. I'm not sure what percentage of users use trim silence, but my guess is that many people download episodes, which would then switch them to `EffectsPlayer` even if trim silence is not enabled.

I'm keen to hear from @emilylaguna or @leandroalonso on this issue before submitting a PR.  

As per description above

_No response_

iPhone 14 Pro, iOS 17.1, Pocket Casts 7.50.0

rviljoen:
Fixes #1184 

For downloaded episodes, only use EffectsPlayer if trim silence is enabled, else use DefaultPlayer (which can already handle  Speed and Volume Boost)

## To test
1. Run current build
2. Download episode
3. Play episode with no effects applied
4. Notice that EffectsPlayer is used (can be seen in device log file)
5. Apply change and repeat steps above
6. Notice that DefaultPlayer is now used
7. Enable trim silence
8. Notice that app switches to EffectsPlayer
9. Disable trim silence
10. Notice that app switches back to DefaultPlayer
11. Enable and disable other effect settings (Speed and Volume Boost) as well to ensure that everything works

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -1083,7 +1083,7 @@ class PlaybackManager: ServerPlaybackDelegate {
                 return possiblePlayers // for Google Cast, only the Google Cast player is allowed
             }
 
-        if !playingOverAirplay(), !currEpisode.videoPodcast(), (currEpisode.downloaded(pathFinder: DownloadManager.shared) && effects().trimSilence != .off) || currEpisode.bufferedForStreaming() {
+            if !playingOverAirplay(), !currEpisode.videoPodcast(), (currEpisode.downloaded(pathFinder: DownloadManager.shared) && effects().trimSilence != .off) || currEpisode.bufferedForStreaming() {
                 possiblePlayers.append(EffectsPlayer.self)
             }
         #endif
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -1083,7 +1083,7 @@ class PlaybackManager: ServerPlaybackDelegate {
                 return possiblePlayers // for Google Cast, only the Google Cast player is allowed
             }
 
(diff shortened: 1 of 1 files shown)


leandroalonso (approved):
Thank you so much, @rviljoen!

PR #1188, opened 2023-10-30
Disable new show notes and RSS artwork

leandroalonso:
Unfortunately, we've been seeing some crashes related to the way we retrieve the new show notes. This PR disables it so we can measure and see if this is really an issue.

Also, given the RSS artwork depends on the new show notes, it has also been disabled.

## To test

1. Run the app
2. ✅ Make sure the show notes loads correctly

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 7.51
 -----
-
+- Improvements on battery consumption when trim silence is not enabled [#1186]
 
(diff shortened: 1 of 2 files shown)


emilylaguna (approved):
I tested various episodes and all the show notes loaded correctly in episode details, and the player. I also tested on iOS 15 and iOS 16

PR #1189, opened 2023-10-30
Playback 2023: Plus flow

leandroalonso:
|:---:|:---:|

Add the purchase flow for Plus stories. It also updates logic so the old login flow is not shown. 

https://github.com/Automattic/pocket-casts-ios/assets/7040243/6e07380a-2257-4fb9-96f8-e8caf82fcd00

Ps.: This does not implement the 2022 syncing.

## Correct login flow

1. Make sure you're logged in to an account that has a few episodes listened
3. Log out
3. Go back to Profile
4. Tap the EOY image
5. ✅ The current onboarding flow should be displayed — not the old one

## Paid story for non-paid users

1. Make sure you're logged in to an account that has a few episodes listened
3. Go back to Profile
4. Tap the EOY image
5. Skip to the 8th story
6. Tap "Start your free trial"
7. ✅ The purchase flow should be displayed
8. Complete the purchase
9. ✅ The flow should dismiss and the story should be available
10. Repeat the steps above with `patron` feature flag enabled

## Checklist


podcasts/End of Year/Stories/Views/PaidStoryWallView.swift
```diff
@@ -0,0 +1,53 @@
...
+                    ],
+                    startPoint: UnitPoint(x: 0.5, y: 0.24),
+                    endPoint: UnitPoint(x: 0.5, y: 1.04)
+                )
+
+                VisualEffectView(effect: UIBlurEffect(style: .systemThinMaterialDark))
+            }
+            .allowsHitTesting(false)
+            .ignoresSafeArea(edges: [.top, .bottom])
+        )
```
emilylaguna: I feel like we can just use the `ultraThinMaterial` as the background without the gradient which allows more of the background to show through but still ensures the text is legible:

| Before | After |
|:---:|:---:|
|||


```suggestion
        .background(.ultraThinMaterial)
        .environment(\.colorScheme, .dark)
```

podcasts/Onboarding/Models/PlusPurchaseModel.swift
```diff
@@ -140,6 +140,18 @@ private extension PlusPurchaseModel {
...
+        if OnboardingFlow.shared.currentFlow == .endOfYearUpsell {
+            if FeatureFlag.patron.enabled {
+                parentController.dismiss(animated: true)
+            } else {
+                parentController.presentingViewController?.dismiss(animated: true)
+            }
+
+            return
+        }
+
```
emilylaguna: Should we make this more generalized and add it as a flag like we did with the [shouldDismiss check in the OnboardingFlow](https://github.com/Automattic/pocket-casts-ios/blob/ae46fed9d26748f106c04db3cc041851fb2ef503/podcasts/Onboarding/OnboardingFlow.swift#L123)?

That way in the future if there are more flows that we want to do this with it will be easier to do.

podcasts/End of Year/Stories/Views/PaidStoryWallView.swift
```diff
@@ -0,0 +1,53 @@
...
+        GeometryReader { geometry in
+            PodcastCoverContainer(geometry: geometry) {
+                StoryLabelContainer(geometry: geometry) {
+                    SubscriptionBadge(tier: .plus, displayMode: .gradient, foregroundColor: .black)
+                    StoryLabel(L10n.eoyTheresMore, for: .title, geometry: geometry)
+                    StoryLabel(L10n.eoySubscribeToPlus, for: .subtitle, color: Color(hex: "8F97A4"), geometry: geometry)
+                }
+                .padding(.bottom, geometry.size.height * 0.06)
+
+                Button(L10n.eoyStartYourFreeTrial) {
```
emilylaguna: This is a tricky one since there is a chance the user isn't eligible for the free trial anymore, or we stop offering it. So this would say start free trial, but in the upgrade view it won't be an option. 

The fix for this is to use the `PlusPricingInfoModel` and check the `pricingInfo.hasFreeTrial` value:

```
struct PaidStoryWallView: View {
    @StateObject private var model = PlusPricingInfoModel()

    var body: some View {
        ...
        Button(model.pricingInfo.hasFreeTrial ? L10n.eoyStartYourFreeTrial : L10n.upgradeToPlan(L10n.pocketCastsPlusShort)) {
        }
        ...
    } 
}
```

emilylaguna (changes requested):
Nice work. Left a couple comments.

podcasts/End of Year/Stories/Views/PaidStoryWallView.swift
leandroalonso: cc @ashiagr

podcasts/Onboarding/Models/PlusPurchaseModel.swift
leandroalonso: Done in 885769d66270ebaac7f7313b30a3b643b0e893ef

podcasts/End of Year/Stories/Views/PaidStoryWallView.swift
leandroalonso: Done in ecf79bc7895f5716eae9d5ed13efdef8448f4e1b

leandroalonso:
@emilylaguna could you please 👀 again?

emilylaguna (approved):

PR #1191, opened 2023-10-30
Playback 2023: add logic to sync 2022 and 2023 listening history

leandroalonso:
| 📘 Part of: #1142 |
|:---:|

This PR adds logic for:

* Syncing 2022 and 2023 data when the user is a Plus user
* Syncing 2022 and 2023 data after a successful purchase from the Stories
* Syncing 2022 and 2023 logic after a user successfully purchases outside Stories

## Purchase outside Stories

1. Make sure you're logged in to an account that has a few episodes listened to this year and last year
4. Put a breakpoint on `let syncedWithSuccess = sync?()`
5. Open the stories
6. ✅ The breakpoint should hit
7. Mark yourself as plus or patron
8. Open the stories again
9. ✅ The breakpoint should hit (meaning the 2022 data will be synced)
10. Dismiss the stories
11. Open it again
12. ✅ The breakpoint should not hit

## Purchase from stories

1. In Xcode, enable the StoreKit configuration (Edit Scheme > Run > StoreKit Config)
1. Make sure you're logged in to an account that has a few episodes listened to this year and last year
3. Make sure to mark yourself as not Plus
4. Open stories
5. Go to the 8th story
6. Purchase Plus or Patron
7. ✅ You should see a screen with a spinner, meaning your 2022 data is syncing
8. After it finishes, you should see the year over year story
9. Tap next
10. ✅ You should see the completion rate story

## Checklist


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
```diff
@@ -17,7 +17,7 @@ class EndOfYearDataManager {
                                            """
 
     private lazy var listenedEpisodesPreviousYear = """
-                                            lastPlaybackInteractionDate IS NOT NULL AND lastPlaybackInteractionDate BETWEEN strftime('%s', '2021-01-01') and strftime('%s', '"2023-01-01"')
+                                            lastPlaybackInteractionDate IS NOT NULL AND lastPlaybackInteractionDate BETWEEN strftime('%s', '2022-01-01') and strftime('%s', '2023-01-01')
```
emilylaguna: Do we still need this and the `listenedEpisodesThisYear` if the `listenedEpisodes` does the same thing?

emilylaguna (approved):
This is working as described. I left a comment about improving the code.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift
leandroalonso: Hmm probably not.

I'll write that down and refactor it on a subsequent PR just to avoid any merge conflict.

rviljoen:
@leandroalonso I took at peak at the 2023 stories. Looks very sleek, well done! I noticed one issue on this story:

CleanShot 2023-11-08 at 15 11 49

It is not picking up my 2022 listen history. I had a quick look at the code and narrowed it down to this bit of SQL:

https://github.com/Automattic/pocket-casts-ios/blob/1a25a43a82227f76fdf6fdae69412726505b4414/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EndOfYearDataManager.swift#L19-L21

Looking at the SQL data directly, I can see many episodes where `lastPlaybackInteractionDate` is NULL, yet `playedUpTo` has a value.  While I don't recall doing so, I guess it's entirely possible that I may have cleared the listening history at some point, which does set `lastPlaybackInteractionDate` to NULL.

https://github.com/Automattic/pocket-casts-ios/blob/1a25a43a82227f76fdf6fdae69412726505b4414/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift#L428-L436

Either way, it may be a good idea to only display this story if 2022 listening time is greater than 0? It should cater for cases where listening history was cleared, as well as new users that only switched to Pocket Casts in 2023.

leandroalonso:
Hey @rviljoen, there are a few implications in doing so with the current way the stories are handled. You're right, clearing your listening history prevents us from checking the data.

I'll pass along the feedback to the team so we can see what we can do.

PR #1195, opened 2023-11-01
Make Pocket Casts compatible with the favourite (star) functionality introduced in iOS 17.1

### LINKED ISSUE #1194: Star functionality not compatible with iOS 17.1
### Description

This issue is related to #194 but differs in subtle ways, so I have opted to open this new issue to address what we can at this point and then leave #194 open to fix the rest down the line, should Apple make the updates we require.

In iOS 17.1, Apple [introduced expanded favourite functionality](https://support.apple.com/en-za/HT213781) for the Music app. As part of this change, a favourite (star) icon is now rendered on the lock screen and control centre widgets for Music.

[Now Playing](https://developer.apple.com/documentation/mediaplayer/becoming_a_now_playable_app) apps can add support for this favourite functionality by registering the [likeCommand](https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter/1618998-likecommand). Pocket Casts already does this when the _Extra Playback Actions_ setting is enabled. The difference between the iOS 17 implementation and pre iOS 16 is the fact that the skip back button is no longer removed, but an extra star icon is rendered. When enabling the existing option in the app, the following is rendered:

The current implementation has the following issues:
- While clicking the star will result in the episode being starred in the app, the Now Playing widget status is not updated (i.e. the star is not filled)
- If you start playing an episode which was starred previously, it shows up as starred in the app but not on the widgets
- Because of the above, there is no way to remove a star from an episode via the widgets
- If you star an episode from the widget, the star status is updated in the episode list, but not in the shelf of the full screen player (assuming you have the star as one of the configured options for the shelf)

1. Enabled _Extra Playback Actions_ in settings
2. Notice that an unfilled star show up in the Now Playing widgets (when an episode is playing)
3. Click the star. Notice that the episode is starred in the app, but the widgets are not updated

_No response_

iPhone 14 Pro, iOS 17.1, Pocket Casts 7.51

rviljoen:
Fixes #1194

Make Pocket Casts compatible with the favourite functionality introduced in iOS 17.1.

## To test
1. Disable _Extra Playback Actions_ in settings.
2. Play an episode. Note that star icon is _not present_ on native iOS Now Playing widget on lock screen or control centre.
3. Enable _Extra Playback Actions_ in settings.
4. Play an episode. Note that star icon _is present_ on Now Playing widget on lock screen or control centre.
5. Tap star icon on widget. Star should animate and remain filled.
6. Go to episode in app. Ensure star is applied.
7. Go to full screen player. Add Star to shelf if not present. Note that episode is starred in shelf.
8. Click star on shelf. Go back to Now Playing widgets and ensure that star is not filled.
9. Disable _Extra Playback Actions_ in settings. Note that star icon is no longer visible on Now Playing widgets.

## Checklist


podcasts/PlaybackManager.swift
rviljoen, line 1: With the previous code, the target was called multiple times resulting in a starred episode being unstarred immediately. Looking at the Apple documentation and sample app, using removeTarget(nil) ensures that all previous targets are removed and seems to be the safer implementation. The reason why this worked previously, is that we only supported "like" and not unlike. I.e. multiple calls to star always yielded the same result of star, whereas with the new functionality we now support unstar as well to conform to the new Apple UI.

podcasts/PlaybackManager.swift
```diff
@@ -81,6 +81,7 @@ class PlaybackManager: ServerPlaybackDelegate {
         NotificationCenter.default.addObserver(self, selector: #selector(updateExtraActions), name: Constants.Notifications.extraMediaSessionActionsChanged, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(updateNowPlayingInfo), name: Constants.Notifications.userEpisodeUpdated, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(updateAllNowPlayingData), name: .episodeEmbeddedArtworkLoaded, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(handleEpisodeStarredChanged), name: Constants.Notifications.episodeStarredChanged, object: nil)
```
emilylaguna: Instead of adding a notification, we can look at using 
[`PlaybackManager.nowPlayingStarredChanged`](https://github.com/Automattic/pocket-casts-ios/blob/5866f0f53731c790f9c60d1717e26929ff8aea27/podcasts/PlaybackManager.swift#L1797) which should be triggered in the same way.

podcasts/PlaybackManager.swift
```diff
@@ -1567,22 +1568,31 @@ class PlaybackManager: ServerPlaybackDelegate {
...
             markPlayedCommand.isEnabled = false
 
-            starCommand.removeTarget(self)
+            starCommand.removeTarget(nil)
             starCommand.isEnabled = false
         }
     }
+    
+    @objc private func handleEpisodeStarredChanged() {
+        guard let episode = self.currentEpisode() else {return}
```
emilylaguna: [Nit] Small formatting nitpicks. 

```suggestion
        guard let episode = currentEpisode() else { return }
```

emilylaguna (changes requested):
Hey @rviljoen!! Thanks for tackling this 💪, I'm very happy to have my star back on the lock screen! 🥳 

This is working well in my testing but I did notice a few things:

- When switching episodes from one that is starred -> not starred (and vice versa) the lock screen starred state doesn't update
- If I change the star state on another device and sync the change:
   - the star in the shelf doesn't update
   - the lock screen star doesn't update
- If I play a custom file which doesn't support being starred, the shelf action disappears correctly, but the lock screen star doesn't.

podcasts/PlaybackManager.swift
rviljoen: Good catch! Updated in https://github.com/Automattic/pocket-casts-ios/pull/1195/commits/752451be3d68167b8fdca8f07c77d9c75fd7c3ea

rviljoen:
@emilylaguna Thanks for the thorough testing and feedback. I think have managed to address everything. Ready for another review 🙏🏻

rviljoen:
@emilylaguna Something else worth thinking about for the future. In the current implementation, the star button is behind a user setting, which I believe is disabled by default. This made sense when the user had to give up the Skip Back button to get the like (star) functionality. Now that the star has it's own dedicated slot on the Now Playing widgets, it may be a good idea to expose this all users on iOS 17.1 and higher. I don't think we should do this right away, but in future once we are comfortable with the functionality being stable. Thoughts?

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 7.52
 -----
-
+- Fixes star functionality on native Now Playing widgets (lock screen and control centre) for iOS 17.1 and later [#1195]
```
emilylaguna: Can you move this into the 7.54 section?

emilylaguna (changes requested):
Sorry this took so long. I gave this another test and it seems to be working well now! 

I left 1 comment to move the release notes item into the new section, and after that we'll be good to go!

CHANGELOG.md
rviljoen: Resolved in https://github.com/Automattic/pocket-casts-ios/pull/1195/commits/9386f710482a22d3b3b1b5d56a1b2a2a8845a45d

rviljoen:
Thanks @emilylaguna. I think we are good to go 🥳

emilylaguna (approved):

PR #1197, opened 2023-11-02
Fix scroll indicator behind player for long episode lists

### LINKED ISSUE #1072: Bug: Scroll indicator behind player
When scrolling down in the episodes list of an podcast the scroll indicator disappears behind the player.

image

rviljoen:
Fixes #1072 

When scrolling long episode lists, the vertical scroll bar indicator no longer disappears behind the mini player when you get to the bottom of the list

## To test
1. Open current build and find long episode list (100+ should do)
2. As you scroll to the bottom, notice that the vertical scroll indicator moves beyond the bottom of the episode list. If you scroll down far enough, it will disappear completely.

IMG_5171

3. Build with the changes in this PR and repeat the steps above. Notice that the scroll indicator now stays above the mini player

image

4. Repeat the same test with episode list in multi-select mode enabled

## Checklist


leandroalonso (changes requested):
@rviljoen this is a step forward in fixing this issue, however when the mini player is not enabled this feels like a bug to me:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/3ab74245-f7b3-474a-86ca-bec8b05bd38d

For SwiftUI we created the [MiniPlayerPadding](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/SwiftUI/MiniPlayerPaddingModifier.swift) view modifier.

I wonder if we can apply something similar here?

Ps.: we also have the same issue on Profile > Starred/Downloaded/Listening History. We don't necessarily need to tackle this here but it doesn't fix the issue in all places.

rviljoen:
@leandroalonso Good feedback, thanks. I'll take a look.

emilylaguna:
👋 If it helps, there is a common UITableView extension that applies the mini player padding in other areas in the app: 

https://github.com/Automattic/pocket-casts-ios/blob/cde718fe203f1c52e10e8d7c3c455194a4eb5b5d/podcasts/UITableView%2BMiniPlayer.swift#L4

rviljoen:
I played around some more and it turns out that fixing this may require a bit more effort and testing than what it is worth for fixing a relatively small visual glitch. I have closed the PR for now and may relook at it in future.

PR #1215, opened 2023-11-14
Playback 2023: fix hanging loading the episodes

leandroalonso:
| 📘 Part of: #1142 |
|:---:|

There's a bug that can occur if the user has no episodes to sync.

The sync code ends up on `totalEpisodesDispatchGroup.wait()` and it will never leave. That's why I didn't handle the case where there are no episodes to sync. 🤡 

## To test

1. Do a clean install
3. Go to Profile and tap the End of Year card
4. ✅ Your data should load correctly

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
@@ -95,6 +95,9 @@ class SyncYearListeningHistoryTask: ApiBaseTask {
                 performRequest(yearToSync: year, token: token, shouldSync: true)
             } else {
                 success = true
+
+                // If there are no episodes to sync we leave the dispatch group
+                totalEpisodesDispatchGroup.leave()
             }
         } catch {
             print("SyncYearListeningHistory had issues decoding protobuf \(error.localizedDescription)")
--- a/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
@@ -31,36 +31,21 @@ public class SyncYearListeningProgress: ObservableObject {
 class SyncYearListeningHistoryTask: ApiBaseTask {
     private var token: String?
 
-    private let yearsToSync: [Int32]
-
-    private var totalEpisodesDispatchGroup = DispatchGroup()
+    private let yearToSync: Int32
 
     var success: Bool = false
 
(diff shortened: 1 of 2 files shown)


Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
```diff
@@ -95,6 +95,9 @@ class SyncYearListeningHistoryTask: ApiBaseTask {
                 performRequest(yearToSync: year, token: token, shouldSync: true)
             } else {
                 success = true
+
+                // If there are no episodes to sync we leave the dispatch group
+                totalEpisodesDispatchGroup.leave()
```
leandroalonso: There's a way to improve this (unfortunate) collection of dispatch groups. However, I think that for the release branch, we should be careful so I'm making as less changes as possible.

podcasts/End of Year/Stories/EpilogueStory.swift
```diff
@@ -77,13 +77,15 @@ struct EpilogueStory: ShareableStory {
...
-                VStack {
-                    Spacer()
-                    HStack {
-                        Spacer()
-                        Image("logo")
-                            .padding(.bottom, geometry.size.height * 0.06)
+                if !renderForSharing {
+                    VStack {
                         Spacer()
+                        HStack {
```
leandroalonso, line 83: I also added this since is a small change.

To test:

1. Go to the last story
2. Trigger a screenshot
3. Tap "Share"
4. Save the image
5. ✅ Verify there is just a single logo at the bottom

emilylaguna (changes requested):
While testing I noticed that if there are multiple year requests, and one of them fails the app will be stuck loading:

To force reproduce this state, I changed SyncYearListeningHistoryTask.swift:69 from

`let url = ServerConstants.Urls.api() + "history/year"` 
to 
`let url = yearToSync == 2023 ? ServerConstants.Urls.api() + "history/year" : "https://localhost/"`

leandroalonso:
@emilylaguna good catch! I decided to change a bit this code to get rid of all of these `DispatchGroup`. The changes seem a little bit bigger but it's not:

* I reverted `SyncYearListeningHistoryTask` to how it was before handling multiple years. Now it handles just one year;
* I used a `DisaptchGroup` outside of it. Given we've kept the previous behavior, the operation will always succeed or fail
* This cleaned a lot of the mess using the `DispatchGroup`

I've also updated the testing steps.

emilylaguna (approved):
I tested this with both scenarios and they're both handled now, I also tested the success state, no internet, logged out to purchasing, and all is working. 

I also tested the screenshot detection and no longer see the double PC logo.

PR #1218, opened 2023-11-16
Playback 2023: design tweaks

leandroalonso:
| 📘 Part of: #1142 |
|:---:|

This addresses some of the design feedback. I tried to make every change in a single commit so it can be easily reviewed:

1. Reduce longest episode duration and add bounce (747f1382)
2. Reduce top 1 to top 5 transition duration (a494c8f9)
3. Increase all stories duration to 7 seconds (829eac95)
4. Add PC logo and Playback logo with shadow on the first story (acacaa3c)
5. Adjust tap and drag gesture so they don't collide (96352809)
6. Multiple tweaks on Year over Year (4f01f28f)

## To test

2. Go to Profile and tap the End of Year card
3. ✅ The first story should contain Pocket Casts logo and Playback 2023 should have a light shadow
4. ✅ Notice that the first story (and all others) now takes 7s to complete (instead of 5s)
5. Go to the top 1 story
6. Tap to go to the next story
7. ✅ The transition should be quicker now
8. Go to the longest episode story
9. ✅ The animation should be quicker and with bounce
10. Tap and hold for just a few moments and let go
11. The story should go to the next/back
12. Tap to go to the next/previous story
13. ✅ It should go back/next as it was before

### Year over Year

Go to `EndOfYearDataManager.swift` and change `eturn YearOverYearListeningTime(totalPlayedTimeThisYear: listeningTimeThisYear, totalPlayedTimeLastYear: listeningTimePreviousYear)` to:

```
return YearOverYearListeningTime(totalPlayedTimeThisYear: listeningTimeThisYear, totalPlayedTimeLastYear: 0)
```

1. Run the app and go to the Year over year story
2. ✅ Notice how the bar is not underneath the "Share story"
3. ✅ Check that there are widows on the title
4. ✅ Check that the font is slightly smaller on the bars 

## Checklist


podcasts/End of Year/Stories/CompletionRateStory.swift
```diff
@@ -3,7 +3,7 @@ import PocketCastsServer
 import PocketCastsDataModel
 
 struct CompletionRateStory: ShareableStory {
-    var duration: TimeInterval = 5.seconds
+    let duration: TimeInterval = EndOfYear.defaultDuration
```
emilylaguna: What if instead of storing the same duration in each story, we just define it as a default property on the Story protocol?

```
extension Story {
    var duration: TimeInterval { EndOfYear.defaultDuration }
}
```

Then we can remove all the duplicated references.

emilylaguna (approved):
Left a small comment about improving the duration code, but it's non-blocking for this PR.

podcasts/End of Year/Stories/CompletionRateStory.swift
leandroalonso: Done in c33577bb44d1ecef1bea4d15d57ede1d7ad07a40

PR #1223, opened 2023-11-17
Playback 2023: tweak parallax on the intro story

leandroalonso:
| 📘 Part of: #1142 |
|:---:|

This PR makes two changes to the parallax effect:

1. It calculates the pitch/roll based on the initial value. This means that the "2023" won't jump on the screen as previously;
2. It defines a boundary so the pitch and roll values will always be between -1.4 and 1.4. This is because tilting aggressively can sometimes cause the numbers to "jump"

## To test

Test using a real device.

2. Make sure yo have your device fully on the vertical
3. Go to Profile and tap the End of Year card
4. ✅ Notice that the numbers don't jump
5. Tilt your device and the numbers should follow
6. Go to the next story and back
7. ✅ The numbers won't jump

## Checklist

### DIFF
--- a/podcasts/End of Year/Stories/IntroStory.swift
+++ b/podcasts/End of Year/Stories/IntroStory.swift
@@ -103,9 +103,11 @@ private struct TwentyThreeParallaxModifier: ViewModifier {
     var rollMultiplier: Double = 4
     var pitchMultiplier: Double = 40
 
+    private let rollAndPitchBoundary = -1.4..<1.5
+
     func body(content: Content) -> some View {
-        let roll = manager.roll.betweenOrClamped(to: -1.4..<1.5) * 10
-        let pitch = manager.pitch.betweenOrClamped(to: -1.4..<1.5)
+        let roll = manager.roll.betweenOrClamped(to: rollAndPitchBoundary) * 10
+        let pitch = manager.pitch.betweenOrClamped(to: rollAndPitchBoundary)
         return content
             .offset(x: roll * rollMultiplier, y: pitch * pitchMultiplier)
             .onAppear() {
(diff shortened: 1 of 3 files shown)


podcasts/End of Year/Stories/IntroStory.swift
```diff
@@ -99,14 +99,16 @@ struct IntroStory: ShareableStory {
 
 private struct TwentyThreeParallaxModifier: ViewModifier {
     @Environment(\.accessibilityReduceMotion) private var reduceMotion
-    @StateObject var manager: MotionManager = .init()
+    @StateObject var manager: MotionManager = .init(relativeToWhenStarting: true)
     var rollMultiplier: Double = 4
     var pitchMultiplier: Double = 40
 
+    private let rollAndPitchBoundary = -1.4..<1.5
```
leandroalonso, line 106: Even with these boundaries, I feel that if you do crazy tilting the numbers jump crazy. However, I'm considering this an edge case for now 🤷 

Any ideas are very appreciated though.

emilylaguna: This is oddly annoying and difficult to fix, and something I spent far too long trying to. 

A problem with this solution is if the user is lying in bed, with their phone upside then the effect breaks. Unfortunately I don't have a better solution though. 

https://github.com/Automattic/pocket-casts-ios/assets/793774/57d5fd7c-6f76-46d3-8b22-65dd5c426d2c

leandroalonso: I might be wrong here but I think this issue is caused even without the changes on this PR?

So at least we can have the "jump" issue fixed with it.

emilylaguna (approved):

podcasts/End of Year/Stories/IntroStory.swift
emilylaguna: true!!

PR #1232, opened 2023-11-27
End of Year: fix thread-safety access in iOS 17

leandroalonso:
This PR ensures the changes to `podcastsToUpdate` are done in a thread-safe manner by using an `NSLock`. This portion of code can crash due to a data racing condition.

While I'm not able to directly reproduce that (because it needs two threads accessing the same var at the exact same time) it's possible to reproduce this issue by changing `updateEpisodes` to [this](https://gist.github.com/leandroalonso/ea4b0a06c361f38429f94f1171bcfb6b) — running this code a few times will result in a crash.

## To test

1. Change `EndOfYearStoriesBuilder.swift` if in line 44 to be `if true` (forcing the sync to always happen)
2. Run the app
3. Check your stories
4. ✅ It should display normally

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Sync/SyncYearListeningHistoryTask.swift
@@ -102,6 +102,8 @@ class SyncYearListeningHistoryTask: ApiBaseTask {
     }
 
     private func updateEpisodes(updates: [Api_HistoryChange]) {
+        let lock = NSLock()
+
         var podcastsToUpdate: Set<String> = []
 
         // Get the list of missing episodes in the database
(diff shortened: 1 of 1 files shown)


emilylaguna (approved):
Before applying this PR I was able to reproduce the original issue using a breakpoint in the simulator that automatically continued after running a debugger command.





## Tested:
For each of the tests I:
- Verified that I can reproduce the original issue by:
   - Commenting out the locks
   - Launched the app and did the EOY sync process until the app crashed
- Verified the PR fixes it:
   - Uncommented locks
   - Launched the app and did the EOY sync about ~10 times
   - Disabled the breakpoint
   - Launched the app a bunch

✅  iOS 15.5
✅  iOS 16.4
✅  iOS 17.1
✅  iOS 17.2

PR #1238, opened 2023-11-29
Playback: small fixes for next year

leandroalonso:
* Changes the title of the created top podcasts lists to not contain the year (so we don't run into the issue of it saying 2022 as it happened this year)
* Adds a lock in PodcastExistsHelper to prevent data race conditions¶

¶ Ideally the sync of podcasts should be rewritten to use async/await and get rid of those manual thread manipulations/locks. However, it's good to have this in place so this won't cause crashes.

## To test

1. Change `EndOfYearStoriesBuilder.swift` if in line 44 to be `if true` (forcing the sync to always happen)
2. Run a clean install of the app
3. Check your stories
4. ✅ It should display normally
5. On your top five podcasts story, tap share
6. ✅ Check that the created lists contain the title "My top podcasts of the year!"

## Checklist


emilylaguna (approved):
Works for me!

PR #1249, opened 2023-12-01
Issue 17 – Buffer Logs to Reduce Filesystem Writes

### LINKED ISSUE #17: Battery Usage
### User report

>Just sent a Twitter DM as well but I’m finding the app uses a ton of power. Is this expected? Was surprised to find my new iPhone 13 Pro doesn’t have that much better battery life than my aging iPhone XS, and seems Pocket Casts is the primary culprit. 
image1 image0

### Additional info

From Chip Snyder:

I spent some time looking onto this and I have a few take-aways for this user. 

They're on `7.18.1` and they have `132` podcasts that they subscribe to. Also a few settings that stand out to me: 
> Auto Download On: yes
> Auto Download Only on Wifi: yes
> Auto Download Up Next:  yes

They have Auto-Download on but our logs don't provide insight into how many are included in the Auto-Download. We might want to add that to the meta-data if these issues persist. I wasn't familiar with the app "No Cell Coverage" but [apparently](https://discussions.apple.com/thread/7421202) this is a tool that Apple will use to help boost the signal in low connectivity cases. 

**Thoughts**
Given that they have so many podcasts and that they Auto Download I wonder if their battery is being effected by episodes downloading under cellular data. If they're downloading on a cell network (especially a poor cellular network) then they'll definitely have a greater impact on their battery life. 

With that bug fixed I'd be interested to see how their battery usage impact changes because they have all of the settings I would encourage them to use. Maybe turn off `Auto Download` for shows that they don't listen to regularly 🤷 and only download them when added to up next using `Auto Download Up Next`

----- 
# Other Research

I did some deep diving into this over the last two days so I want to document some of my findings before the week is out. 

I pulled up some of our battery stats and found a few interesting cases. Our biggest impact is while in the background which should be expected given the nature of the app.

| Average Users | Heavy Users |
| --- | --- |
|  |  |

For our average users our impact is minimal and we shouldn't worry too much IMO. We average about 1.3% of their daily battery usage which is down from the previous release. 

Our heavy users use 6.65% on average (this aligns pretty closely to my own use.) Audio is high (2.59%) but that's expected and I haven't found a way to be more efficient there yet. Processing is nearly just as high though. 

----- 

### FileLog

I believe a big contributor to the processing impact is due to the FileLog. Each time an event fires a loggable event the process is
- Open the log file
- Seek to the end
- Write the log
- Close

This kind of I/O can be expensive. A few [recommendations from Apple](https://developer.apple.com/library/archive/documentation/Performance/Conceptual/EnergyGuide-iOS/MinimizeIO.html#//apple_ref/doc/uid/TP40015243-CH42-SW1):

> Minimize data writes. Write to files only when their content has changed, and aggregate changes into a single write whenever possible. Avoid writing out an entire file if only a few bytes have changed. **If you frequently change small portions of large files, consider using a database to store the data instead.**

While streaming an episode our energy impact is on the high side.

Running the same test with the file log disabled moved us back into the low category. 

Batching these events and/or moving them to the DB could also be a viable approach to saving us energy usage. However this needs more validation because it's not a 1 to one comparison given the network calls that happen around these screen shots with various intervals. Which raises that we might be able to batch requests as well. 

More info: #4529585-zen

mylesbarros:
Fixes #17 

This is a partial fix for [Issue 17 Battery Usage](https://github.com/Automattic/pocket-casts-ios/issues/17). I don't have access to the performance data from Apple but implemented the `FileLog` buffering described in the issue.

There are three primary changes to the `FileLog`.

The first and most significant change is that we no longer write each log message to the filesystem each time that `addMessage(_ message: String)` is called. Instead we store the log entry in a `logBuffer` collection. Each time that we reach a number of log messages specified in the `init` (defaults to 100) we batch write the cached log entries to the log file. We also forcibly flush the cache when the app enters the background.

The second change is the delegation of the log file rotation and the log writing to separate protocols `FileRotating` and `PersistentTextWriting` respectively. While the abstraction is nice this has mostly been done to implement unit tests for the `FileLog` to ensure that its flushing the log buffer.

The final change is that we no longer invoke `FileLog.shared.setup()` at app start. Instead the `LogFileWriter` that implements `PersistentTextWriting` in production will create the directories in which we store the logs if writing to the log fails with `NSFileNoSuchFileError`.

## To test

1. Navigate around the app and start a podcast to generate log entries.
2. Tap the Profile tab
3. Tap the gear to access Settings
4. Scroll to the bottom and tap Help & Feedback
5. Scroll to the bottom and tap Get in touch
6. Tap Get Support
7. Tap Attached Logs at the bottom
8. Verify that Debug Logs contains the expected log output such as for the podcast that was started

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Logging/FileLog.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Logging/FileLog.swift
@@ -38,16 +38,16 @@ public final class FileLog {
     private let bufferThreshold: UInt
     private let logPersistence: PersistentTextWriting
     private let logRotator: FileRotating
+    private let logQueue: DispatchQueueing
     private let logger: Logger?
 
-    private lazy var entries: [LogEntry] = [] {
+    private lazy var logBuffer: [LogEntry] = [] {
         didSet {
-            if entries.count >= bufferThreshold {
+            if logBuffer.count >= bufferThreshold {
                 writeLogBufferToDisk()
             }
         }
     }
-    private let logQueue: DispatchQueueing
 
     init(
         logPersistence: PersistentTextWriting,
@@ -74,7 +74,7 @@ public final class FileLog {
         // if it's important enough to log to file, write it to the debug console as well
         logger?.log("\(message, privacy: .public)")
 
-        entries.append(LogEntry(message))
+        logBuffer.append(LogEntry(message))
     }
 
     // Just a shortcut for `addMessage` to be used specifically for
@@ -90,21 +90,8 @@ public final class FileLog {
     }
 
     public func forceFlush() {
-        logger?.log("\(Self.self) forcibly flushing to disk.")
+        logger?.debug("\(Self.self) forcibly flushing to disk.")
         writeLogBufferToDisk()
-    }
-
-    private func writeLogBufferToDisk() {
-        let logTimestampFormatter = DateFormatHelper.sharedHelper.localTimeJsonDateFormatter
-        let newLogChunk = entries.reduce(into: "") { resultChunk, logEntry in
-            let formattedTimestamp = logTimestampFormatter.string(from: logEntry.timestamp)
-            let logLine = "\(formattedTimestamp) \(logEntry.message)\n"
-
-            resultChunk.append(logLine)
-        }
-
-        entries.removeAll(keepingCapacity: true)
-        appendStringToLog(newLogChunk)
     }
 
     public func loadLogFileAsString(completion: @escaping (String) -> Void) {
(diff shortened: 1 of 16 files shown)


Modules/Utils/Sources/PocketCastsUtils/Logging/FileLog.swift
```diff
@@ -102,36 +128,26 @@ public class FileLog {
         .eraseToAnyPublisher()
     }
 
-    private func appendStringToLog(_ line: String) {
+    private func writeLogBufferToDisk() {
+        let logTimestampFormatter = DateFormatHelper.sharedHelper.localTimeJsonDateFormatter
+        let newLogChunk = logBuffer.reduce(into: "") { resultChunk, logEntry in
+            let formattedTimestamp = logTimestampFormatter.string(from: logEntry.timestamp)
+            let logLine = "\(formattedTimestamp) \(logEntry.message)\n"
```
leandroalonso: Maybe we can move this formatation logic to `LogEntry`?

leandroalonso (commented):
@mylesbarros a PR with unit tests created by someone with a Donkey Kong profile picture? Sign me in! 😁

Tested and it works just as before. I left one small suggestion, not a blocker though.

Modules/Utils/Sources/PocketCastsUtils/Logging/FileLog.swift
mylesbarros: Good idea – the `DateFormatter` is static so it's nice isolating it to `LogEntry`. Updated ✅

leandroalonso (approved):
Thank you so much for your contribution @mylesbarros!

PR #1251, opened 2023-12-01
Fix episodes not showing up on the app

leandroalonso:
We have an issue on iOS that prevents episodes from appearing if:

1. A user is subscribed to the podcast
2. The podcast author change the dates of the episodes (this can happen)

The app ends up in a state where episodes are not added but it does have the latest episode. This causes those "in-between" episodes never to be added.

## To test

Preferably, it would be good to have a database with this state. If you don't have one, please ping me.

1. Open the app
2. Go to a podcast that had the dates changed ([it seems that this almost always happens with this one](https://pca.st/k09G))
3. ✅ The missing episodes should be added and their states will be synced

## Additional information

The additional update made in this fix will happen only if:

1. The episode has missing episodes that are not the latest ones. This means that there are "holes": missing episodes that are in between already added episodes;
2. It just syncs enough to get the missing episodes

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Podcast.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Podcast.swift
@@ -56,7 +56,7 @@ public class Podcast: NSObject, Identifiable {
 
     // if set to true, all podcast episodes will be refreshed in the next sync
     // useful for when episodes are missing
-    public var forceRefreshEpisodes: Bool = false
+    public var forceRefreshEpisodeFrom: String? = nil
 
     public func autoDownloadOn() -> Bool {
         autoDownloadSetting == AutoDownloadSetting.latest.rawValue
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Podcast.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Podcast.swift
@@ -54,8 +54,8 @@ public class Podcast: NSObject, Identifiable {
     // transient not saved to database
     public var cachedUnreadCount = 0
 
-    // if set to true, all podcast episodes will be refreshed in the next sync
-    // useful for when episodes are missing
+    // if set to an episode UUID, all podcast episodes after the given
+    // UUID will be updated
     public var forceRefreshEpisodeFrom: String? = nil
 
     public func autoDownloadOn() -> Bool {
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Podcast.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Podcast.swift
@@ -54,6 +54,10 @@ public class Podcast: NSObject, Identifiable {
     // transient not saved to database
     public var cachedUnreadCount = 0
 
+    // if set to true, all podcast episodes will be refreshed in the next sync
+    // useful for when episodes are missing
+    public var forceRefreshEpisodes: Bool = false
+
     public func autoDownloadOn() -> Bool {
         autoDownloadSetting == AutoDownloadSetting.latest.rawValue
     }
--- a/Modules/Server/Sources/PocketCastsServer/Public/Refresh/MainServerHandler.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Refresh/MainServerHandler.swift
@@ -237,7 +237,7 @@ public class MainServerHandler {
         var jsonRequest = jsonWithStandardParams(uniqueId: uniqueId)
         jsonRequest["push_sound"].string = "11" // for legacy reasons, this is always the push sound we send, since it's no longer configurable
         jsonRequest["podcasts"].string = podcasts.map(\.uuid).joined(separator: ",")
-        jsonRequest["last_episodes"].string = podcasts.map { $0.forceRefreshEpisodes ? "" : $0.latestEpisodeUuid ?? "" }.joined(separator: ",")
+        jsonRequest["last_episodes"].string = podcasts.map { $0.forceRefreshEpisodeFrom ?? $0.latestEpisodeUuid ?? "" }.joined(separator: ",")
         jsonRequest["push_messages_on"].string = podcasts.map { (pushEnabled && $0.pushEnabled) ? "1" : "0" }.joined()
         if let pushToken = ServerSettings.pushToken() {
             jsonRequest["push_token"].string = pushToken
--- a/Modules/Server/Sources/PocketCastsServer/Public/Refresh/MainServerHandler.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Refresh/MainServerHandler.swift
@@ -237,7 +237,7 @@ public class MainServerHandler {
         var jsonRequest = jsonWithStandardParams(uniqueId: uniqueId)
         jsonRequest["push_sound"].string = "11" // for legacy reasons, this is always the push sound we send, since it's no longer configurable
         jsonRequest["podcasts"].string = podcasts.map(\.uuid).joined(separator: ",")
-        jsonRequest["last_episodes"].string = podcasts.map { $0.latestEpisodeUuid ?? "" }.joined(separator: ",")
+        jsonRequest["last_episodes"].string = podcasts.map { $0.forceRefreshEpisodes ? "" : $0.latestEpisodeUuid ?? "" }.joined(separator: ",")
         jsonRequest["push_messages_on"].string = podcasts.map { (pushEnabled && $0.pushEnabled) ? "1" : "0" }.joined()
         if let pushToken = ServerSettings.pushToken() {
             jsonRequest["push_token"].string = pushToken
--- a/Modules/Server/Sources/PocketCastsServer/Public/Refresh/RefreshManager.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Refresh/RefreshManager.swift
@@ -30,6 +30,22 @@ public class RefreshManager {
         refreshQueue.addOperation(UpNextSyncTask())
     }
 
+
+    /// Updates a podcast and all the associated information.
+    ///
+    /// Note that this will force all the episodes to be updated.
+    /// - Parameter podcast: a `Podcast` object
+    public func refresh(podcast: Podcast) {
+        podcast.forceRefreshEpisodes = true
+        refresh(podcasts: [podcast]) {
+            if SyncManager.isUserLoggedIn() {
+                guard let episodes = ApiServerHandler.shared.retrieveEpisodeTaskSynchronouusly(podcastUuid: podcast.uuid) else { return }
+
+                DataManager.sharedManager.saveBulkEpisodeSyncInfo(episodes: DataConverter.convert(syncInfoEpisodes: episodes))
+            }
+        }
+    }
+
     public func refreshPodcasts(forceEvenIfRefreshedRecently: Bool = false) {
         if !forceEvenIfRefreshedRecently {
             if let lastRefreshStartTime = ServerSettings.lastRefreshStartTime(), fabs(lastRefreshStartTime.timeIntervalSinceNow) < RefreshManager.minTimeBetweenRefreshes {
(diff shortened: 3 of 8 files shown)


Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
```diff
@@ -104,8 +108,15 @@ extension ServerPodcastManager {
                 continue
             }
 
-            // for subscribed podcasts, only add older episodes, newer ones are handled by a refresh
+            // For subscribed podcasts, only add older episodes, newer ones are handled by a refresh
+            // so stuff like auto-download, auto add to up next works as expected
+            // However, if this is not a new episode we force update the podcast episodes
             if podcast.isSubscribed(), episodeDate.timeIntervalSinceNow >= podcast.latestEpisodeDate?.timeIntervalSinceNow ?? TimeInterval.greatestFiniteMagnitude {
+                if !missingEpisodesThatIsNotTheLastOne {
```
leandroalonso: This algorithm is terrible, it deserves a refactor. However, I'd like to get a review and thoughts about the idea before jumping into refactors/unit testing.

It just checks if the episodes that are missing are "in-between" or if they are the latest ones.

Modules/Server/Sources/PocketCastsServer/Public/Refresh/RefreshManager.swift
```diff
@@ -30,6 +30,22 @@ public class RefreshManager {
...
+
+    /// Updates a podcast and all the associated information.
+    ///
+    /// Note that this will force all the episodes to be updated.
+    /// - Parameter podcast: a `Podcast` object
+    public func refresh(podcast: Podcast) {
+        podcast.forceRefreshEpisodes = true
+        refresh(podcasts: [podcast]) {
+            if SyncManager.isUserLoggedIn() {
+                guard let episodes = ApiServerHandler.shared.retrieveEpisodeTaskSynchronouusly(podcastUuid: podcast.uuid) else { return }
```
leandroalonso, line 42: We update the episode info because the user might have listened to some episodes on another platform.

If we don't do that we risk adding the episodes without their data.

Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
```diff
@@ -104,8 +108,15 @@ extension ServerPodcastManager {
                 continue
             }
 
-            // for subscribed podcasts, only add older episodes, newer ones are handled by a refresh
```
leandroalonso, line 107: At first, I wasn't sure why that, but now it's clear: the refresh handles auto-download, auto-add to Up Next, etc. That's why we shouldn't update these episodes here.

Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
```diff
@@ -155,6 +166,10 @@ extension ServerPodcastManager {
             // for subscribed podcasts remove any non-interacted with episodes that aren't in the server JSON
             cleanupDeletedEpisodes(podcast: podcast, serverEpisodes: episodesJson)
         }
+
+        if missingEpisodesThatIsNotTheLastOne {
+            RefreshManager.shared.refresh(podcast: podcast)
```
geekygecko: This looks like it will download all the podcast episodes sync information. Android uses the following call to download the single episode sync data. It's handy for spots like the Up Next where you might not be subscribed to the podcast but require the user sync information. 


```
    @GET("/mobile/podcast/findbyepisode/{podcastUuid}/{episodeUuid}")
    fun getPodcastAndEpisodeSingle(@Path("podcastUuid") podcastUuid: String, @Path("episodeUuid") episodeUuid: String): Single<PodcastResponse>
```

leandroalonso: @geekygecko we don't even need to make an additional request if we don't want to, given that during this refresh we already have the episode data. This wasn't done for a few reasons:

1. There was a comment explicitly stating that new episodes for subscribed podcasts are skipped by this logic because they might need to be auto-downloaded and auto-added to Up Next. A "podcast refresh" handles that, this logic doesn't;
2. If we just add the episode — without any other call — we have the possibility of adding an episode without any status. So, let's say a user started or even played this podcast on another platform, we won't update it accordingly.

This logic should be triggered if, and **only if**:

1. A podcast has episodes "in the middle" that weren't added before;
2. This should affect pretty much ONLY podcasts that change their publish dates leading the app to "ignore" some episodes

In terms of load on the server, I don't expect much change because is a very specific case.

However, we can take a different approach:

1. If the episode is "in the middle" we just add it right away. No additional calls, no sync.

The downside is that if the user has any special setting for this podcast (like Auto Download and/or Add to Up Next), we won't do it. Also, if they interacted with this episode on another platform we won't sync that either (like never).

It's still better as the current state (in which the episode NEVER appears) but it sounds like a half-fix. 😬

Let me know your thoughts!

pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/37">7.54</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/Refresh/RefreshManager.swift
```diff
@@ -44,14 +60,20 @@ public class RefreshManager {
             }
         }
 
+        refresh(podcasts: DataManager.sharedManager.allPodcasts(includeUnsubscribed: false))
+    }
+
+    private func refresh(podcasts: [Podcast], completion: (() -> Void)? = nil) {
         UserDefaults.standard.set(Date(), forKey: ServerConstants.UserDefaults.lastRefreshStartTime)
-        let podcasts = DataManager.sharedManager.allPodcasts(includeUnsubscribed: false)
+        let podcasts = podcasts
```
emilylaguna:
```suggestion
```

Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
```diff
@@ -155,6 +174,14 @@ extension ServerPodcastManager {
             // for subscribed podcasts remove any non-interacted with episodes that aren't in the server JSON
             cleanupDeletedEpisodes(podcast: podcast, serverEpisodes: episodesJson)
         }
+
+        // If there are episode missing that are not the recent ones this mean that
+        // likely this show changed the episodes dates. This might lead to some
+        // episodes never being added, so we add them here
+        if missingEpisodesThatIsNotTheLatestOnes, let latestAvailableEpisodeUuid {
+            FileLog.shared.addMessage("[ServerPodcastManager] Updating \(podcast.title ?? "") from episode UUID \(latestAvailableEpisodeUuid)")
+            RefreshManager.shared.refresh(podcast: podcast, from: latestAvailableEpisodeUuid)
```
emilylaguna, line 190: One concern I have about this code is that its used by the download queue which could result in us triggering a sync multiple times in quick succession. 

Would it make sense to only trigger the refresh if it's being loaded from the podcast details?

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Podcast.swift
```diff
@@ -54,6 +54,10 @@ public class Podcast: NSObject, Identifiable {
     // transient not saved to database
     public var cachedUnreadCount = 0
 
+    // if set to true, all podcast episodes will be refreshed in the next sync
+    // useful for when episodes are missing
```
emilylaguna: Does this comment need to be updated?

Modules/Server/Sources/PocketCastsServer/Public/Refresh/MainServerHandler.swift
```diff
@@ -237,7 +237,7 @@ public class MainServerHandler {
         var jsonRequest = jsonWithStandardParams(uniqueId: uniqueId)
         jsonRequest["push_sound"].string = "11" // for legacy reasons, this is always the push sound we send, since it's no longer configurable
         jsonRequest["podcasts"].string = podcasts.map(\.uuid).joined(separator: ",")
-        jsonRequest["last_episodes"].string = podcasts.map { $0.latestEpisodeUuid ?? "" }.joined(separator: ",")
+        jsonRequest["last_episodes"].string = podcasts.map { $0.forceRefreshEpisodeFrom ?? $0.latestEpisodeUuid ?? "" }.joined(separator: ",")
```
emilylaguna, line 240: Should we reset this flag after it's used?

emilylaguna (changes requested):
Nice, this is working using my test database. I left some comments below.

pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/37">7.54</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Podcast.swift
leandroalonso: Done in 8cff3f6130778b8edf07f1181c246b9c1a9920d4

Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
leandroalonso: Very good point. Changed to be only in the podcast details here: 0baaa779392e306e1e8679ddf1d1b6702927e623

Modules/Server/Sources/PocketCastsServer/Public/Refresh/MainServerHandler.swift
leandroalonso: The `forceRefreshEpisodeFrom`? I thought we shouldn't because this is not persisted and next time the `Podcast` shouldn't have it... but maybe we should?

pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/37">7.54</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
```diff
@@ -17,16 +24,16 @@ extension ServerPodcastManager {
...
             guard let strongSelf = self else { return }
 
-            strongSelf.update(podcast: podcast, podcastInfo: podcastInfo, lastModified: lastModified)
+            strongSelf.update(podcast: podcast, podcastInfo: podcastInfo, lastModified: lastModified, addMissingEpisodes: addMissingEpisodes)
             completion?()
         }
     }
 
-    private func update(podcast: Podcast, podcastInfo: [String: Any], lastModified: String?) {
+    private func update(podcast: Podcast, podcastInfo: [String: Any], lastModified: String?, addMissingEpisodes: Bool) {
```
emilylaguna, line 36: `addMissingEpisodes` isn't checked in the body, so we're still triggering the refresh even if it's false.

pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/37">7.54</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

emilylaguna (approved):
Working in my tests! 🤞

pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/37">7.54</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

emilylaguna (approved):

Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
leandroalonso: @geekygecko just to keep you in the loop: we're merging this but it's behind a remote feature flag. The logic was changed to refresh only the necessary to add the missing episodes. Plus it's only triggered in a very specific path. If we run into any issues we can just disable the flag.

PR #1253, opened 2023-12-04
Change WatchOS minimum deployment target to 8

### LINKED ISSUE #1217: Apple Watch: Can't install on WatchOS 8.8.1
### Description

We've had users on WatchOS 8.8.1 report that they can no longer install Pocket Casts on their Watch. The issue started from 7.47.0.1. The app is listed with an 'Install' icon, but the installation fails without an error message.

Discussion about it: p1696264891612029-slack-C02A333D8LQ

1. Use an older Apple Watch which uses WatchOS 8.8.1
2. Try to install the Pocket Casts app on it

https://github.com/Automattic/pocket-casts-ios/assets/16253818/d6fbd126-56fe-4b01-9ad2-bde039ae38d9

Apple Watch Series 3, WatchOS 8.8.1 and Pocket Casts v7.52.0.2

7304140-zd-a8c

leandroalonso:
Fixes #1217

This attempts to fix the issue of the app not installing on WatchOS 8.8.1. [As per](https://developer.apple.com/forums/thread/740140):

> Thanks to a tip from the developers of another app, we found that changing the deployment target of the watch app from watchOS 7 to watchOS 8, allowed the app to be installed on devices running watchOS 8.
Since all devices that supported watchOS 7 also support watchOS 8, this was an acceptable workaround and we have now released an app update with this change.

## To test

1. Run the Pocket Casts Watch App target
2. ✅ It should run

## Checklist


emilylaguna (approved):
🤞

PR #1299, opened 2023-12-14
Issue 12 – Replace MaterialBottomSheet

### LINKED ISSUE #12: Replace MaterialBottomSheet with our own implementation
As of June 2021, Material components for iOS are now in maintenance mode. This means that we need to look at replacing our use of this with our own component (or a different third party one). Here are all the places we use it:

### Additional information

turns out iOS 15 has a built in implementation of bottom sheets - `UISheetPresentationController`. I tested it on our effects panel on the player like so:
```
private func showEffectsPanel() {
	let effectsController = EffectsViewController()

	if #available(iOS 15.0, *) {
	    if let sheet = effectsController.sheetPresentationController {
	        sheet.detents = [.medium()]
	        sheet.largestUndimmedDetentIdentifier = .medium
	        sheet.widthFollowsPreferredContentSizeWhenEdgeAttached = true
	    }
	}

	present(effectsController, animated: true, completion: nil)
}
```

And the behavior is virtually identical, maybe even nicer than the material bottom sheet. It also looks the same which is nice:

Once we switch to iOS 15+ we can change to this approach.

mylesbarros:
Fixes #12 [Replace MaterialBottomSheet with our own implementation](https://github.com/Automattic/pocket-casts-ios/issues/12)

This PR replaces usages of `MaterialComponents/BottomSheet` with Apple's [UISheetPresentationController](https://developer.apple.com/documentation/uikit/uisheetpresentationcontroller) and removes the 3rd-party dependency.

## To test

The steps for testing will be different for each sheet in the app. These are the steps to test the Sleep Timer for anyone that would like to preview the new sheet.

1. Tap a Podcast to enter the list of episodes.
2. Tap an episode from the list to start playing one of the podcast episodes.
3. Tap the "now playing" footer at the bottom to go to the fullscreen detail for the podcast episode.
4. Tap the 💤 button in the bottom footer. The Sleep Timer bottom sheet will appear in the new bottom sheet.

## Checklist

- [x] I have considered adding unit tests for my changes. (UI changes, no unit tests)

## Before & After

Note that a video is included for the Overflow Actions behavior as, when the user taps the Edit button, that sheet enlarges and becomes undismissible via swipe actions.

| **App Scene** | **Before** | **After** |
|-----------|---------|---------|
| **Effects Sheet** | effects_sheet_before | effects_sheet_after |
| **Multiselect Actions** | multiselect_sheet_before | multiselect_sheet_after |
| **Multiselect Actions (Expanded)** | multiselect_sheet_expanded_before | multiselect_sheet_expanded_after |
| **Overflow Actions** | overflow_shelf_actions_before | overflow_shelf_actions_after |
| **Overflow Actions Edit Behavior** | <video src="https://github.com/Automattic/pocket-casts-ios/assets/4069415/c1e01ce9-a90e-4fa9-95ab-ea2bfbdd8012"/> | <video src="https://github.com/Automattic/pocket-casts-ios/assets/4069415/8a9d1f96-998e-40ff-a759-66903f5b4eb8"/> |
| **SwiftUI Wrapper Example (EoY)** | presentModally(year_end_example)_sheet_before | presentModally(year_end_example)_sheet_after |
| **Profile** | profile_sheet_before | profile_sheet_after |
| **Sleep Timer** | sleep_timer_before | sleep_timer_after |


podcasts/End of Year/Stories/MDCSwiftUIWrapper.swift
```diff
@@ -81,24 +80,20 @@ class MDCSwiftUIWrapper<ContentView: View>: UIViewController {
...
-        bottomSheet.isScrimAccessibilityElement = true
-        bottomSheet.scrimAccessibilityLabel = L10n.accessibilityDismiss
-        bottomSheet.ignoreKeyboardHeight = true
-
-        viewController.present(bottomSheet, animated: true, completion: nil)
+    func presentModally(in viewController: UIViewController, detents: [UISheetPresentationController.Detent] = [.medium()]) {
+
+        if let sheetController = self.sheetPresentationController {
+            sheetController.detents = detents
+            sheetController.prefersGrabberVisible = false
```
SergioEstevao: It will be nice to have the option to show the grabber as an argument to the method defaulting to false.

mylesbarros: Good idea, added a `showingGrabber` argument that defaults to false and added a comment explaining why false is the default 🙂

SergioEstevao:
@mylesbarros do you mind merging with the latest changes from trunk. Just to have a more updated baseline before I review the PR.

david-gonzalez-a8c:
Thanks, @mylesbarros, what a great change!

This happened with the old component too, but watching all the screens together I'm noticing the different bottom margins (e.g., there's a lot of white space in the playback controls sheet and very little in the Playback sheet), so I'm wondering if we can fix that with the new component and make these more consistent.

Some small amount of white space, similar to what we had in the old Playback sheet looks ideal to me. What do you think?

image

SergioEstevao:
On the iPad there is some fundamental design differences between the way MBS shows things ( always using a bottom sheet) and the way the system shows bottom sheet.
As you can see bellow system converts the bottom sheet to a popover when on the .regular horizontal size class

| MBS | System |
| - | - |
| sleep_timer_mbs | sleep_timer_system |
| playback_timer_mbs |  playback_timer_system |
|  more_actions_mbs | more_actions_system |

@david-gonzalez-a8c what do you think about the above ^

I also found a bug on the width of the playback effect setting  when running on iOS 15

david-gonzalez-a8c:
> @david-gonzalez-a8c what do you think about the above ^

I believe this is what the OS does normally, right? In that case it's a good change in my opinion.

mylesbarros:
Hey @david-gonzalez-a8c thanks for the praise!

I agree, the bottom padding could use improvement. Unfortunately there isn't a good way to ensure we use the same padding for each sheet as we don't have the size of the `UIView` until after it renders due to our use of constraints.

What I did instead then was use custom detents to increase the sizing for the playback effects, sleep timer, and multiselect actions. Where we really need the extra space is in the `BottomSheetSwiftUIWrapper` that was used to display the 2023 playback. I didn't make any changes there as we no longer have the playback UI to test it but if you're confident in it I can add a bottom padding to the wrapper itself to provide the spacing.

Here's an updated comparison, let me know if you think it's enough

| **Device** | **Before** | **After (v2)** |
|-----------|---------|---------|
| **iPhone 15** | 290460850-396c00b9-0268-4d74-8214-fbcbc3e78a39 | playback_effects-v2 |
| **iPhone 15** |  290461117-10c6a538-da90-426c-b5ba-94bad152c30c | multiselect_episodes_v2 |
| **iPhone 15** |  290463953-b1095b5e-c70c-46a3-922a-3b6807ab7754 | sleep_timer-v2 |

david-gonzalez-a8c:
Looks good to me! Thanks for the changes.

podcasts/NowPlayingPlayerItemViewController+Shelf.swift
```diff
@@ -333,16 +346,39 @@ extension NowPlayingPlayerItemViewController: NowPlayingActionsDelegate {
...
-        roundCorners(bottomSheet: bottomSheet)
-        present(bottomSheet, animated: true, completion: nil)
+
+        if let sheetController = effectsController.sheetPresentationController {
+            if #available(iOS 16.0, *) {
+                let hostingControllerHeight = view.bounds.height
+                let sheetDetentHeight = hostingControllerHeight * 0.5
+                sheetController.detents = [.custom(resolver: { _ in sheetDetentHeight })]
+            } else {
+                sheetController.detents = [.medium()]
```
SergioEstevao, line 374: As you can see from the screenshot above  for smaller devices, like the iPhone SE, we need to adapt the height to fit all the content.

This happens because unlike the MDC wrapper that sizes the bottomsheet using the VC preferred  size the system UI uses a static size that does not adapt to the size.


We can do the following for smaller devices (width is smaller than 375pt)
- iOS 16 and above: increase the height ratio to `0.7`
- IOS 15: use the `[.large]` detent

SergioEstevao: Another issue that we can see in iOS 15 is that the Effects does not fit the full width of the BottomSheet.



This again was that MDC adapted the width of the bottom sheet to the preferred size of the VC, but the system BottomSheet doesn't.

This can be fixed by adding the following line on `viewDidLoad` method  of EffectsViewController:

`view.translatesAutoresizingMaskIntoConstraints = false`

SergioEstevao (changes requested):
@mylesbarros thanks for your patience and work on this! We are almost there, please check my comments about the EffectsViewController.

When you have those fixed I will give it another check.

SergioEstevao:
@mylesbarros are you still working on this?

SergioEstevao:
This PR was take over by this one: https://github.com/Automattic/pocket-casts-ios/pull/1472.

Thanks for your contribution @mylesbarros

PR #1343, opened 2024-01-11
Update widget design

jgcaruso:
Updates to widgets based on new designs Lkt7fuf9Nq3XvfAFPCfpTD-fi-995_4119

Also introduces a `PCWidgetColorScheme` struct to store different collections of colours to be passed to the different versions of the widgets to support a new style and a "contrast" style that is more like the current widgets.

cc @david-gonzalez-a8c 

### The widget chooser

https://github.com/Automattic/pocket-casts-ios/assets/4081020/41fcc829-876a-481f-beee-6c8494270b90

### iPhone 13 mini in dark mode (widget doesn't change based on light/dark mode)
| new large | new small / medium | old/contrast large | old/contrast small / medium |
| --- | --- | --- | --- |
| IMG_0004 | IMG_0003 | IMG_0002 | IMG_0001 |

## iPhone 15
| default text small / medium | default text large | max text size small/medium | max text size large |
| --- | --- | --- | --- |
| Simulator Screenshot - iPhone 15 - 2024-02-08 at 16 01 09 | Simulator Screenshot - iPhone 15 - 2024-02-08 at 16 00 49 | Simulator Screenshot - iPhone 15 - 2024-02-08 at 15 22 23 | Simulator Screenshot - iPhone 15 - 2024-02-08 at 15 21 47 |

## iPhone 15 - other widgets
| empty state | empty state | filter widgets new style | filter widgets contrast |
| --- | --- | --- | --- |
| Simulator Screenshot - iPhone 15 - 2024-01-11 at 16 12 03 | Simulator Screenshot - iPhone 15 - 2024-01-11 at 16 12 06 | Simulator Screenshot - iPhone 15 - 2024-02-08 at 16 06 36 | Simulator Screenshot - iPhone 15 - 2024-02-08 at 16 06 49 |

## Desktop widgets

SCR-20240208-nwyd

## ios 16 non-interactive 
| small / medium | large | now playing (empty) and filter | large filter |
| --- | --- | --- | --- |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-02-08 at 16 12 35 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-02-08 at 16 12 30 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-02-08 at 16 13 34 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-02-08 at 16 13 38 |

### Test on an iOS 17 device (for interactive widgets) and a lower iOS version (for non interactive widgets)

1. View all of the widgets in the widget viewer, they should be ordered smallest to largest, new style design first and "contrast" styles second
2. add all widget types to your homescreen
3. Ensure you have a bunch of podcasts in your "up next" queue
4. All widget sizes should match the designs
5. For iOS 17 interactive widgets, try playing the podcast with the play buttons, they should start and update instantly
6. Clicking anywhere else on a row should open the podcast in the app (all iOS versions)
7. clear out your now playing list, leave 1 episode in "now playing"
8. widgets should match the designs
9. Remove the currently playing episode
10. Go to your filters and make sure that the top filter has some episodes in it
11. view the widgets again, they should look pleasing (but not exactly like designs, these weren't designed i just applied some common colours to them because there is too much shared code between the widgets to leave them the way they were)
12. Clear episodes from the top filter
13. View the widgets again, they should look pleasing and show the empty widget state

### Test on an iOS 17 device, linked to a MacOS Sonoma mac for Desktop widgets
- ensure your mac and test phone are on the same icloud account
- if you have multiple phones you can choose which phone is used for widgets in `System Settings ➝ Desktop & Dock`, and in the "‌Widgets‌" section, click the ‌iPhone‌ dropdown
- add widgets to your desktop by right-clicking on the desktop -> Edit Widgets, search for Pocket Casts then add the widgets

### Update experience
1. remove all pocket casts widgets
2. return to trunk, build and run
3. add all pocket casts widgets to your home screen
4. switch back to this branch, build and run
5. all widgets should still be on the home screen and should be using the "contrast" style, not the new style

## Checklist


pocketcasts:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso:
> widget doesn't change based on light/dark mode

Is this intentional? Showing white while the rest of the system is dark is pretty bad IMHO. Also, if we look at all Apple widgets (such as Calendar, News, Notes) they have a white background if light mode is set and a dark background if dark mode is set. I think this is a pattern we should follow. @david-gonzalez-a8c

leandroalonso:
@jgcaruso it seems that on iOS 16- devices the color of the subtitle on the red widget is the wrong color (check the "now playing" on the first widget vs the 3rd):

Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2024-02-13 at 14 04 07

david-gonzalez-a8c:
> I think this is a pattern we should follow. @david-gonzalez-a8c

Good catch, I agree. Honestly, that didn't cross my mind. I've updated with the [specs](https://www.figma.com/file/Lkt7fuf9Nq3XvfAFPCfpTD/Widgets?type=design&node-id=1094%3A2337&mode=design&t=jE605cpXVsZklGkQ-1) with a dark-theme version of the white background widgets, as I believe it would only affect those.

image

It would be great if we can implement these too 🙏

WidgetExtension/Now Playing/NowPlayingEntryView.swift
```diff
@@ -7,65 +7,79 @@ struct NowPlayingWidgetEntryView: View {
 
     @Environment(\.showsWidgetContainerBackground) var showsWidgetBackground
 
+    @Environment(\.widgetColorScheme) var colorScheme
+
     var body: some View {
```
bjtitus, line 19: I think this would be more readable broken into either separate `View`s or several `ViewBuilder` functions. Maybe an `Artwork`, `EpisodeTitle`, and `PlayingToggle`?

jgcaruso:
@leandroalonso  Latest commits have a fix for the iOS 16 text colour in the Now Playing widget and a return of dark mode for the "Contrast" style widgets

cc @david-gonzalez-a8c 

## iOS 16
| light | dark |
| -- | -- |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-03-28 at 15 32 34 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-03-28 at 15 32 48 |

## iOS 17
| light | dark |
| -- | -- |
| Simulator Screenshot - iPhone 15 - 2024-03-28 at 15 38 03 | Simulator Screenshot - iPhone 15 - 2024-03-28 at 15 37 50 |

david-gonzalez-a8c:
We've been getting a lot of harsh feedback about the changes in the new Android widgets, specially regarding the red backgrounds, so I think we should tread very carefully with these. I'm less concerned here since we provide a more neutral option and people with the current widget will get those, but now I'm reconsidering the dark theme L version. 

Can we please make adjustments and replace the red (#D9201C) for this dark grey (#292B2E) in this version only? I feel like it looks better now.

image

Sorry if I'm missing something or if we've discussed this already, but I see that the M version doesn't match the specs.

image

Thank you!

jgcaruso:
Hey @david-gonzalez-a8c 👋 

> Can we please make adjustments and replace the red (#D9201C) for this dark grey (#292B2E) in this version only? I feel like it looks better now.

That should be pretty quick to take care of. I'll try to swap those out in the next day or 2.

> Sorry if I'm missing something or if we've discussed this already, but I see that the M version doesn't match the specs.

I think I left this out for now because it was a brand new layout, and I wanted to make sure all the other big changes being made were OK before doing something new.

While we're on the subject though, I don't remember if it was decided if this new layout should replace the existing 2 podcast list, or if it should be a new one added in addition to it? We can have both of them if you want to keep it around. One will just be a Medium "now playing" vs a Medium "up next" widget.

jgcaruso:
I've updated the PR with the new grey colour instead of red for the dark mode widgets

| large | medium / small |
| -- | -- |
| Simulator Screenshot - iPhone 15 Plus - 2024-04-24 at 17 49 59 | Simulator Screenshot - iPhone 15 Plus - 2024-04-24 at 17 50 04 |

david-gonzalez-a8c:
> I think I left this out for now because it was a brand new layout, and I wanted to make sure all the other big changes being made were OK before doing something new.

Given the feedback on Android when we replaced people's existing widgets, it would be great having both so there's less change.

leandroalonso:
@jgcaruso given this:

> Given the feedback on Android when we replaced people's existing widgets, it would be great having both so there's less change.

Is this something you'd like to tackle or should we? Just want to know before we start reviewing that.

jgcaruso:
> Is this something you'd like to tackle or should we? Just want to know before we start reviewing that.

I'm OK to add the new medium widget type, it shouldn't be too big of a job. I feel like it should be something I can follow up with quickly after this. But I'd still like to do it after this PR is wrapped up incase there are any fundamental changes requested in this PR if that makes sense?

leandroalonso:
> But I'd still like to do it after this PR is wrapped up incase there are any fundamental changes requested in this PR if that makes sense?

Ok!

leandroalonso:
@jgcaruso just one thing that popped into my head and I think you probably thought about it too: for people already using the widgets, we should probably keep the previous appearance. 

And, of course, have the new one as an option.

I think this should even be one of the testing steps: upgrade and ensure your widgets' appearance is similar to the one before.

Do you think the code will change much after doing these changes? Or it should be ok to review this one?

WidgetExtension/Up Next/EpisodeView.swift
```diff
@@ -62,27 +55,57 @@ struct EpisodeView: View {
...
+                    .foregroundStyle(colorScheme.topButtonTextColor)
                 :
-                    Image("icon-play")
+                Image("icon-play")
                     .resizable()
-                    .foregroundStyle(.black)
+                    .foregroundStyle(colorScheme.topButtonTextColor)
+            }
+            .frame(width: 28, height: 28)
+            // TODO: Something fun - create a timeline that counts down by the minute instead of showing "now playing"
```
jgcaruso: I just noticed I left this here... it is more of a "this would be fun if we could do this" activity, but not something that can easily be done right now.

Comment should probably be removed, but I'll leave it here for the moment incase anyone has thoughts on it.

jgcaruso:
> @jgcaruso just one thing that popped into my head and I think you probably thought about it too: for people already using the widgets, we should probably keep the previous appearance.
> 
> And, of course, have the new one as an option.
> 
> I think this should even be one of the testing steps: upgrade and ensure your widgets' appearance is similar to the one before.
> 
> Do you think the code will change much after doing these changes? Or it should be ok to review this one?

@leandroalonso  So the way we thought to handle this was to give them the "contrast" theme, which is very close to the original, just with some Red detail on top rows vs Grey (for the lists). It is very close but not exactly the same. The new "2 tone" red variation is something you need to choose on purpose.

If you want to re-introduce the original widget styles, and just have the 2 new ones as options... I think we'd need to then provide all 3 as options in the widget chooser (instead of leaving the current style as the default, but not let new users select it). Unless there is some technique you have in the app for excluding certain things for users who are new vs legacy past a certain version?

As far as how much the code would change... it would just be a couple new widget entries in `WidgetExtension/PocketCastsWidgetBundle.swift`, a new `Widget[xyz].swift` file for each of the sizes (so 3 I think) and 2 new styles in `WidgetExtension/Up Next/UpNextStyles.swift` (for light and dark mode). It shouldn't fundamentally change this PR though, just add some code for the new styles. So you should be able to review this as-is and I can add it in later, if that is what you are asking.

leandroalonso:
> It is very close but not exactly the same.

That's what I was mentioning!

> If you want to re-introduce the original widget styles

No no no, that's not my intention.

I saw that this PR already includes the "contrast" version so it's ready. Sorry for the back and forth!

WidgetExtension/Up Next/UpNextStyles.swift
```diff
@@ -4,6 +4,128 @@ import SwiftUI
...
+    let filterViewIconAssetName: String
+}
+
+let rgbMax = 255.0
+let widgetRedDark = Color(.sRGB, red: 217/rgbMax, green: 32/rgbMax, blue: 28/rgbMax)
+let widgetRedLight = Color(.sRGB, red: 244/rgbMax, green: 62/rgbMax, blue: 55/rgbMax)
+let widgetBlack = Color(.sRGB, red: 22/rgbMax, green: 23/rgbMax, blue: 24/rgbMax)
+let widgetCoolGrey = Color(.sRGB, red: 41/rgbMax, green: 43/rgbMax, blue: 46/rgbMax)
+
+let widgetColorSchemeBold = PCWidgetColorScheme(
```
bjtitus: Could this be a `static` property on `PCWidgetColorScheme`? Then we could call with `PCWidgetColorScheme.bold`.

WidgetExtension/Up Next/UpNextStyles.swift
```diff
@@ -4,6 +4,128 @@ import SwiftUI
...
+    topBackgroundColor: widgetRedLight,
+    bottomBackgroundColor: widgetRedDark,
+    topButtonBackgroundColor: .white,
+    bottomButtonBackgroundColor: Color(.sRGB, red: 1, green: 1, blue: 1, opacity: 0.2),
```
bjtitus: Can `Color.white.opacity(0.2)` be used here?

WidgetExtension/Up Next/UpNextStyles.swift
```diff
@@ -4,6 +4,128 @@ import SwiftUI
...
 let lightBackgroundColor = Color(.clear)
 
+struct PCWidgetColorScheme {
+    let topBackgroundColor: Color
+    let bottomBackgroundColor: Color
+    let topButtonBackgroundColor: Color
+    let bottomButtonBackgroundColor: Color
+    let topTextColor: Color
+    let bottomTextColor: Color
+    let topButtonTextColor: Color // foreground color?
```
bjtitus: Can the comments or question marks be removed here? I guess I'm not sure whether these are notes or actually describe the colors in some way.

WidgetExtension/Up Next/EpisodeView.swift
```diff
@@ -62,27 +55,57 @@ struct EpisodeView: View {
...
+            RoundedRectangle(cornerRadius: 100)
+                .foregroundColor(colorScheme.topButtonBackgroundColor)
+        )
+    }
+}
+
+struct WidgetPlayToggleStyle: ToggleStyle {
+    let colorScheme: PCWidgetColorScheme
+
+    func makeBody(configuration: Configuration) -> some View {
```
bjtitus, line 91: Should this be marked `@ViewBuilder` in case the view is not a single View? I think the protocol method is marked ViewBuilder, so maybe that's automatically applied?

WidgetExtension/Up Next/EpisodeView.swift
```diff
@@ -62,27 +55,57 @@ struct EpisodeView: View {
...
-        ZStack {
-            configuration.label
-                .truncationMode(.tail)
+struct WidgetFirstEpisodePlayToggleStyle: ToggleStyle {
+    let colorScheme: PCWidgetColorScheme
 
-            Circle()
-                .foregroundStyle(.white)
-                .frame(width: 24, height: 24)
+    func makeBody(configuration: Configuration) -> some View {
```
bjtitus, line 61: Should this be marked @ViewBuilder in case the view is not a single View? I think the protocol method is marked ViewBuilder, so maybe that's automatically applied?

WidgetExtension/Up Next/UpNextStyles.swift
```diff
@@ -4,6 +4,128 @@ import SwiftUI
...
+    let bottomButtonTextColor: Color // foreground color?
+    let iconAssetName: String
+    let filterViewBackgroundColor: Color
+    let filterViewTextColor: Color
+    let filterViewIconAssetName: String
+}
+
+let rgbMax = 255.0
```
bjtitus: I wonder if this could be added to an extension initializer on `Color`? It would make this a little more self contained and avoid polluting the `WidgetExtension` with this value.

```swift
extension Color {
    /// A convenience initializer that takes RGB values 0-255 and converts to the 0-1 range expected by Color.
    /// - Parameters:
    ///   - red: A value 0-255 for the red component
    ///   - green: A value 0-255 for the green component
    ///   - blue: A value 0-255 for the blue component
    init(red: Int, green: Int, blue: Int) {
        let rgbMax = 255
        Color(.sRGB, red: red/rgbMax, green: green/rgbMax, blue: blue/rgbMax)
    }
}
```

bjtitus (changes requested):
Thanks for this, @jgcaruso!
I left a few small cleanup comments for code review and I'll test the widgets out more tomorrow!

WidgetExtension/Up Next/EpisodeView.swift
SergioEstevao: Instead of the TODO comment add an issue on the repo as an improvement/suggestion to do in the future.

WidgetExtension/Up Next/EpisodeView.swift
```diff
@@ -62,27 +55,57 @@ struct EpisodeView: View {
...
+
+struct WidgetPlayToggleStyle: ToggleStyle {
+    let colorScheme: PCWidgetColorScheme
+
+    func makeBody(configuration: Configuration) -> some View {
+        HStack {
+            ZStack {
+                Circle()
+                    .foregroundStyle(colorScheme.bottomButtonBackgroundColor)
+                    .frame(width: 24, height: 24)
```
SergioEstevao, line 96: Could this 24 be refactored to an enum constant?

WidgetExtension/Up Next/UpNextStyles.swift
```diff
@@ -4,6 +4,128 @@ import SwiftUI
 let darkBackgroundColor = Color(UIColor.tertiarySystemFill)
 let lightBackgroundColor = Color(.clear)
 
+struct PCWidgetColorScheme {
```
SergioEstevao: Maybe it will be better to have a separate file for  `PCWidgetColorScheme`

SergioEstevao:
Tested this in the iPad and iOS using iOS 17 and its working correctly!

https://github.com/Automattic/pocket-casts-ios/assets/651601/95e154ca-8f2a-4833-b2f0-d0b53e128c35

SergioEstevao:
I've also tested the Sonoma/iPhone integration and they are also looking good.
One thing that I noticed is that sometimes the widgets can get a bit out of sync from the phone status, specially if you change the upnext or playing status on the phone it can take a while to update on MacOS. Is this expected?

Screenshot 2024-06-11 at 13 13 39

SergioEstevao (changes requested):
In general is working correctly, just left some code reviews and a comment about MacOS widgets syncronization.

WidgetExtension/Up Next/UpNextStyles.swift
jgcaruso: Absolutely, I like that suggestion. Also makes using it way less verbose since I can just use `.bold` in places where a PCWidgetColorScheme is expected.

WidgetExtension/Up Next/UpNextStyles.swift
jgcaruso: I think that was a question for myself... if I should rename `TextColor` to `ForegroundColor` 😅 I'll remove the comment.

WidgetExtension/Up Next/EpisodeView.swift
jgcaruso: I think its fine to not mark it because as you said the protocol method is marked `@ViewBuilder` (among other things). The sample code in the documentation also doesn't specify it https://developer.apple.com/documentation/swiftui/togglestyle/makebody(configuration:)

WidgetExtension/Up Next/EpisodeView.swift
jgcaruso: same as above (tldr; I _**think**_ its fine)

WidgetExtension/Now Playing/NowPlayingEntryView.swift
jgcaruso: I went with a couple of functions, just to break up the large wall of swiftui.

If the `playToggleOrPlaybackLabel` isn't granular enough, I'm happy to break it down further.

WidgetExtension/Up Next/EpisodeView.swift
jgcaruso: Do you mean just for these 2 or for all the random magic numbers in these new views?

A lot of these were here from before, but maybe it is worth going over them and properly naming them?

jgcaruso:
> I've also tested the Sonoma/iPhone integration and they are also looking good. One thing that I noticed is that sometimes the widgets can get a bit out of sync from the phone status, specially if you change the upnext or playing status on the phone it can take a while to update on MacOS. Is this expected?

So I'm not sure it is expected, but I'm also not sure we're doing anything special in order to get them to appear there. It is just leveraging whatever iOS already does to make widgets available on desktop.

WidgetExtension/Common/PCWidgetColorScheme.swift
```diff
@@ -0,0 +1,122 @@
+import SwiftUI
+
+let widgetRedDark = Color(red: 217, green: 32, blue: 28)
```
SergioEstevao: Instead of creating global constants you can do the following:

```
Extension Color {
  static let widgetRedDark = Color(red: 217, green: 32, blue: 28)
}
```
And then in any place you need it you just do: `.widgetRedDark`

leandroalonso:
> One thing that I noticed is that sometimes the widgets can get a bit out of sync from the phone status, specially if you change the upnext or playing status on the phone it can take a while to update on MacOS. Is this expected?

I faced the same while testing with the old widget code. As @jgcaruso mentioned I don't think we do anything special...

WidgetExtension/Common/PCWidgetColorScheme.swift
jgcaruso: Done in [36762aa](https://github.com/Automattic/pocket-casts-ios/pull/1343/commits/36762aa353bdc94b9a933004dafaf6cff274899c)

SergioEstevao (approved):
From my checks on iOS17 it's all looking and code is now updated with my suggestions.
So this is a green for me, but lets wait for @bjtitus and @leandroalonso to approve too.

leandroalonso:
I'm gonna test this a bit more but I propose we leave that for 7.68 given 7.67 has already some important changes.

leandroalonso (approved):
All good from my side.

@bjtitus would you mind taking a final look?

@jgcaruso could you please waiting to merge this one so it goes to 7.68?

jgcaruso:
> @jgcaruso could you please waiting to merge this one so it goes to 7.68?

Sure thing! How will I know it is safe to merge?

Alternatively, feel free to merge this in when appropriate if there are no other requested changes. I'm not planning any additional changes at this point. Only thing planned at this point is a separate PR once this is merged in for the new medium size "Now Playing" widget

bjtitus (approved):
Looks good 👍🏻 

Thank you so much for all of your hard work and being receptive of our feedback @jgcaruso! 🙇

bjtitus:
I just merged this for the 7.68 release now that 7.67 has been frozen.

leandroalonso:
@bjtitus I believe we should add this to the `CHANGELOG`.

bjtitus:
@leandroalonso 👍 Submitted in https://github.com/Automattic/pocket-casts-ios/pull/1864

PR #1349, opened 2024-01-15
Fix tap on tab bar to display search

### LINKED ISSUE #1345: Second tap on tab bar doesn't focus on search
On iOS there's a somewhat common pattern to bring focus on search when tapping the tab bar.

Eg.: if you are already on Discover, another tap on Discover would bring the focus to the search (displaying keyboard, etc).

This used to work but I believe after iOS 17 it stopped.

#7565431-zen

leandroalonso:
Fixes #1345

Due to a change on iOS, the tap on the tab bar to display search stopped working on iPhone.

This happened because the value returned by `view.safeAreaInsets.top` is slightly different than before: `97.66666666666667` instead of `97.66666666666666`. This makes the `if` never succeed thus never showing the search. I just rounded the numbers to avoid this issue.

I'm targeting the frozen branch since it's a small fix.

## To test

1. Run the app on a device/simulator with the Dynamic Island
2. Go to DIscover
3. Scroll down
4. Tap "Discover" on the tab bar
5. ✅ Discover section should scroll to the top
6. Tap it again
7. ✅ The Discover search should appear
8. Dismiss it and go to "Podcasts"
9. Scroll down
10. Tap "Podcasts" on the tab bar
11. ✅ Podcasts will scroll to the top with the search bar visible
12. Tap "Podcasts" again
13. ✅ Podcasts search should appear

Repeat these tests on iPhone SE and iPad.

## Checklist

### DIFF
--- a/podcasts/DiscoverViewController.swift
+++ b/podcasts/DiscoverViewController.swift
@@ -83,7 +83,7 @@ class DiscoverViewController: PCViewController {
         guard let index = notification.object as? Int, index == tabBarItem.tag else { return }
 
         let defaultOffset = -PCSearchBarController.defaultHeight - view.safeAreaInsets.top
-        if mainScrollView.contentOffset.y.rounded() > defaultOffset.rounded() {
+        if mainScrollView.contentOffset.y.rounded(.down) > defaultOffset.rounded(.down) {
             mainScrollView.setContentOffset(CGPoint(x: 0, y: defaultOffset), animated: true)
         } else {
             searchController.searchTextField.becomeFirstResponder()
--- a/podcasts/DiscoverViewController.swift
+++ b/podcasts/DiscoverViewController.swift
@@ -83,7 +83,7 @@ class DiscoverViewController: PCViewController {
         guard let index = notification.object as? Int, index == tabBarItem.tag else { return }
 
         let defaultOffset = -PCSearchBarController.defaultHeight - view.safeAreaInsets.top
-        if mainScrollView.contentOffset.y > defaultOffset {
+        if mainScrollView.contentOffset.y.rounded() > defaultOffset.rounded() {
             mainScrollView.setContentOffset(CGPoint(x: 0, y: defaultOffset), animated: true)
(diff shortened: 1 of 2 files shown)


podcasts/DiscoverViewController.swift
```diff
@@ -83,7 +83,7 @@ class DiscoverViewController: PCViewController {
         guard let index = notification.object as? Int, index == tabBarItem.tag else { return }
 
         let defaultOffset = -PCSearchBarController.defaultHeight - view.safeAreaInsets.top
-        if mainScrollView.contentOffset.y > defaultOffset {
+        if mainScrollView.contentOffset.y.rounded() > defaultOffset.rounded() {
```
bjtitus: My first reaction was that we simply don't care about the fractional value, so maybe rounding [`down`](https://developer.apple.com/documentation/swift/floatingpointroundingrule/down) or simply casting to `Int` would make more sense?

In using `rounded` here, I believe the default rounding method is [toNearestOrAwayFromAero](https://developer.apple.com/documentation/swift/floatingpointroundingrule/tonearestorawayfromzero) and I just wonder if we would ever run into a case where a > `.5` value might be compared with < `.4` and we really don't care about that.

Here are the examples:
```
(5.2).rounded(.toNearestOrAwayFromZero)
// 5.0
(5.5).rounded(.toNearestOrAwayFromZero)
// 6.0
(-5.2).rounded(.toNearestOrAwayFromZero)
// -5.0
(-5.5).rounded(.toNearestOrAwayFromZero)
// -6.0
```

I think this is probably very nitpicky so take it or leave it – just my initial reaction.

bjtitus (approved):
Looked good in testing!. Had one probably very nitpicky comment about the `rounded` use.

podcasts/DiscoverViewController.swift
leandroalonso: That's a good point! Changed in dfe88068e62e6b09290f9bde91cd5ab8de974ee9

PR #1326, opened 2024-01-04
Move Danger to GitHub Actions

iangmaia:
This PR moves the current Buildkite step setup to GitHub Actions (as discussed on paaHJt-5Qn-p2).

Running Danger on Buildkite is still an option, but it would require a GitHub action (or an intermediate server) to forward to Buildkite Pull Request Events such as `labeled`, `unlabeled`, `milestoned` and so on. This was implemented as a prototype on https://github.com/woocommerce/woocommerce-android/pull/10385.

If we find problems using GitHub Actions, we can always revert this change and go back to Buildkite, as the setup works in general (just not as fast and as simple as GHA alone).

## SwiftLint
In addition to moving Danger to GHA, this PR changes the SwiftLint setup: it's running it as an independent step in Buildkite.
I discussed some of the issues involved [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/22408#issuecomment-1896238267), but the main point for the change being that SwiftLint was in fact failing in PocketCasts iOS for some time and there isn't a practical container solution that would allow us to run SwiftLint via Danger, without resorting to `mac` VMs (expensive and not very fast on GHA).

Any SwiftLint violation will also be reported in Buildkite with an Annotation:
Screenshot 2024-01-19 at 16 40 28

This is a follow up PR with additional SwiftLint updates: https://github.com/Automattic/pocket-casts-ios/pull/1361

## How to test
Make sure CI is 🟢 and Danger runs on labels/milestones/reviewers changes as well as on code changes.

PR #1361, opened 2024-01-18
SwiftLint setup update and code fixes

iangmaia:
This PR builds on https://github.com/Automattic/pocket-casts-ios/pull/1326 and implements the following SwiftLint setup changes:

- Makes any SwiftLint violation to fail the build (`--strict` SwiftLint option)
- Upgraded SwiftLint Pod to align with the one used on CI -- I assume developers use it locally, otherwise we can get rid of it
- Removes all `disabled_rules` so that we don't have additional exceptions
- Fixes all SwiftLint code violations found at this point
- **UPDATE**: added a Build Phase to run SwiftLint locally, reporting warnings to Xcode

## To test

Any SwiftLint violation will now fail the build. I've committed d1c664597d7e14d49206d5d75f2c9fa2fe0250bb (and reverted it) so that a failure would be reported:
Screenshot 2024-01-18 at 15 35 52

The code fixes were simple enough and there shouldn't be any problem, but it would be nice to check the UI of some of the affected views using the test build.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -21,7 +21,7 @@ steps:
   - group: "Linters"
     steps:
       - label: ":swift: SwiftLint"
-        command: .buildkite/commands/run-swiftlint.sh
+        command: .buildkite/commands/run-swiftlint.sh --strict
         plugins:
           - docker#v5.8.0:
               image: "ghcr.io/realm/swiftlint:0.54.0"
--- a/podcasts/DateHeadingView.swift
+++ b/podcasts/DateHeadingView.swift
@@ -25,7 +25,7 @@ class DateHeadingView: UIView {
 
         titleLabel = ThemeableLabel(frame: CGRect(x: 20, y: 10, width: bounds.width - 20, height: 30))
         titleLabel?.font = UIFont.systemFont(ofSize: 22, weight: UIFont.Weight.bold)
-        titleLabel?.textAlignment = .left
+        titleLabel?.textAlignment = .natural
         titleLabel?.text = title
         titleLabel?.translatesAutoresizingMaskIntoConstraints = false
         addSubview(titleLabel!)
--- a/podcasts/WhatsNewPageView.swift
+++ b/podcasts/WhatsNewPageView.swift
@@ -91,11 +91,11 @@ class WhatsNewPageView: ThemeableView {
                     let button = WhatsNewLinkButton(url: url)
                     button.setTitle(Bundle.main.localizedString(forKey: text, value: text, table: nil), for: .normal)
                     button.titleLabel?.font = UIFont(name: "SFProText-Medium", size: 18)
-                    button.titleLabel?.textAlignment = .left
+                    button.titleLabel?.textAlignment = .natural
                     button.buttonStyle = .primaryUi01
                     button.textStyle = .primaryInteractive01
                     button.contentMode = .center
-                    button.contentHorizontalAlignment = .left
+                    button.contentHorizontalAlignment = .leading
                     button.titleLabel?.adjustsFontSizeToFitWidth = true
                     button.translatesAutoresizingMaskIntoConstraints = false
                     stackView.addArrangedSubview(button)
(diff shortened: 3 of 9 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/pipeline.yml
```diff
@@ -21,7 +21,7 @@ steps:
   - group: "Linters"
     steps:
       - label: ":swift: SwiftLint"
-        command: .buildkite/commands/run-swiftlint.sh
+        command: .buildkite/commands/run-swiftlint.sh --strict
```
AliSoftware, line 24: I know that the PR description says that this is intentional… but this will make BOTH errors and warnings be reported as errors (like the `TREAT_WARNINGS_AS_ERRORS` build setting in Xcode).

While I love the commitment of adhering to the zero-warning policy, I'm just wondering if that's something that devs have agreed on?

For example:
 - Some people find this too strict, especially if they're working on a feature iteratively, using multiple PRs all targeting the feature branch to shape the code one step at a time, and would want to allow warnings in the intermediate PR, and only get rid of the warnings and clean up things on the last steps, etc.
 - Other people prefer this strict mode to not let the tech dept slip.
 - A compromise could be to only enforce `--strict` if the PR is targeting `trunk` or `release/*`, but still allow PRs with swiftlint warnings to land if they're targeting an intermediate feature branch, for example 🤔

podcasts/DateHeadingView.swift
```diff
@@ -25,7 +25,7 @@ class DateHeadingView: UIView {
 
         titleLabel = ThemeableLabel(frame: CGRect(x: 20, y: 10, width: bounds.width - 20, height: 30))
         titleLabel?.font = UIFont.systemFont(ofSize: 22, weight: UIFont.Weight.bold)
-        titleLabel?.textAlignment = .left
+        titleLabel?.textAlignment = .natural
```
AliSoftware, line 28: Someone should probably test this view using an RTL language in a simulator and confirm that this change doesn't conflict with other aspects of the layout of screens where this view is used.

In theory, this is indeed the right thing to use, so that it's adaptive to the LTR vs RTL layout. But if other parts of the screen layout—like manual frame computations, alignment of the date text with other parts of the screen…—were based on the supposition that this was always left-aligned for example, that could benefit from adjusting accordingly if needed.

podcasts/WhatsNewPageView.swift
AliSoftware, line 1: Same remark for this whole view as above:
 - In theory this is totally the right thing to do, so as to support RTL locales as well as LTR
 - In practice, this needs to be tested in a simulator, on the screens this view is used on, to confirm that this change doesn't create unintended design or layout changes with the rest of the screen (of if it does, fix the parent layout to be flexible and support RTL too)

AliSoftware (commented):
Approving in principle, but we need someone with knowledge of the app (and where those two views are used) to do a quick smoke test of the app on device (or simulator) in any RTL language to validate the layout still look right.

.buildkite/pipeline.yml
iangmaia: > While I love the commitment of adhering to the zero-warning policy, I'm just wondering if that's something that devs have agreed on?

Given we're making changes to the SwiftLint setup and this is also how I'd personally prefer it set up, I used this PR also as a way to raise the discussion of a zero-warning policy with the team. Thanks for highlighting it so that's clearer!
I do understand that perhaps this isn't for all teams and that's why I also liked your suggestion:

> A compromise could be to only enforce --strict if the PR is targeting trunk or release/*, but still allow PRs with swiftlint warnings to land if they're targeting an intermediate feature branch, for example 

I can see some teams preferring it that way, that's a great suggestion!

AliSoftware: What about a RFC P2? Otherwise not sure the whole PCiOS team will see this PR and thus consider it as a kind of poll on the question… 🤔

iangmaia: > What about a RFC P2? Otherwise not sure the whole PCiOS team will see this PR and thus consider it as a kind of poll on the question… 🤔

An RFC P2 is a good idea 👍 my initial point was just having a pair of eyes from the team in the PR, not exactly a poll. But I agree it's good also to bring awareness and clarity to the entire team. Thanks!

mokagio: The RFC was posted at pdeCcb-4qy-p2

podcasts/DateHeadingView.swift
leandroalonso: Just tested that and it's working just fine. :)

podcasts/WhatsNewPageView.swift
```diff
@@ -91,11 +91,11 @@ class WhatsNewPageView: ThemeableView {
                     let button = WhatsNewLinkButton(url: url)
                     button.setTitle(Bundle.main.localizedString(forKey: text, value: text, table: nil), for: .normal)
                     button.titleLabel?.font = UIFont(name: "SFProText-Medium", size: 18)
-                    button.titleLabel?.textAlignment = .left
+                    button.titleLabel?.textAlignment = .natural
```
leandroalonso, line 94: This is not used anymore so these changes are ok. I wrote it down so we can remove this in the future.

leandroalonso (approved):
From the code changes POV: all good!

PR #1370, opened 2024-01-24
Refactor IAP related code

SergioEstevao:
Project #1373

This PR refactor the IAP related code to use enum better (avoid rawValues uses) and group related types on a project group.
It also adds a unit test to validate current code

## To test

Steps to test your PR.

- Check if the flows to Purchase a subscription show the correct values for payments
- Go to Profile
- Tap Account
- Check if values show for Plus and Patron are correct
- Tap Subscribe to Plus
- Check if values for Plus monthly and yearly are correct
- Swipe to Patron
- Check if values for Patron monthly and yearly are correct
- Tap on Subscribe and check if IAP flow starts correctly

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">View files have been modified, but no screenshot is included in the pull request. Consider adding some for clarity.</td>
    </tr>
  </tbody>
</table>

</p>

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -0,0 +1,78 @@
...
+
+final class IAPHelperTests: XCTestCase {
+
+    override func setUpWithError() throws {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
```
leandroalonso: Can we delete those?

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        let expectation = XCTestExpectation(description: "Fetch Products")
+        NotificationCenter.default.addObserver(forName: ServerNotifications.iapProductsUpdated, object: nil, queue: nil) { notification in
+            expectation.fulfill()
+        }
+        helper.requestProductInfo()
+
+        wait(for: [expectation], timeout: 5)
+        XCTAssert(helper.hasLoadedProducts)
+
+        let _ = helper.getProduct(for: .monthly)
```
leandroalonso, line 41: Do we need this line? We're already calling `requestProductInfo()` directly

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        session.disableDialogs = true
+
+        let helper = IAPHelper()
+        let expectation = XCTestExpectation(description: "Fetch Products")
+        NotificationCenter.default.addObserver(forName: ServerNotifications.iapProductsUpdated, object: nil, queue: nil) { notification in
+            expectation.fulfill()
+        }
+        helper.requestProductInfo()
+
+        wait(for: [expectation], timeout: 5)
```
leandroalonso: 5s seems a lot. 😬 Is there any special reason why? Or can we reduce it to 2, maybe 1s?

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        defer {
+            SKPaymentQueue.default().remove(helper)
+        }
+        let expectation = XCTestExpectation(description: "Fetch Products")
+        NotificationCenter.default.addObserver(forName: ServerNotifications.iapProductsUpdated, object: nil, queue: nil) { notification in
+            expectation.fulfill()
+        }
+        helper.requestProductInfo()
+
+        wait(for: [expectation], timeout: 5)
```
leandroalonso: Same comment as above about the timeout.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        XCTAssertEqual(price, "$3.99")
+
+        let priceWithFrequency = helper.getPriceWithFrequency(for: .monthly)
+        XCTAssertEqual(priceWithFrequency, "$3.99 per month")
+
+        let paymentFrequency = helper.getPaymentFrequency(for: .monthly)
+        XCTAssertEqual(paymentFrequency, "month")
+    }
+
+    func testPurchase() throws {
```
leandroalonso, line 55: This test fails to me when running standalone and on a clean build because `ServerSettings.syncingEmail()` is `nil`. It's probably succeeding in CI because the `EndOfYearStoriesBuilderTests` set it on the `setUp`.

Also, there's a transaction after this test run. Is it possible to delete/clean that?

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        XCTAssert(helper.hasLoadedProducts)
+
+        let buyExpectation = XCTestExpectation(description: "Buy Product")
+        NotificationCenter.default.addObserver(forName: ServerNotifications.iapPurchaseCompleted, object: nil, queue: nil) { notification in
+            buyExpectation.fulfill()
+        }
+        let buyResult = helper.buyProduct(identifier: .monthly)
+        XCTAssert(buyResult)
+
+        wait(for: [buyExpectation], timeout: 500)
```
leandroalonso: Also wondering if we can reduce this `timeout`.

leandroalonso (changes requested):
@SergioEstevao the app is crashing to me:

Screenshot 2024-01-26 at 11 41 02

Apparently is something related to `Tracks`, it seems that the `purchase_cancelled` event is receiving the `IAPProductID`, not the string.

It would be worth double-checking the Analytics events.

To reproduce this:

1. Open the purchase flow
2. Dismiss it
3. Wait for a few seconds
4. 💥 

I also ran into a weird behavior: after logging-in and choosing a plan, I instantly bought it (no confirmation modal). 🤔 I think this was some Xcode issue though as I can't reproduce it consistently.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: We are just checking if we get a product without an error after doing a request. If you notice the signature method of getProductInfo has a ! on the return type. I'm not happy with that and I'm considering to change it for a ? and check if a get a non nil.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: Let me test with a smaller time, when doing this type of tests on another project we sometime run into delays in the CI builds.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: This was a mistake :)

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: Will add the code to setup a fake account.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: Will try it with 1s.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: I'm adding code to clean up the session after the test.

SergioEstevao:
@leandroalonso fixed the tracks code, can you please give it another look?

Regarding the auto-purchase it's strange, unless for some reason the Test session code was interfering with the real app. 
The session cleanup code I added to tests should help if that is the case.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    override func setUpWithError() throws {
+        // Pretend we're logged in
+        email = ServerSettings.syncingEmail()
+        ServerSettings.setSyncingEmail(email: "test@test.com")
+    }
+
+    override func tearDownWithError() throws {
+        // Restore previous login
+        ServerSettings.setSyncingEmail(email: email)
+    }
```
leandroalonso, line 20: [nit] I wonder if we can move this to a `LoggedInXCTestCase` and just inherit from it?

leandroalonso (approved):

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: I will do this in a separate PR.

Issue #1366, opened 2024-01-24
Project: Update libraries


leandroalonso:
> Starting in spring 2024, you must include the privacy manifest for any SDK listed below when you submit new apps in App Store Connect that include those SDKs, or when you submit an app update that adds one of the listed SDKs as part of the update. Signatures are also required in these cases where the listed SDKs are used as binary dependencies. Any version of a listed SDK, as well as any SDKs that repackage those on the list, are included in the requirement.
— https://developer.apple.com/support/third-party-SDK-requirements/

## Dependencies that need privacy manifests

- [x] FMDB #1368
- [x] Lottie #1367 
- [x] SwiftyGif #1374
- [x] SwiftyJSON #1376
- [x] Kingfisher (Cocoapods) #1375

### Firebase and subdependencies

Firebase has a lot of dependencies that are listed in Apple's article. However, even though Firebase already has an `xcprivacy`, not all subdependencies have it. This section documents each one.

- [x] Firebase #1584
- [ ] Abseil (`xcprivacy` file not available)
- [ ] BoringSSL / openssl_grpc (removed with Firebase update)
- [ ] GoogleDataTransport (`xcprivacy` file not available)
- [ ] GoogleUtilities (`xcprivacy` file not available)
- [ ] gRPC (`xcprivacy` file not available)
- [ ] GTMSessionFetcher (`xcprivacy` file not available, PR draft opened)
- [ ] nanopb (`xcprivacy` file not available)
- [ ] Promises (`xcprivacy` file not available, PR opened and approved)
- [ ] leveldb (`xcprivacy` file not available)

### Google Sign in and subdepencencies

- [x] GoogleSignIn #1583
- [ ] AppAuth (`xcprivacy` file not available)
- [ ] GTMAppAuth (`xcprivacy` file not available)
- [ ] GTMSessionFetcher (`xcprivacy` file not available, PR draft opened)
- [ ] GoogleUtilities (`xcprivacy` file not available)

## Dependencies that (in theory) don’t need privacy manifests

(if they repackage any of the SDKs mentioned above they do need a privacy manifest)

- [x] Agrume https://github.com/Automattic/pocket-casts-ios/pull/1374
- [x] AutomattickTracksiOS #1382
- [x] BuildkiteTestCollector
- [x] DifferenceKit
- [x] Fuse
- [x] GoogleAppMeasurement
- [x] Sentry
- [x] Sodium
- [x] SwiftProtobuf
[...]

PR #1375, opened 2024-01-26
[Library Update] Kingfisher

leandroalonso:
Part of #1366

This PR updates Kingfisher and moves it to Swift Package Manager.

We had an issue in the past in which importing Kingfisher through SPM was causing crashes on Series 3 watches. I don't have this specific device to test so I think we can try this change and see if the issues happen during the beta period.

## To test

1. ✅ Check that Kingfisher has a `xcprivacy` file in Package Dependencies > Kingfisher > Sources 
2. Run the main app and the watch app and make sure everything works as expected ¶

¶ Keep in mind that Xcode has an issue with running on the watch simulator: https://developer.apple.com/documentation/xcode-release-notes/xcode-15_1-release-notes

> Running a WatchApp that requires the companion iOS app to be installed will result in an error if the run destination is set to a Watch via iPhone Simulator pair. (119640671)
> Workaround: Select a singular Watch Simulator run destination and then Run.

## Checklist

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -8,20 +8,9 @@ inhibit_all_warnings!
 
 app_ios_deployment_target = Gem::Version.new('15.0')
 
-def kingfisher
-  # Any version compatible with 7.6, starting from 7.6.2 to ensure
-  # compatibility with Xcode 14.3.
-  #
-  # Notice that 7.6.2 is not necessarily the first version compatible with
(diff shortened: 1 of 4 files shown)


bjtitus (approved):
✅ Verified xcprivacy manifest file
✅ Verified image loading on iPhone and Apple Watch (Series 8)

PR #1376, opened 2024-01-26
[Library Update] SwiftyJSON

leandroalonso:
Part of #1366

This PR removes SwiftyJSON.

### MainServerHandler.swift

Given this method has some external dependencies I wasn't able to easily unit test it, but we can manually test that the changes here are working as expected.

1. Go to `trunk`
2. Add a breakpoint on `createRefreshRequest` `return` statement
3. Run the app
4. Once the breakpoint is hit, print the contents of the generated JSON: `po String.init(data: try! jsonRequest.rawData(), encoding: .utf8)`
5. Save it somewhere
6. Now run this branch, repeat the same steps, and once again print the contents of the generated JSON: `po String.init(data: data, encoding: .utf8)`
7. Now compare the old and new generated JSONs using a tool like https://www.jsondiff.com
8. ✅ The contents should be the same, even though the order might differ

### ServerHelper

For `ServerHelper` I added a unit test to test that the conversion is working. You can copy this file to `Server/Tests` and run it. On `trunk` you should see a few unit tests failing because it sets some fields to an empty String, instead of `nil`. Based on my tests though, that doesn't make any difference.

1. Copy an "old database" to your simulator's app container (so we can simulate episodes updating)
2. Run the app
3. ✅  Check that episodes are correctly updated

### ApiServerHandler

2. Go to `APIServerHandler+Account.swift` and remove lines `67-68` (the ones setting username/password)
3. Add a breakpoint in `APIServerHandler+Account.swift` `extractErrorResponse` (inside the first `do` block)
4. Run a clean install of the app
5. Create a new account
7. ✅ The breakpoint should hit
8. Notice that the return will be `APIError.PASSWORD_INVALID`

## Checklist


Modules/Server/Tests/PocketCastsServerTests/ServerHelperTests.swift
```diff
@@ -0,0 +1,125 @@
...
+        XCTAssertEqual(secondPodcastUpdates?[0].episodeType, "full")
+        XCTAssertEqual(secondPodcastUpdates?[0].seasonNumber, 0)
+        XCTAssertEqual(secondPodcastUpdates?[0].episodeNumber, 0)
+        XCTAssertEqual(secondPodcastUpdates?[0].publishedDate, "2024-01-26 15:00:52")
+    }
+}
+
+private extension ServerHelperTests {
+    func dataJSON() -> Data {
+        """
```
leandroalonso: I wasn't able to use this content inside a JSON file, so I just copied it here.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><code>Modules/Server/Package.swift</code> was changed without updating its corresponding <code>Package.resolved</code>. Please resolve the Swift packages in Xcode.</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso:
@SergioEstevao This one is a bit more involved to manually test so I requested a review from @emilylaguna given she knows the steps to test episode updates.

Anyway, feel free to review if you'd like to!

SergioEstevao (approved):
:shipit:

Modules/Server/Tests/PocketCastsServerTests/ServerHelperTests.swift
SergioEstevao: I'm curious about this, why couldn't you read it from a file?

Modules/Server/Sources/PocketCastsServer/Public/Refresh/MainServerHandler.swift
```diff
@@ -292,8 +291,8 @@ public class MainServerHandler {
         }
 
         var jsonRequest = jsonWithStandardParams(uniqueId: uniqueId)
-        jsonRequest["podcast_uuid"].string = podcast.uuid
-        guard let data = try? jsonRequest.rawData() else {
+        jsonRequest["podcast_uuid"] = podcast.uuid
+        guard let data = try? JSONSerialization.data(withJSONObject: jsonRequest, options: []) else {
```
emilylaguna:
```suggestion
        guard let data = try? JSONSerialization.data(withJSONObject: jsonRequest) else {
```

Modules/Server/Sources/PocketCastsServer/Public/Refresh/MainServerHandler.swift
```diff
@@ -235,15 +234,15 @@ public class MainServerHandler {
...
+        jsonRequest["last_episodes"] = podcasts.map { $0.forceRefreshEpisodeFrom ?? $0.latestEpisodeUuid ?? "" }.joined(separator: ",")
+        jsonRequest["push_messages_on"] = podcasts.map { (pushEnabled && $0.pushEnabled) ? "1" : "0" }.joined()
         if let pushToken = ServerSettings.pushToken() {
-            jsonRequest["push_token"].string = pushToken
+            jsonRequest["push_token"] = pushToken
         }
-        jsonRequest["push_on"].string = pushEnabled ? "true" : "false"
-        guard let data = try? jsonRequest.rawData() else {
+        jsonRequest["push_on"] = pushEnabled ? "true" : "false"
+        guard let data = try? JSONSerialization.data(withJSONObject: jsonRequest, options: []) else {
```
emilylaguna:
```suggestion
        guard let data = try? JSONSerialization.data(withJSONObject: jsonRequest) else {
```

Modules/Server/Tests/PocketCastsServerTests/ServerHelperTests.swift
emilylaguna: I agree, can we move this to a file?

emilylaguna (changes requested):
I tested all scenarios except for the `ApiServerHandler` because the steps aren't triggering the breakpoint for some reason. 

The ones I tested are working correctly. 

I left some comments below to clean up the code a bit.

Modules/Server/Tests/PocketCastsServerTests/ServerHelperTests.swift
leandroalonso: Is just not that simple, unfortunately.

This test is on the `Server` module and that's what makes this complicated, I can't just simply access the file it as I would by testing on the app test target.

I can move it to a file and make some changes on the `Package.swift` but then the file will be copied as part of module, even though it's not needed by the app/module. See the first answer here: https://stackoverflow.com/questions/67662742/getting-file-in-path-spm-swift

I can try the second answer there. But it involves a lot of additional code for something that, in theory, should be simple.

What do we prefer? Keep this way? Copy it to the app even though is not needed? Try adding additional code for that?

leandroalonso:
@emilylaguna I updated the `ApiServerHandler` testing steps.

Modules/Server/Tests/PocketCastsServerTests/ServerHelperTests.swift
leandroalonso: I did the additional code here, let me know what do you think: 60bb34050234693e5f9028c1f227e58dd593197e

SergioEstevao:
@leandroalonso The sepate file and fixture is looking good for me! :shipit:

Modules/Server/Tests/PocketCastsServerTests/ServerHelperTests.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        url = url.appendingPathComponent("\(thisModuleName).bundle")
+
+        guard let bundle = Bundle(url: url) else {
+            fatalError("Foundation.Bundle.module could not load resource bundle: \(url.path)")
+        }
+
+        return bundle
+    }
+}
```
emilylaguna: We can use `Bundle.module` to get the resource path for the tests, and remove the extra code here:

```suggestion
public class JSONHelper {
    public static func loadJSON(withFile fileName: String, subdirectory: String = "Fixtures") -> Data? {
        let bundle = Bundle.module

        if let url = bundle.url(forResource: fileName, withExtension: "json", subdirectory: subdirectory) {
            do {
                return try Data(contentsOf: url)
            } catch {
                debugPrint(error)
            }
        } else {
            debugPrint("Could not find the json file: \(fileName) in bundle: \(bundle)")
        }
        return nil
    }
}
```

emilylaguna (changes requested):
Was able to test with the new test steps, left a comment about reducing the Bundle boilerplate

Modules/Server/Tests/PocketCastsServerTests/ServerHelperTests.swift
emilylaguna: We can use `Bundle.module` for accessing resources in a Swift Package: https://developer.apple.com/documentation/xcode/bundling-resources-with-a-swift-package#Access-a-resource-in-code

The caveat here is that you need to have at least 1 file being copied to the test target for the code to be generated.

emilylaguna (approved):

PR #1390, opened 2024-01-31
Refactor the IAPHandler to not make requests to the production API during testing

emilylaguna:
While looking into an issue I noticed during unit testing the IAPHelper is sending requests to the production server. 
- After the products request comes in the IAPHelper checks for the [trial eligibility](https://github.com/Automattic/pocket-casts-ios/blob/7c34f752cceb04ffe716fd437ec0dd8611b6c075/podcasts/IAPHelper.swift#L130)
- After making a purchase the IAPHelper [sends the receipt](https://github.com/Automattic/pocket-casts-ios/blob/4dd2fd2966ec8387395e46bb9ad38a2501d7c472/podcasts/IAPHelper.swift#L362-L368) to the server for validation

To fix this, I initially set out to inject the `ApiServerHandler` but in order to subclass those methods I would have had to refactor a lot more since they are contained in a different module and would need to be set as `open`, but since they are defined in an Extension we can't do that without reorganizing that code. 

Instead, I created a simple proxy class ServerHandler (name could be better) that interfaces with the "non-injectable" classes: `ApiServerHandler` and `ServerSettings`. We can then inject this class into IAPHelper and be able to subclass it for unit testing. 

While updating the `IAPHelperTests`, I also did a refactor of the common code for setting up/tearing down the `SKTestSession` and `IAPHelper`. Also, since I'm sure this will come up more as we add more unit tests, I made a small extension to be able to wait for notifications to be fired in the test using a single line. 

## To test

1. 🟢 CI
2. Open `APIServerHandler+TrialEligibility.swift` and put a breakpoint on line 6
3. Open `ApiServerHandler+Account.swift` and put a breakpoint on line 236
4. Open `ServerSettings.swift` and put a breakpoint on lines 152, and 156
5. Open the Scheme Editor, and set the StoreKit Configuration file to Pocket Casts Configuration
6. Launch the app while sign into a non-plus account
7. ✅ Verify the `checkTrialEligibility` breakpoint is hit
8. Tap the Podcasts tab
9. Tap the locked folders icon
10. Continue following the steps to purchase any subscription
11. After finalizing the purchase
12. ✅ Verify the `ServerSettings` line 156 breakpoint is hit
13. Continue
14. ✅ Verify the `ServerSettings` line 152 breakpoint is hit
15. Continue
16. ✅ Verify the `ApiServerHandler+Account` line 236 breakpoint is hit

### Verify no API calls are made during testing
1. Following the steps from above
2. Open Proxyman or Charles
3. Run the tests in IAPHelperTests 
4. ✅ Verify no breakpoints are hit
5. ✅ Verify no API calls are made

## Checklist

### DIFF
--- a/PocketCastsTests/Extensions/XCTestCase+WaitForNotification.swift
+++ b/PocketCastsTests/Extensions/XCTestCase+WaitForNotification.swift
@@ -1,0 +1,12 @@
+import XCTest
+
+extension XCTestCase {
+    /// Reduces the boilerplate of waiting for a notification expectation to fire into 1 line
+    /// wait(for: .helloWorld, decription: "Hello, of worlds")
+    func wait(for notification: Notification.Name, object: Any? = nil, notificationCenter: NotificationCenter = .default, timeout: TimeInterval = 1, description: String) {
+        let expectation = XCTNSNotificationExpectation(name: notification, object: object, notificationCenter: .default)
+        expectation.expectationDescription = description
+
+        wait(for: [expectation], timeout: timeout)
+    }
+}
--- a/PocketCastsTests/Tests/IAP/IAPHelperTests.swift
+++ b/PocketCastsTests/Tests/IAP/IAPHelperTests.swift
@@ -27,12 +27,6 @@ final class IAPHelperTests: XCTestCase {
     }
 
     func testRequestInfo() throws {
-        let session = try SKTestSession(configurationFileNamed: configurationFile)
-        session.clearTransactions()
-        session.resetToDefaultState()
-        session.disableDialogs = true
-
-        let helper = IAPHelper()
         let expectation = XCTestExpectation(description: "Fetch Products")
         NotificationCenter.default.addObserver(forName: ServerNotifications.iapProductsUpdated, object: nil, queue: nil) { notification in
             expectation.fulfill()
@@ -52,21 +46,9 @@ final class IAPHelperTests: XCTestCase {
 
         let paymentFrequency = helper.getPaymentFrequency(for: .monthly)
         XCTAssertEqual(paymentFrequency, "month")
-
-        session.clearTransactions()
     }
 
     func testPurchase() throws {
-        let session = try SKTestSession(configurationFileNamed: configurationFile)
-        session.clearTransactions()
-        session.resetToDefaultState()
-        session.disableDialogs = true
-
-        let helper = IAPHelper()
-        SKPaymentQueue.default().add(helper)
-        defer {
-            SKPaymentQueue.default().remove(helper)
-        }
         let expectation = XCTestExpectation(description: "Fetch Products")
         NotificationCenter.default.addObserver(forName: ServerNotifications.iapProductsUpdated, object: nil, queue: nil) { notification in
             expectation.fulfill()
@@ -85,7 +67,6 @@ final class IAPHelperTests: XCTestCase {
 
         wait(for: [buyExpectation], timeout: iapTestTimeout)
 
-        session.clearTransactions()
     }
 }
 
--- a/PocketCastsTests/Tests/IAP/IAPHelperTests.swift
+++ b/PocketCastsTests/Tests/IAP/IAPHelperTests.swift
@@ -6,17 +6,11 @@ import PocketCastsServer
 
 final class IAPHelperTests: XCTestCase {
 
-    var email: String?
 
     override func setUpWithError() throws {
-        // Pretend we're logged in
-        email = ServerSettings.syncingEmail()
-        ServerSettings.setSyncingEmail(email: "test@test.com")
     }
 
     override func tearDownWithError() throws {
-        // Restore previous login
-        ServerSettings.setSyncingEmail(email: email)
     }
 
     let configurationFile = "Pocket Casts Configuration"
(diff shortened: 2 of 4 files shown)


PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -5,40 +5,30 @@ import StoreKitTest
...
-        let expectation = XCTestExpectation(description: "Fetch Products")
-        NotificationCenter.default.addObserver(forName: ServerNotifications.iapProductsUpdated, object: nil, queue: nil) { notification in
-            expectation.fulfill()
-        }
         helper.requestProductInfo()
-
-        wait(for: [expectation], timeout: iapTestTimeout)
-        XCTAssert(helper.hasLoadedProducts)
-
-        let _ = helper.getProduct(for: .monthly)
```
emilylaguna, line 41: I checked the code, but I didn't see a reason for this to be called. Let me know if I missed something.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -5,40 +5,30 @@ import StoreKitTest
...
-        // Pretend we're logged in
-        email = ServerSettings.syncingEmail()
-        ServerSettings.setSyncingEmail(email: "test@test.com")
+        session = try SKTestSession(configurationFileNamed: configurationFile)
+        session.clearTransactions()
+        session.resetToDefaultState()
+        session.disableDialogs = true
+
+        helper = IAPHelper(serverHandler: MockIAPHandler())
+        SKPaymentQueue.default().add(helper)
```
emilylaguna, line 24: The payment queue isn't technically used in both of the tests, but if we add more tests then it'll most likely be needed again.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -48,40 +38,50 @@ final class IAPHelperTests: XCTestCase {
...
-            buyExpectation.fulfill()
-        }
         let buyResult = helper.buyProduct(identifier: .monthly)
         XCTAssert(buyResult)
 
-        wait(for: [buyExpectation], timeout: iapTestTimeout)
+        wait(for: ServerNotifications.iapPurchaseCompleted, timeout: iapTestTimeout, description: "Buy Products")
 
-        session.clearTransactions()
+        XCTAssertEqual(session.allTransactions().first?.productIdentifier, IAPProductID.monthly.rawValue)
```
emilylaguna, line 57: I added this check to verify the purchased product is the the correct one

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: It was just a way to confirm we got the product after the request is completed.

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
SergioEstevao: Good addition!

PocketCastsTests/Extensions/XCTestCase+WaitForNotification.swift
```diff
@@ -0,0 +1,12 @@
+import XCTest
+
+extension XCTestCase {
+    /// Reduces the boilerplate of waiting for a notification expectation to fire into 1 line
+    /// wait(for: .helloWorld, decription: "Hello, of worlds")
+    func wait(for notification: Notification.Name, object: Any? = nil, notificationCenter: NotificationCenter = .default, timeout: TimeInterval = 1, description: String) {
+        let expectation = XCTNSNotificationExpectation(name: notification, object: object, notificationCenter: .default)
```
SergioEstevao, line 7: TIL: XCTNSNotificationExpectation 
:)

podcasts/IAPHelper.swift
```diff
@@ -417,3 +422,28 @@ private extension IAPHelper {
         Analytics.track(event, properties: properties)
     }
 }
+
+// MARK: - ServerHandler
+
+extension IAPHelper {
+    /// Acts as a proxy to the `ServerSettings` static methods the IAPHelper uses, and the `ApiServerHandler` calls
+    /// This allows for these to be injected and test, but keep the functionality the same.
+    class ServerHandler {
```
SergioEstevao: This class approach is totally valid, but couldn't we create a protocol that has these methods, and the have an implemention forthe real calls and another for the mock. That way we will avoid subsclassing this on the tests/mocks

SergioEstevao (approved):
I tested all the scenarios proposed and it's working great.

I just left some comments about the way the injection/handler was implemented but feel free to ignore it.

leandroalonso (approved):

podcasts/IAPHelper.swift
emilylaguna: Makes sense. Using a protocol also prevents the possibility of the tests using production data. 

Since I wasn't happy with the naming, I decided to split the dependencies into 2 protocols: Settings and Networking

Done here: 
- 7be4ab20b245339e494a0b1e2123b14f933bea9e
- 056b8c017bc6f509a78065834dac1c85ea759248

PR #1398, opened 2024-02-02
Disable the IAPHelperTests

emilylaguna:
The test is still being flaky, so we'll disable it for now. 

## To test

🟢 CI

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/IAP/IAPHelperTests.swift
+++ b/PocketCastsTests/Tests/IAP/IAPHelperTests.swift
@@ -28,7 +28,7 @@ final class IAPHelperTests: XCTestCase {
         session.clearTransactions()
         SKPaymentQueue.default().remove(helper)
     }
-
+/*
     func testRequestInfo() throws {
         helper.requestProductInfo()
         wait(for: ServerNotifications.iapProductsUpdated, timeout: iapTestTimeout, description: "Fetch Products")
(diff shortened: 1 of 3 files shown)


PocketCastsTests/UnitTests.xctestplan
```diff
@@ -52,6 +52,10 @@
   },
   "testTargets" : [
     {
+      "skippedTests" : [
+        "IAPHelperTests",
```
bjtitus, line 56: Isn't this first one enough to disable the whole file?

PocketCastsTests/Tests/IAP/IAPHelperTests.swift
```diff
@@ -28,7 +28,7 @@ final class IAPHelperTests: XCTestCase {
         session.clearTransactions()
         SKPaymentQueue.default().remove(helper)
     }
-
+/*
```
bjtitus, line 31: Does this need to be commented out when the whole file is disabled from the test plan?

bjtitus (approved):
Left a few comments but it fixes the currently broken test so I'll approve. ✅

PocketCastsTests/UnitTests.xctestplan
emilylaguna: I used the test plan to skip it in my first commit, but it still ran 🤷‍♂️ .

bjtitus: Weird, it skipped for me locally. Better to be safe than sorry, I guess! 😅

PR #1405, opened 2024-02-04
Update protobuf files for Settings sync

bjtitus:
Fixes #1404

Just updates the protobuf files to latest with new NamedSettings values and types.

## To test

🟢 CI should be green

1. Run the app
2. Login
3. Subscribe to podcast and make sure syncing works

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

Issue #1400, opened 2024-02-02
Settings Sync Project


bjtitus:
Project Issue for syncing all user settings between devices.

Project Thread: pdeCcb-31c-p2

```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/pocket-casts-ios/issues/1402
- [ ] https://github.com/Automattic/pocket-casts-ios/issues/1404
- [x] Update local settings storage for Modified Date
- [ ] https://github.com/Automattic/pocket-casts-ios/issues/1461
```

```[tasklist]
### Settings Screens
- [x] General Settings
- [x] Podcast settings
- [x] Notification settings
- [x] Appearances settings
- [x] Auto Archive settings
- [x] Auto Add to Up Next
- [x] Storage and Data Usage
- [x] Headphone Controls
- [x] Files
- [x] Podcasts ••• Menu
- [x] Playback Effects
- [x] Player Shelf
```

BKrnk1337:
Sounds great. Having all settings synced between iPhone and Apple Watch would be nice. I have so many individual settings per podcast. Does this "project" include this aswell?

PR #1406, opened 2024-02-05
Settings sync storage

bjtitus:
| 📘 Part of: #1400 | Depends on #1405|
|:---:|:---:|

* Adds storage for app settings (to be synced with server) with a modified date attached to each property.
* Converts `openLinks` and `rowAction` General settings to use this new storage mechanism

#### Overview
* `CodableStore` is a simple property wrapper for `UserDefaults` that handles encode/decode of a `JSONCodable`.
* `ModifiedDate` is a simple property wrapper to add a `modifiedAt` property which is updated when the value is set.

* `AppSettings` provides a model to represent our synced settings with appropriate primitive types and `JSONCodable` conformance
* `SettingsStore` wraps both the `AppSettings` model and `UserDefaults` to coordinate encoding.

I've added more details below about design thinking and one change to note.

#### Storage

`AppSettings` is a model specific to synced settings and is `JSONCodable` for persistence to `UserDefaults`. This seemed much cleaner than juggling individual `UserDefaults` keys and the data is unlikely to ever get large enough to cause a problem with `UserDefaults`. 

`SettingsStore` wraps `AppSettings` (or any other `JSONCodable` instance) and handles encoding the object and saving to a `UserDefaults` (with `CodableStore`). This indirection is needed so that different stores can be configured with different keys (like for individual podcasts). Simply using a key for the property wrapper in `CodableStore` or `AppStorage` wouldn't work because it can't be dynamic at runtime. `SettingsStore` also provides a centralized location which can be made an `ObservableObject` at a later time. The `dynamicMember` subscript lookup function allows the `AppSettings` properties to be accessed directly, like `store.openLinks` rather than `store.settings.openLinks` but also provides a single point of access where a future Observable/Publisher construct can be added.

#### Notable change

With the feature flag enabled, the cached version of `rowAction` is not used because we don't need to be caching `UserDefaults`.

https://github.com/Automattic/pocket-casts-ios/blob/4a7501b5bd1e1ce239aa3f2eedfaae132333d9f4/podcasts/Settings.swift#L132-L141

[The docs](https://developer.apple.com/documentation/foundation/userdefaults) indicate this is already done by `UserDefaults` automatically:

> At runtime, you use `UserDefaults` objects to read the defaults that your app uses from a user's defaults database. `UserDefaults` caches the information to avoid having to open the user's defaults database each time you need a default value. When you set a default value, it's changed synchronously within your process, and asynchronously to persistent storage and other processes.

## To test

* Run the app
* Enable the `settingsSync` feature flag
* Restart the app
* In General settings, change "Open Links in Browser" (`openLinks`) and "Row Action" (`rowAction`)
* Restart the app
* Make sure the settings persist

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/AppSettings.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/AppSettings.swift
@@ -1,0 +1,16 @@
+import PocketCastsUtils
+
+/// Model type for synced & stored App Settings
+public struct AppSettings: JSONCodable {
+    @ModifiedDate public var openLinks: Bool
+
+    @ModifiedDate public var rowAction: PrimaryRowAction
+
+    static var defaults: AppSettings {
+        return AppSettings(openLinks: true, rowAction: .download)
+    }
+}
+
+extension SettingsStore<AppSettings> {
+    public static let appSettings = SettingsStore(key: "app_settings", value: AppSettings.defaults)
+}
--- a/Modules/Server/Sources/PocketCastsServer/Public/SettingsStore.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/SettingsStore.swift
@@ -1,0 +1,27 @@
+import Foundation
+import PocketCastsUtils
+
+/// A storage type for a `Codable` settings value.
+/// This handles storage to `UserDefaults` (optionally overridable)
+@dynamicMemberLookup
+public final class SettingsStore<Value: JSONCodable> {
+    public let userDefaults: UserDefaults
+
+    public init(userDefaults: UserDefaults = .standard, key: String, value: Value) {
+        self.userDefaults = userDefaults
+        _settings = CodableStore(wrappedValue: value, key)
+    }
+
+    @CodableStore var settings: Value
+
+    /// Access any property from `settings` without direct access to settings.
+    /// Avoids having to type `appSettings.settings` and allows for future ObservableObject / publisher adoption in this method
+    public subscript<T>(dynamicMember keyPath: WritableKeyPath<Value, T>) -> T {
+        get {
+            settings[keyPath: keyPath]
+        }
+        set {
+            settings[keyPath: keyPath] = newValue
+        }
+    }
+}
--- a/Modules/Server/Tests/PocketCastsServerTests/SettingsStoreTests.swift
+++ b/Modules/Server/Tests/PocketCastsServerTests/SettingsStoreTests.swift
@@ -4,35 +4,58 @@ import PocketCastsUtils
 
 class SettingsStoreTests: XCTestCase {
 
-    struct TestType: JSONCodable {
+    private struct TestType: JSONCodable {
         var name: String
     }
 
     private let userDefaultsSuiteName = "PocketCastsServer-SettingsStoreTests"
+    private let defaultsKey = "app_settings"
+    private let initialName = "Initial"
+    private let changedName = "Changed"
 
     override func setUp() {
         super.setUp()
         UserDefaults.standard.removePersistentDomain(forName: userDefaultsSuiteName)
     }
 
-    func testStore() throws {
+    /// Tests the initial Empty state of the UserDefaults suite
+    func testEmptyStore() throws {
         let defaults = try XCTUnwrap(UserDefaults(suiteName: userDefaultsSuiteName), "User Defaults suite should load")
-        let defaultsKey = "app_settings"
 
         XCTAssertNil(defaults.data(forKey: defaultsKey), "User Defaults data should not exist yet for \(defaultsKey)")
+    }
 
-        let initialName = "Hello"
+    /// Tests the initial value taken from the SettingsStore.value
+    func testInitialValue() throws {
+        let defaults = try XCTUnwrap(UserDefaults(suiteName: userDefaultsSuiteName), "User Defaults suite should load")
 
         let store = SettingsStore(userDefaults: defaults, key: defaultsKey, value: TestType(name: initialName))
 
-        let changedName = "Changed"
-
         XCTAssertEqual(store.settings.name, initialName, "Accessed initial value should match")
+    }
+
+    /// Tests updating values to a single SettingsStore instance
+    func testValueUpdate() throws {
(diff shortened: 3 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">No reviewers have been set for this PR yet. Please request a review from <strong>@ Automattic/pocket-casts-ios</strong>.</td>
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/SettingsStore.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+import PocketCastsUtils
+
+/// A storage type for a `Codable` settings value.
+/// This handles storage to `UserDefaults` (optionally overridable)
+@dynamicMemberLookup
```
leandroalonso, line 6: Clever!

Modules/Server/Tests/PocketCastsServerTests/SettingsStoreTests.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    private let userDefaultsSuiteName = "PocketCastsServer-SettingsStoreTests"
+
+    override func setUp() {
+        super.setUp()
+        UserDefaults.standard.removePersistentDomain(forName: userDefaultsSuiteName)
+    }
+
+    func testStore() throws {
+        let defaults = try XCTUnwrap(UserDefaults(suiteName: userDefaultsSuiteName), "User Defaults suite should load")
```
leandroalonso, line 23: What do you think about breaking down this into more descriptive tests?

I like the ArrangeActAssert pattern because it makes super easy-to-read tests: http://wiki.c2.com/?ArrangeActAssert

Even though you'll probably repeat some code between them, I think the main function of unit tests is serving the readability, so duplicated code is more than fine IMHO.

This way we can have a:

1. Test that asserts that the `UserDefaults` is indeed empty
2. A test that asserts that an initial value is correctly saved
3. A test that asserts that changing the value works as expected
4. And last, but not least, a test that asserts that an additional store returns the correct value

In this case, if any of those tests fails we'll know why just by its name. :)

Modules/Utils/Tests/PocketCastsUtilsTests/ModifiedDateTests.swift
```diff
@@ -0,0 +1,20 @@
+import XCTest
+@testable import PocketCastsUtils
+
+class ModifiedDateTests: XCTestCase {
+    struct TestType {
+        @ModifiedDate var name: String
+    }
+
+    func testModifiedDate() throws {
+        let initialName = "Hello"
```
leandroalonso: The same comment as before, I believe we can have two tests here: for the initial `nil` data and the updated data. Thoughts?

Modules/Server/Sources/PocketCastsServer/Public/AppSettings.swift
```diff
@@ -0,0 +1,16 @@
+import PocketCastsUtils
+
+/// Model type for synced & stored App Settings
+public struct AppSettings: JSONCodable {
+    @ModifiedDate public var openLinks: Bool
+
+    @ModifiedDate public var rowAction: PrimaryRowAction
+
+    static var defaults: AppSettings {
+        return AppSettings(openLinks: true, rowAction: .download)
```
leandroalonso, line 10: Shouldn't this return whatever the default value the user had before? I have an issue in which the values are changed in this scenario:

1. I have the flag off and `openLinks` is `false` and `rowAction` is `play`
2. I turn the `syncingSettings` flag on
3. Re-run the app
4. 💥  `openLinks` is now `true` and `rowAction` is `download`, which are values different from what I have before

leandroalonso (changes requested):
@bjtitus nice work on that!

Left a couple of comments, let me know what do you think.

Modules/Server/Sources/PocketCastsServer/Public/AppSettings.swift
bjtitus: Yeah, I was planning to add this in a follow up PR since this is just the base storage change. I believe those changes will be completely additive so the core design here will remain the same. Do you think it would be better to include this now so settings aren't lost if someone turns the flag on? Though, the flag is off by default and the original `UserDefaults` settings will still be around.

Modules/Server/Tests/PocketCastsServerTests/SettingsStoreTests.swift
bjtitus: I think this is a great idea!

I do worry that removing the intermediate assertions could make it hard to track down where something has broken since these tests are so order dependent but I guess the prior tests should cover those cases. I love the improved readability, though, so I think it's worth making this change!

I've updated the branch with the separate tests and more strictly follow `ArrangeActAssert`, please give it a look!

Modules/Server/Sources/PocketCastsServer/Public/AppSettings.swift
bjtitus: If you're curious about what this might look like, I pushed a commit here: https://github.com/Automattic/pocket-casts-ios/commit/36951d8957e86539b76250f480944e35a3994988

I think I'd like to submit that separately, though, and make sure I can include a modification to default sync times for those settings as described here: https://github.com/Automattic/pocket-casts-android/pull/1798

leandroalonso: > Do you think it would be better to include this now so settings aren't lost if someone turns the flag on? Though, the flag is off by default and the original UserDefaults settings will still be around.

I think we're good, not an issue! Just wanted to make sure. :)

leandroalonso (approved):

PR #1410, opened 2024-02-05
Feature/updade account upgrade flows

SergioEstevao:
Fixes #

This PR updates the upgrade prompt on Account to use the same cards that we have on the onboarding flow.
This allows us to go directly to Apple IAP flow from the account screen.

In order to test what option result in better conversion this feature will be controlled trough a feature flag: `new_account_upgrade_prompt_flow`

| 1 | 2 | 3 | 4 |
| - | - | - | - |
| Simulator Screenshot - iPhone 15 Pro - 2024-02-06 at 17 05 13 |  Simulator Screenshot - iPhone 15 Pro - 2024-02-06 at 17 05 19 |  Simulator Screenshot - iPhone 15 Pro - 2024-02-06 at 17 05 23 |  Simulator Screenshot - iPhone 15 Pro - 2024-02-06 at 17 05 28 | 

## To test

 - Edit the podcast scheme to use the "Pocket Casts Configuratio.storekit" file for the StoreKitConfiguration
 - Enable the FF `new_account_upgrade_prompt_flow` in the developer settings
 - Go to the Profile tap ( you must have an account with no active subscription)
 - Tap on Account
 - Check if the cards show correctly for the upgrade options like the screenshots above
 - Switch between the Yearly and Monthly options and see if it shows correctly
 - Tap on Subscribe to Plus
 - Check if the IAP flow starts correctly
 - You can repeat the tests above for Patron 

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso:
> Check if a modal is show and match the design above ( on smaller devices it should show a taller modal)

@SergioEstevao which modal is being referred to in this step? The IAP one?

podcasts/Onboarding/Plus/Account Prompt/UpgradePrompt.swift
```diff
@@ -0,0 +1,61 @@
+import SwiftUI
+import PocketCastsServer
+
+struct UpgradePrompt: View {
```
leandroalonso, line 4: Can we name it `UpgradePromptView`? There are some other views without the suffix `View` already but I wonder if that makes it more descriptive? 🤔

podcasts/Onboarding/Plus/PlusLandingViewModel.swift
```diff
@@ -136,6 +137,6 @@ extension PlusLandingViewModel {
     @ViewBuilder
     private static func view(with viewModel: PlusLandingViewModel) -> some View {
         UpgradeLandingView(viewModel: viewModel)
-            .setupDefaultEnvironment()
+            .setupDefaultEnvironment(theme: Theme.init(previewTheme: .light))
```
leandroalonso, line 140: Is this change intentional? Or it was committed by mistake?

leandroalonso (changes requested):
@SergioEstevao the events being triggered don't have the `flow` (`plus_account_upgrade`): `select_payment_frequency_next_button_tapped ["flow": "none", "product": "com.pocketcasts.plus.yearly"]`

Also, the account header spacing is super tight:

| this branch | trunk | without a subscription |
| ---- | ------------ | ----------------------- |
| Simulator Screenshot - iPhone 15 - 2024-02-08 at 13 00 20 | Simulator Screenshot - iPhone 15 - 2024-02-08 at 13 04 30 | Simulator Screenshot - iPhone 15 - 2024-02-08 at 13 11 16 |

podcasts/Onboarding/Plus/Account Prompt/UpgradePrompt.swift
SergioEstevao: I can do, this is something that even Apple is not very consistent. They have loads of controls that are not named view: Button, Text, etc...

podcasts/Onboarding/Plus/PlusLandingViewModel.swift
SergioEstevao: Yes it was intentional, to show the UpgradeLandingView, that is used on onboarding and has a coloured background   always using the light theme.

simulator_screenshot_6EBA8A9F-CFD9-4EC3-A163-6435E80C39E4

SergioEstevao:
@leandroalonso 

> @SergioEstevao the events being triggered don't have the flow (plus_account_upgrade): 

Regarding the above, from what I see in the code the flow is set on the OnboardingFlow `begin` method.
But in this scenario we don't use that method, is it ok for me to change the OnboardindFlow to set the flow variable directly or some other method that creates a model but do not return a ViewController?

podcasts/Onboarding/Plus/Account Prompt/UpgradePrompt.swift
leandroalonso: Oh, that makes sense. Nevermind!

emilylaguna:
> Regarding the above, from what I see in the code the flow is set on the OnboardingFlow begin method.
> But in this scenario we don't use that method, is it ok for me to change the OnboardindFlow to set the flow variable directly or some other method that creates a model but do not return a ViewController?

This is a bit tricky, but I think that is the better option for this. I would just also make sure it gets reset if the user leaves that view without tapping anything. It should use the same flow and source that we do currently for the account prompt.

SergioEstevao:
> @SergioEstevao which modal is being referred to in this step? The IAP one?

Just realized that I mixed up some steps with the ones from the previous flow, I updated the text in the PR description with the correct steps.

SergioEstevao:
> > Regarding the above, from what I see in the code the flow is set on the OnboardingFlow begin method.
> > But in this scenario we don't use that method, is it ok for me to change the OnboardindFlow to set the flow variable directly or some other method that creates a model but do not return a ViewController?
> 
> This is a bit tricky, but I think that is the better option for this. I would just also make sure it gets reset if the user leaves that view without tapping anything. It should use the same flow and source that we do currently for the account prompt.

Do you think we could just call: `let _ = OnboardingFlow.shared.begin(flow: .plusAccountUpgrade, in: model.parentController, source: PlusAccountPromptViewModel.Source.profile.rawValue, context: nil)`

and ignore the controller returned?

SergioEstevao:
@leandroalonso I improved the spacing on the header and add a solution for the event tracking ( not sure if this is the best approach for it) 
Let me know what do you think.

podcasts/Onboarding/Plus/Account Prompt/PlusAccountPromptTableCell.swift
```diff
@@ -12,9 +12,17 @@ class PlusAccountPromptTableCell: ThemeableCell {
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
 
-        let view = PlusAccountUpgradePrompt(viewModel: model, contentSizeUpdated: { [weak self] size in
-            self?.contentSizeUpdated?(size)
-        }).themedUIView
+        let view: UIView
+        if FeatureFlag.newAccountUpgradePromptFlow.enabled {
+            let _ = OnboardingFlow.shared.begin(flow: .plusAccountUpgrade, in: model.parentController, source: PlusAccountPromptViewModel.Source.profile.rawValue, context: nil)
```
SergioEstevao, line 17: Just calling this method to get the flow and source correctly set for the tracking events, but then I'm ignoring the VC because we don't need it here.

leandroalonso (commented):
@SergioEstevao almost there! The events are fixed, but the spacing is still a bit weird:

| Before | Now |
| ------ | ----- |
| Simulator Screenshot - iPhone 15 - 2024-02-08 at 13 04 30 | Simulator Screenshot - iPhone 15 - 2024-02-09 at 11 32 29 |

Notice how the label top space and bottom space are different.

SergioEstevao:
@leandroalonso just improved the spacing around the badge, can you have a look?

leandroalonso:
edited: I ran the wrong branch.

leandroalonso:
@SergioEstevao it seems we still have a different spacing between the elements:



We should have the same spacing between the avatar and the email, the email and the badge, the badge and the rest.

SergioEstevao:
@david-gonzalez-a8c wanted the space to be tighter in order to have more space for the upgrade cards... I can adjust for the case when no cards is showing.

SergioEstevao:
@leandroalonso I just update the spaces to be consistent across all elements:
| No Upgrade | With Upgrade |
| - | - |
| simulator_screenshot_6375F75B-2B33-4304-A099-36B30E524F76 | simulator_screenshot_FFC05676-33C9-4065-B469-3C154C59F225 |

leandroalonso (approved):
@SergioEstevao thanks for the patience on this back and forth! Even though the spacing is a bit different from what we had before I confirmed with David that this is intended! :shipit:

PR #1369, opened 2024-01-24
[Library Update] Firebase

leandroalonso:
Part of #1366

## Context

We're updating our libraries in "batches" so we can meet Apple's requirement of including a privacy manifest for SDKs.

This first batch will target `release/7.56` so we can give it a test on the current beta. I'm aiming at 4-5 libs per release.

This PR updates Firebase. However, a lot of Firebase dependencies are listed as needing privacy manifest but haven't added it yet.

## To test

There are two main things we need to test: events and remote feature flags. The app also uses Performance but we're looking into removing that so it's not part of the testing steps.

### Remote Feature Flags

1. Delete the app from your simulator/device
2. Go to `Constants.swift` and change `newPlayerTransitionDefault` to `false`
3. Run the app
4. Go to Profile > Settings > Beta Features
5. ✅ `newPlayerTransition` should be enabled, which means remote feature flags are working as expected

### Events

1. Go to `AnalyticsHelper.swift` and change `private extension AnalyticsHelper` to `extension AnalyticsHelper`
2. On `ProfileViewController.swift` add `AnalyticsHelper.logEvent("testing_testing")` on the end of `viewDidAppear`
3. Run the app
4. Go to profile a few times
5. Put the app on background
6. ✅  Go to [this link](https://console.firebase.google.com/u/0/project/singular-vector-91401/analytics/app/ios:au.com.shiftyjelly.podcasts/overview/reports~2Fdashboard%3Fparams%3D_u.comparisonOption%253Ddisabled%2526_u.date00%253D20231227%2526_u.date01%253D20240124%2526_r..dimension-value%253D%257B%2522dimension%2522%253A%2522eventName%2522%252C%2522value%2522%253A%2522testing_testing%2522%257D&r%3Devents-overview&fpn%3D910146533958), change the date to today and check that the event has been triggered in the past 30 minutes (might take some time)

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1418, opened 2024-02-06
Dangermattic checks update

iangmaia:
This PR adapts the `Dangerfile` to the latest [Dangermattic](https://github.com/Automattic/dangermattic) version, using newly added plugins, adapting the code as needed and implementing a few improvements.

The main changes:
- Improvements to the Rubocop check call
- Use `github_utils` plugin instead of locally declared functions
- Added message report when we're skipping checks due to the `Releases` label
- Check for assigned reviewers only when the PR isn't draft
- Use officially distributed `danger-dangermattic` Gem

## Testing
The main validation is to make sure CI is 🟢.

To run Danger yourself:
1. Checkout this branch 
2. Run `bundle install`
3. Run:
```ruby
DANGER_GITHUB_API_TOKEN=<github_token> bundle exec danger pr <PR_URL>
```
You should get in the console the same errors / warnings reported in the PR.


Dangerfile
```diff
@@ -1,46 +1,30 @@
...
 milestone_checker.check_milestone_due_date(days_before_due: 2)
 
-# `files: []` forces rubocop to scan all files, not just the ones modified in the PR
-rubocop.lint(files: [], inline_comment: true, fail_on_inline_comment: true, include_cop_names: true)
-
-warn('PR is classed as Work in Progress') if wip_feature?
+warn('PR is classed as Work in Progress') if github_utils.wip_feature?
 
-warn("No reviewers have been set for this PR yet. Please request a review from **@\u2028Automattic/pocket-casts-ios**.") unless requested_reviewers?
+warn("No reviewers have been set for this PR yet. Please request a review from **@\u2060Automattic/pocket-casts-ios**.") unless github_utils.requested_reviewers? || github.pr_draft?
```
dangermattic, line 30: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-ios/blob/a5d7b9d2d6e2133baad215aaf91ad2dc691dca8a/Dangerfile#L30"></span>Style/IfUnlessModifier: Modifier form of <code>unless</code> makes the line too long.</td>
    </tr>
  </tbody>
</table>

dangermattic, line 30: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-ios/blob/a5d7b9d2d6e2133baad215aaf91ad2dc691dca8a/Dangerfile#L30"></span>Layout/LineLength: Line is too long. [181/180]</td>
    </tr>
  </tbody>
</table>

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.github/workflows/run-danger.yml
```diff
@@ -6,6 +6,6 @@ on:
 
 jobs:
   dangermattic:
-    uses: Automattic/dangermattic/.github/workflows/reusable-run-danger.yml@trunk
+    uses: Automattic/dangermattic/.github/workflows/reusable-run-danger.yml@v1.0.0
```
iangmaia, line 11: @leandroalonso: we can use this PR to add what you mentioned [here](https://github.com/Automattic/pocket-casts-ios/pull/1369#issuecomment-1910063436): skipping Danger on drafts, either completely or on some of the checks.

I think a Draft PR can also serve as a way to make sure CI will succeed, so my first instinct is to keep most checks there. But this is totally up to the teams and their own process, so let me know what you think makes more sense 🙂

leandroalonso: > we can use this PR to add what you mentioned https://github.com/Automattic/pocket-casts-ios/pull/1369#issuecomment-1910063436: skipping Danger on drafts, either completely or on some of the checks.

Can we only skip the checks like labels, milestones, etc? The type of thing you are not worried about until the PR is ready. I think the rest is perfectly fine to leave!

iangmaia: I've made some changes here 314261d7f085ca852002892fbce003437eca1054, avoiding some checks and also to prevent the job to run unnecessarily while the PR is a Draft.

Dangerfile
```diff
@@ -1,46 +1,31 @@
...
 
-def wip_feature?
-  has_wip_label = github.pr_labels.any? { |label| label.include?('WIP') }
-  has_wip_title = github.pr_title.include?('WIP')
+github.dismiss_out_of_range_messages
 
-  has_wip_label || has_wip_title
-end
+# `files: []` forces rubocop to scan all files, not just the ones modified in the PR
+rubocop.lint(files: [], force_exclusion: true, inline_comment: true, fail_on_inline_comment: true, include_cop_names: true)
```
mokagio, line 6: Maybe it's possible to do what in Swift would be an `extension` and I think Ruby calls [monkey patch](https://en.wikipedia.org/wiki/Monkey_patch) to add a method to the RuboCop plugin that would go like:

```ruby
rubocop.lint_all_files(force_exclusion: true, inline_comment: true, fail_on_inline_comment: true, include_cop_names: true)
```

This is of course a Dangermattic-level idea, which is out of scope for this PR.

Dangerfile
```diff
@@ -1,46 +1,31 @@
...
 
-# these are reviewers actively providing feedback and potentially changing the state of the PR (approved, changes-requested)
-def active_reviewers?
-  repo_name = github.pr_json['base']['repo']['full_name']
-  pr_number = github.pr_json['number']
+manifest_pr_checker.check_all_manifest_lock_updated
 
-  !github.api.pull_request_reviews(repo_name, pr_number).empty?
+# skip remaining checks if we're during the release process
```
mokagio: Nitpick:

```suggestion
# skip remaining checks if we're in a release-process PR
```

Dangerfile
```diff
@@ -1,46 +1,31 @@
...
 
-# requested_teams / requested_reviewers are users initially requested to review a PR, who haven't reacted yet
-def requested_reviewers?
-  has_requested_reviews = !github.pr_json['requested_teams'].to_a.empty? || !github.pr_json['requested_reviewers'].to_a.empty?
-  has_requested_reviews || active_reviewers?
-end
+view_changes_checker.check
 
-return if github.pr_labels.include?('Releases')
+pr_size_checker.check_diff_size(max_size: 500)
```
mokagio: Should we log a note about skipped checks like we do for release PRs?

The opposite suggestion would also apply. If we don't log a message for drafts, we might want to avoid it for release PRs, too. 

I can't put my finger on why, but I feel like not logging a message for release PRs would be inappropriate—maybe it's just that I'm used to see it there, though. Under that rationale, I'd like to have consistency and log a message here, too, but it's not something I feel strongly about.

Dangerfile
```diff
@@ -1,46 +1,31 @@
...
-
 milestone_checker.check_milestone_due_date(days_before_due: 2)
 
-# `files: []` forces rubocop to scan all files, not just the ones modified in the PR
-rubocop.lint(files: [], inline_comment: true, fail_on_inline_comment: true, include_cop_names: true)
-
-warn('PR is classed as Work in Progress') if wip_feature?
-
-warn("No reviewers have been set for this PR yet. Please request a review from **@\u2028Automattic/pocket-casts-ios**.") unless requested_reviewers?
+warn('PR is classed as Work in Progress') if github_utils.wip_feature?
```
mokagio, line 34: Another Dangermattic-level suggestion. I like `github_utils.wip_feature?`. Have you thought about implementing a similar `github_utils.release_pr?` syntax-sugar/helper method?

.github/workflows/run-danger.yml
```diff
@@ -2,10 +2,12 @@ name: ☢️ Danger
 
 on:
   pull_request:
-    types: [opened, synchronize, edited, review_requested, review_request_removed, labeled, unlabeled, milestoned, demilestoned]
+    types: [opened, reopened, ready_for_review, synchronize, edited, review_requested, review_request_removed, labeled, unlabeled, milestoned, demilestoned]
```
mokagio: Do we need the `review_requested` and `review_request_removed` events? 

I can't come up with something that might have changed in the PR and that wouldn't have already triggered a run that one of those two events would pick up.

Then again, it's cheap to run Dangermattic, so we might as well leave them.

mokagio (approved):

.github/workflows/run-danger.yml
iangmaia: Thanks for noticing -- I actually planned to remove them as I figured the "reviewers check" isn't necessary when PRs are auto-assigned to code owners, but forgot to remove the events.
Changed on 856bf8a12410b254a5e9f652c22a80164ffa2de5.

Dangerfile
iangmaia: Good suggestion! I've added an item to the backlog [here](https://github.com/orgs/Automattic/projects/775?pane=issue&itemId=53114939).

Dangerfile
iangmaia: Thanks! I've added it to the same item in backlog I [mentioned above](https://github.com/orgs/Automattic/projects/775?pane=issue&itemId=53114939).

Dangerfile
iangmaia: My rationale for having it in release mode but not on Drafts is mainly due to the scope of what's being skipped: most of the checks are skipped if we are in the release process, so it makes sense to make it clear. Just a couple of more superficial checks (e.g. labels) are skipped for Drafts, to avoid unnecessary noise while the PR creator just wants to try some things out.

I understand the question though. I think we can log it just incase, and if we feel it's unnecessary we can remove it later.

PR #1413, opened 2024-02-05
Fix `SyncManager` variable name in non-debug builds

bjtitus:
Fixes `SyncManager.shouldUseNewSettingsSync` variable name issue in `AppDelegate`

## To test

* Build & Run Release scheme
* Ensure that project compiles

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1419, opened 2024-02-07
Additional pipeline job for running release builds

bjtitus:
This compiles the Release configuration and produces an (unused) IPA file as a result.

## Background

Last week I added a variable which was used inside of an ifdef expression dependent on the DEBUG flag being false and required a follow-up fix (https://github.com/Automattic/pocket-casts-ios/pull/1413).

https://github.com/Automattic/pocket-casts-ios/blob/c795e4025756319d80991fca1cbc06f4dfb94286/podcasts/AppDelegate.swift#L312

The CI jobs all passed successfully and I failed to test the Release configuration locally. Since we have about 6 places where the DEBUG flag is checked, I thought it would be useful to update CI to compile this configuration.

## Q&A

**Should this produce an IPA or is that unnecessary?**
I'm not sure it adds much to the build process so I just left it with default behavior.

**Should the Fastlane lanes for building this job and building the release build (`build_and_upload_app_store_connect`) be combined?**
I didn't think so because the lane includes symbols and other things which probably aren't necessary.

**Should the produced IPA be made available in Buildkite's Artifacts?**
I don't know that it's particularly useful since it isn't signed for ad-hoc installation.

## To test

🟢 CI should be green
* CI Build should include new job for building Release configuration

## Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -129,7 +129,7 @@ platform :ios do
   ENV['FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT'] = '120'
 
   desc 'Run the unit tests'
-  lane :test do |options|
+  lane :test do
     run_tests(
       scheme: SCHEME
     )
@@ -139,11 +139,11 @@ platform :ios do
   lane :build do |options|
     configure_code_signing
 
-    CONFIGURATION = options[:configuration]
-    if CONFIGURATION
+    configuration = options[:configuration]
+    if configuration
       build_app(
         scheme: SCHEME,
-        configuration: CONFIGURATION,
+        configuration: configuration,
         include_bitcode: false,
         export_method: 'app-store'
       )
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -129,10 +129,30 @@ platform :ios do
   ENV['FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT'] = '120'
 
   desc 'Run the unit tests'
-  lane :test do
+  lane :test do |options|
     run_tests(
       scheme: SCHEME
     )
+  end
+
+  desc 'Build a given (or default) configuration'
+  lane :build do |options|
+    configure_code_signing
+
+    CONFIGURATION = options[:configuration]
+    if CONFIGURATION
+      build_app(
+        scheme: SCHEME,
+        configuration: CONFIGURATION,
+        include_bitcode: false,
+        export_method: 'app-store'
+      )
+    else
+      build_app(
+        scheme: SCHEME,
+        include_bitcode: false
+      )
+    end
   end
 
   desc 'This lane downloads and configures the code signing certificates and profiles.'
--- a/.buildkite/commands/build.sh
+++ b/.buildkite/commands/build.sh
@@ -32,5 +32,15 @@ install_gems # see bash-cache Automattic's Buildkite plugin
 echo "--- Install Pods"
 install_cocoapods # see bash-cache Automattic's Buildkite plugin
 
-echo "--- Build & Test"
-bundle exec fastlane test
+# If a build configuration is provided, we'll add build that config without testing
+if [ $# -eq 1 ]; then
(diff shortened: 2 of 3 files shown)


fastlane/Fastfile
```diff
@@ -129,12 +129,32 @@ platform :ios do
   ENV['FASTLANE_XCODEBUILD_SETTINGS_TIMEOUT'] = '120'
 
   desc 'Run the unit tests'
-  lane :test do
+  lane :test do |options|
```
dangermattic, line 132: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-ios/blob/0d9adb5eba61feebfcfef9fdfef2a2ae16d90e3d/fastlane/Fastfile#L132"></span>Lint/UnusedBlockArgument: Unused block argument - <code>options</code>. You can omit the argument if you don't care about it.</td>
    </tr>
  </tbody>
</table>

fastlane/Fastfile
```diff
@@ -129,12 +129,32 @@ platform :ios do
...
     run_tests(
       scheme: SCHEME
     )
   end
 
+  desc 'Build a given (or default) configuration'
+  lane :build do |options|
+    configure_code_signing
+
+    CONFIGURATION = options[:configuration]
```
dangermattic, line 142: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-ios/blob/0d9adb5eba61feebfcfef9fdfef2a2ae16d90e3d/fastlane/Fastfile#L142"></span>Lint/ConstantDefinitionInBlock: Do not define constants this way within a block.</td>
    </tr>
  </tbody>
</table>

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso (changes requested):
@bjtitus Even though there are 5 places where we use `if DEBUG` the only one your change would test is the `#if !DEBUG` inside `updateRemoteFeatureFlags()`.

I think there's an easier way to prevent issues like that and catch them even before committing, we can do something like that:

```swift
    private var isDebug: {
        #if DEBUG
        true
        #else
        false
        #endif
    }
```

And then instead of using the macro, use a regular Swift `if` checking the `isDebug`. If we had this in place the issue on your PR wouldn't have happened. :)

Also, this is running the pipeline twice: one for testing and one for Release. Even though both seem to happen simultaneously I do worry about the costs of pretty much doubling our CI usage — we'd need to involve people from Apps Infra if we want to move forward with this solution.

bjtitus:
@leandroalonso Thanks for the quick check!

> instead of using the macro, use a regular Swift if checking the isDebug. If we had this in place the issue on your PR wouldn't have happened

Ah, this is an interesting solution. I think that could be a good compromise here. The other `DEBUG` usage is really just adjusting strings and lists of types so no code divergence.

> this is running the pipeline twice: one for testing and one for Release. Even though both seem to happen simultaneously I do worry about the costs of pretty much doubling our CI usage — we'd need to involve people from Apps Infra if we want to move forward with this solution.

I was planning to publish a P2 post and get some feedback from Apps Infra before moving forward. I'm not sure about the CI costs but the 7 minute builds are significantly lower than the 20+ minute builds we had on Tumblr.

Maybe we should start with the approach you mentioned and revisit if we have further need for a Release build in the future?

leandroalonso:
> Maybe we should start with the approach you mentioned and revisit if we have further need for a Release build in the future?

If you agree, let's do it!

bjtitus:
Closing in favor of https://github.com/Automattic/pocket-casts-ios/pull/1420

PR #1426, opened 2024-02-09
[Deselecting Chapters] Playback

leandroalonso:
| 📘 Part of: #1396 | Depends on #1399 |
|:---:|:---:|

Handles skipping chapters when they are deselected.

https://github.com/Automattic/pocket-casts-ios/assets/7040243/ba3be641-1182-4e8e-ba4a-5e6c72b5ebdf

A piece of important information regarding chapters:

1. Chapters can overlap. All they need is to have a start time and a duration. Some chapters might even end after the total duration of the episode if the producer wants;
2. Chapters can be visible or not. We only show visible chapters

Also, for this PR I decided to not change how chapters work in `PlaybackManager` to avoid introducing issues. I'll revisit that in a future PR.

Also, even though the changes here are not under a feature flag, they should only work if a chapter has the `shouldPlay` property set to `false`, which is only possible when the feature is enabled.

## To test (Clublife)

For this podcast, the last chapter doesn't end the episode.

1. Play an episode with chapters (use any episode of `Clublife`)
4. On the full player, go to "Chapters" tab and deselect some chapters
5. Go back to the "Now Playing Tab"
6. ✅ Use the chapter controls to move forward/backwards and check that the chapters you skipped are not played
7. Restart the playback
8. Deselect all chapters except the first and let the first chapter play
9. ✅ After the first chapter finishes the playback moves forward but doesn't finish the episode — this is expected, the last chapter doesn't finish at the end of the episode

## To test (ATP)

1. Play any episode of [ATP](https://pca.st/accidentaltech)
4. On the Chapters tab, toggle one chapter to play, and the other to not
5. Go back to "Now Playing"
6. ✅ Use the chapters controls and make sure the deselected chapters are not played

## Additional tests

Please do any manual additional tests you can think of. :)

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

Issue #1396, opened 2024-02-02
Deselecting Chapters


leandroalonso:
Add the ability to deselect chapters and sync them between devices.

### Basics

- [x] Feature Flag #1397
- [x] Update the chapter UI to include way to choose which chapters to play #1399
- [x] While playing an episode and coming to the end of a chapter, skip to the next available chapter position #1426
- ~[ ] If the user scrubs to a playback position that is contained in a deselected chapter, playback should proceed as normal until reaching the end of that chapter~
- [x] Final design #1460 

### Paid Feature

- [x] Add the locked UI if the feature is not available #1449
- [x] Show the upsell when tapped #1449
- [x] Enable the feature for Patron under early access #1449
- [x] Then Plus for full access #1449

### Syncing

- [x] Send the selections to the server #1480
- [x] Apply incoming selections from the server #1480

### Watch

- [x] Skip deselected chapters on the watch when tapping the chapter next/previous buttons #1480

### Other

- [x] Add What’s New #1473
- [x] Add the feature to the plus features list #1467 
- [x] Tracks #1462

rviljoen:
@leandroalonso Nice job. This is working very well 👌🏻

PR #1431, opened 2024-02-12
[Deselecting Chapters] Persistence

leandroalonso:
| 📘 Part of: #1396 | Depends on #1426 |
|:---:|:---:|

Persists users' deselected chapters on the database.

These are podcasts with chapters you can use to test:

* Clublife: https://pca.st/rLzZCv
* ATP: https://pca.st/accidentaltech

## To test

2. Run the app
3. Play any episode with chapters
4. Deselect some chapters
5. Close and reopen the app
6. ✅ Your chapter list should reflect your previous selection
7. Toggle between episode with chapters and change their selections
8. ✅ Your chapter selection should be correctly reflected

## Checklist


podcasts/PlayerChapterCell.swift
```diff
@@ -111,6 +112,16 @@ class PlayerChapterCell: UITableViewCell {
...
         toggleChapterButton.currentlyOn.toggle()
+
+        if let currentEpisode = PlaybackManager.shared.currentEpisode(), let index = chapter?.index {
+            if chapter?.shouldPlay == true {
+                currentEpisode.select(chapterIndex: index)
+            } else {
+                currentEpisode.deselect(chapterIndex: index)
+            }
+
+            DataManager.sharedManager.save(episode: currentEpisode)
```
leandroalonso, line 123: I thought about moving this call to `BaseEpisode`. However, the UI design will change and I believe the logic on where we save users' selection will probably change, so I decided to just leave that here for now.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
```diff
@@ -656,6 +656,16 @@ class DatabaseHelper {
             }
         }
 
+        if schemaVersion < 43 {
```
leandroalonso, line 659: Ideally, we would put this behind a feature flag. However, for deselecting chapters, we have a remote feature flag, so this is not possible. We could, for example, run into the flag being enabled after migration has happened and the feature would be enabled without its corresponding database change.

Since this is a simple column addition, I think it's safe to have it.

SergioEstevao (approved):
I figured out that you needed to activate a feature flag to see the selection, but after that all is working well!
:shipit:

leandroalonso:
@SergioEstevao ops! Sorry for the missing step. 🙈

PR #1439, opened 2024-02-15
[What's New] Change how the announcements are return to allow remotely enabling them

leandroalonso:
The previous What's New code relied on checking the previous opened version, which meant that an announcement that was enabled after the app was updated would never show.

This PR updates this behavior.

## To test

1. Do a clean install of `trunk` and run it
2. Check this branch and run it
3. ✅ No announcement should show
4. Go to Profile > Settings > Beta Features and enable `slumber`
6. Re-run the app
7. ✅ The What's New should show
8. Re-run the app
9. ✅ No announcement should show

## Checklist

### DIFF
--- a/podcasts/Whats New/WhatsNew.swift
+++ b/podcasts/Whats New/WhatsNew.swift
@@ -76,6 +76,8 @@ class WhatsNew {
         // Don't show any announcements if this is the first run of the app,
         // or if we've already checked the what's new for this version
         guard let previousOpenedVersion else {
+            // Set the lastWhatsNewShown so it doesn't run after the app is reopened
+            Settings.lastWhatsNewShown = currentVersion
             return nil
         }
 
--- a/podcasts/Whats New/WhatsNew.swift
+++ b/podcasts/Whats New/WhatsNew.swift
@@ -75,7 +75,7 @@ class WhatsNew {
     var visibleAnnouncement: Announcement? {
         // Don't show any announcements if this is the first run of the app,
         // or if we've already checked the what's new for this version
-        guard let previousOpenedVersion, previousOpenedVersion != currentVersion else {
+        guard let previousOpenedVersion, lastWhatsNewShown != currentVersion else {
             return nil
         }
 
@@ -87,7 +87,7 @@ class WhatsNew {
             .last(where: {
                 $0.isEnabled() &&
                 $0.version != lastWhatsNewShown &&
-                $0.version.inRange(of: previousOpenedVersion, upper: currentVersion)
+                $0.version.inRange(of: lastWhatsNewShown ?? "0", upper: currentVersion)
             })
     }
 }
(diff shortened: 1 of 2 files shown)


podcasts/Whats New/WhatsNew.swift
```diff
@@ -87,7 +87,7 @@ class WhatsNew {
             .last(where: {
                 $0.isEnabled() &&
                 $0.version != lastWhatsNewShown &&
-                $0.version.inRange(of: previousOpenedVersion, upper: currentVersion)
+                $0.version.inRange(of: lastWhatsNewShown ?? "0", upper: currentVersion)
```
leandroalonso: The `?? 0` feels a bit odd, any suggestion is very much appreciated.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/40">7.57 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

emilylaguna (changes requested):
It looks like the changes are causing the CI tests to fail, which I can also reproduce on my local machine. 

Another thing I noticed is that if I installed `7.56` and then update to this version, I see the bookmarks announcement when previously I did not. 

I have an alternate idea that keeps our existing logic intact, but also gives us the flexibility to show the what's new after the first launch after an update. 

The idea is: 
- Add a "forcefully show what's new" flag whose value is the what's new version that we want to show
- When we retrieve the remote feature flag, we'll check to see if the local flag was enabled (before == false, after == true) and if it was, we'll then check if the slumber what's new hasn't been shown yet
- If we haven't, set the forcefully show version to the slumber version
- On the next launch, the what's new will check if the force show flag is set, ignore the rest of the logic, and will instead retrieve the announcement for the forced version.
   - I think we should also check if it's enabled here, just in case
- After showing a forced announcement, the whats new will then clear the force show flag to prevent showing repeatedly,  defaulting back to the original logic.

What do you think?

emilylaguna (commented):
I pushed up a quick concept of this here: https://github.com/Automattic/pocket-casts-ios/compare/release/7.57...emily/concept/whats-new-force

emilylaguna (approved):
Nice work. This is working in all my tests:

- [x] Fresh install 7.56 → 7.57 with the remote flag disabled doesn't show the what's new
- [x] Remote flag enabled, relaunching 7.57 shows the what's new
- [x] Fresh install 7.57 with the remote flag enabled doesn't show the what's new
- [x] Fresh install 7.57 with remote flag disabled doesn't show the what's new
- [x] Relaunch after fresh install and the remote flag enabled doesn't show the what's new

PR #1441, opened 2024-02-15
Settings Sync: Privacy Flags

bjtitus:
| 📘 Part of: #1400 |
|:---:|

* Adds synced settings flag for privacy (crash & log tracking), Marketing Opt-In (Newsletter), and Free Gift Acknowledgement flags.
* Moves `FeatureFlags` handling to `Utils`. This isn't strictly necessary for this PR, but it does make replacing the settings a bit easier. It will be necessary for future up next settings.

### Privacy

1. D1: Navigate to Settings > Privacy
2. D1: Change the "Collect Information" switch
3. D1: Go back to Profile and refresh
4. D2: Refresh the Profile
5. D2: Navigate to Settings > Privacy
6. D2: Check the "Collect Information" switch

### Marketing

1. D1: Navigate to Account
2. D1: Change the "Pocket Casts Newsletter" switch a bit below the fold
3. D1: Go back to Profile and refresh
4. D2: Refresh the Profile
5. D2: Navigate to Account
6. D2: Check the "Pocket Casts Newsletter" switch

### Gift Acknowledgement

This setting doesn't really have any visible effect and isn't read from the device.

### Import Old Settings

1. Reinstall the app from `trunk`, or this branch also works as long as the app is reinstalled.
2. Change one of the settings mentioned above (Collect Information for example)
3. Build from this branch if build from trunk
4. Enable the `settingsSync` flag
5. Restart the app
6. Ensure that your settings change remains

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso (changes requested):
@bjtitus the project is not building:

Screenshot 2024-02-15 at 16 45 27

bjtitus:
@leandroalonso Sorry, just pushed two changes to fix app and unit test compilation.

leandroalonso (changes requested):
@bjtitus I'm facing the same issue as the shelf PR:

1. I run `trunk`
2. Set Collection information to off (Profile > Privacy)
3. Run this branch with the flag enabled
4. It's turned on

I think you mentioned about doing this part of "carrying" the current option in the future. But I do wonder if that's the best course of action because testing each property will be a lot.

For each property, I think we should:

1. Ensure it's syncing correctly (testing steps already written in the PRs)
2. Ensure when it's updated it reflects the correct behavior
3. Ensure when a user is migrating to this version with the flag disabled that their previous value is still the same
4. Ensure when a user is migrating to this version with the flag enabled that their previous value is still the same

I mention that because one thing that infuriates users is messing with their options/customizations. 😬

Let me know what do you think it's the best way to move forward!

bjtitus:
@leandroalonso I believe this is due to the import only running a single time after the feature flag has been enabled. My thought was that no user would need another import once we have completely moved all settings over and that we wouldn't enable this flag until that was complete. Trying to avoid a situation where there's a bug with the flag value and settings are overwritten by their older counterparts.

https://github.com/Automattic/pocket-casts-ios/blob/783c83b140b043f85bab5ae1a5f346646716d6aa/podcasts/AppDelegate%2BDefaults.swift#L129-L134

For testing, this means you need to reinstall the app (or wipe `MigrateToSyncedSettings`) to retrigger the old defaults import.

I do think this behavior can be updated to reset `MigrateToSyncedSettings` when the `settingsSync` flag is _disabled_ so the import has a chance to run again since we don't want a situation where the flag is disabled in production and settings changes are lost when the `settingsSync` flag is re-enabled. However, this doesn't really help with a problem where some settings need to be changed or re-imported at a later time. Presumably, this would be a new migration job in `AppDelegate+Defaults` but I don't want to make one for every single PR.

I'll start work on clearing the `MigrateToSyncedSettings` when the `settingsSync` flag is disabled if that makes sense to you.

leandroalonso:
@bjtitus 

> I believe this is due to the import only running a single time after the feature flag has been enabled. My thought was that no user would need another import once we have completely moved all settings over and that we wouldn't enable this flag until that was complete.

Oh got it! 😌 

>  since we don't want a situation where the flag is disabled in production and settings changes are lost when the settingsSync flag is re-enabled.

This is a very good observation!

> I'll start work on clearing the MigrateToSyncedSettings when the settingsSync flag is disabled if that makes sense to you.

Up to you! I think just adding the context about that on the PRs is enough for now.

bjtitus:
> I think just adding the context about that on the PRs is enough for now.

Sounds good! I just added an "Import Old Settings" section to be clear about those steps. I'll try to add some more unit tests for these like I did for Player Shelf. It is easy to forget to add the setting to several of these spots. 😅
Maybe I should get a macro added sooner rather than later. 😆

Modules/Server/Sources/PocketCastsServer/Public/ServerSettings.swift
```diff
@@ -129,12 +129,19 @@ public class ServerSettings {
     // MARK: Marketing Opt In
 
     public class func setMarketingOptIn(_ value: Bool) {
+        if FeatureFlag.settingsSync.enabled {
+            SettingsStore.appSettings.marketingOptIn = value
+        }
         UserDefaults.standard.set(value, forKey: ServerConstants.UserDefaults.marketingOptInKey)
```
leandroalonso, line 135: Is this not under a `else` because this value is what actually opts-in/out for the marketing communication?

leandroalonso (approved):
@bjtitus everything worked as expected however I ran into an issue with the "migration":

1. Run the app on D1 with feature flag enabled
2. Set both flags to false
3. Sync
4. Fresh-run the app on D2
5. Keep both flags disabled
6. Enable the feature flag
7. Re-run the app
8. Check that the flags are enabled
9. Sync
10. Go to D1, sync
11. Check that the flags didn't reflect the enabled one

I was able to change the flags only after toggling them twice. Not a blocker but if you can reproduce it worths creating an issue for it.

Modules/Server/Sources/PocketCastsServer/Public/ServerSettings.swift
bjtitus: I've been keeping all settings writing outside of the flag. This allows us to disable the flag and not have to migrate from new to old settings, only old to new.

leandroalonso: Oh nice!

bjtitus:
> 5. Keep both flags disabled
> 6. Enable the feature flag
> 7. Re-run the app
> 8. Check that the flags are enabled

In testing, I didn't see any issues with the flags syncing after the FF was enabled. However, I think there is an issue with importing the previous values from `UserDefaults.bool` when it returns `false` for unset keys. There are many values which just default to `false`, but this one is inverting the value so that the `false` value `UserDefaults` returns in the case of `nil` is not correct.

I filed an issue (https://github.com/Automattic/pocket-casts-ios/issues/1461) and will address separately since this affects several other flags.

PR #1444, opened 2024-02-16
Settings Sync: Player Shelf

bjtitus:
| 📘 Part of: #1400 |
|:---:|

* Adds synced setting for Player Shelf actions.
* Adds an `Option` type with a `known` and `unknown` case to represent values which we know about and those which we don't. We want to preserve the unknown values to send back to the server.

https://github.com/Automattic/pocket-casts-ios/blob/82d6e59ff96fad321899fc43060156e3df5bc7be/Modules/DataModel/Sources/PocketCastsDataModel/Public/Enums.swift#L64-L67

* The `PlayerAction` enum has been changed to use a `String` `rawValue`, so also include an initializer to convert the old values.

https://github.com/Automattic/pocket-casts-ios/blob/82d6e59ff96fad321899fc43060156e3df5bc7be/Modules/DataModel/Sources/PocketCastsDataModel/Public/Enums.swift#L93-L131

## To test

1. D1: Navigate to the Player
2. D1: Tap the ⋯ icon to rearrange shelf actions
3. D1: Move actions in and out of the top set and around in the list
4. D1: Return to Podcasts list & pull to refresh
5. D2: Pull to refresh Podcasts list
6. D2: Open the player
7. D2: Ensure that shelf actions match

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/41">7.58</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

leandroalonso (changes requested):
When the flag is disabled and I move an item to be inside the "Shortcut on Player" it never works:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/5e36fdf2-1630-4bc1-8168-d734f5713b10

Also, if I run `trunk` and change the order of the items and then run this branch with the flag enabled, all my customizations are gone.

bjtitus:
@leandroalonso Both of these should be fixed.

> [!NOTE]
> You will need to reinstall the app because the import is a one-time job at the moment which runs a single time after the feature flag is change. See [this comment](https://github.com/Automattic/pocket-casts-ios/pull/1441#issuecomment-1948913082) for more details.

I added the old `intValue` for the old setting reading code and a unit test to cover it.

https://github.com/Automattic/pocket-casts-ios/blob/f8c514c93fc6f6ec451ea8e90617ac47a5a8fc1c/podcasts/Enumerations.swift#L205-L228

https://github.com/Automattic/pocket-casts-ios/blob/f8c514c93fc6f6ec451ea8e90617ac47a5a8fc1c/podcasts/Settings.swift#L641

Also added the defaults import and unit tests for that as well.

https://github.com/Automattic/pocket-casts-ios/blob/f8c514c93fc6f6ec451ea8e90617ac47a5a8fc1c/podcasts/AppSettings%2BImportUserDefaults.swift#L25

leandroalonso (approved):
@bjtitus all works as described. I did an additional test that I just want to make sure it's working as intended:

1. Run D1 with feature flag enabled
2. Re-arrange the self
3. Sync
4. Run D2 without feature flag enabled
5. Change the shelf
6. Enable the feature flag
7. Re-run the app on D2
8. Sync
9. Go back to D1
10. Sync
11. ✅ The previous settings are still there

The reason I tested that is that I think we can potentially run into someone running Pocket Casts on a device they don't use very frequently and we need to be smart enough to not "reset" al their settings.

Based on my tests it seems to work that way though?

bjtitus:
> ✅ The previous settings are still there

@leandroalonso The previous settings here is D1s changed settings, right? So the imported D2 settings are overridden by the D1 changed setting. If that's what you saw, it sounds like the expected behavior. Any settings modified after the `settingsSync` enabled update should win as they are the only settings we have modified dates for.

To quote some discussions we've had internally on this. This came down to two different options:

> * Either we’re treating Device A (the first device to update to the new version) as the source of truth, and replacing the settings of any subsequent devices that are updated
> * Or each subsequent device updated is treated as the source of truth, and replaces the settings of the previously updated devices settings

I do think this brings up some confusion that the setting was changed on D2 more recently on D1 but we have no way of knowing this prior to these changes since there is no Modified Date associated with the change. Maybe there's more user-facing messaging we can do to differentiate versions which have this syncing applied and those that don't?

PR #1448, opened 2024-02-16
Settings Sync: Headphone Controls

bjtitus:
| 📘 Part of: #1400 |
|:---:|

Adds synced setting for headphone controls

## To test

> [!IMPORTANT]
> Staging configurations must be used. These settings are new and will not sync in production.

### Import Old Settings

1. Reinstall the app from `trunk`, or this branch also works as long as the app is reinstalled.
2. Change one of the settings mentioned above (Collect Information for example)
3. Build from this branch if build from trunk
4. Enable the settingsSync flag
5. Restart the app
6. Ensure that your settings change remains

### Setting toggles

1. Navigate to Settings
2. Change the "Headphone Controls" actions
3. Ensure that settings stick
4. Perform "Headphone controls" test if you wish

### Setting Sync

1. D1: Navigate to Settings
2. D1: Change the "Headphone Controls" actions
3. D1: Navigate back to the Profile & Pull to Refresh
5. D2: Pull to Refresh
6. D2: Navigate to Settings
7. D2: Verify the "Headphone Controls" actions

### Headphone controls

After changing the settings above, pair bluetooth headphones and ensure the correct actions are performed.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/ServerEnums.swift
```diff
@@ -66,3 +66,11 @@ public enum PrimaryRowAction: Int32, Codable {
 public enum PrimaryUpNextSwipeAction: Int32, Codable {
     case playNext = 0, playLast = 1
 }
+
+public enum HeadphoneControl: Int32, Codable {
```
leandroalonso, line 70: In the case a new action is added here, does the sync handle that well?

leandroalonso (approved):
@bjtitus I had an issue while testing but I wasn't able to reproduce it. After I installed and logged into a new account with the flag enabled, the headphone controls weren't synced. Other settings synced just fine.

After I changed the headphone settings on the other devices and synced, then I was able to see it. Any idea what could cause this?

Maybe my initial syncing on the other device never really actually happened?

Modules/Server/Sources/PocketCastsServer/Public/ServerEnums.swift
bjtitus: I made the conversion between `HeadphoneControl` and the internal `HeadphoneControlAction` exhaustive so it should be clear when a new value is added to update these.

https://github.com/Automattic/pocket-casts-ios/blob/16a548773f58259ae282ea7a0b663d9de30eb888/podcasts/Settings.swift#L1050-L1080

The values do need to be identical between devices which is just defined in our internal documentation. To be clear about the values, [I've added](https://github.com/Automattic/pocket-casts-ios/pull/1448/commits/47a5d31afe4f952b272c73921d9e5f86ea3e45a3) the `rawValue` so it's more obvious if we add them in different orders.

bjtitus:
> After I installed and logged into a new account with the flag enabled, the headphone controls weren't synced. Other settings synced just fine.

> After I changed the headphone settings on the other devices and synced, then I was able to see it. Any idea what could cause this?

I tried this a few times on devices and wasn't able to reproduce. Maybe the defaults weren't synced?

In any case, I don't think it's specific to these settings. I'll follow up on this as part of our larger E2E testing with all settings.

PR #1454, opened 2024-02-19
Settings Sync: Up Next

bjtitus:
| 📘 Part of: #1400 |
|:---:|

Adds General and Podcast specific Up Next settings.

## Changes

Most of this is similar to the rest of the synced settings changes. I wanted to call out a few things to get feedback on, particularly where SQL queries are replaced with manual iteration.

#### AutoAddCandidates

* The Query for `AutoAddCandidates` has been updated to fetch `settings` instead of the `autoAddToUpNext` column.

https://github.com/Automattic/pocket-casts-ios/blob/c730a29b01ba942394f990efb795249d83a8ea2b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/AutoAddQueueDataManager.swift#L57-L70

* The `AutoAddCandidate` initializer has been updated to handle this:

https://github.com/Automattic/pocket-casts-ios/blob/c730a29b01ba942394f990efb795249d83a8ea2b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/AutoAddQueueDataManager.swift#L120-L126

#### PodcastDataManager

These changes switch from making a SQL `UPDATE` to manually iterating and saving the Podcast Settings. This _could_ impact performance with a large number of subscribed podcasts. SQLite (as of iOS 16+) does have [JSON functions](https://www.sqlite.org/json1.html), including an insert, which could be used for extra speed.

* The `saveAutoAddToUpNext(podcastUuid:)` method has been updated to just write to a podcast instead of making a SQL query.

https://github.com/Automattic/pocket-casts-ios/blob/df74ad3fffedb2d9b4e15d95e2f5bfe2aa0b5e27/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift#L376-L389

* The `saveAutoAddToUpNextForAllPodcasts‎ ` method has been updated to iterate over podcasts and save each one with the modified settings values instead of an `UPDATE` SQL query.

https://github.com/Automattic/pocket-casts-ios/blob/df74ad3fffedb2d9b4e15d95e2f5bfe2aa0b5e27/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift#L431-L443

* The `updateAutoAddToUpNext` method has been updated to iterate over podcasts and save each one with the modified settings values instead of an `UPDATE` SQL query.

https://github.com/Automattic/pocket-casts-ios/blob/df74ad3fffedb2d9b4e15d95e2f5bfe2aa0b5e27/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift#L451-L463

#### Podcast Settings

1. Log in to two devices or simulators
2. D1: Enable Add to Up Next setting in a podcast
3. D1: Pull to refresh in the Podcast list
4. D2: Pull to refresh in the Podcast list
5. D2: Verify that the setting appears as expected

#### General Settings

1. Log in to two devices or simulators
2. D1: Navigate to the "Auto Add to Up Next" section in Settings
3. D1: Change properties in this section
4. D1: Pull to refresh in the Profile or Podcast list
5. D2: Pull to refresh in the Profile Podcast list
6. D2: Navigate to the "Auto Add to Up Next" section in Settings
7. D2: Verify that the setting appears as expected

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

Modules/DataModel/Tests/PocketCastsDataModelTests/AutoAddCandidatesDataManagerTests.swift
```diff
@@ -0,0 +1,93 @@
...
+import PocketCastsUtils
+import SQLite3
+@testable import PocketCastsDataModel
+
+final class AutoAddCandidatesDataManagerTests: XCTestCase {
+
+    private var overriddenFlags = [FeatureFlag: Bool]()
+
+    private func override(flag: FeatureFlag, value: Bool) throws {
+        overriddenFlags[flag] = flag.enabled
```
leandroalonso: There's a `FeatureFlagMock` we can use that does that. I think we can use this among the different unit tests.

Modules/DataModel/Tests/PocketCastsDataModelTests/AutoAddCandidatesDataManagerTests.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    private func reset(flag: FeatureFlag) throws {
+        if let oldValue = overriddenFlags[flag] {
+            try FeatureFlagOverrideStore().override(flag, withValue: oldValue)
+        }
+    }
+
+    private func setupDatabase() throws -> DataManager {
+        let dbPath = (DataManager.pathToDbFolder() as NSString).appendingPathComponent("podcast_testDB.sqlite3")
```
leandroalonso: Do we need to expose `pathToDbFolder` to be used here? Can't we just use any other path?

Modules/Server/Sources/PocketCastsServer/Public/ServerSettings.swift
```diff
@@ -295,21 +295,35 @@ public class ServerSettings {
 
     private static let autoAddLimitKey = "AutoAddToUpNextLimit"
     public class func autoAddToUpNextLimit() -> Int {
-        UserDefaults.standard.integer(forKey: autoAddLimitKey)
+        if FeatureFlag.settingsSync.enabled {
+            return Int(SettingsStore.appSettings.autoUpNextLimit)
+        } else {
+            return UserDefaults.standard.integer(forKey: autoAddLimitKey)
+        }
```
leandroalonso: [nit]

```suggestion
        if FeatureFlag.settingsSync.enabled {
           Int(SettingsStore.appSettings.autoUpNextLimit)
        } else {
           UserDefaults.standard.integer(forKey: autoAddLimitKey)
        }
```

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/AutoAddQueueDataManager.swift
```diff
@@ -95,7 +115,15 @@ public struct AutoAddCandidatesDataManager {
         public let episodeUuid: String
 
         init?(from resultSet: FMResultSet) {
-            let setting = resultSet.int(forColumn: Constants.autoAddSettingColumnName)
+
+            let setting: Int32
+            if FeatureFlag.settingsSync.enabled {
+                let settingsString = resultSet.string(forColumn: Constants.settingsColumnName)
+                let settings: PodcastSettings? = DBUtils.convertData(value: settingsString?.data(using: .utf8))
```
leandroalonso: This makes me a bit worried, that for every candidate we'd convert the `PodcastSettings`. We have some users with a crazy amount of podcasts, so I wonder the impact this might have.

I was thinking how we could potentially solve that... question: the `AutoAddCandidates` table will still be updated, right? Do we need to update the code here to use the new settings?

What I'm wondering is if it would be possible to keep things here as they were and mainly keep `AutoAddCandidates` in sync with the new setting. The new setting would serve just to sync, but the logic in how it works would be kept the same (using the `AutoAddCandidates` table and the current query).

I think this will probably mitigate the risk of performance hits.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
```diff
@@ -415,6 +428,18 @@ class PodcastDataManager {
     }
 
     func saveAutoAddToUpNextForAllPodcasts(autoAddToUpNext: Int32, dbQueue: FMDatabaseQueue) {
+        if FeatureFlag.settingsSync.enabled {
+            if let setting = AutoAddToUpNextSetting(rawValue: autoAddToUpNext) {
+                let podcasts = DataManager.sharedManager.allPodcasts(includeUnsubscribed: false)
+                // Could be replaced with SQLite JSON operation in iOS 16+
```
leandroalonso: I'm not super worried about that as it is an operation done just once when we change settings.

However, maybe we can go ahead and add the iOS 16 operation? If it's too much work I guess we can leave a comment with more info here (a link to docs or something like that)

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/AutoAddQueueDataManager.swift
bjtitus: > the AutoAddCandidates table will still be updated, right? Do we need to update the code here to use the new settings?

We could continue to use the `autoAddToUpNext` but that would deviate from storing everything in the `settings` property and cause problems when we go to remove the old settings properties in the future.

`AutoAddCandidates` is kept up to date with which podcasts are available to Up Next, but the position for the Up Next queue is fetch from the `SJPodcasts` table.

I did run some performance tests and it's about 5 times slower on a set of 500 podcasts. Although, the fetch time is still about half a second, so I'm not sure if it's premature optimization.

bjtitus: Updating this to use `json_extract` gets the `candidates` fetch down to 1.5 times slower over 500 podcasts.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
leandroalonso: Can we change here to use `json_extract`?

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
```diff
@@ -423,6 +448,13 @@ class PodcastDataManager {
             do {
                 let uuids = podcasts.map { $0.uuid }
 
+                if FeatureFlag.settingsSync.enabled {
+                    // Could be replaced with SQLite JSON operation in iOS 16+
```
leandroalonso: Same as above.

leandroalonso (changes requested):
@bjtitus There are still comments on using `json_extract` on the code. I guess we can make this change now.

I re-reviewed and everything looks good on the code, I'm gonna let these changes run on some real devices to make sure everything is working as expected.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
bjtitus: 👍 Updated to use `json_patch`

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
bjtitus: 👍 Updated to use `json_patch`

leandroalonso (approved):
@bjtitus thanks for the patience on this one. I tested the syncing and the episodes being added, both with the flag enabled and without the flag, it all works as described.

I still faced some issues as in other branches: sometimes some values wouldn't update. I have to go there, change it again, and only then it's synced. I'm still not able to provide realiable steps to reproduce that though. :(

PR #1456, opened 2024-02-19
 Update search for podcast code to handle polling results

### LINKED ISSUE #1457: Adding new feed sometimes fails saying no search results
### Description

When adding a new feed on the app the adding fails and is disguised as a failed search

1. Go to Podcasts tab
2. Add a new podcast url on the search field ( preferably something large with a lot of episodes/item)
3. Wait for a bit
4. See that the search failed
5. Try again and you will see that the search will show up with results ( meanwhile the server was able to complete it)

_No response_

iPhone 15 Pro

SergioEstevao:
Fixes #1457 

Update search for podcast code to handle polling results
The poll can happen when a rss feed is so large that our servers takes a while to process it.

## To test

- Start the app
- Open the Podcasts tab
- Insert the url of a new podcast to be added to the server. The best way of doing this is to get a large RSS feed file from a know podcast ( the Daily for example) and store in a service like [Pastebin](https://pastebin.com) or a Gist (you can copy [this one](https://gist.github.com/SergioEstevao/e1ab9e690c3de504c34f6c43620d276a)). Then copy/past the URL for the raw of the file in the search box and tap Enter
- Wait for the search to conclude, if all goes well the podcast should be added to our server and show as a result
- If you want to see if the poll process happens put a break point on `PodcastSearchTask` line 97.

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Refresh/PodcastSearchOperation.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Refresh/PodcastSearchOperation.swift
@@ -57,7 +57,7 @@ class PodcastSearchOperation: Operation {
         dispatchGroup.enter()
         URLSession.shared.dataTask(with: request) { data, _, error in
             guard let data = data, error == nil else {
-                shouldRetry = false
+                shouldRetry = true
                 self.completion(PodcastSearchResponse.failedResponse())
                 self.dispatchGroup.leave()
 
--- a/Modules/Server/Sources/PocketCastsServer/Public/Refresh/PodcastSearchOperation.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Refresh/PodcastSearchOperation.swift
@@ -58,9 +58,7 @@ class PodcastSearchOperation: Operation {
         URLSession.shared.dataTask(with: request) { data, _, error in
             guard let data = data, error == nil else {
                 shouldRetry = true
-                self.completion(PodcastSearchResponse.failedResponse())
                 self.dispatchGroup.leave()
-
                 return
             }
 
--- a/Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
@@ -5,10 +5,6 @@ struct PodcastsSearchEnvelope: Decodable {
     let status: String
     let message: String?
     let result: PodcastsSearchEnvelopeResult
-
-    static var failedResult: PodcastsSearchEnvelope {
-        return PodcastsSearchEnvelope(status: "failed", message: "", result: .empty)
-    }
 }
 
 struct PodcastsSearchEnvelopeResult: Decodable {
@@ -20,10 +16,6 @@ struct PodcastsSearchEnvelopeResult: Decodable {
 
     /// The poll uuid if the result is still being processed on the server
     let pollUuid: String?
-
-    static var empty: PodcastsSearchEnvelopeResult {
-        return PodcastsSearchEnvelopeResult(podcast: nil, searchResults: [], pollUuid: nil)
-    }
 }
 
 public struct PodcastFolderSearchResult: Codable, Hashable {
@@ -81,7 +73,7 @@ public class PodcastSearchTask {
     }
 
     public func search(term: String) async throws -> [PodcastFolderSearchResult] {
-        var envelope: PodcastsSearchEnvelope = .failedResult
+        var envelope: PodcastsSearchEnvelope?
         var retry = true
         var pollCount = 0
         while retry {
@@ -89,7 +81,7 @@ public class PodcastSearchTask {
             // Check if status of search is poll, if it's polled we will repeat the call after x amount of secs.
             pollCount += 1
             let backOffTime = pollBackoffTime(pollCount: pollCount)
-            guard envelope.status == "poll", backOffTime > 0 else {
+            guard envelope?.status == "poll", backOffTime > 0 else {
                 retry = false
                 continue
             }
@@ -97,10 +89,10 @@ public class PodcastSearchTask {
             try await Task.sleep(nanoseconds: backOffTime)
         }
 
-        if let podcast = envelope.result.podcast {
+        if let podcast = envelope?.result.podcast {
             return [podcast]
         } else {
-            return envelope.result.searchResults ?? []
+            return envelope?.result.searchResults ?? []
         }
     }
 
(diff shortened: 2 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/41">7.58</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
```diff
@@ -66,18 +79,53 @@ public class PodcastSearchTask {
...
+            envelope = try await search(term: term)
+            // Check if status of search is poll, if it's polled we will repeat the call after x amount of secs.
+            pollCount += 1
+            let backOffTime = pollBackoffTime(pollCount: pollCount)
+            guard envelope.status == "poll", backOffTime > 0 else {
+                retry = false
+                continue
+            }
+
+            try await Task.sleep(nanoseconds: backOffTime)
```
SergioEstevao, line 89: Good place to a breakpoint to see if the poll code is being used.

leandroalonso:
@SergioEstevao can you update the `CHANGELOG` please?

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
```diff
@@ -2,15 +2,28 @@ import Foundation
 import PocketCastsDataModel
 
 struct PodcastsSearchEnvelope: Decodable {
+    let status: String
```
leandroalonso, line 5: Are we sure this is a `String` that will always be there? In `PodcastSearchResponse` this seems to be optional, it might cause decoding issues although it seems everything is working as expected.

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
```diff
@@ -66,18 +79,53 @@ public class PodcastSearchTask {
     public init(session: URLSession = .shared) {
         self.session = session
     }
+
     public func search(term: String) async throws -> [PodcastFolderSearchResult] {
+        var envelope: PodcastsSearchEnvelope = .failedResult
```
leandroalonso: Do we need that? Can't this just be optional?

This way we can remove the added static stuff.

leandroalonso (changes requested):
@SergioEstevao works like a charm! Left a few comments about some improvements. :)

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
SergioEstevao: I was trying to have the same results as the old operation in order to be able to replace the operation code later on.

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
SergioEstevao: From my tests, there always there, is the an API spec that we could double check? or maybe a link to the server code?

Modules/Server/Sources/PocketCastsServer/Public/Refresh/PodcastSearchOperation.swift
```diff
@@ -57,7 +57,7 @@ class PodcastSearchOperation: Operation {
         dispatchGroup.enter()
         URLSession.shared.dataTask(with: request) { data, _, error in
             guard let data = data, error == nil else {
-                shouldRetry = false
+                shouldRetry = true
```
leandroalonso, line 60: @SergioEstevao This does not fix the issue because in the line below we call the completion handler:



By the way, I just re-tested this again and it seems that the issue is with the `ServerPodcastManager.shared.addFromUuid` call:

Screenshot 2024-02-19 at 20 51 35

leandroalonso (changes requested):

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
SergioEstevao: @geekygecko confirmed that we always have a status value.

Modules/Server/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
SergioEstevao: I update the code to use nil instead of the failed response.

Modules/Server/Sources/PocketCastsServer/Public/Refresh/PodcastSearchOperation.swift
SergioEstevao: I removed the completion call, and I'm now looking at that call.

SergioEstevao: Implemented a retry logic for that call so the cache can have time to return the data.

Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
```diff
@@ -27,7 +27,32 @@ public class ServerPodcastManager: NSObject {
...
+    /// If The poll mechanism is to be used in cases when the podcast was just added to server and the first call might fail because the server didn't have time to update
+    /// The call will try a maximum of three time with each call having an exponetial backoff period of base 2 seconds for each try
+    /// - Parameters:
+    ///   - podcastUuid: the uuid of the podcast to caache
+    ///   - subscribe: if we should subscribe to the podcast after adding
+    ///   - tries: the number of tries already done
+    ///   - completion: the code to execute on completion
+    public func addFromUuidWithRetries(podcastUuid: String, subscribe: Bool, tries: UInt = 0, completion: ((Bool) -> Void)?) {
+        var pollbackCounter = tries
+        addFromUuid(podcastUuid: podcastUuid, subscribe: subscribe) { [weak self]success in
```
leandroalonso:
```suggestion
        addFromUuid(podcastUuid: podcastUuid, subscribe: subscribe) { [weak self] success in
```

Modules/Server/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
```diff
@@ -27,7 +27,32 @@ public class ServerPodcastManager: NSObject {
...
+                completion?(success)
+                return
+            }
+            pollbackCounter += 1
+            if pollbackCounter < 3 {
+                Thread.sleep(forTimeInterval: pow(2, Double(pollbackCounter)))
+                self?.addFromUuidWithRetries(podcastUuid: podcastUuid, subscribe: subscribe, tries: pollbackCounter, completion: completion)
+                return
+            }
+            completion?(success)
```
leandroalonso: I wonder if we can just do:

```suggestion
            completion?(false)
```

leandroalonso (approved):
@SergioEstevao tested multiple times and I can confirm the polling works as expected. :)

Added just a few nitpicky comments, up to you to change or not!

PR #1464, opened 2024-02-22
Settings Sync: Appearance

bjtitus:
| 📘 Part of: #1400 |
|:---:|

Adds synced setting for Appearance properties `useSystemTheme`, `theme`, `lightThemePreference`, `darkThemePreference`, `useEmbeddedArtwork`, `useDarkUpNextTheme`.

## To test

> [!IMPORTANT]
> Staging build configuration must be used. These settings are not all available in production.

### Preferred Themes

1. D1: Navigate to the Settings > Appearance screen
2. D1: Set the preferred Light Theme to a non-default, like rosé
3. D1: Pull to refresh or hit "Refresh Now" on Profile tab
4. D2: Pull to refresh or hit "Refresh Now" on Profile tab
5. D2: Notice that Theme changes
6. D2: Navigate to Settings > Appearance and make sure "Light Theme" is set.

### Sync System Appearance

1. D1: Make sure "Use iOS Light/Dark Mode" is enabled
2. D1: Change the preferred Dark Theme to a non-default using something like "Dark Contrast"
3. D1: Set device to Dark Mode
4. D1: Ensure theme changes
5. D2: Navigate to Settings > Appearance
6. D2: Disable "Use iOS Light/Dark Mode"
7. D2: Pull to refresh or hit "Refresh Now" on Profile tab
8. D1: Pull to refresh or hit "Refresh Now" on Profile tab
9. D1: Notice that theme changes to preferred light theme
10. D1: Navigate to Settings > Appearance
11. D1: Check that "Use iOS Light/Dark Mode" is disabled

## Checklist


podcasts/ServerSyncManager.swift
```diff
@@ -73,6 +73,7 @@ class ServerSyncManager: ServerSyncDelegate {
         Analytics.shared.refreshRegistered()
         DispatchQueue.main.async {
             PlaybackManager.shared.effectsChangedExternally()
+            Theme.sharedTheme.toggleTheme()
```
bjtitus, line 76: Theme is refreshed here. In testing, it didn't seem to matter if this was called each refresh vs. only when the theme changes.

Modules/Server/Sources/PocketCastsServer/Public/ServerEnums.swift
```diff
@@ -74,3 +74,70 @@ public enum HeadphoneControl: Int32, Codable {
...
+    case extraDark = 2
+    case electric = 7
+    case classic = 8
+    case indigo = 4
+    case radioactive = 9
+    case rosé = 3
+    case contrastLight = 6
+    case contrastDark = 5
+
+    public init(old: Old) {
```
leandroalonso, line 91: Can you add a comment about this conversion?

I think now it makes sense but I'm pretty sure someone will look at that in the future and wonder what's going on because everything is the same. 😅

podcasts/ServerSyncManager.swift
leandroalonso: Even though there's really no visual difference I believe we should check if the theme really changed. This call fires a notification that is listened in a bunch of different places that will trigger layout stuff that shouldn't be necessary to do...

podcasts/Settings.swift
```diff
@@ -624,11 +624,18 @@ class Settings: NSObject {
...
         UserDefaults.standard.set(value, forKey: Constants.UserDefaults.shouldFollowSystemThemeKey)
     }
 
     class func shouldFollowSystemTheme() -> Bool {
-        UserDefaults.standard.bool(forKey: Constants.UserDefaults.shouldFollowSystemThemeKey)
+        if FeatureFlag.settingsSync.enabled {
+            return SettingsStore.appSettings.useSystemTheme
+        } else {
+            return UserDefaults.standard.bool(forKey: Constants.UserDefaults.shouldFollowSystemThemeKey)
+        }
```
leandroalonso:
```suggestion
        if FeatureFlag.settingsSync.enabled {
           SettingsStore.appSettings.useSystemTheme
        } else {
           UserDefaults.standard.bool(forKey: Constants.UserDefaults.shouldFollowSystemThemeKey)
        }
```

podcasts/Settings.swift
```diff
@@ -885,7 +892,18 @@ class Settings: NSObject {
 
     static var loadEmbeddedImages: Bool {
         get {
-            UserDefaults.standard.bool(forKey: Constants.UserDefaults.loadEmbeddedImages)
+            if FeatureFlag.settingsSync.enabled {
+                return SettingsStore.appSettings.useEmbeddedArtwork
```
leandroalonso: We can remove the `return` here too :)

podcasts/Settings.swift
```diff
@@ -967,11 +985,19 @@ class Settings: NSObject {
 
     static var darkUpNextTheme: Bool {
         get {
-            Constants.UserDefaults.appearance.darkUpNextTheme.value
+            if FeatureFlag.settingsSync.enabled {
+                return SettingsStore.appSettings.useDarkUpNextTheme
```
leandroalonso: Same as the other comments about `return`. Let's kill it! 🔪

podcasts/Settings.swift
```diff
@@ -967,11 +985,19 @@ class Settings: NSObject {
...
+                return Constants.UserDefaults.appearance.darkUpNextTheme.value
+            }
         }
 
         set {
-            Constants.UserDefaults.appearance.darkUpNextTheme.save(newValue)
+            if FeatureFlag.settingsSync.enabled {
+                SettingsStore.appSettings.useDarkUpNextTheme = newValue
+            } else {
+                Constants.UserDefaults.appearance.darkUpNextTheme.save(newValue)
```
leandroalonso: Should this really be inside the `else`? If we disable `settingsSync` we might have different values, no?

podcasts/Theme.swift
```diff
@@ -161,15 +169,27 @@ class Theme: ObservableObject {
...
 
     class func setPreferredDarkTheme(_ preferredType: ThemeType, systemIsDark: Bool, userInitiated: Bool = false) {
-        UserDefaults.standard.setValue(preferredType.rawValue, forKey: preferredDarkThemeKey)
+
+        if FeatureFlag.settingsSync.enabled {
+            SettingsStore.appSettings.darkThemePreference = preferredType
+        } else {
+            UserDefaults.standard.setValue(preferredType.rawValue, forKey: preferredDarkThemeKey)
+        }
+        UserDefaults.standard.setValue(preferredType.old.rawValue, forKey: preferredDarkThemeKey)
```
leandroalonso: If `settingsSync` is disabled, the second `else` will always be overridden by the last `setValue` call.

podcasts/Theme.swift
```diff
@@ -181,15 +201,28 @@ class Theme: ObservableObject {
...
 
     class func setPreferredLightTheme(_ preferredType: ThemeType, systemIsDark: Bool) {
-        UserDefaults.standard.setValue(preferredType.rawValue, forKey: preferredLightThemeKey)
+
+        if FeatureFlag.settingsSync.enabled {
+            SettingsStore.appSettings.lightThemePreference = preferredType
+        } else {
+            UserDefaults.standard.setValue(preferredType.rawValue, forKey: preferredLightThemeKey)
+        }
+        UserDefaults.standard.setValue(preferredType.old.rawValue, forKey: preferredLightThemeKey)
```
leandroalonso: Same as the other comment. The `else` will always be overriden by the last `setValue`.

leandroalonso (changes requested):

podcasts/ServerSyncManager.swift
bjtitus: @leandroalonso Would it be alright to change `toggleTheme` to check this instead?

```swift
func toggleTheme() {
    let newTheme = toggledThemed()
    if activeTheme != newTheme {
        activeTheme = toggledThemed()
    }
}
```

leandroalonso: I guess so!

bjtitus:
@leandroalonso Ready for review!

leandroalonso (changes requested):
@bjtitus I'm having issues when changing the theme. Parts of the screen remains on the previous one (flag toggled off) and the remaining on the new one, you can see that on Profile (Rose on the avatar + Indigo on the rest):


https://github.com/Automattic/pocket-casts-ios/assets/7040243/82392c48-27c1-43fe-a225-703ece3c639b

Shouldn't the value be the same when I toggle between the flag on/off?

bjtitus:
> I'm having issues when changing the theme. Parts of the screen remains on the previous one

I'm not sure what may be causing the differing themes. An install before https://github.com/Automattic/pocket-casts-ios/pull/1464/commits/358edc704e1755fdeb99adb91386ca1e6a70f377 would have different themes or the import bug fixed in https://github.com/Automattic/pocket-casts-ios/pull/1464/commits/b8530546fdc5804f628087d7a278efcc003b733b might have failed to import the theme, although I'd expect that to be one of the defaults.

The profile avatar bug seems to be unrelated since I can reproduce this on `trunk` and it happens when toggling system appearance:

https://github.com/Automattic/pocket-casts-ios/assets/3250/3d69a226-02b6-41a8-9898-c909afa653fe

leandroalonso (approved):
@bjtitus re-tested with a clean install and all good!

PR #1468, opened 2024-02-22
Fix/playback skipping

SergioEstevao:
Fixes #

This PR makes all episodes that are currently played, downloaded to the local filesystem and cached.

The rational behind this is that skipping was happening because the episode streaming could change from one moment to the other because of changing ads on the stream.

With this approach the episode is downloaded locally and the seek position should be valid all the the time.

There could always be issues if user change listening devices a lot, because the cache will be different between device and the seek position different, but we hope this will minimize the issue for the majority of our users.

## To test

- Start the app
- Open a new podcast that you don't have on auto-download
- Select any episode
- Ensure that you don't have any effect active (Speed change, Trim silence, Volume Boost)
- Tap on Play
- Check the following:
  - if the audio plays correctly to the end
  - you see in the console: `Using EffectsPlayer`
  - Open the app folder in the simulator and check the following folder: `Documents/podcasts_buffered/`
  - It should have at least one file
  - When the episode finishes the file should be deleted 

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bjtitus (approved):
✅ Verified playback
✅ Verified removal of buffered episode upon completion

:shipit:

PR #1472, opened 2024-02-23
Feature/issue 12 replace material bottom sheet take over

### LINKED ISSUE #12: Replace MaterialBottomSheet with our own implementation

SergioEstevao:
This PR is a takeover of this issue #1299 the text bellow is copied from there.

I've updated the branch to the latest version of trunk and fixed the missing issues [I refer on the original PR](https://github.com/Automattic/pocket-casts-ios/pull/1299#pullrequestreview-1822028500).

Fixes #12 [Replace MaterialBottomSheet with our own implementation](https://github.com/Automattic/pocket-casts-ios/issues/12)

This PR replaces usages of `MaterialComponents/BottomSheet` with Apple's [UISheetPresentationController](https://developer.apple.com/documentation/uikit/uisheetpresentationcontroller) and removes the 3rd-party dependency.

## To test

The steps for testing will be different for each sheet in the app. These are the steps to test the Sleep Timer for anyone that would like to preview the new sheet.

1. Tap a Podcast to enter the list of episodes.
2. Tap an episode from the list to start playing one of the podcast episodes.
3. Tap the "now playing" footer at the bottom to go to the fullscreen detail for the podcast episode.
4. Tap the 💤 button in the bottom footer. The Sleep Timer bottom sheet will appear in the new bottom sheet.

## Checklist

- [x] I have considered adding unit tests for my changes. (UI changes, no unit tests)

## Before & After

Note that a video is included for the Overflow Actions behavior as, when the user taps the Edit button, that sheet enlarges and becomes undismissible via swipe actions.

| **App Scene** | **Before** | **After** |
|-----------|---------|---------|
| **Effects Sheet** | effects_sheet_before | effects_sheet_after |
| **Multiselect Actions** | multiselect_sheet_before | multiselect_sheet_after |
| **Multiselect Actions (Expanded)** | multiselect_sheet_expanded_before | multiselect_sheet_expanded_after |
| **Overflow Actions** | overflow_shelf_actions_before | overflow_shelf_actions_after |
| **Overflow Actions Edit Behavior** | <video src="https://github.com/Automattic/pocket-casts-ios/assets/4069415/c1e01ce9-a90e-4fa9-95ab-ea2bfbdd8012"/> | <video src="https://github.com/Automattic/pocket-casts-ios/assets/4069415/8a9d1f96-998e-40ff-a759-66903f5b4eb8"/> |
| **SwiftUI Wrapper Example (EoY)** | presentModally(year_end_example)_sheet_before | presentModally(year_end_example)_sheet_after |
| **Profile** | profile_sheet_before | profile_sheet_after |
| **Sleep Timer** | sleep_timer_before | sleep_timer_after |


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/NowPlayingPlayerItemViewController+Shelf.swift
```diff
@@ -224,10 +223,24 @@ extension NowPlayingPlayerItemViewController: NowPlayingActionsDelegate {
...
-        roundCorners(bottomSheet: bottomSheet)
 
-        present(bottomSheet, animated: true, completion: nil)
+        if let sheetController = shelfController.sheetPresentationController {
+            if #available(iOS 16.0, *) {
+                // We create a custom detent height of a bit more than half the hosting VC to try and
+                // ensure that all of the shelf content is visible in the sheet. We offer a large detent
+                // as a fallback so that the user can pull the sheet up if any content is ever cut off.
+                let hostingControllerHeight = view.bounds.height
+                let sheetDetentHeight = hostingControllerHeight * 0.675
```
bjtitus: I noticed there's a bunch of extra space for several sheets on larger devices. Could this be calculated using `sizeThatFits` to reduce the height based on the contents of the table view?

I also wonder if we should remove the `large` detent here. Previously, we didn't allow the sheet to expand and I think that makes since in most of these cases.

Here's an example of the Playback Effects on the 15 Pro Max:

| Before | After |
| ------ | ----- |
| CleanShot 2024-02-26 at 11 08 24@2x | CleanShot 2024-02-26 at 11 08 40@2x |

| Before | After |
| ------ | ------ |
| <video src="https://github.com/Automattic/pocket-casts-ios/assets/3250/888e5bc1-8993-40a0-9216-f0a5ef1f1236" loop autoplay> | <video src="https://github.com/Automattic/pocket-casts-ios/assets/3250/41dec1dd-dffe-45ac-90df-3672feb66634" loop autoplay> |

```suggestion
                let maxWidth = sheetController.containerView?.bounds.width ?? .greatestFiniteMagnitude
                let sheetDetentHeight = sheetController.presentedViewController.view.sizeThatFits(CGSizeMake(maxWidth, .greatestFiniteMagnitude)).height
```

This would apply to all of the custom detent providers.

bjtitus (changes requested):
Thanks for making the API update! I had one comment about detent sizing but everything seems to be working well!

podcasts/NowPlayingPlayerItemViewController+Shelf.swift
SergioEstevao: I had look, and in general this approach works but on smaller device ( iPhone SE) we can get content cropped if for example it has scrollable content.

Simulator Screenshot - iPhone SE (3rd generation) - 2024-02-26 at 19 58 24


The overflow content had scroll disabled but if I enabled that users will be able to go to other options, proabably it will be better to have it scrollable to future proof new additions to this screen.

SergioEstevao:
@bjtitus ready for another look.

podcasts/NowPlayingPlayerItemViewController+Shelf.swift
bjtitus: > it will be better to have it scrollable to future proof new additions to this screen

Yeah, we should definitely have this so we don't have to worry about the count changing.

Would it make sense to use a larger size for smaller devices so it's less likely to be cut off? We could always check the trait collection in the custom detent provider.

bjtitus (approved):
:shipit: Looked good in testing!

rviljoen:
@SergioEstevao I've noticed the following warning pop up in the debugger when opening the Sleep Timer sheet:

```
Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x28301ab20 UILayoutGuide:0x282aa3640'UIViewSafeAreaLayoutGuide'.bottom == UIView:0x104c4e730.bottom   (active)>",
    "<NSLayoutConstraint:0x2830182d0 V:[UIView:0x104c4e730]-(0)-|   (active, names: '|':UIView:0x104c543b0 )>",
    "<NSLayoutConstraint:0x283018e10 'UIViewSafeAreaLayoutGuide-bottom' V:[UILayoutGuide:0x282aa3640'UIViewSafeAreaLayoutGuide']-(34)-|   (active, names: '|':UIView:0x104c543b0 )>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x28301ab20 UILayoutGuide:0x282aa3640'UIViewSafeAreaLayoutGuide'.bottom == UIView:0x104c4e730.bottom   (active)>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKitCore/UIView.h> may also be helpful.
```

I'm not 100% sure, but I don't think I saw these with the previous implementation.

PR #1474, opened 2024-02-23
Fix setting the new upgrade account flow feature flag

SergioEstevao:
Fixes #

Fix setting the new upgrade account flow feature flag.
Value was not been read from the RemoteConfig..

## To test

 - Comment the guard line on `updateRemoteFeatureFlags` code on line 304 of `AppDelegate.swift`
 -  Put a break point on line 324 of the same method
 - Start the app  and check that the `newAccountUpgradePromptFlow` feature flag is overridden with the value on the Remote server. ( it should be false)

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -80,6 +80,8 @@ public enum FeatureFlag: String, CaseIterable {
         switch self {
         case .deselectChapters:
             "deselect_chapters"
+        case .newAccountUpgradePromptFlow:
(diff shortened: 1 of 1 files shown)


leandroalonso (approved):
@SergioEstevao in order to test I also had to uncomment this line:

```swift
guard BuildEnvironment.current != .debug else { return }
```

I think it worth adding this to the PR description because otherwise the breakpoint is never hit.

SergioEstevao:
> @SergioEstevao in order to test I also had to uncomment this line:

@leandroalonso isn't that line 304 of the AppDelegate? The one that I refer on the testing steps?

leandroalonso:
@SergioEstevao I'm stupid! lol

Sorry, I missed that.

PR #1483, opened 2024-02-27
[Slumber] Make it available only for lifetime and yearly subscriptions

leandroalonso:
The Slumber benefit is available only for:

1. Lifetime subscriptions
2. Yearly subscriptions

This PR changes this behavior in:

1. The Upgrade modal
2. The What's new

## To test

Change `WhatsNew.swift` `init` to:

```swift
    init(announcements: [Announcement] = Announcements().announcements, previousOpenedVersion: String? = UserDefaults.standard.string(forKey: Constants.UserDefaults.lastRunVersion), currentVersion: String = Settings.appVersion(), lastWhatsNewShown: String? = Settings.lastWhatsNewShown) {
        self.announcements = announcements
        self.previousOpenedVersion = "7.57"
        self.currentVersion = "7.57"
        self.lastWhatsNewShown = "7.56"
    }
```
 
1. Go to Profile > Settings > Beta Features > enable `slumber`
1. Hit <kbd>Command</kbd>+<kbd>Shift</kbd>+<kbd>,</kbd> to open the Scheme editor
2. Click the options tab
3. Select the `Pocket Casts Configuration.storekit` file in the StoreKit configuration dropdown
4. Run the app and login to an account
5. You should see the What's New
6. Tap the upgrade button
7. Toggle between Yearly/Monthly plus
8. ✅ Slumber is only show for Yearly
9. Subscribe to monthly
10. ✅ You should not see the promocode
11. Subscribe to Yearly
12. ✅ You should now see the promocode
13. Re-run the app
14. ✅ You should see the promocode
15. Delete and reinstall the app
16. Login to an account with lifetime (it can be your A8C one)
17. ✅ You should see the promocode

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1487, opened 2024-02-28
Merge #1483 into 7.58

leandroalonso:
Same as #1483


leandroalonso:
Too many conflicts, applying changes one by one to avoid issues

PR #1506, opened 2024-03-13
Settings Sync: Remove unused shouldUseNewSettingsSync param

bjtitus:
Now that FeatureFlags are available in the Server component, we no longer need the indirection of `shouldUseNewSettingsSync`. This cleans up an unused `shouldUseNewSettingsSync` on `SyncManager` and `SyncSettingsTask`. It also immediately applies the Feature Flag change instead of requiring a restart.

## To test

* Make sure CI compiles 🟢 
* Test that settings syncing works:
    * Install on two devices/simulators
    * Enable the `settingsSync` feature flag
    * Make sure that changing settings on one and refreshing shows the same setting on the other
    * You can verify that the `Syncing new settings: ` message is logged to the console as an easy method to check that syncing is happening.

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1507, opened 2024-03-13
Settings Sync: `new_settings_storage` Feature Flag

bjtitus:
Adds a `new_settings_storage` feature flag for migrating settings to new JSON based format.

The original `settings_sync` flag was trying to accomplish two things:
1. Ensure we can fall back to the same old settings logic in case something breaks
2. Ensure we can enable/disable syncing between devices.

This separates the settings storage changes from the syncing changes. This way, we can disable syncing without disabling storage changes, retaining any modified dates as long as the `new_settings_storage` flag is enabled.

This fixes issues with settings being dropped by simply disabling the syncing logic instead of reverting to the old user defaults. The behavior now matches Android.

I feel that the new flag is still useful to have in case something gets messed up with settings conversion, we have a way to back out ALL of the storage related changes _just in case_.

## To test

* Install this branch on two devices/simulators
* Enable the `new_settings_storage` AND `settings_sync` feature flags & restart the app (this restart won't be needed after https://github.com/Automattic/pocket-casts-ios/pull/1506)
* D1: Change a setting and refresh
* D2: Refresh and ensure setting is synced
* D1: Disable `settings_sync` flag + restart
* D1: Change a setting and refresh
* D2: Refresh and note that the setting remains unchanged
* D2: Change a setting and refresh
* D1: Enable `settings_sync` flag + restart
* D1: Refresh to be sure settings are synced
* D2: Refresh and notice that changes made while `settings_sync` was disabled are reflected now 

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -28,6 +28,9 @@ public enum FeatureFlag: String, CaseIterable {
     /// Enable selecting/deselecting episode chapters
     case deselectChapters
 
+    /// Store settings as JSON in User Defaults (global) or SQLite (podcast)
+    case newSettingsStorage
+
     /// Syncing all app and podcast settings
     case settingsSync
 
@@ -75,6 +78,8 @@ public enum FeatureFlag: String, CaseIterable {
             false
         case .cachePlayingEpisode:
             true
+        case .newSettingsStorage:
+            false
         }
     }
 
@@ -88,6 +93,8 @@ public enum FeatureFlag: String, CaseIterable {
             "new_account_upgrade_prompt_flow"
         case .cachePlayingEpisode:
             "cache_playing_episode"
+        case .newSettingsStorage:
+            "new_settings_storage"
         default:
             nil
         }
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -71,7 +71,7 @@ public enum FeatureFlag: String, CaseIterable {
         case .deselectChapters:
             false
         case .settingsSync:
-            false
+            false // `newSettingsStorage` also needs to be `true` for syncing to function
         case .slumber:
             false
         case .newAccountUpgradePromptFlow:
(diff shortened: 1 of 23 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bjtitus:
cc @leandroalonso This should fix the problems with settings not being retained while the Feature Flag is off. Instead of re-running the import, I separated the flag for syncing logic and storage logic. This allows each to be turned off independently (we would probably leave `new_settings_storage` enabled unless there is some issue identified with the new storage) and better matches what Android does.

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -75,6 +78,8 @@ public enum FeatureFlag: String, CaseIterable {
             false
         case .cachePlayingEpisode:
             true
+        case .newSettingsStorage:
+            false
```
leandroalonso, line 82: Wdyt about adding a comment on `settingsSync` that it also needs this flag to be `true`? Similar to this:

```swift
case .episodeFeedArtwork:
            false // To be enabled, newShowNotesEndpoint needs to be too
```

leandroalonso (approved):
Nice work @bjtitus!

✅ Tested setting syncing between devices
✅ Tested setting not syncing if flag is disabled
✅ Tested that the latest change will be the one kept

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/43">7.60</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
bjtitus: Good idea! I wondered whether these two flags should be tied together but I think it's better for us to be able to independently control them. In theory, `settingsSync` should never be enabled without `newSettingsStorage`.

PR #1508, opened 2024-03-13
Settings Sync: Add remote feature flag for `settings_sync`

bjtitus:
| 📘 Part of: #1400 | Depends on #1545 |
|:---:|:---:|

* Adds a `settings_sync` remote feature flag key.
* Sets the default value of `new_settings_storage` to `true` **when not running from the App Store**
* Sets the default value of `settings_sync` to `true` **when not running from the App Store**
* Only syncs the Podcast settings when `settings_sync` is enabled

CleanShot 2024-03-19 at 15 35 35@2x

### Enabled
* Delete previous install and install a new build
* Ensure that `newSettingsStorage` and `settingsSync` are enabled under Beta Features
* Test syncing some settings between two devices with the flags enabled

### Disabled
* Set a breakpoint in AppDelegate.checkDefaults:129
* Install with build configuration "Release"
* Ensure the `FeatureFlag.newSettingsStorage.enabled` flagged off portion is skipped

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -70,16 +70,16 @@ public enum FeatureFlag: String, CaseIterable {
             false
         case .deselectChapters:
             false
+        case .newSettingsStorage:
+            true
         case .settingsSync:
-            false // `newSettingsStorage` also needs to be `true` for syncing to function
+            BuildEnvironment.current != .appStore
         case .slumber:
             false
         case .newAccountUpgradePromptFlow:
             false
         case .cachePlayingEpisode:
             true
-        case .newSettingsStorage:
-            false
         }
     }
 
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -95,6 +95,8 @@ public enum FeatureFlag: String, CaseIterable {
             "cache_playing_episode"
         case .newSettingsStorage:
             "new_settings_storage"
+        case .settingsSync:
+            "settings_sync"
         default:
             nil
         }
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -71,9 +71,9 @@ public enum FeatureFlag: String, CaseIterable {
         case .deselectChapters:
             false
         case .newSettingsStorage:
-            BuildEnvironment.current != .appStore
+            shouldEnableSyncedSettings
         case .settingsSync:
-            BuildEnvironment.current != .appStore
+            shouldEnableSyncedSettings
         case .slumber:
             false
         case .newAccountUpgradePromptFlow:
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/43">7.60 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -70,16 +70,16 @@ public enum FeatureFlag: String, CaseIterable {
             false
         case .deselectChapters:
             false
+        case .newSettingsStorage:
+            true
```
leandroalonso: Isn't this the part that can cause more issues?

From what we've been seeing recently related to issues it seems the conversion is what is causing some issues, not necessarily the sync.

I think it would be safer to keep this disabled but enable it through the remote feature flag? 

Also, maybe we should enable this only on TestFlight/debug?

leandroalonso (changes requested):
@bjtitus leave a comment about a question. There are no issues with the PR but I just want to make sure we're on the same page!

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
bjtitus: > I think it would be safer to keep this disabled but enable it through the remote feature flag?

Ideally, I agree that we should use the remote feature flag but I don't think there's a good way for us to target only TestFlight builds outside of a code change right now.

> maybe we should enable this only on TestFlight/debug

I think it would be reasonable to keep `newSettingsStorage` and `settingsSync` the same here and only enable both for TestFlight + debug. We can always change this prior to submitting an App Store release.

bjtitus: > I think it would be safer to keep this disabled but enable it through the remote feature flag?

Another thing to note: It may be safer to keep these enabled at launch whenever possible because some settings may affect episode listings, notifications, up next, etc.

So I think I'd prefer to only use the remote override as a last resort just in case of runtime issues.

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -95,6 +95,8 @@ public enum FeatureFlag: String, CaseIterable {
             "cache_playing_episode"
         case .newSettingsStorage:
             "new_settings_storage"
+        case .settingsSync:
+            "settings_sync"
```
leandroalonso: I think for these two we can do:

```suggestion
        case .newSettingsStorage:
           BuildEnvironment.current != .appStore ? "new_settings_storage" : nil
        case .settingsSync:
            BuildEnvironment.current != .appStore ? "settings_sync" : nil
```

This way any remote feature flag change will only affect the non-App Store builds. Wdyt?

I know this is becoming a bit complex but I think it will keep things safer without us having to change code after merging that.

bjtitus: Good point. I combined these checks into a single bool so they don't deviate for now. We can always override individually if needed.

https://github.com/Automattic/pocket-casts-ios/pull/1508/commits/bbcdd03297817267a3cd585e02637b17aeafd8eb#diff-496cbacd6eecba67165682b12d72e4938d110a5fba96e1873858c637fb37905dR86-R89

leandroalonso (approved):

PR #1554, opened 2024-03-21
Update project and CI for xcode15.3

SergioEstevao:
Fixes #

Updates the project in order to build on Xcode 15.3 and use the matching CI

## To test

- CI should be green
- Build it locally on your machine and run in a simulator
- Just smoke test the app to see if nothing is broken because of the update

## Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -6,7 +6,7 @@ common_params:
   # Common environment values to use with the `env` key.
   - &common_env
     # If you update the image to a newer Xcode version, don't forget to also update the badge in the README.md file accordingly for consistency
-    IMAGE_ID: xcode-15.2
+    IMAGE_ID: xcode-15.3
   # Common agents values to use with the `agents` key.
   - &common_agents
     queue: mac
--- a/.buildkite/release-builds.yml
+++ b/.buildkite/release-builds.yml
@@ -7,7 +7,7 @@ common_params:
     - automattic/a8c-ci-toolkit#2.15.1
   # Common environment values to use with the `env` key.
   - &common_env
-    IMAGE_ID: xcode-15.2
+    IMAGE_ID: xcode-15.3
 
 steps:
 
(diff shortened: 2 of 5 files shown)


danielebogo (approved):
I did some tests and the new Xcode builds and runs the app correctly.
I'm wondering if we can merge [this](https://github.com/Automattic/pocket-casts-ios/commit/27e9063e4f070204f68c3586dd80f433250ddca1) branch into this PR and change the `.xcode-version` to also gain the security check of the supported version.

@Automattic/pocket-casts-ios wdyt?

leandroalonso:
@danielebogo does this prevent a build in case someone is using, let's say, Xcode 15.2 while we moved to Xcode 15.3? What if someone is in a beta Xcode version, does it build?

If yes, I'm not entirely against it but I'd like to point out a few things:

1. Since I joined the project I remember seeing an issue related to Xcode only once, and it was between major versions when Apple introduced support for dynamic widgets;
2. However, the build only started to fail once we added code to support dynamic widgets;
3. The build itself failing was some sort of an indicator that you needed to update Xcode;

The only reason I'm a bit on the fence about this change is: what value does it add? Or what issue does it prevent?

A few caveats I think of:

1. Multiple branches might be on different Xcode versions. I always keep the latest Xcode versions on my machine but opening different stuff on different Xcode versions is something I'd like to avoid if possible 😬;
2. Sometimes I might be in a hurry to fix something or to submit a PR only to see that I can't build anymore because I'm not on the correct version. TBH I'd remove the build phase temporarily just so I don't disrupt my workflow waiting to download/unzip Xcode;

Just sharing my 2 cents. I'm pretty sure there might be a good reason to use that but at the moment I'm seeing more cons than pros. 😅

danielebogo:
Thanks @leandroalonso! Usi are right. The xcode version check prevents a contributor uses a different version than the one specified in the xcode-version. Beta included.

Usually, even if this check is not present, if you update your Xcode version, you should keep the old one for the whole new release cycle anyway, as the prod version is built on that. Same for the different branches. Each time you vet a new version, a rebase is needed. That's also needed to spread the pipeline changes.

That said it's not mandatory to have it. It's an additional check to make sure everybody uses the same version. And also that's why I'd love to have feedback from the whole team ☺️

SergioEstevao:
@danielebogo should I merge this PR and then we do a separate PR for the other change when we get a consensus about this?

danielebogo:
@SergioEstevao absolutely! I didn't intend to block this PR with that!

leandroalonso:
@danielebogo I still like the idea though. It would be possible to convert the build failure into a warning?

danielebogo:
@leandroalonso good question 🤔 I think it is. If we can get rid of all our warnings, it's also a good solution. @SergioEstevao maybe I'm wrong, but did you push a PR to get rid of our warnings?

PR #1558, opened 2024-03-22
Bookmark share

SergioEstevao:
Fixes #

Implements a sharing option for bookmarks, that creates a sharing link that can be send to other users.

https://github.com/Automattic/pocket-casts-ios/assets/651601/6db4697a-e1ba-4f62-89bf-b79dced48cfb

| Podcast Bookmarks | Profile Bookmarks | Player Bookmarks |
| - | - | - |
Simulator Screenshot - iPhone 15 Pro - 2024-03-22 at 10 01 46 | Simulator Screenshot - iPhone 15 Pro - 2024-03-22 at 09 59 16 |  Simulator Screenshot - iPhone 15 Pro - 2024-03-22 at 10 01 30 |

## To test

- Start the app using a user with a Plus or Patron Account on a device
- If you still don't have any Bookmarks please create some on the account on multiple podcasts
- Go to Profile
- Open Bookmarks
- Tap on the more button 
- Select one bookmark
- Tap on Share
- Copy the link
- Build and run the app on another device/simulator, switch to Safari on that device
- Open in Safari the link that you copied from the other device
- Tap on Open In Pocket Casts
- Check that the link is opened on the correct Podcast -> Episode -> Time location
- Repeat the test above on the other locations where Bookmarks list are show
 - Player full screen, tab bookmarks
 - Podcast detail, tab bookmarks  

## Checklist


danielebogo (approved):
LGTM! I tested the feature and the bookmark worked as described.
Code looks good. Not a blocker but I was wondering if we can improve the protocol implementation a bit to avoid some repetitions. It seems that the `func bookmarkShare(_ bookmark: Bookmark)` implementation is the same in all vc. Perhaps some default implementation with a protocol constraint to VC only? Just something to keep in mind as future improvement.

CookieyedCodes:
The order dosen't really make sense to me, personally I would do edit, delete, share 😅

PR #1563, opened 2024-03-25
Standard Tags: save episode metadata JSON into the database

leandroalonso:
This does the same as #1557 but with a few changes discussed in more detail below.

## Technical discussion

In the original PR once the app loads a show info it:

1.  Parses the whole JSON, in which some cases might contain [up to 2000 episodes](https://cache.pocketcasts.com/mobile/show_notes/full/4eb5b260-c933-0134-10da-25324e2a541d)
2. Maps the return to each `Episode` in the database
3. Change show notes/image for each model
4. Saves it back to the database

To add context, in the past, we had issues manipulating some big JSONs. While 2000 episodes might be ok for some devices, for others might be the difference between using too much RAM (and crashing) and everything going well.

Here I changed the approach a bit. Once the app loads a show info it:

1. Uses `JSONSerialization` which is faster than Decodable;
2. Extract each episode JSON `String`
3. Save the JSON `String` into the database

We only parse the JSON unless is needed — and when we do that we parse only the requested episode JSON, which is way lighter than the whole response.

In terms of code, this simplifies a lot of our lives: for upcoming tags such as transcripts and funding, all we'd need to change is the `Episode.Metadata` struct.

Another benefit of this solution is that episode metadata and episodes are decoupled, so this avoids the issue of not having the actual episodes to save the metadata and not having it later.

Of course, this is not only rainbow and butterflies. We're saving the episode JSON metadata into the database, so if we need to query this data it can be twice as slow as opposed to having each metadata in its column.

Anyway, I think that in the meantime this offers the best benefit in terms of code maintainability but happy to be convinced otherwise. :)

### With the flag disabled

1. Change `newShowNotesEndpoint` to `false`
2. Re-run the app
3. Go to Discover > tap on any podcast > tap on an episode
4. ✅ You should see the show notes appear
5. Play the episode
6. Open the full player (by tapping the miniplayer) > tap "Details"
7. You should see the show notes

### Notes

Go to Profile > Settings > Beta Features > enable `newShowNotesEndpoint` and `episodeFeedArtwork`

2. Go to Discover > tap on any podcast > tap on an episode
9. On the app, dismiss the episode screen
14. Play the episode
15. Open the player and tap "Details"
16. ✅ You should see the notes
19. Stop the app, turn off your Wi-fi (or go airplane mode)
20. Open the app and then tap on the miniplayer > go to "Details"
21. ✅ The notes should shown

### Embedded artwork images

1. Go to Profile > Settings (cog icon) > Appearance > enable "Use Embedded Artwork"
2. Search for "The Bulwark Podcast" or "Clublife" (both have embedded XML artwork)
3. Play any episode
4. ✅ After a while the image should change to the embedded episode artwork
5. Download any episode from "The Bulwark Podcast" or "Clublife"
6. Close the app
7. Go offline
8. Reopen the app and play any of those episodes
9. ✅ You should see the episode artwork
10. ✅ Also check that you see the episode artwork on CarPlay and Now Playing screen

### Downloading

1. Open a podcast with many episodes (such as "The Bulwark Podcast")
2. Tap "..." > "Download All"
3. ✅ 100 episodes should be added to the queue
4. Wait for a few of them to download
5. Go offline
6. Play any episode
7. ✅ The episode artwork should show

## Checklist


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
```diff
@@ -1014,3 +1014,84 @@ extension EpisodeDataManager {
...
+                        Task {
+                            let metadata = await self.getShowInfo(for: metadataData)
+                            continuation.resume(returning: metadata)
+                        }
+                    } else {
+                        continuation.resume(returning: nil)
+                    }
+                } catch {
+                    FileLog.shared.addMessage("EpisodeDataManager.loadMultiple Episode error: \(error)")
+                    continuation.resume(returning: nil)
```
SergioEstevao: Is hiding the error and returning nil the best approach here? Shouldn't we make the method throw and decide on the calling site? Or is this an approach that we are following on all of this database methods?

podcasts/EpisodeArtwork.swift
```diff
@@ -45,12 +46,17 @@ class EpisodeArtwork {
     }
 
     private func loadEpisodeArtworkFromUrl(podcastUuid: String, episodeUuid: String) {
-        CacheServerHandler.shared.loadEpisodeArtworkUrl(podcastUuid: podcastUuid, episodeUuid: episodeUuid) { [weak self] imageUrl in
+        if !FeatureFlag.newShowNotesEndpoint.enabled && !FeatureFlag.episodeFeedArtwork.enabled {
```
SergioEstevao: Maybe a guard statement here?

SergioEstevao (commented):
Tested the scenarios and it look is working well but I have the following questions:

- Using the Clublife podcast I could see different artwork for each episode but only on the player and mini-player. Not on the episode details screen
- For the `The Bullwark Podcast` while the show notes is working correctly I don't see any specific artwork for each episode, it's always the same image.

If the above is expected all is good!

podcasts/EpisodeArtwork.swift
leandroalonso: Good one! Done in 610917957a91cbe7f56886d3602c00787c50783e

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
leandroalonso: Good question. This is the first time we're using an `async` approach here.

I changed the code to throw exceptions in ecee87fca0aa4c6c4f11dee3d0e2d1f16c46e5ec and 7efe1ceb82e67f1328e4df6c727684d54231ec67

leandroalonso:
@SergioEstevao 

> Not on the episode details screen

This is coming in a subsequent PR. :)

> For the The Bullwark Podcast while the show notes is working correctly I don't see any specific artwork for each episode, it's always the same image.

Oh, it seems they changed that. The latest episodes don't have custom artwork, but some old ones have it (January and before).

SergioEstevao (approved):
:shipit:

PR #1576, opened 2024-03-28
Fallback to default player when length is not available

SergioEstevao:
Refs #900

This PR detects goes around one of the issue that can affect the EffectPlayer: the failure to read the frame count on file, that resulted in a exception being throw and the file being skipped. 
With this change if the Effects Player fails to load it will default to the DefaultPlayer and try to play it.

There are still files that even using the default player, they will fail with some foundation error and are unable to play if the file is downloaded locally.

From my tests I saw this error: `AVErrorFourCharCode='optm'` and from my investigation on the documentation it looks related [to this error](https://developer.apple.com/documentation/audiotoolbox/kaudiofilestreamerror_notoptimized).

## To test

- Start the app
- Open the following episode: [https://pca.st/22iph0yt](https://pca.st/22iph0yt)
- Tap play
- Check that it plays correctly to the end

Steps to test your PR.

## Checklist

### DIFF
--- a/podcasts/EffectsPlayer.swift
+++ b/podcasts/EffectsPlayer.swift
@@ -109,11 +109,15 @@ class EffectsPlayer: PlaybackProtocol, Hashable {
                 if strongSelf.cachedFrameCount == 0 {
                     // we haven't cached a frame count for this episode, do that now
                     strongSelf.cachedFrameCount = strongSelf.audioFile!.length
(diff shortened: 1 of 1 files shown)


bjtitus (approved):
👍 Playback worked in testing, although there was a pretty significant delay before playback started. I think it's from downloading the larger 3 hour file?

PR #1577, opened 2024-03-28
Use contentType to set extension for downloaded files

### LINKED ISSUE #900: An episode can't be played when downloaded, but will play when streamed.

SergioEstevao:
Fixes #900

Some podcasts use file URLs with an extension that does not match with the contentType of the files. This changes saves the content file provided by the header response from servers and use that to use as an extension

## To test

- Search for the following podcast: `ondacer` on the discovery section
- Subscribe to the podcast
- Open the podcast and download one of the episodes
- Tap play
- The episode should play correctly
- Smoke test other podcasts, to see if this change doesn't brake anything on that front

## Checklist


leandroalonso (approved):

leandroalonso:
@SergioEstevao could you please add this to the `CHANGELOG.md`? We missed that

PR #1582, opened 2024-03-29
Standard Tags: fixes issues with episode artwork

### LINKED ISSUE #1579: Episode Artwork: wrong artwork displayed on Up Next
When enabling Episode Artwork and opening the Up Next for the first time, the wrong episode artworks were loaded for some episodes:

I believe the episode artwork loading code is being triggered even if the cell is out of screen, which should not happen.

leandroalonso:
Fixes #1579

This fixes a few things for episode artwork:

* Fix an issue where the Up Next was "blinking" with a different image when showing episode artwork;
* Correctly set up cache so the original image is cached and we don't need to request it more than once;
* Ensure only the correct image is loaded

## To test

It's better to run this on a real device and also ensure on Profile > Settings > Appearance > enable "Use Episode Artwork"

1. Ensure your Up Next queue has many items with episode artwork;
2. Open it
3. ✅ Images should load correctly, you should not see any "blink"
4. Go to a filter with many podcasts
5. Scroll down as fast as you can, then go up
6. ✅ Ensure each episode has loaded its correct image

## Checklist

### DIFF
--- a/podcasts/PodcastImageView.swift
+++ b/podcasts/PodcastImageView.swift
@@ -6,6 +6,8 @@ class PodcastImageView: UIView {
     private var shadowView: UIView?
     var imageView: UIImageView?
 
+    var currentEpisode: Episode? = nil
+
     override init(frame: CGRect) {
         super.init(frame: frame)
 
@@ -28,8 +30,21 @@ class PodcastImageView: UIView {
     func setEpisode(_ episode: Episode, size: PodcastThumbnailSize) {
         guard let imageView = imageView else { return }
 
+        currentEpisode = episode
+
         Task {
             if FeatureFlag.episodeFeedArtwork.enabled, Settings.loadEmbeddedImages, let episodeArtworkUrl = try? await ShowInfoCoordinator.shared.loadEpisodeArtworkUrl(podcastUuid: episode.parentIdentifier(), episodeUuid: episode.uuid) {
+
+                ImageManager.sharedManager.setPlaceholder(imageView: imageView, size: size)
+                imageView.kf.cancelDownloadTask()
+
+                // The app might run into the case where the episode changed but there's still
+                // a pending task to display the image of another episode
+                // This check prevent this from happening.
+                guard episode.uuid == currentEpisode?.uuid else {
+                    return
+                }
+
                 ImageManager.sharedManager.loadImage(url: episodeArtworkUrl, imageView: imageView, size: size, showPlaceHolder: true)
                 adjustForSize(size)
             } else {
--- a/podcasts/PodcastImageView.swift
+++ b/podcasts/PodcastImageView.swift
@@ -40,6 +40,7 @@ class PodcastImageView: UIView {
 
                 // The app might run into the case where the episode changed but there's still
                 // a pending task to display the image of another episode
+                // This can happen when dequeing a cell, for example.
                 // This check prevent this from happening.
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/44">7.61</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/PodcastImageView.swift
```diff
@@ -28,8 +30,21 @@ class PodcastImageView: UIView {
...
+
         Task {
             if FeatureFlag.episodeFeedArtwork.enabled, Settings.loadEmbeddedImages, let episodeArtworkUrl = try? await ShowInfoCoordinator.shared.loadEpisodeArtworkUrl(podcastUuid: episode.parentIdentifier(), episodeUuid: episode.uuid) {
+
+                ImageManager.sharedManager.setPlaceholder(imageView: imageView, size: size)
+                imageView.kf.cancelDownloadTask()
+
+                // The app might run into the case where the episode changed but there's still
+                // a pending task to display the image of another episode
+                // This check prevent this from happening.
```
leandroalonso: I wonder if this comment is conveying the message in a good way, feedback is appreciated.

Basically, this prevents something like:

1. We start to load episode artwork for episode A
2. The user scroll downs and the cell get changed to episode B
3. We already have episode B metadata and image cached
4. Episode B image is shown
5. Episode artwork for episode A finishes loading, so the executing proceeds to load its image
6. 💥 Episode artwork for episode A is shown while what's being shown is episode B

bjtitus: I would add something about cell dequeuing here. I think this is a common enough occurrence that just mentioning that makes it clear what issue you're trying to prevent.

bjtitus (approved):
Overall this looks good and loading all works.

When I clear the image caches and scroll quickly down, I see a ten or so second delay while loading catches up. I also get a low memory warning about a few hundred megabyte memory usage spike.

https://github.com/Automattic/pocket-casts-ios/assets/3250/88dd0547-a082-4467-a303-098b2a4a5c66

I see in Instruments that a bunch of the Swift Tasks are queueing up rapidly and seem to far exceed the number of current cells. I'm guessing this is because the database tasks are queueing up and all of the tasks are effectively blocked and remained un-cancelled?

CleanShot 2024-04-01 at 13 12 54@2x

I tried playing around with checking the task cancellation but didn't get anywhere, I believe because the `dbQueue.inDatabase` call has no real way of cancelling. I'm not sure if there's anything we can do for this if we can't cancel the db query from inside of the task but just thought I'd bring it up.

leandroalonso:
> When I clear the image caches and scroll quickly down, I see a ten or so second delay while loading catches up. I also get a low memory warning about a few hundred megabyte memory usage spike.

Thanks for pointing that out! I'll check that on another PR. I believe we can decouple the need to wait for storing in the database (thus fixing the behavior you pointed), but I'll need to look deeper into the memory issue.

PR #1602, opened 2024-04-05
Categories Redesign: Pills

bjtitus:
### Changes

Adds a scrolling set of pills/chips to the top of Discover to navigate categories.

<br/>

### Code Changes

* `CategoriesSelectorView` loads the categories from the Discover source and displays loading state
* `CategoriesPillsView` contains the pill layout and specific buttons
* `CategoryButtonStyle` encapsulates the style of these buttons with border, pressed and selected states
* `CategoriesSelectorViewController` is a `UIHostingController` which handles sizing, storing the observable for updating the Discover Item, and implementing `DiscoverSummaryProtocol`
* A new `DiscoverServerHandler.discoverCategories` async method is added to better interface with SwiftUI
* The `content_v2.json` file is used when the `categoriesRedesign` is enabled
* The new `pills` summaryStyle is used to differentiate this style from the old `category` style, which could still be returned if needed.

## To test

* Build using Staging configuration
* Enable the `categoriesRedesign` feature flag in Beta Features
* Navigate to Discover - If Discover doesn't reload for some reason, go to Developer menu and `Force Reload Discover`
* Verify that Category pills appear and they are not shown at the bottom of Discover
* Verify that tapping a pill changes state and hides other pills

### Known Issues

* Close button needs to be rounded and some other small styling tweaks will probably need to be made.

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
@@ -51,6 +51,20 @@ public class DiscoverServerHandler {
         }
     }
 
+    public func discoverCategories(source: String) async -> [DiscoverCategory] {
+        return await withCheckedContinuation { continuation in
+            DiscoverServerHandler.shared.discoverCategories(source: source, completion: { discoverCategories in
+                DispatchQueue.main.async {
+                    guard let discoverCategories = discoverCategories else {
+                        continuation.resume(returning: [])
+                        return
+                    }
+                    continuation.resume(returning: discoverCategories)
+                }
+            })
+        }
+    }
+
     public func discoverCategoryDetails(source: String, completion: @escaping (DiscoverCategoryDetails?) -> Void) {
         discoverRequest(path: source, type: DiscoverCategoryDetails.self) { categoryDetails, _ in
             completion(categoryDetails)
--- a/Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
@@ -1,5 +1,6 @@
 import Combine
 import Foundation
+import PocketCastsUtils
 
 public class DiscoverServerHandler {
     enum DiscoverServerError: Error {
@@ -28,7 +29,14 @@ public class DiscoverServerHandler {
     }
 
     public func discoverPage(completion: @escaping (DiscoverLayout?, Bool) -> Void) {
-        discoverRequest(path: ServerConstants.Urls.discover() + "ios/content.json", type: DiscoverLayout.self) { discoverItems, cachedResponse in
+        let contentPath: String
+        if FeatureFlag.categoriesRedesign.enabled {
+            contentPath = "ios/content_v2.json"
+        } else {
+            contentPath = "ios/content.json"
+        }
+
+        discoverRequest(path: ServerConstants.Urls.discover() + contentPath, type: DiscoverLayout.self) { discoverItems, cachedResponse in
             completion(discoverItems, cachedResponse)
         }
     }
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/DiscoverViewController.swift
```diff
@@ -190,7 +193,7 @@ class DiscoverViewController: PCViewController {
                 addSummaryController(CollectionSummaryViewController(), discoverItem: discoverItem)
             case ("network_list", _, _):
                 addSummaryController(NetworkSummaryViewController(), discoverItem: discoverItem)
-            case ("categories", _, _):
+            case ("categories", "category", _):
```
leandroalonso, line 196: Why do we need this change?

podcasts/Categories/CategoriesSelectorViewController.swift
```diff
@@ -0,0 +1,41 @@
+import SwiftUI
+import PocketCastsServer
+
+class CategoriesSelectorViewController: UIHostingController<CategoriesSelectorView>, DiscoverSummaryProtocol {
+
+    class Observable: ObservableObject {
```
leandroalonso: Can we rename that to `DiscoverItemObservable`?

I think it conveys a better name because `Observable` is super generic and it took me a while to understand why we were using this.

Same on `CategoriesSelectorView`.

Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
```diff
@@ -28,7 +29,14 @@ public class DiscoverServerHandler {
     }
 
     public func discoverPage(completion: @escaping (DiscoverLayout?, Bool) -> Void) {
-        discoverRequest(path: ServerConstants.Urls.discover() + "ios/content.json", type: DiscoverLayout.self) { discoverItems, cachedResponse in
+        let contentPath: String
+        if FeatureFlag.categoriesRedesign.enabled {
+            contentPath = "ios/content_v2.json"
```
leandroalonso, line 34: To properly test the PR I had to change to `Staging`. Not sure if this is needed, but can you add it to the PR description?

leandroalonso (changes requested):
@bjtitus compared to the designs the spacing between the search and the pills is too big:

Screenshot 2024-04-08 at 15 07 28

I think is fine to address that in a future PR so just let me know how you'd like to handle it.

Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
bjtitus: Sorry, yes this is only on staging at the moment. Paths will be the same in production.

podcasts/DiscoverViewController.swift
bjtitus: It's to differentiate from the new "pills" summary style [added here](https://github.com/Automattic/pocket-casts-ios/pull/1602/files#diff-7ace00507c7ae070e7e49bb9884e38981c21f38977964a3ba9ace1adb32b26adR182). I didn't like that the order could matter here, so added the old summary style to be sure.

Do you think I should change it back just in case the summary style is missing or different somewhere and let the ordering decide?

podcasts/Categories/CategoriesSelectorViewController.swift
bjtitus: > Can we rename that to DiscoverItemObservable?

Sure! I was thinking it more as the observable object tied to the view controller so it could have other things than the Discover Item but we can change it if that ever happens.

bjtitus:
> compared to the designs the spacing between the search and the pills is too big

This was intentional because other non-light themes left the pills un-centered with the color difference. I'll address matching the color of the search bar separately since it introduces some issues with the color of the over-scroll area which are going to require extra code to fix.

leandroalonso (approved):

PR #1604, opened 2024-04-06
Categories Redesign: Discover loading changes

bjtitus:
Includes the changes necessary to reload Discover when selecting a category.

https://github.com/Automattic/pocket-casts-ios/assets/3250/91e80a75-d75e-40ff-b192-4c31cd0cd9e1

### Code Changes

* Adds an `NSDiffableDataSourceSnapshot` based `apply` method to `DiscoverViewController` - eventually, I think this screen should switch to using a `UICollectionView`, so this is a stepping stone in that direction.
* Adds a `showExpanded(item: DiscoverItem, category: DiscoverCategory?)` method to `DiscoverDelegate` for the `CategoriesSelectionViewController` to call.
* Adds a `reload(except: [DiscoverItem], category: DiscoverCategory)` method to `DiscoverViewController` which creates the `CategoryPodcastsViewController` and a new snapshot to reload Discover.
* Adds some animations by setting `alphas` of the added views to 0 and then animating them during the reload process.

## To test

* Build using the `Staging` build configuration
* Ensure the `categoriesRedesign` flag is on
* Navigate to the Discover tab
* Tap a Category Pill
* Notice that the Discover screen changes to a Category list but the Category Pills remain at the top and properly animate

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerStructs.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerStructs.swift
@@ -256,7 +256,7 @@ public struct DiscoverRegion: Decodable {
     public var flag: String
 }
 
-public struct DiscoverItem: Decodable {
+public struct DiscoverItem: Decodable, Equatable {
     public var id: String?
     public var uuid: String?
     public var title: String?
@@ -281,7 +281,7 @@ public struct DiscoverItem: Decodable {
     }
 }
 
-public struct CarouselSponsoredPodcast: Decodable {
+public struct CarouselSponsoredPodcast: Decodable, Equatable {
     public var position: Int?
     public var source: String?
 }
--- a/Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerStructs.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerStructs.swift
@@ -257,6 +257,7 @@ public struct DiscoverRegion: Decodable {
 }
 
 public struct DiscoverItem: Decodable {
+    public var id: String?
     public var uuid: String?
     public var title: String?
     public var type: String?
@@ -276,7 +277,7 @@ public struct DiscoverItem: Decodable {
         case isSponsored = "sponsored"
         case sponsoredPodcasts = "sponsored_podcasts"
         case expandedTopItemLabel = "expanded_top_item_label"
-        case type, title, source, regions, curated, uuid, popular
+        case type, title, source, regions, curated, uuid, popular, id
     }
 }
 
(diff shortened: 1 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerStructs.swift
```diff
@@ -276,11 +277,27 @@ public struct DiscoverItem: Decodable {
...
+        self.type = type
+        self.summaryStyle = summaryStyle
+        self.expandedStyle = expandedStyle
+        self.source = source
+        self.sponsoredPodcasts = sponsoredPodcasts
+        self.expandedTopItemLabel = expandedTopItemLabel
+        self.curated = curated
+        self.regions = regions
+        self.isSponsored = isSponsored
+        self.popular = popular
```
leandroalonso: Can't we just do:

```suggestion
    public init(id: String? = nil, title: String? = nil, source: String? = nil, regions: [String]) {
        self.id = id
        self.title = title
        self.source = source
        self.regions = regions
```

Since it's what is really used by the code and the rest we can omit?

podcasts/DiscoverViewController.swift
```diff
@@ -145,13 +145,29 @@ class DiscoverViewController: PCViewController {
         }
     }
 
+    /// Reloads discover, keeping the items listed in `exclude`
+    /// - Parameters:
+    ///   - items: Items to exclude from the reload process. These items will REMAIN in Discover
+    ///   - category: The `DiscoverCategory` to add to the layout. This is sort of an artifical `DiscoverLayout`.
+    func reload(except items: [DiscoverItem], category: DiscoverCategory) {
+        let categoryVC = CategoryPodcastsViewController(category: category)
+        categoryVC.registerDiscoverDelegate(self)
```
leandroalonso: [nit] Maybe we can just expose `delegate` and do the usual `categoryVC.delegate = ...`?

leandroalonso (approved):
Works as described!

Left some minor comments to improve code.

Modules/Server/Sources/PocketCastsServer/Public/Sharing/Structs/ServerStructs.swift
bjtitus: Thanks! Good catch 👍

podcasts/DiscoverViewController.swift
bjtitus: Sounds good 👍 

I had initially conformed to `DiscoverSummaryProtocol` but not needed any more.

PR #1606, opened 2024-04-08
Standard Tags: ensure metadata is updated when it needs to do so

leandroalonso:
I discovered an issue with the cache system we're using to show metadata, it prevents new episodes from ever updating until the `URLCache` expires.

For the user, what it causes is that episode metadata (such as artwork, show notes) never updates. It works like this:

1. The app fetches the metadata and store in the `URLCache`
2. When a new episode is released the app requests `ShowInfoDataRetriever` for the metadata
3. It looks into the `URLCache` and returns the cached data there
4. The cached data don't contain the new episode

All I did was to remove the `URLCache`, given we do not need it anymore, as we're persisting episode metadata into the database.

## To test

To see the bug in action use the app installed on your phone, go to Podcasts > Sort By Episode Release Date > open any podcast new episode and see that the show notes never displays (you need to have loaded the show notes before, otherwise it will just work).

Anyway, the PR can still be tested by:

1. Run the app
2. Open any episode
3. ✅ Ensure the show notes (Details) appear correctly

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Cache/ShowInfoDataRetriever.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Cache/ShowInfoDataRetriever.swift
@@ -2,13 +2,9 @@ import Foundation
 
 /// Request information about an episode using the show notes endpoint
 public actor ShowInfoDataRetriever {
-    private let showNotesUrlCache: URLCache
-
     private var dataRequestMap: [String: Task<Data, Error>] = [:]
 
-    public init() {
-        showNotesUrlCache = URLCache(memoryCapacity: 1.megabytes, diskCapacity: 10.megabytes, diskPath: "show_notes")
-    }
+    public init() { }
 
     public func loadShowInfoData(
         for podcastUuid: String
@@ -20,14 +16,8 @@ public actor ShowInfoDataRetriever {
         let url = ServerHelper.asUrl(ServerConstants.Urls.cache() + "mobile/show_notes/full/\(podcastUuid)")
         let request = URLRequest(url: url, cachePolicy: .reloadRevalidatingCacheData)
 
-        if let cachedResponse = showNotesUrlCache.cachedResponse(for: request) {
-            return cachedResponse.data
-        }
-
         let task = Task<Data, Error> {
-            let (data, response) = try await URLSession.shared.data(for: request)
-            let responseToCache = CachedURLResponse(response: response, data: data)
-            showNotesUrlCache.storeCachedResponse(responseToCache, for: request)
+            let (data, _) = try await URLSession.shared.data(for: request)
             dataRequestMap[podcastUuid] = nil
             return data
         }


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/44">7.61 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/Cache/ShowInfoDataRetriever.swift
```diff
@@ -20,14 +16,8 @@ public actor ShowInfoDataRetriever {
         let url = ServerHelper.asUrl(ServerConstants.Urls.cache() + "mobile/show_notes/full/\(podcastUuid)")
         let request = URLRequest(url: url, cachePolicy: .reloadRevalidatingCacheData)
 
-        if let cachedResponse = showNotesUrlCache.cachedResponse(for: request) {
```
geekygecko, line 23: Do you still only redownload the content if the last modified or etag have changed? Below is an example of the call. It calls the server to get the static immutable file location. Ideally you would only process this final json `/show_notes/e7a6f7d0-02f2-0133-1c51-059c869cc4eb/episodes_1712289890.json` once to save on downloads, client side processing, and server costs. 
```
wget -S https://cache.pocketcasts.com/mobile/show_notes/full/e7a6f7d0-02f2-0133-1c51-059c869cc4eb
--2024-04-09 07:06:13--  https://cache.pocketcasts.com/mobile/show_notes/full/e7a6f7d0-02f2-0133-1c51-059c869cc4eb
Resolving cache.pocketcasts.com (cache.pocketcasts.com)... 3.217.52.176, 3.217.9.18, 3.216.204.171
Connecting to cache.pocketcasts.com (cache.pocketcasts.com)|3.217.52.176|:443... connected.
HTTP request sent, awaiting response...
  HTTP/1.1 302 Found
  Date: Mon, 08 Apr 2024 21:36:14 GMT
  Content-Length: 0
  Connection: keep-alive
  vary: origin
  Location: https://shownotes.pocketcasts.com/show_notes/e7a6f7d0-02f2-0133-1c51-059c869cc4eb/episodes_1712289890.json
Location: https://shownotes.pocketcasts.com/show_notes/e7a6f7d0-02f2-0133-1c51-059c869cc4eb/episodes_1712289890.json [following]
--2024-04-09 07:06:14--  https://shownotes.pocketcasts.com/show_notes/e7a6f7d0-02f2-0133-1c51-059c869cc4eb/episodes_1712289890.json
Resolving shownotes.pocketcasts.com (shownotes.pocketcasts.com)... 104.22.71.132, 104.22.70.132, 172.67.25.191
Connecting to shownotes.pocketcasts.com (shownotes.pocketcasts.com)|104.22.71.132|:443... connected.
HTTP request sent, awaiting response...
  HTTP/1.1 200 OK
  Date: Mon, 08 Apr 2024 21:36:15 GMT
  Content-Type: application/json; charset=utf-8
  Transfer-Encoding: chunked
  Connection: keep-alive
  Last-Modified: Fri, 05 Apr 2024 04:04:52 GMT
  ETag: W/"8caf53682ef105234da3b896ab5382b6"
  x-amz-server-side-encryption: AES256
  CF-Cache-Status: DYNAMIC
  Server: cloudflare
  ```

leandroalonso: We handle the cache ourselves now, but this is outside this entity: https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/ShowInfoCoordinator.swift#L71

Client-side processing was our main goal when redesigning the solution for full notes (episode metadata), given that with the previous solution, we had an increase in memory crashes due to the size of the JSON and `JSONDecoder` performance.

I removed the `URLCache` from here because it's not necessary anymore. :)

geekygecko: I might be reading the Swift code wrong, but it only downloads the latest show notes file if you can't find an episode in your cache. Is that to reduce the amount the JSON is parsed? 

My only concern was that if any information changes about an older episode, the cache won't be updated, but I believe the previous implement did handle this using the `Last-Modified` header and 304 status code.

leandroalonso: You are right!

> Is that to reduce the amount the JSON is parsed?

This is to reduce the whole thing (network and processing).

> My only concern was that if any information changes about an older episode, the cache won't be updated, but I believe the previous implement did handle this using the Last-Modified header and 304 status code.

How often does that happen?

I can change this behavior for `7.62` — for this version, it's a bit risky as it won't be that simple and we won't have enough time to test IMHO. How's Android handling that? Before it was straightforward because the show notes were displayed only in two places:

1. Episode detail
2. Full player

But now we also have episode artwork, and these can be displayed on any list. Is Pocket Casts Android requesting the endpoint whenever an episode artwork is shown?

bjtitus (approved):
✅  Episode Detail loading looked good to me in testing

I didn't see any cache hits using `URLSession.shared.configuration.cache`, though.

Modules/Server/Sources/PocketCastsServer/Public/Cache/ShowInfoDataRetriever.swift
geekygecko: > But now we also have episode artwork, and these can be displayed on any list. Is Pocket Casts Android requesting the endpoint whenever an episode artwork is shown?

I would need to take a look when I'm back on Android. It feels like these calls should happen at the end of a refresh and only download the files that have changed, rather when you are navigating through the app. 🤔

leandroalonso: > It feels like these calls should happen at the end of a refresh and only download the files that have changed

I think this makes sense! Refresh and when loading a podcast, right?

PR #1613, opened 2024-04-11
UpNext on tab bar

SergioEstevao:
Fixes #

This PR implements the display of the UpNext on the tab bar and showing the profile link on the podcasts tab.

https://github.com/Automattic/pocket-casts-ios/assets/651601/ce5c13c3-1b0f-4716-90b3-639a1b3aaeef

| Light |  Dark |  Rose |
| - | - | - |
| Simulator Screenshot - iPhone 15 Pro - 2024-04-14 at 14 04 38 |  Simulator Screenshot - iPhone 15 Pro - 2024-04-14 at 14 05 00 | Simulator Screenshot - iPhone 15 Pro - 2024-04-14 at 14 05 17 |
|  Simulator Screenshot - iPhone 15 Pro - 2024-04-14 at 16 04 35 | Simulator Screenshot - iPhone 15 Pro - 2024-04-14 at 16 04 18 | Simulator Screenshot - iPhone 15 Pro - 2024-04-14 at 16 04 03 |

## To test

 - Start the app
 - Ensure that the FF UpNext on the tab bar is enabled
 - Restart the app if you need to enable it
 - Check if Up Next shows on the tab on the second position
 - Tap on it
 - Check if the Up Next shows correctly and you can interact with the item correctly
 - Go to the Podcasts tab
 - Check if on the top left you see the profile icon with the user avatar ( the default avatar will be show if the user is not logged in or does not have an avatar set)
 - Tap on the profile button
 - Check that the profile view shows up by being pushed to the navigation of podcasts
 - Test the navigation to profile with this navigation
   - Open the bookmarks screens when you don't have any bookmarks, tap on headphone settings and check if it navigates correctly  

-Check analytics events
 - Enable tracks logging on Profile -> Beta Developer
 - Tap on UpNext on the tab bar
 - Check if you see: `up_next_tab_opened`

## Checklist

When updating the spreadsheet I noticed that the format for tab events in the the `[tab name]_tab_opened` but on the app we are using this format `[tab name]_tab_open`. What is the correct one?


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/45">7.62</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

david-gonzalez-a8c:
Looking good.

I've noticed there's some transparency in the Up Next icon in the tab bar when not active, let's remove that and have all elements at the same opacity:

image

leandroalonso:
@SergioEstevao 

> When updating the spreadsheet I noticed that the format for tab events in the the [tab name]_tab_opened but on the app we are using this format [tab name]_tab_open. What is the correct one?

`tab_opened` is the correct one, can you change it there?

podcasts/PCViewController.swift
```diff
@@ -93,14 +99,15 @@ class PCViewController: SimpleNotificationsViewController {
     }
 
     @objc func refreshRightButtons() {
-        if supportsGoogleCast {
+        if supportsGoogleCast || !extraRightButtons.isEmpty {
```
leandroalonso, line 102: This logic seems wrong to me. Or maybe I'm missing something.

Given `supportsGoogleCast` is `false` but `extraRightButtons` is not empty. The Chromecast icon will be displayed even though it should not.

podcasts/UpNextViewController.swift
```diff
@@ -260,10 +261,18 @@ class UpNextViewController: UIViewController, UIGestureRecognizerDelegate {
             navigationItem.leftBarButtonItem = UIBarButtonItem(title: L10n.cancel, style: .plain, target: self, action: #selector(cancelTapped))
         } else if !isMultiSelectEnabled, PlaybackManager.shared.queue.upNextCount() > 0 {
             navigationItem.rightBarButtonItem = UIBarButtonItem(title: L10n.select, style: .plain, target: self, action: #selector(selectTapped))
-            navigationItem.leftBarButtonItem = UIBarButtonItem(title: L10n.done, style: .plain, target: self, action: #selector(doneTapped))
+            if showDone {
+                navigationItem.leftBarButtonItem = UIBarButtonItem(title: L10n.done, style: .plain, target: self, action: #selector(doneTapped))
+            } else {
+                navigationItem.leftBarButtonItem = nil
+            }
```
leandroalonso: Do we really need the `else`?

podcasts/UpNextViewController.swift
```diff
@@ -260,10 +261,18 @@ class UpNextViewController: UIViewController, UIGestureRecognizerDelegate {
...
         } else {
             navigationItem.rightBarButtonItem = nil
-            navigationItem.leftBarButtonItem = UIBarButtonItem(title: L10n.done, style: .plain, target: self, action: #selector(doneTapped))
+            if showDone {
+                navigationItem.leftBarButtonItem = UIBarButtonItem(title: L10n.done, style: .plain, target: self, action: #selector(doneTapped))
+            } else {
+                navigationItem.leftBarButtonItem = nil
+            }
```
leandroalonso: Same as before. I think we can omit it.

leandroalonso (changes requested):
@SergioEstevao there are still issues with these changes, in terms of functionality and design.

### Profile icon it's too big and the color is different from the rest of tab bar

The default profile icon is bigger than the rest of the items in the navigation bar. Also, the color it's using is different. Some examples:

| Example 1 | Example 2 | Example 3 |
| ---------- | ---------- | ----------- |
| Simulator Screenshot - iPhone 15 - 2024-04-12 at 11 46 15 | Simulator Screenshot - iPhone 15 - 2024-04-12 at 11 46 43 | Simulator Screenshot - iPhone 15 - 2024-04-12 at 11 47 04 |

## Dark Up Next

In Settings > Appearance if you have "Use Dark Up Next Theme" enabled it will use the dark theme anyways, and the final result is weird:



### "Playing icon" doesn't appear when not using dark up next

This is not due to your changes but something that should not be hard to fix. The animated "playing icon" doesn't appear when the theme is not dark on Up Next. Check it here on the first cell:



## Crash

We have multiple places relying on `switchToTab(.profile)`. All of those are gonna crash with the new addition. One example is the empty Bookmarks screen when you have Plus. We need to fix those and add a test step for each of them to make sure everything works as expected:

Screenshot 2024-04-12 at 11 54 02

podcasts/PCViewController.swift
SergioEstevao: Thanks for detecting this! Fix done!

podcasts/UpNextViewController.swift
SergioEstevao: I kept it because the button logic run multiple times so wanted to ensure that the state is always correct

SergioEstevao:
@leandroalonso and @david-gonzalez-a8c  thanks for your reviews! It's ready for another round!

podcasts/UpNextViewController.swift
```diff
@@ -66,7 +68,9 @@ class UpNextViewController: UIViewController, UIGestureRecognizerDelegate {
             upNextTable.register(UINib(nibName: "NothingUpNextCell", bundle: nil), forCellReuseIdentifier: UpNextViewController.noUpNextCell)
             upNextTable.register(UINib(nibName: "UpNextNowPlayingCell", bundle: nil), forCellReuseIdentifier: UpNextViewController.nowPlayingCell)
             upNextTable.backgroundView = nil
-
+            if showingInTab {
+                upNextTable.applyInsetForMiniPlayer()
```
leandroalonso: I think we should move this to `viewDidAppear`.

If a user has an empty Up Next and opens the app, then fills it with a lot of items, the last one will be hidden till the app is restarted. If we move to `viewDidAppear` it should fix this.

leandroalonso (changes requested):
@SergioEstevao all works as described.

Left one comment about a minor issue.

podcasts/UpNextViewController.swift
SergioEstevao: Good call I moved this to viewDidAppear as you suggested.

leandroalonso (changes requested):
@SergioEstevao for some reason the applied padding now is bigger after the change:



I guess it's adding up.

leandroalonso (approved):

PR #1615, opened 2024-04-12
Settings Sync: Fix Up Next settings

bjtitus:
Fixes a minor issue when importing the Up Next setting. Previously this passed through the new update mechanism which would set the `ModifiedDate` to the current date, so this setting would be overwritten by the last device to import, which was not the original guidelines. We want any imported values to be assigned an epoch date (done in https://github.com/Automattic/pocket-casts-ios/pull/1616), not the current date, so they are clearly delineated from changed settings.

## To test

* Change the return value of `shouldEnableSyncedSettings` in `FeatureFlag.swift` to false
* Subscribe to a podcast
* Enable "Add to Up Next" in the Podcast settings
* Enable the `newSettingsStorage` and `settingsSync` feature flags
* Restart the app
* Ensure your "Add to Up Next" setting remains the same

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/PodcastSettings+OldSettings.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/PodcastSettings+OldSettings.swift
@@ -23,9 +23,7 @@ extension PodcastSettings {
 
 public extension AutoAddToUpNextSetting {
     var enabled: Bool {
-        get {
-            return self != .off
-        }
+        return self != .off
     }
 
     var position: UpNextPosition? {
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/PodcastSettings+OldSettings.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/PodcastSettings+OldSettings.swift
@@ -13,16 +13,29 @@ extension PodcastSettings {
             }
         }
         set {
-            switch newValue {
-            case .addFirst:
-                addToUpNext = true
-                addToUpNextPosition = .top
-            case .addLast:
-                addToUpNext = true
-                addToUpNextPosition = .bottom
-            case .off:
-                addToUpNext = false
+            addToUpNext = newValue.enabled
+            if let position = newValue.position {
+                addToUpNextPosition = position
             }
+        }
+    }
+}
+
+public extension AutoAddToUpNextSetting {
+    var enabled: Bool {
+        get {
+            return self != .off
+        }
+    }
+
+    var position: UpNextPosition? {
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/46">7.63</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/PodcastSettings+OldSettings.swift
```diff
@@ -13,16 +13,29 @@ extension PodcastSettings {
...
-                addToUpNext = true
-                addToUpNextPosition = .top
-            case .addLast:
-                addToUpNext = true
-                addToUpNextPosition = .bottom
-            case .off:
-                addToUpNext = false
+            addToUpNext = newValue.enabled
+            if let position = newValue.position {
+                addToUpNextPosition = position
```
leandroalonso, line 18: Not sure if it matters, but shouldn't `addToUpNextPosition` be optional? In case it has no position, is nil.

As it stands it will be the previous value that had a `position`.

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/PodcastSettings+OldSettings.swift
```diff
@@ -13,16 +13,29 @@ extension PodcastSettings {
...
         }
     }
 }
+
+public extension AutoAddToUpNextSetting {
+    var enabled: Bool {
+        get {
+            return self != .off
+        }
+    }
```
leandroalonso:
```suggestion
    var enabled: Bool {
        self != .off
    }
```

leandroalonso (changes requested):
@bjtitus this seems to not be working for me:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/840c8175-a31c-4b81-baab-d72592b115ce

If I change the flag to `false`, the correct values return.

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/PodcastSettings+OldSettings.swift
bjtitus: It can't be optional in the Protobuf settings so I was thinking we shouldn't make it optional in our local model.

PR #1620, opened 2024-04-15
Animation + style tweaks on Discover Categories

bjtitus:
Fixes a problem with reloading the categories selector view when navigating back away from a selected category. This was a regression from a previous PR.

| Before | After |
| -- | -- |
| Before | CleanShot 2024-04-15 at 16 14 18 |

Adjusts the close button to be fully circular:

Adds maximum number of 3 lines to sponsored Podcasts:

## To test

Build using `Staging` configuration

* Enable the `categoriesRedesign` feature flag.
* Force reload discover from the Developer menu
* Try tapping one of the categories in the selector
* Ensure the animations works properly - it shouldn't fade the selector out, it should animate the selected pill
* Tap the X button to close
* Ensure the animations works properly - it shouldn't fade the selector out, it should animate the selected pill

### Sponsored Podcast

* Navigate to the "Sports" category
* Ensure that the Sponsored Podcast on top is truncated to only 3 lines

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/45">7.62 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/Categories/CategoryButtonStyle.swift
```diff
@@ -33,24 +34,65 @@ struct CategoryButtonStyle: ButtonStyle {
...
+                if #available(iOS 16.0, *) {
+                    $0
+                        .padding(cornerStyle == .circle ? 3 : 0)
+                        .overlay(
+                        cornerStyle.shape
+                            .stroke(isSelected ? selectedBackground : border, lineWidth: 1)
+                    )
+                    .clipShape(cornerStyle.shape)
+                } else {
+                    $0.overlay(
```
bjtitus, line 96: I'm just falling back to RoundedRect here since `AnyShape` is iOS 16+

I don't think it's a huge deal to use the non-circular style and it didn't seem worth the effort to work around since we're likely to drop iOS 15 soonish.

podcasts/DiscoverViewController+DiscoverDelegate.swift
```diff
@@ -7,7 +7,7 @@ extension DiscoverViewController: DiscoverDelegate {
         if let category {
             reload(except: [item], category: category)
         } else {
-            reloadDiscoverTapped(NSObject())
+            reload(except: [item], category: nil)
```
bjtitus, line 10: Reload needs to be different than a full reload so we should use the same mechanism as reloading a category.

If we switch this to a `CollectionView` we may be able to rely on the data source diffing to animate more cleanly between states.

leandroalonso (approved):
@bjtitus works as described. Two nit though:

The separator between the ad and the podcasts is not aligned in the center (is more distant from the ad and close to the first podcast):



When selecting a category that is close to the border of the screen, it will bounce: it will move to its original position and then go to the beginning:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/0b3c1de8-acbb-4aea-b2c2-ec543f93cd19

I believe this one might be tricky to fix so if that's the case just ignore it.

bjtitus:
I fixed the first one by adjusting the spacing specifically for `CategoryPodcastViewController`s added to Discover.

There's still a lot of space above the sponsored cell but I'll take a look at that separately when I add the Most Popular cell as well since it will be at the top.

@leandroalonso Could you take a quick peak at the code now that it's changed?

| Before | After |
| -- | -- |
| Simulator Screenshot - iPhone 15 Pro Max - 2024-04-16 at 12 06 43 | Simulator Screenshot - iPhone 15 Pro Max - 2024-04-16 at 12 05 03 |

> When selecting a category that is close to the border of the screen, it will bounce

Yeah, I'm not sure where to begin with this so I think we should take a look at it separately.

leandroalonso (approved):

PR #1627, opened 2024-04-17
Sleep Timer Improvements: shake device to restart sleep timer

leandroalonso:
This PR adds the ability to shake the device so the sleep timer is restarted. This way, users can restart the sleep timer without unblocking their phones.

## Additional info

The traditional way of checking a device shake through `motionEnded` doesn't work if the app is in the background. The only way to achieve that is by using `CMMotionManager`.

I created an entity called `BackgroundShakeObserver` that only starts observing device motion while:

1. A timer is going on
2. The app is in the background

My tests show energy consumption shouldn't be higher, but we'll have to follow up.

## To test

As you can imagine, test on a real device. :)

Go to Profile > Settings > Beta features > enable tracksLogging.

1. Run the app
2. Start a timer
3. Shake the device
4. ✅ Timer should restart and you should see a toast saying "Sleep timer restarted due to device shake"
5. ✅ You should see `playerSleepTimerRestarted` being fired with `reason: device_shake`
7. Put the app on background and lock the phone
8. Shake the device
9. ✅ You should see `playerSleepTimerRestarted` being fired with `reason: device_shake`
10. ✅ A subtle confirmation sound should play
11. Open the app again and check the timer
12. ✅ It should have restarted

## Checklist


podcasts/SleepTimerManager.swift
```diff
@@ -39,6 +49,16 @@ class SleepTimerManager {
         }
     }
 
+    func restartSleepTimer() {
+        if let setting = Settings.sleepTimerLastSetting {
+            if let duration = setting.duration {
```
leandroalonso, line 71: We only restart when there's a timer going on with duration, as this wouldn't make much sense for the episode end.

podcasts/SleepTimerManager.swift
```diff
@@ -39,6 +49,16 @@ class SleepTimerManager {
...
+                PlaybackManager.shared.setSleepTimerInterval(duration)
+                Analytics.shared.track(.playerSleepTimerRestarted, properties: ["time": duration, "reason": "device_shake"])
```
leandroalonso, line 73: @mebarbosa I added a `reason: device_shake` in this event in case we want to differentiate from the rest in the future.

mebarbosa: I will update Android to match with it 👌🏽

bjtitus (approved):
:shipit: Worked really well!

I did notice some resetting close together, probably due to how I'm shaking the phone 😆 . Probably not a big deal in practice since it's basically the same amount of time.

PR #1628, opened 2024-04-18
New podcasts grid design

SergioEstevao:
Fixes #

This PR updates the layout for the Podcast grid adding spacing and a new badge design.

https://github.com/Automattic/pocket-casts-ios/assets/651601/2244b8e9-df6e-451a-91ef-77d6d3ad6db2

| Large Grid | Small Grid | List |
| - | - | - |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 13 09 00 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 13 09 03 |  Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 13 09 07 |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 13 08 36 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 13 08 33 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 13 08 27 |

## To test

 - Start the app
 - Open the Podcasts tab
 - Test the following
 - Check if the new layout is working correctly, by scrolling around on you podcast view
 - Open folders ( make a folder if you don't have one)
 - Tap on the top right more button, and switch between the different layout types: Small Grid, Large Grid, List. 
 - On each grid types check the different badge options: Off, Unfinished Episodes, Only Latest Episode
 - Check with different themes
 - Check on the iPad

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/FolderListCell.swift
```diff
@@ -49,5 +49,6 @@ class FolderListCell: ThemeableCollectionCell {
         } else {
             unplayedBadge.isHidden = true
         }
+        unplayedBadge.updateColors()
```
leandroalonso, line 53: [nit] (adds breakline)

```suggestion

        unplayedBadge.updateColors()
```

podcasts/FolderViewController.swift
```diff
@@ -56,6 +56,8 @@ class FolderViewController: PCViewController, UIGestureRecognizerDelegate {
         mainGrid.addGestureRecognizer(longPressGesture)
         longPressGesture.delegate = self
 
+        miniPlayerStatusDidChange()
```
leandroalonso, line 59: Do we need that here given it's on `viewDidAppear`?

podcasts/FolderViewController.swift
```diff
@@ -96,6 +98,10 @@ class FolderViewController: PCViewController, UIGestureRecognizerDelegate {
         let titleColor = ThemeColor.filterText01(filterColor: folderColor)
         let iconColor = ThemeColor.filterIcon01(filterColor: folderColor)
         let backgroundColor = ThemeColor.filterUi01(filterColor: folderColor)
+
+        if let themeableCollectionView = mainGrid as? ThemeableCollectionView {
+            themeableCollectionView.backgroundColor = backgroundColor
```
leandroalonso: This change causes changes such as this (Default Dark):

| Before | After |
| ------ | ----- |
| Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-04-19 at 13 17 57 | Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-04-19 at 13 15 46 |

Before looks better for me because the cells almost merge with the background.

podcasts/PodcastListCell.swift
```diff
@@ -42,6 +42,7 @@ class PodcastListCell: ThemeableCollectionCell {
         } else {
             unplayedBadge.isHidden = true
         }
+        unplayedBadge.updateColors()
```
leandroalonso, line 46: [nit]

```suggestion

        unplayedBadge.updateColors()
```

podcasts/PodcastListViewController.swift
```diff
@@ -87,14 +87,17 @@ class PodcastListViewController: PCViewController, UIGestureRecognizerDelegate,
         longPressGesture.delegate = self
 
         gridHelper.configureLayout(collectionView: podcastsCollectionView)
+        if let themeableCollectionView = podcastsCollectionView as? ThemeableCollectionView {
```
leandroalonso: [nit]

```suggestion

        if let themeableCollectionView = podcastsCollectionView as? ThemeableCollectionView {
```

leandroalonso (changes requested):
@SergioEstevao all good in the code, I did find some issues testing with different themes:

### Animation

Is this animation made on purpose? I don't see it on iPad and it seems weird that the podcasts appear from left to right:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/336988f0-2091-46b3-8716-c284de991891

### Margin on the bottom

When scrolling all the way through the bottom, the bottom margin almost doesn't exist:



### Podcasts background color

The podcasts background color matches the navigation bar for some themes. This is a bit weird, if you take a look at all the other screens this doesn't happen. Also, for classic, the badge red is the same one as the background so it's almost invisible:

| Example 1 | Example 2 |
| ---------- | ---------- |
| Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-04-19 at 13 23 47 | Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-04-19 at 13 25 29 |

podcasts/FolderViewController.swift
SergioEstevao: On the list view the before is the correct one, I will address this.

SergioEstevao:
@leandroalonso thanks for you review!

Regarding your comments:

- Animation: not on purpose I think this could be because of some discrepancy on the insets when opening the view, I will give it a look
- Margin on the bottom: I will increase it a bit to compensate the extra insets for spacing between cells
- Podcasts colors: This was according to design but I agree with you that it does not look good on those themes. I will let @david-gonzalez-a8c comment on that one to see if we want to go back to the previous background colors that is different from the navigation bar background.

david-gonzalez-a8c:
>  I will let @david-gonzalez-a8c comment on that one to see if we want to go back to the previous background colors that is different from the navigation bar background.

Agree, let's have it match Filters or Discover with the background color (not sure which color variable is used there, but the specs show $primary-ui-02)

SergioEstevao:
@david-gonzalez-a8c  looking at the code we are using $primary-ui-04 for background of collection/lists and the $primary-ui-02 for the cells background.

Here is a screenshot going back to $primary-ui-04

Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 20 23 22

podcasts/FolderViewController.swift
SergioEstevao: Yes or else we see the inset changing animating.

SergioEstevao:

david-gonzalez-a8c:
Can we see how it looks with $primary-ui-02?

SergioEstevao:
> Can we see how it looks with $primary-ui-02?

@david-gonzalez-a8c here you go, I think the issue of using this color is that we will need to have or a different background color for lists or update the cell colors for list

| 1 | 2 | 3 | 4 |
| - | - | - | - |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 22 34 52 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 22 34 41 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 22 34 25 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-19 at 22 34 06 |

leandroalonso (changes requested):
@SergioEstevao the colors are much better now! Tested the mentioned issues and they're gone. One last final touch:



Given we reduced the size of each square, the font for a folder in this case looks too big. I think we should slightly reduce it, wdyt?

SergioEstevao:
> Given we reduced the size of each square, the font for a folder in this case looks too big. I think we should slightly reduce it, wdyt?

@leandroalonso I agree with you, @david-gonzalez-a8c what do you think?

david-gonzalez-a8c:
Agree, this looks a bit tight, also in terms of the margins to the outer container. I suggest we reduce the covers a bit too.

Let's try that first, but if we can't adjust it successfully, I can work on an alternative version (only applied to the small grid folder) with a different design, maybe something like the covers stacked up.

SergioEstevao:
@leandroalonso and @david-gonzalez-a8c I tweaked the tile size and font size a bit here are some screenshots.

Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-22 at 18 24 00
Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-22 at 18 23 55

What do you say?

david-gonzalez-a8c:
Much better, nice job!

I'd push the text up a little bit so the top and bottom margins are more even:

image

SergioEstevao:
@david-gonzalez-a8c I updated the margins here is the result:

simulator_screenshot_1542EE53-C94E-4566-9BA4-EDFBC2C93FC4
simulator_screenshot_733C57AD-2F32-4E79-A62E-3CCA6F5D8B45

leandroalonso (changes requested):
For the 4-per-row grid the folder title it's aligned at the top:

SergioEstevao:
@leandroalonso what device are you using, if you see my screenshot above the layout looks good... maybe it's some calculation issue on smaller/larger devices...

leandroalonso:
@SergioEstevao iPhone 15 (iOS 17.2), Simulator

SergioEstevao:
@leandroalonso improved the label positioning across devices as you can see here:

| A | B |
| - | - |
| Simulator Screenshot - iPhone 15 Pro - 2024-04-23 at 16 56 15 | Simulator Screenshot - iPhone 15 Pro - 2024-04-23 at 16 56 01 |

leandroalonso (approved):

deaxes:
Please Revert - this is a horrible UI design that hurts my eyes.

PR #1629, opened 2024-04-18
Sleep Timer Improvements: fade out 5s before pausing playback

leandroalonso:
When a Sleep Timer is 5s from finishing, the audio has faded out.

I did some research and found that for human hearing a logarithmic fade-out (instead of an exponential one) sounds more natural, so I went ahead and implemented it.

## To test

Go to `SleepTimerViewController.swift` and change `fiveMinutesTapped` to `10.seconds`, so it's easier to test.

### DefaultPlayer

1. Go to `PlaybackManager` and comment lines `1155` to `1157` (to ensure `EffectsPlayer` is not used)
2. Run the app
3. Play any episode
4. Open the full player, tap the sleep timer icon (zZz), tap "5 minutes"
5. ✅ On the 5s mark the audio should start to fade out until it's paused
6. Play the episode again
7. When it starts fading, pause the playback
8. Resume it
9. ✅ The remaining time should play without fade out

### EffectsPlayer

1. Remove the changes you made on `PlaybackManager`
2. Download an episode
3. Play the downloaded episode
4. Open the full player, tap the sleep timer icon (zZz), tap "5 minutes"
5. ✅ On the 5s mark the audio should start to fade out until it's paused
6. Play the episode again
7. When it starts fading, pause the playback
8. Resume it
9. ✅ The remaining time should play without fade-out

## Checklist


podcasts/FadeOutManager.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+
+class FadeOutManager {
```
leandroalonso, line 3: I don't like this name at all, but couldn't think of something better.

SergioEstevao: Tricky one maybe:`VolumeControl` or `FaderMixer`

SergioEstevao (approved):
Working great! Tested with the DefaultPlayer and EffectsPlayer and the fade out is pretty smooth.

PR #1637, opened 2024-04-22
Fix: Ensure tab bar icons respect increase contrast setting for all combinations of PC themes and iOS appearance

bikrrr:
Set the `tabBar`'s `userInterfaceStyle` to match Pocket Casts' Theme type so iOS's **Increase Contrast** setting always increases icon contrast.

## Problem
When a _single_ Pocket Cast theme is selected and doesn't change based on iOS light/dark mode, iOS's **Accessibility › Display & Text Size › Increase Contrast** setting doesn't always increase the contrast of the tabBar icons. If the PC theme is light and the iOS theme is dark, the icons will actually get lighter instead of darker. The same is true when light/dark are reversed.

Here's a summary of the different combinations. Notice the **increase contrast** cells aren't correct (❌) when the Pocket Casts theme is _opposite_ the iOS appearance.

| iOS Appearance/Contrast| Pocket Cast "Default Light" | "Default Dark" |
|--------|--------|--------|
| Light/Standard | Screenshot 2024-04-22 at 11 24 24@2x | Screenshot 2024-04-22 at 11 25 23@2x |
| Light/Increased | ✅ Screenshot 2024-04-22 at 11 24 37@2x | ❌ Screenshot 2024-04-22 at 11 25 35@2x |
| Dark/Standard | Screenshot 2024-04-22 at 11 24 10@2x | Screenshot 2024-04-22 at 11 25 49@2x |
| Dark/Increased | ❌ Screenshot 2024-04-22 at 11 23 07@2x | ✅ Screenshot 2024-04-22 at 11 26 45@2x |

**Video of Issue**

https://github.com/Automattic/pocket-casts-ios/assets/2607653/1e23403a-a77e-45c2-a62c-99d7c0ec8926

## Fix
This fix sets the overrideUserInterfaceStyle so it matches the PC theme. When **Increase Contrast** is enabled, it always makes it darker with a light theme and lighter with a dark theme, regardless of whether iOS is dark or light (since the background color isn't changing with iOS).

**Video of Fix**

https://github.com/Automattic/pocket-casts-ios/assets/2607653/1fd9f4f7-7a99-4e77-9e59-d45631162184

### Single Pocket Casts Theme

1. **Pocket Casts › Settings › Appearance**, turn off **Use iOS Light/Dark Mode**
2. Select **Default Light** theme
3. Set iOS **Appearance** to **Light**
[...]

PR #1639, opened 2024-04-23
Fix: Ensure nav bar icons respect 'increase contrast' for all combos of PC theme and iOS appearance 

bikrrr:
Set the `navigationBar`'s `userInterfaceStyle` to match Pocket Casts' Theme type so iOS's **Increase Contrast** setting always increases icon contrast.

## Problem
When a _single_ Pocket Cast theme is selected and doesn't change based on iOS light/dark mode, iOS's **Increase Contrast** accessibility setting doesn't always increase the contrast of the navigationBar icons. If the PC theme is light and the iOS theme is dark, the icons will actually get _lighter_ instead of darker. And it's also a problem when light/dark are reversed.

Here's a summary of the different combinations. Notice the **increase contrast** cells aren't correct (❌) when the Pocket Casts theme is _opposite_ the iOS appearance.

| iOS Appearance / Contrast| Pocket Cast "Default Light" | "Default Dark" |
|--------|--------|--------|
| Light / Standard | DefaultLight01 | DefaultDark01 | 
| Light / Increased | ✅ DefaultLight02 | ❌ DefaultDark04 |
| Dark / Standard | DefaultLight01 | DefaultDark03 |
| Dark / Increased | ❌ DefaultLight04  | ✅  DefaultDark02 |

**Video of Issue**

https://github.com/Automattic/pocket-casts-ios/assets/2607653/ece04c52-b621-417a-91b7-d894d67d6e8e

## Fix
This fix sets the `overrideUserInterfaceStyle` on the `navigationBar` so it matches the Pocket Casts' theme. When **Increase Contrast** is enabled, it always makes it darker with a light theme and lighter with a dark theme, regardless of whether _iOS_ is dark or light (since the background color doesn't change with iOS appearance).

**Video of Fix**

https://github.com/Automattic/pocket-casts-ios/assets/2607653/b3a6fbb1-6209-4669-b113-67b2a59bdbcc

### Single Pocket Casts Theme

1. In Pocket Casts, click **Profile menu › Settings › Appearance**, turn OFF **Use iOS Light/Dark Mode** and select **Default Light** theme
2. Go back to **Profile** screen
3. Confirm iOS **Appearance** is set to **Light** (**Cmd+Shift+A** in simulator)
4. Toggle on/off iOS **Settings › Accessibility › Display & Text Size › Increase Contrast** (simulator menu **Features › Toggle Increased Contrast**)
5. ✅ In the top navigation bar, confirm the **left chevron** and **gear** icons get _darker_ when **Increase Contrast** is on
6. Set iOS **Appearance** to **Dark**
7. Toggle on/off iOS **Settings › Accessibility › Display & Text Size › Increase Contrast** (simulator menu **Features › Toggle Increased Contrast**)
8. ✅ Confirm the **left chevron** and **gear** icons get _darker_ when **Increase Contrast** is on

### Separate Pocket Casts Themes for iOS Light/Dark mode
**Note**: Issue doesn't affect this use case, but need to test to make sure it still works properly
1. In Pocket Casts, click **Profile menu › Settings › Appearance**, turn ON **Use iOS Light/Dark Mode**
2. For **Light Theme**, select **Default Light**
3. For **Dark Theme**, select **Default Dark**
4. Go back to **Profile** screen
5. Set iOS **Appearance** to **Light**
6. Toggle on/off iOS **Settings › Accessibility › Display & Text Size › Increase Contrast** (simulator menu **Features › Toggle Increased Contrast**)
7. ✅ Confirm the tab bar icons get _darker_ when **Increase Contrast** is on
8. Set iOS **Appearance** to **Dark**
9. Toggle on/off iOS **Settings › Accessibility › Display & Text Size › Increase Contrast** (simulator menu **Features › Toggle Increased Contrast**)
10. ✅ Confirm the tab bar icons get _lighter_ when **Increase Contrast** is on

## Checklist

### DIFF
--- a/podcasts/PCNavigationController.swift
+++ b/podcasts/PCNavigationController.swift
@@ -57,6 +57,9 @@ class PCNavigationController: UINavigationController, UIGestureRecognizerDelegat
         labelAppearance.adjustsFontSizeToFitWidth = true
         labelAppearance.minimumScaleFactor = 0.75
         labelAppearance.baselineAdjustment = .none
+
+        // Link UserInterfaceStyle to Theme type so iOS's Increase Contrast does the right thing
+        navigationBar.overrideUserInterfaceStyle = Theme.isDarkTheme() ? .dark : .light
     }
 
     override var preferredStatusBarStyle: UIStatusBarStyle {


bikrrr:
This is a similar fix as [PR #1637](https://github.com/Automattic/pocket-casts-ios/pull/1637), but for the top navigation bar.

bikrrr:
+@leandroalonso 

I also want to tackle the icons in the **Profile**, **Settings**, etc. screen `tableView`s as they exhibit the same problem, but I think I'd just be repeating code in many places. I'm new at this but know enough that that's a red flag ([DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)). Perhaps overriding `UIImage` with an extension that sets `overrideUserInterfaceStyle` at a higher level would be a better solution? 🤔

I'm still learning and this codebase is large so that may be a bigger challenge than what I'm ready for. But if someone could point me in the right direction, I wouldn't mind trying it.

bjtitus (approved):
Thanks for the excellent write-up @bikrrr!

:shipit: Everything seemed to work in testing and the icons look much better!

PR #1675, opened 2024-04-30
Standard Tags: disable episode artwork on cells and rely on `URLCache`

leandroalonso:
For users with big databases, saving one episode of metadata takes ~1ms. This doesn't look like a big problem, but if you have a huge podcast (such as The Daily) with 2k episodes, this will take ~2s.

All the other queries need to wait for this one to finish, and if you're scrolling through the Filters, for example, this can go exponentially high to the point it's impossible to use the app.

The logic here is changed so we rely on `URLCache` until we're sure we don't have performance issues. It also disabled enabling artwork on episode cells.

## To test

Go to `FeatureFlag.swift` and change `newShowNotesEndpoint`, `episodeFeedArtwork` and `rssChapters` to `true`.

1. Run an instance of the app with a big database 
2. Check this branch and run it ¶
3. Go to a Filter (preferably the one with the most number of episodes)
4. Scroll as fast as you can
5. ✅ You shouldn't experience big freezes
6. Play an episode from a podcast that has artwork (like Clublife)
7. ✅ Episode artwork should appear

¶ Without the debugger attached (you can just close and reopen the app). I experienced that the debug can slow down the app a lot for huge databases.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/Cache/ShowInfoDataRetriever.swift
```diff
@@ -14,11 +18,30 @@ public actor ShowInfoDataRetriever {
         }
 
         let url = ServerHelper.asUrl(ServerConstants.Urls.cache() + "mobile/show_notes/full/\(podcastUuid)")
-        let request = URLRequest(url: url, cachePolicy: .reloadRevalidatingCacheData)
+        var request = URLRequest(url: url, cachePolicy: .reloadRevalidatingCacheData)
+
+        if let cachedResponse = cache.cachedResponse(for: request) {
```
leandroalonso: I expected `URLSession` to take care of this for us, but guess what: it doesn't.

I'm not sure if this is because the `mobile/show_notes/full/` redirects to another URL but the cache never worked for some podcasts (while it did work for some others).

SergioEstevao: Could it because of the policy specified on the request: `.reloadRevalidatingCacheData` does our server works correctly with those?

SergioEstevao: I also noticed that we use `URLSession.shared` for our requests, this has a cache but it's being shared with all other network request so maybe it can get easily recycled.

SergioEstevao (approved):
I tested this with my own database and the large database and things went smoothly I couldn't notice any hangs on the filters views with all objects and also by duration.

SergioEstevao (commented):
I had another go using the big database that Leandro shared on Slack.
I deleted completely the app and started fresh with that database.
The app started well but if I go immediately to filters open All filters and scroll a bit I see some initial hangs but after some seconds it stabilizes.
After that initial hangs if I go around on other filters things work correctly.
There must be some initial steps done on first launch: saving on database and initial cache misses that makes things slow with too much writes on the DB.

SergioEstevao:
Curiously when trying to do it again, using the same large database,  on the same simulator/device with a clean install, I couldn't notice the hangs again that makes me believe there is also some device network calls cache making a difference

bjtitus (approved):
✅ Episode artwork still works

PR #1692, opened 2024-05-02
7.63 Release: Merge changes from 7.63.0.2 into `trunk`

pocketcasts:
Merge changes from 7.63 (7.63.0.2) to `trunk`.

## To test

- Ensure the build is 🟢
- The code changes here were tested in their own PRs


leandroalonso:
I won't merge that into `trunk` until we're sure this solves the issues.

PR #1702, opened 2024-05-03
Performance: remove bookmark check

leandroalonso:
Every episode that is shown is checked to whether they have a bookmark or not. This query is not expensive, however, if there's any other more expensive query in the queue, this will block the whole app.

I temporarily disabled this call here and this causes a regression: the bookmark icon is not shown anymore. Of course, we will re-add this back in a non-blocking way soon.

## To test

1. Run the app with a huge database
2. Go to a long list of episodes
3. Scroll down
4. ✅ Scroll should be as smooth as possible

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Episode.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Episode.swift
@@ -47,7 +47,12 @@ public class Episode: NSObject, BaseEpisode {
     @objc public var deselectedChaptersModified = 0 as Int64
 
     public var hasBookmarks: Bool {
-        DataManager.sharedManager.bookmarks.bookmarkCount(forEpisode: uuid) > 0
+        // This wil cause a regression in which the bookmarks won't be displayed
+        // for episodes with bookmarks.
+        // However, this call is happening on the main thread and can block the whole app.
+        // We will re-add this again in a way that's not a blocker
+        //DataManager.sharedManager.bookmarks.bookmarkCount(forEpisode: uuid) > 0
+        false
     }
 
     public var isUserEpisode: Bool {


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

bjtitus:
@leandroalonso Would you rather remove this or use this optimized `hasBookmarks` query?

```swift
/// Returns a boolean indicating whether bookmarks exist for the given episode and can optionally include deleted items in the query
public func hasBookmarks(forEpisode episodeUuid: String, includeDeleted: Bool = false) -> Bool {
	let deletedWhere: String? = includeDeleted ? nil : "\(Column.deleted) = 0"

	let whereString = [deletedWhere, "\(Column.episode) = ?"]
		.compactMap { $0 }.joined(separator: " AND ")

	let query = "SELECT EXISTS(SELECT 1 FROM \(Self.tableName) WHERE \(whereString)) AS record_exists;"

	var hasBookmarks = false
	dbQueue.inDatabase { db in
		do {
			let resultSet = try db.executeQuery(query, values: [episodeUuid])
			resultSet.next()
			hasBookmarks = resultSet.bool(forColumnIndex: 0)
			resultSet.close()
		} catch {
			FileLog.shared.addMessage("BookmarkManager.bookmarkCount failed: \(error)")
		}
	}

	return hasBookmarks
}
```

I can submit separately when we re-enable if we'd just like to avoid for 7.63.

SergioEstevao (approved):
Working butter smooth with a big database opening The Daily!

leandroalonso:
@bjtitus The problem with ANY query is that if it happens on the main thread and there's an expensive query going on somewhere else, it will freeze. It doesn't matter how fast the query is.

The main problem here is not the query per-se, but the fact that is happening on the Main Thread. We can switch to yours + move it to a background thread.

PR #1708, opened 2024-05-04
Improve folders performance

SergioEstevao:
Fixes #

This PR implements a folder cache of it's top podcasts uuids

## To test

- Start app using a large database with folders with lot of podcasts episodes
- Open podcasts view
- Scroll down ( still some slow down can happen while the cache is being fulfilled)
- Scroll back up and scroll performance should be optimal.
- Open a folder
- Change folder properties like: sort order, name, color
- Go back to the root level
- Check if folder image is updated
- Search in the podcasts view for some folder name, see if the results cells show the folders correctly

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/PodcastImageView.swift
```diff
@@ -44,9 +44,9 @@ class PodcastImageView: UIView {
     }
 
     func setTransparentNoArtwork(size: PodcastThumbnailSize) {
+        imageView?.kf.cancelDownloadTask()
```
leandroalonso, line 47: Why do we need this change?

Asking because very recently we had issues with images blinking so unless this is strictly necessary I think we should let this out for now.

leandroalonso (changes requested):
@SergioEstevao there's still a regression when compared to the latest version: when I unsubscribed to a podcast it still appears at the folder icon until I re-run the app.

podcasts/PodcastImageView.swift
SergioEstevao: I added this because if the ImageView was being used for a async loading of image, we could have a scenario where we set the image to nil, but meanwhile the download finishes and we set it to the the podcast image previously being loaded.

leandroalonso: I see that this method is only used in `FolderPreviewView.swift` so we should be fine?

SergioEstevao: Even with only that use this can happen, because we know are reusing the image views for folder,  so if you set to transparent, but on the previous use of the FolderPreviewView it was being loaded we need to cancel that download.

SergioEstevao:

leandroalonso (commented):
@SergioEstevao even though this code does some sort of cache, I wonder if our users would benefit more from doing the `FolderGridCell` population async which will solve the slowdown?

This way we won't cause any lag while the database operations are going on. 

Take `FolderPreviewView.populateFrom`:

```swift
    func populateFrom(folder: Folder) {
        let podcastUuids = DataManager.sharedManager.allPodcastsInFolder(folder: folder).map(\.uuid)
        setup(folderName: folder.name, folderColor: folder.color, topPodcastUuids: podcastUuids)
    }
```

If we can do `DataManager.sharedManager.allPodcastsInFolder(folder: folder).map(\.uuid)` async and only then call `setup` we'll probably eliminate this issue.

SergioEstevao:
> f we can do DataManager.sharedManager.allPodcastsInFolder(folder: folder).map(\.uuid) async and only then call setup we'll probably eliminate this issue.

It's not that linear, because we will need to have some way to create a placeholder folder image while the async call is ongoing. 
I will try to do placeholder that will use only the folder color and 4 transparent images while the podcasts images load.
But I still think it's worth to have the cache because you will don't need to run this query when scrolling around and locking up the Database queue.

leandroalonso:
@SergioEstevao I agree we should also have the cache.

But I think the placeholder can be easily accomplished with the code we already have, take this as an example:

```swift
    func populateFrom(folder: Folder) {
        let podcastUuids = DataManager.sharedManager.allPodcastsInFolder(folder: folder).map(\.uuid)
        setup(folderName: folder.name, folderColor: folder.color, topPodcastUuids: [])
        DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
            self.setup(folderName: folder.name, folderColor: folder.color, topPodcastUuids: podcastUuids)
        }
    }
```

SergioEstevao:
> But I think the placeholder can be easily accomplished with the code we already have, take this as an example:

I got to the same conclusion just needed to add some small adjustments to ensure all worked correctly, can you give it another look?

podcasts/FolderPreviewView.swift
```diff
@@ -16,18 +16,29 @@ class FolderPreviewView: UIView {
     private var gradientLayer: CAGradientLayer?
     private var nameLabel: UILabel?
     private var nameLabelBottomConstraint: NSLayoutConstraint?
+    private var currentFolderUuid: String?
 
     func populateFrom(folder: Folder) {
-        let podcastUuids = DataManager.sharedManager.allPodcastsInFolder(folder: folder).map(\.uuid)
-        setup(folderName: folder.name, folderColor: folder.color, topPodcastUuids: podcastUuids)
+        currentFolderUuid = folder.uuid
```
leandroalonso, line 27: This change will cause a regression for CarPlay. On `CarPlayImageHelper.imageForFolder` we use this to populate the folder preview and to snapshot it.

I believe we can create a new method that is used only on the podcast screen and leave this one as it is so other unintended parts are not affected.

leandroalonso (changes requested):

podcasts/FolderPreviewView.swift
SergioEstevao: Maybe I'm missing something here but if you see the `setImage` code

```
private func setImage(in imageView: PodcastImageView, for uuid: String) {
        if forCarPlay {
            // For CarPlay we just want to grab whatever we have in cache
            imageView.imageView?.image = ImageManager.sharedManager.cachedImageFor(podcastUuid: uuid, size: .list)
        } else {
            imageView.setPodcast(uuid: uuid, size: .list)
        }
    }
```

We already have a special case for carPlay so we shouldn't need to change anything else?

SergioEstevao: Ok, I got deeper in the code and found out what you mean, I will add some extra handling for it.

SergioEstevao:
@leandroalonso Updated the code as you suggested to keep the previous code for CarPlay. Tested it on the CarPlay simulator and it looks all good now.

Simulator Screenshot - iPhone SE (3rd generation) - 2024-05-07 at 18 10 10

podcasts/HomeGridListItem.swift
```diff
@@ -58,6 +58,8 @@ class HomeGridListItem: ListItem {
                 folder.name == otherFolder.name &&
                 folder.color == otherFolder.color &&
                 folder.syncModified == otherFolder.syncModified &&
+                folder.sortType == otherFolder.sortType &&
+                folder.sortOrder == otherFolder.sortOrder &&
```
leandroalonso, line 62: Why do we need this addition?

leandroalonso (approved):
✅ Tested on CarPlay
✅ Tested with a lot of folders/podcasts
✅ Tested folder creation preview

podcasts/HomeGridListItem.swift
SergioEstevao: From my understanding this is for the DifferenceKit algorithm that we are using to distinguish updates on the collection view to work correctly when there are changes on the folder sort order.

PR #1711, opened 2024-05-06
Fix crash audio high filter

SergioEstevao:
Fixes [this crash](https://a8c.sentry.io/issues/4238066024/?project=4503921846583296&query=is:unresolved+issue.priority:%5Bhigh,+medium%5D+release.build:%227.63*%22&statsPeriod=7d&stream_index=0)

This PR replaces the use of `unsafeBitCast` to use Unmanaged as recommended in the documentation [here](https://developer.apple.com/documentation/swift/unsafebitcast(_:to:))

## To test

- Start the app
- Start playing a podcast with Volume Boost enabled and Trim Silence Disabled
- Play around forward, backward, skipping forward and stopping on the player to see if all works correctly

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/47">7.64</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

leandroalonso (changes requested):
@SergioEstevao are we able to reproduce the original crash? It would be good to test in those conditions.

Also, even though your changes seem ok from the code point of view, if anything goes wrong for any reason, it will affect playback which is the core of the app.

We should be extra cautious and put this change behind a remote feature flag.

SergioEstevao:
@leandroalonso While I try it in some different ways  I couldn't reproduce it so I use the Sentry report as reference to find what could be wrong.
I agree with you I will add a remote FF just to be on the safe side.

podcasts/DefaultPlayer.swift
```diff
@@ -396,7 +396,8 @@ class DefaultPlayer: PlaybackProtocol, Hashable {
         let peakLimiterRenderCallback: AURenderCallback = { inRefCon, _, _, _, inNumberFrames, ioData -> OSStatus in
             if ioData == nil { return -1 }
 
-            let referenceToSelf = unsafeBitCast(inRefCon, to: DefaultPlayer.self)
+            let reference = Unmanaged<DefaultPlayer>.fromOpaque(inRefCon)
+            let referenceToSelf = reference.takeUnretainedValue()
```
bjtitus: Since the crash is a `EXC_BAD_ACCESS`, would it make any sense to try `takeRetainedValue` here and `passRetained` in the other calls?

In [this thread](https://forums.swift.org/t/using-unmanaged-and-unsafebitcast-to-interface-with-c/30864), Quinn ways:

> If you can guarantee that the instance of Button outlasts any possibility that widget might call your handler, it’s OK to use an unretained reference. If not, it’s best to use a retained reference (although you’ll then need a way to release that reference when you’re done).

I just wonder if the crash indicates that the player may be going away and taking a retained version could help. Maybe it's best to save that for a future PR if this crash isn't resolved?

SergioEstevao: My fear here is that we can get a retain cycle because we create a new DefaultPlayer every time we load a different episode.
I also realized that on other methods of this class we are already following this approach for example check the `DefaultPlayer::tapProcess` method first line:
```
Unmanaged<DefaultPlayer>.fromOpaque(MTAudioProcessingTapGetStorage(tap)).takeUnretainedValue()
```

bjtitus: > I also realized that on other methods of this class we are already following this approach

Ah, this is good to know! I agree with the retain cycle concern, although I would think that would be easy for us to check. If these other spots aren't showing up in crash logs it's probably best to just copy them for now.

leandroalonso:
@SergioEstevao What do you think about having this on `7.64`?

We have A LOT going on now with the performance improvements, fixing this means we need to keep track of this issue to make sure it's fixed or not and to ensure there are no side-effects on top of everything else.

SergioEstevao:
@leandroalonso All good for me, I just decided to tackle it because there were some user reports of crashes when playing. We can keep an eye on 7.63 and see if there are more reports of this and apply it if necessary.

SergioEstevao:
@bjtitus do you mind to give it another look?

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -42,6 +42,8 @@ public enum FeatureFlag: String, CaseIterable {
 
     case cachePlayingEpisode
 
+    case playerFiltersFix
```
leandroalonso: I'd change that to `defaultPlayerFilterCallbackFix`. Adding a comment linking to the crash would also be good so it gives more context to someone looking at removing/changing that.

leandroalonso (approved):

leandroalonso:
@SergioEstevao [it seems that this didn't fix the issue](https://a8c.sentry.io/issues/4238066024/events/b0a3b0377923422a8f1d279dafb3cc55/) :(

SergioEstevao:
@leandroalonso I think this was only merged on 7.64 release and the sentry link you shared looks like it's for version 7.63.1

leandroalonso:
@SergioEstevao 

Screenshot 2024-05-22 at 17 57 21

[You can also check all versions here](https://a8c.sentry.io/issues/4238066024/tags/release/?referrer=tag-distribution-meter).

PR #1715, opened 2024-05-06
Performance: `VACUUM` the database

leandroalonso:
For old, big databases, `VACUUM` can reduce the database's size and increase performance absurdly.

This is an initial attempt to add it. In theory, we should be smart when running it:

1. Check disk space
2. Run in moments when the user is not actively using the app

However, I think the ideal for now would be to test that ASAP as it's probably the biggest change we can make in terms of performance.

## To test

1. Run another branch with a huge database with complex filters (the one we have with 500mb+ is ideal)
2. Run this branch
3. ✅ You should see a message when opening the app with a spinner
4. It will take a while to disappear (it depends on your hardware)
5. ✅ Once it's gone, interact with the app, filters, playback and everything should be snappier

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
@@ -82,6 +82,10 @@ public class DataManager {
 
             }
         }
+
+        dbQueue.inDatabase { db in
+            try? db.executeUpdate("VACUUM;", values: nil)
+        }
     }
 
     // MARK: - Up Next
--- a/podcasts/MainTabBarController+Alert.swift
+++ b/podcasts/MainTabBarController+Alert.swift
@@ -1,16 +1,17 @@
 extension MainTabBarController {
     func presentLoader() {
         DispatchQueue.main.async { [weak self] in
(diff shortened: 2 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

SergioEstevao:
I tested this with two large databases and it worked correctly. I also tested on an older device (iPhone Xs) and it worked well without taking that much time to clean up.
My final test was killing the app in the middle of the process and when restarted all worked correctly.

SergioEstevao (approved):
Working correctly and on large database the performance looked greatly improved.

bjtitus (commented):
Ran against 500MB database and saw a 50% size decrease. Spend about 2 minutes and 15 seconds performing the vacuum on my iPhone X.

leandroalonso:
@bjtitus performance is not great for huge and old databases.

For the long-term, we need to think about a better solution, such as auto_vacuum. But instead of waiting that long and the app being unusable I think we should at least try this solution.

PR #1743, opened 2024-05-13
Re-add: Use content type to set extension for downloaded files

SergioEstevao:
Re-add [fix to use content type to set file extension for audio files](https://github.com/Automattic/pocket-casts-ios/pull/1577)

I needed to re-add the columns on episode tables for this fix to work correctly.

I also updated the code use an existing MetadataTask class to update the contentType instead of doing it on the end of the upload.

In order to make sure this change didn't affect performance in queries because of Episode table changes I run some performance measurement using the Filters tab and measuring the method: `episodeCount(forFilter: EpisodeFilter, episodeUuidToAdd: String?)`

Here are the results I got [before and after these changes](https://docs.google.com/spreadsheets/d/1n16l_q1j-DKrk_kBxJIrbJFUYX1YRHv3vhHq8lV60kU/edit?usp=sharing) on a iPhone Xs

| Before | After |
| - | - |
|  0.3064338418  |   0.3009749147 | 

## Test if the fix worked

- Search for the following podcast: `ondacer` on the discovery section
- Subscribe to the podcast
- Open the podcast and download one of the episodes
- Tap play
- The episode should play correctly
- Now play an episode without downloading it first and see if it works correctly
- Smoke test other podcasts, to see if this change doesn't brake anything on that front

## Test DB migrations because of fields `contentType` being added again

- Install a 7.62 build then migrate to this version 7.66
- Install a 7.61 build then migrate to this version 7.66
- Install a 7.64 build then migrate to this version 7.66

## Test the performance of the app because of database changes

- Setup the app to use a large database by cherry picking the last 4 commits in this branch: `hang/copy_database`
- Use the large user database with 500mb
- Use a real device
- Start the app
- Go to filters view
- Scroll around

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/47">7.64</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

leandroalonso (changes requested):
@SergioEstevao I'm not sure we should re-add columns to the `SJEpisode` table.

If we want to follow this path, I think we need to answer if the query time average changes for big databases by adding those columns. We should probably do some sort of average time or something like that.

Alternatively, we could store this somewhere.

bjtitus:
I think this is a separate item from this PR but should we switch from `FileTypeUtil` to [`UTType(mimeType:...)`](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype) for parsing the `Content-Type` header and [`UTType.preferredFilenameExtension`](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype) to fetch the extension at some point?

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
```diff
@@ -717,11 +717,29 @@ class DatabaseHelper {
         // Those migrations were some heavy DROP COLUMN that we moved outside of DB startup
         if schemaVersion < 48 {
             schemaVersion = 48
+            do {
+                try db.executeUpdate("ALTER TABLE SJEpisode DROP COLUMN contentType", values: nil)
```
SergioEstevao: I'm not sure if we should put this back here but my rational is:
- Some users maybe be on versions where where the  clean up process is not run, so they will have the new columns, so we need to drop it and then create it again on version 50.
- The reason for the above is that cleanUp code is run after this update code.

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
```diff
@@ -849,7 +843,12 @@ public class DataManager {
     }
 
     public func episodeCount(forFilter: EpisodeFilter, episodeUuidToAdd: String?) -> Int {
-        filterManager.episodeCount(forFilter: forFilter, episodeUuidToAdd: episodeUuidToAdd, dbQueue: dbQueue)
+        var count = 0
```
SergioEstevao: This should be reverted after the tests.

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
leandroalonso: I think what we might be able to do is to just remove the `DROP TABLE` from everywhere.

Then on `50` you can do:

```
try? db.executeUpdate("ALTER TABLE SJEpisode ADD COLUMN contentType TEXT;", values: nil)
try? db.executeUpdate("ALTER TABLE SJUserEpisode ADD COLUMN contentType TEXT;", values: nil)
```

If the column exists, this will fail but we're using `try?`. If not, it will be created. This way we avoid the unnecessary work of removing and re-adding for those users.

I have to test this though.

SergioEstevao: @leandroalonso good suggestion, just changed the code to use try?.

leandroalonso (approved):
Tested on the following scenarios:

✅ Version without the column. Column is created and the downloaded file plays correctly;
✅ Version with the column. Column was still there and the downloaded file plays correctly
✅ Pretty old version without the column. Column is created and the downloaded file plays correctly;

PR #1749, opened 2024-05-14
Automation refactors to better support Enterprise builds

mokagio:
I'm looking at setting up _prototype builds_, which we'll use for internal testing and distribute via Automattic's Enterprise account. Internal ref pdnsEh-1D4-p2.

This PR does some refactors that aims to make it easier to add support for this upcoming new build.

Highlights:

- Use `api_key` instead of `api_key_path`.
- Extract extension names so we can easily define the bundle identifiers for the prototype build version once ready.
- Rename some constants and lanes to include "production" do distinguish them from the prototype build version ([staging](https://github.com/Automattic/pocket-casts-ios/blob/8db56037cc66747be8088ebdd453d2350b2bcc34/config/PocketCasts.staging.xcconfig)?) in the future

## To test

If CI is green, then the refactor is correct.

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary. — N.A.
- [x] I have considered adding unit tests for my changes. — N.A.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics. — N.A.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -20,29 +23,46 @@ EXTRACTED_RELEASE_NOTES_PATH = File.join(RESOURCES_FOLDER, 'release_notes.txt')
...
+  storage_mode: 's3',
+  s3_bucket: 'a8c-fastlane-match',
+  s3_region: 'us-east-2'
+}.freeze
+
+ASC_API_KEY_PRODUCTION = app_store_connect_api_key(
+  key_id: get_required_env("APP_STORE_CONNECT_API_#{TEAM_ID_PRODUCTION}_KEY_ID"),
+  issuer_id: get_required_env("APP_STORE_CONNECT_API_#{TEAM_ID_PRODUCTION}_ISSUER_ID"),
+  key_content: get_required_env("APP_STORE_CONNECT_API_#{TEAM_ID_PRODUCTION}_KEY_CONTENT")
+)
```
mokagio: Works 👌 

image

fastlane/Fastfile
```diff
@@ -941,14 +957,14 @@ inal copy, and try again.")
...
     )
 
     # Add all devices to the provisioning profiles
     add_all_devices_to_provisioning_profiles(
-      team_id: TEAM_ID,
-      app_identifier: MAIN_BUNDLE_IDENTIFIERS
+      team_id: TEAM_ID_PRODUCTION,
+      app_identifier: PRODUCTION_BUNDLE_IDENTIFIERS
     )
   end
```
mokagio, line 963: These last calls come from `register_new_device`, a lane that adds devices to the ASC account. We've long since improved this workflow via internal tooling. @Automattic/pocket-casts-ios is it okay to remove `register_new_device`?

fastlane/Fastfile
AliSoftware: Two notes:

 - One main drawback of having this as a global constant is that this will be evaluated no matter what lane is invoked—even if this is a lane that doesn't need any interaction with the ASC API, including running `fastlane lanes` to list available lanes–and would thus not only be not needed, but would also fail if one of the `get_required_env` vars is not defined locally 💥 
 - Is there any benefit of having our `.mobile-secrets/…/env` (and the `.env-default` file devs use locally) to use those compound env var names for those 3 values, and then having to pass the values explicitly… as opposed to directly use the ENV var names associated with the `ConfigItems` of `app_store_connect_api_key`—so that we don't have to pass any explicit parameter at all and this works out of the box?

---

My suggestion:

 - Update `~/.mobile-secrets/CI/secrets/pocket-casts-ios/env` to `export APP_STORE_CONNECT_API_KEY_KEY_ID="${APP_STORE_CONNECT_API_PZYM8XX95Q_KEY_ID}"` etc, so that those values (defined in `shared-secrets.yaml` are exposed as env vars matching the one from `fastlane action app_store_connect_api_key`.
 - In this `Fastfile`, remove the declaration of this global constant, and just use `api_key: app_store_connect_api_key` in the call sites instead of `api_key: ASC_API_KEY_PRODUCTION`. Given that all the parameters needed by `app_store_connect_api_key` would be provided by the env vars from `./mobile-secrets`, there is no need to provide them explicitly anymore, making its call site simpler.
 - Alternatively, instead of using `api_key: app_store_connect_api_key` in the parameters of the call sites, you can also just call `app_store_connect_api_key` first, and then the action you need without specifying any `api_key:` parameter, as those action propagate their values to one another via Lane Contexts (see https://docs.fastlane.tools/app-store-connect-api/#use-the-shared-value-in-lane-context)

fastlane/Fastfile
AliSoftware: This obsolete lane also has the big drawback that any device registered using it would not be claimed and associated with the Matticspace user the device belongs to (and thus would be removed like any other unclaimed device when we do our yearly cleanup too). So 💯  to remove that lane.

AliSoftware (changes requested):
Overall looks good, except the constant at the global scope having side effects (potentially making the `Fastfile` crash if the env vars are not defined locally… even when running a lane that don't need to interact with the ASC API and won't need that API key at all)

fastlane/Fastfile
mokagio: Thanks @AliSoftware 

> One main drawback of having this as a global constant is that this will be evaluated no matter what lane is invoked

Believe it or not, when I woke up this morning, this was one of the first things that came to mind. _Don't know whether that's a good or a bad thing_ 🤣 

Definitely best to compute it on demand.

> Is there any benefit of having our .mobile-secrets/…/env (and the .env-default file devs use locally) to use those compound env var names for those 3 values, and then having to pass the values explicitly… as opposed to directly use the ENV var names associated with the ConfigItems of app_store_connect_api_key—so that we don't have to pass any explicit parameter at all and this works out of the box?

Clearly, I haven't worked with Enterprise distribution in a while... I was under the impression we would have needed two API key, one for the App Store and one for the Enterprise account. But that's not the case as looking at the automation in other projects show.

As a matter of fact, I couldn't even find an option to create keys in the Enterprise account... and I'm the account holder

image
<sub>Enterprise</sub>

image
<sub>App Store</sub></br></br>

With that in mind, I agree there is no value in having the key namespaced at this level.

Maybe, just maybe, one could argue it might be useful for A12s infrastructure engineers who might need to store keys for multiple App Store accounts. But most of our projects use the local `.env-...` file to set env vars which makes it possible to keep the standard name at this level. We might want to re-evaluate if/when we'll drop the need for those env files polluting a users' home folder.

I'll update. Thanks again.

---

> you can also just call app_store_connect_api_key first, and then the action you need without specifying any api_key: parameter, as those action propagate their values to one another via Lane Contexts

TIL. Nice.

fastlane/Fastfile
mokagio: Addressed in https://github.com/Automattic/pocket-casts-ios/pull/1758

fastlane/Fastfile
mokagio: Addressed in [33b8162](https://github.com/Automattic/pocket-casts-ios/pull/1749/commits/33b81628e501c5e92b875608f0e205f22e2133f9)

AliSoftware (approved):
I didn't test to confirm but I'm assuming you updated the env vars in `.mobile-secrets` too; if so, :shipit:

mokagio:
> I didn't test to confirm but I'm assuming you updated the env vars in .mobile-secrets too; if so, :shipit:

Yes I did. But it doesn't hurt to test in the real world. I should have thought of that.

See https://buildkite.com/automattic/pocket-casts-ios/builds/6900#018f7b31-f65d-437c-91bf-b214ed8101c3

mokagio:
Breadcrumb for future reference. I just realized that CI is fine because it has the env vars, but a developer machine might not.

I'll followup tomorrow with a PR that sets up the `.env` pattern to account for local builds.

PR #1761, opened 2024-05-16
Load `.env` file in `Fastlane` or fail if not available

mokagio:
See env vars discussion with @AliSoftware in https://github.com/Automattic/pocket-casts-ios/pull/1749

I have mixed feelings about this. Part of me wishes to move away from `.env` files to avoid having the home folder filled with them. However, I understand that's a problem only for infra folks and that a `.env` file is easier to set up than requiring every dev who wants to use advanced lanes to have a shell setup that loads environment variables.

Also, that is the current standard used in other apps, and I think it's safe to say we have bigger fishes to fry at the moment than changing how we provision local environments.

## To test

- Check whether you have a `.pocketcasts-ios.env-default` file in your home folder. If you do, `mv` to a backup location. E.g. `mv ~/.pocketcastsios-env.default{,.bkp}`
- Run any lane, e.g. `bundle exe fastlane test`—It should fail

image
_(The screenshot says `cd` but I fixed the code to use the correct `cp`)_

- `cp fastlane/user.env-example /Users/gio/.pocketcastsios-env.default`
- Run the lane again, it should pass (or at least, start)

image

Notice that the template env file has no values, so a lane that requires them will fail. 

image

**Keen to hear your thoughts on whether we should proactively check for required env values for a better UX upon failures, or just rely on folks to add the values they need when they do.**

If you want to restore your original env file: `rm ~/.pocketcastsios-env.default && mv ~/.pocketcastsios-env.default{.bkp,}`

## Checklist


AliSoftware:
> I have mixed feelings about this. Part of me wishes to move away from .env files to avoid having the home folder filled with them.

Same. Short from having the bandwidth and priority to make a complete overhaul of how we handle this, at least I've been thinking about:
 - Starting to move those in `~/.a8c-apps` instead of putting them directly into `~` (similar to how we've started to do the same for `~/.configure/*` files to stop polluting the repo or the `$HOME`
 - Starting to name those properly, especially using just the `.env` file extension instead of this odd `.env-default` extension which doesn't make sense for those files (and prevent IDEs like VSCode to detect the file type properly automatically to provide syntax highlighting when editing those…)

Maybe we can start doing those changes for _new_ projects we start configuring those `.env` files for, even if we don't change the other projects at the same time just yet (… and over time when the occasion arise for each project, we'll slowly start to migrate other projects one by one to use this new convention)?

fastlane/Fastfile
```diff
@@ -134,6 +135,19 @@ before_all do
...
   configure_apply
+
+  unless is_ci || File.file?(USER_ENV_FILE_PATH)
+    example_path = 'fastlane/user.env-example'
+    UI.user_error! <<~MSG
+      #{ENV_FILE_NAME} not found!
+
+      Please copy #{example_path} to #{USER_ENV_FILE_PATH} and fill in the values for the automation you require.
+
+      cd #{example_path} #{USER_ENV_FILE_PATH}
```
AliSoftware:
```suggestion
      cp #{example_path} #{USER_ENV_FILE_PATH}
```

AliSoftware: I'm wondering if warning about this systematically (i.e. in the `before_all`, run for every lane even the ones that might not need any of those env vars) is the right approach after all (I know that's what we do for other apps, but maybe this PR is a good occasion to start introducing a better approach?).

What do you think about making that error only printed when a `get_required_env` fails (and `is_ci` is `false`), so that it's only printed _if_ a required env var is actually missing, but not printed nor required if the lane being called ends up not needing any secret env value at all?

---

An alternative approach that I've started to use in Tumblr is that in addition to `get_required_env` helper, we [also have a `require_env_vars ` helper](https://github.tumblr.net/Tumblr/ios/blob/develop/fastlane/lib/helpers.rb#L18-L20) which can take multiple env var names, and whose role is only to check, [at the very start of each lane](https://github.tumblr.net/Tumblr/ios/blob/da7f978a78a8f755eb1cbaa151cfb2f1072e9cde/fastlane/Fastfile#L95), if all those required env vars that the lane will be needed are defined.

The nice aspect of that approach is that it's immediately clear when you look at each lane in the `Fastfile` which env var this lane will be required to be defined, instead of skimming the code of the lane for all the `get_required_env` calls that it contains… which would not be enough anyway, because that lane might require that env var _directly_ but might call helper functions or other lanes which _do_ require additional env vars, or might be calling actions without explicitly passing some env vars (e.g. not explicitly passing the `token: get_required_env('GITHUB_TOKEN')` parameter to an action but instead relying on the fact that `GITHUB_TOKEN` is already the `env_name` of the corresponding `ConfigItem` in that action and that it will use it automatically).

If you go with that approach of adding `require_env_vars` at the top of your lanes to make it more clear what's required for each, then it would then make sense to implement that test (and printing the instructions for the user to create the `USER_ENV_FILE_PATH` file) in the implementation of `require_env_vars` itself (and only if at least one of the env var was not defined), instead of in `before_all`?

fastlane/Fastfile
```diff
@@ -23,6 +21,9 @@ EXTRACTED_RELEASE_NOTES_PATH = File.join(RESOURCES_FOLDER, 'release_notes.txt')
 SENTRY_ORG_SLUG = 'a8c'
 SENTRY_PROJECT_SLUG = 'pocket-casts-ios'
 
+ENV_FILE_NAME = '.pocketcastsios-env.default'
+USER_ENV_FILE_PATH = File.join(Dir.home, ENV_FILE_NAME)
```
AliSoftware: As suggested in my comment, if we want to start slowly at least improving our conventions one project at a time around that, we might as well start with PCiOS leading the way (even if other apps will only follow only way later).

I'm also suggesting to use the name of the git repo as the basename of the env (hence adding the `-` dashes in addition to fixing the file extension)

```suggestion
ENV_FILE_NAME = '.pocket-casts-ios.env'
USER_ENV_FILE_PATH = File.join(Dir.home, '.a8c-apps-env', ENV_FILE_NAME)
```
Feel free to choose something other than `.a8c-apps-env` if you're more inspired.

AliSoftware (commented):
Eh, sorry, seems that when I posted my review comments, GitHub made some of them as "immediate comments" and others be "pending comments as part of a review" 🤷 

Posting the ones that were still pending now, even though the feedback they're about was already mentioned in the comments that were posted "immediately" (instead of as part of review)

mokagio:
Thanks for your great suggestions @AliSoftware !

I implemented them as follow:

- Moved the `.env` file to `~/.a8c-apps/pocket-casts-ios.env`
- Redefined the `get_required_env` method to print the `.env` file suggestion and include `!` to broadcast it will crash if not found
- Added `require_env_vars!` and added it in the lanes that I thought needed it

A nice win from this work is that it revealed the example `.env` was missing two values.

Fun fact: I renamed the example `.env` to `example.env`. Then, looking at the Tumblr implementation, I saw you [made a similar change](https://github.tumblr.net/Tumblr/ios/commit/5e3d420d77db42307bbf47ea38c449597e342556) 🧠 👈 😄

fastlane/Fastfile
```diff
@@ -134,6 +138,8 @@ before_all do
   # but it has such a negligible overhead that it's worth running it here to
   # keep the individual lanes cleaner.
   configure_apply
+
+  Dotenv.load(USER_ENV_FILE_PATH)
```
mokagio, line 141: I noticed Tumblr offers overload options, which is nice. But I decided to keep it simple (or, rather, less flexible?) here because there are already various changes to discuss. Happy to iterate in followups to define a standard.

fastlane/Fastfile
```diff
@@ -150,6 +156,12 @@ platform :ios do
 
   desc 'Download and configure code signing certificates and provisioning profiles for the App Store build.'
   lane :configure_code_signing_app_store do |options|
+    require_env_vars!(
+      'APP_STORE_CONNECT_API_KEY_KEY_ID',
+      'APP_STORE_CONNECT_API_KEY_ISSUER_ID',
+      'APP_STORE_CONNECT_API_KEY_KEY'
+    )
```
mokagio, line 162: I may extract these in a constant when the time comes to add a lane for the enterprise code signing. Or, move the check in an helper method that will DRY the `sync_code_signing` call.

fastlane/lib/helpers.rb
```diff
@@ -0,0 +1,35 @@
...
+# Use this to ensure all env vars a lane requires are set.
+#
+# The best place to call this is at the start of a lane, to fail early.
+def require_env_vars!(*keys)
+  keys.each { |key| get_required_env!(key) }
+end
+
+# Use this instead of getting values from `ENV` directly. It will throw an error if the requested value is missing.
+def get_required_env!(key, env_file_path: USER_ENV_FILE_PATH)
+  unless ENV.key?(key)
```
mokagio: 🤔 

Maybe this would look nicer as

```ruby
return ENV.fetch(key) if ENV.key?(key)

message = ...
```

Less indentation.

fastlane/Fastfile
```diff
@@ -125,6 +126,9 @@ GLOTPRESS_TO_ASC_METADATA_LOCALE_CODES = {
   'zh-tw' => 'zh-Hant'
 }.freeze
 
+require_relative 'lib/helpers'
+import 'lib/helpers.rb'
```
AliSoftware: I think using both `require_relative` and `import` is redundant?

Though I always forget about the difference between the two (I know `require_relative` is from Ruby core, while iinm `import` is from _fastlane_, and I think each has it's own pros/cons, like `import` knowing about `lanes` and being able to call `UI` while already knowing we're within the `Fastlane::` module, but otoh not providing whatever feature that `require_relative` provides I can't remember which…)

I also know we have a mix of `require_relative` and `import` in e.g. the Tumblr Fastfile and that I've meant to better understand the difference between the two and clean up the mess to make it more consistent at some point (but never got to it)…

fastlane/lib/helpers.rb
AliSoftware: Nice catch; I'm actually surprised that Rubocop doesn't suggest you that change already 😛

AliSoftware (approved):
Apart from the open question around `require_relative` vs `import`, LGTM

SergioEstevao (approved):
:shipit:

fastlane/Fastfile
mokagio: It's been some days, so I don't remember exactly why I used this syntax. I think I copied it from Tumblr after getting a failure when using only `require_relative`.

I tested it locally again today. 

#### `require_relative` only

Fails because it cannot access the `USER_ENV_FILE_PATH` constant, which is defined in `Fastfile`.

```
[11:29:46]: Applied configuration
+---------------------------------------------------------+
|                      Lane Context                       |
+------------------+--------------------------------------+
| DEFAULT_PLATFORM | ios                                  |
| PLATFORM_NAME    | ios                                  |
| LANE_NAME        | ios configure_code_signing_app_store |
+------------------+--------------------------------------+
[11:29:46]: fastlane finished with errors

Looking for related GitHub issues on fastlane/fastlane...

Found no similar issues. To create a new issue, please visit:
https://github.com/fastlane/fastlane/issues/new
Run `fastlane env` to append the fastlane environment to your issue
bundler: failed to load command: fastlane (/Users/gio/Developer/a8c/pcios/vendor/bundle/ruby/3.2.0/bin/fastlane)
/Users/gio/Developer/a8c/pcios/fastlane/lib/helpers.rb:11:in `get_required_env!': [!] uninitialized constant USER_ENV_FILE_PATH (NameError)

def get_required_env!(key, env_file_path: USER_ENV_FILE_PATH)
                                          ^^^^^^^^^^^^^^^^^^
	from /Users/gio/Developer/a8c/pcios/fastlane/lib/helpers.rb:7:in `block in require_env_vars!'
```

#### `import` only

Works.

#### `required_relative` and `import`

Works. _We knew that already_.

So, looks like there's more to learn, at least for me. To be fair, I don't like how my implementation accesses constants that are defined in a different file without importing them. But I cut some corners in the interest of getting this done and having methods that are not too cumbersome to call.

I suppose there's a design we could achieve that moves all the constant at the helper level, with an additional method to call `DotEnv.load`. But... I like having the files in `fastlane/lib` be generic enough to be copy pasted to other process. But... that's a bit of a smell too. If we find ourselves copy pasting, maybe we should consider pushing these helper in `release-toolkit` or a new dedicated helper gem.

PR #1770, opened 2024-05-17
Add "Change Avatar" button to account settings

pinarol:
Fixes # gravatarp2 -> 2024/06/10/pt-integrating-gravatar-into-pocket-casts

Adding the ability to change the avatar. Avatar is coming from user's Gravatar account. So the "Change Avatar" button opens the Gravatar web link via in-app Safari.

## To test

Profile > Account > "Change Avatar"
Observe: In app Safari is presented
   - Inside the browser, if you are not logged in you should see the login first, then you'll be redirected to avatar update page.
   - The avatar change does not always get reflected to the app immediately even though we force refresh the avatar on the mobile side skipping all client-side caches. This is a known limitation. There's a backend cache too and it's not updated immediately. The plan is to show such a message inside the browser:
   
 

If you are not logged in to the app then "Change Avatar" button will not be present for you.

Known issue:

A strange auto-zoom happens when testing in a Simulator but this issue is not reproduced in real device. But let me know if you can reproduce this in a real device.

## Checklist


CookieyedCodes:
Wouldn't it make more sence to just tap the icon?

pinarol:
> Wouldn't it make more sence to just tap the icon?

I think this is an experimental feature that's why, it's not a fully native experience to begin with. We just open a browser to change the avatar. But it's a fair question. cc @Luchadores  and @david-gonzalez-a8c if you want to add anything.

Luchadores:
I think we can try it, not sure if there's any other trigger that Pocket Casts already has on the Avatar itself. If not, we can do it, and see if users tap more directly the Avatar itself.

pinarol:
We did some discussions and for now we'll move fwd with the "Change Avatar" button.

podcasts/AccountViewController+TableView.swift
```diff
@@ -172,6 +182,12 @@ extension AccountViewController: UITableViewDataSource, UITableViewDelegate {
         case .supporterContributions:
             let supporterVC = SupporterContributionsViewController()
             navigationController?.pushViewController(supporterVC, animated: true)
+        case .changeAvatar:
+            guard let email = headerViewModel.profile.email,
+                  let safariViewController = GravatarSafariViewController(destination: .avatarUpdate(email: email)) else { return }
+            safariViewController.modalPresentationStyle = .automatic
+            present(safariViewController, animated: true)
+            Analytics.track(.accountDetailsChangeAvatar)
```
leandroalonso, line 190: We need to update [this spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?gid=0#gid=0) when adding new events. I went ahead and created the entry for this one.

leandroalonso (changes requested):
@pinarol from the code side, all looks good. The only thing missing is updating the `CHANGELOG`.

I have a few questions/suggestions though. I understand that many of them are out of the scope of this task — and it involves changes in web/design. My main goal is just to make the experience as good as possible:

1. The email field has autocomplete and autocorrect on. I _think_ we might want to disable those as it kinda keeps suggesting changes to email:



2. Do we need to enter the email? We already have this information, in theory, we could be directly redirected to the code phase, without requiring the user to type any info. This will likely reduce friction.

3. Not a big deal, but the icon feels a bit different from the rest to me. The border seems slightly thinner and the inside normally has a more subtle blue.




4. I felt an "avatar changed" confirmation is missing (either on the webpage or in the app). After cropping the image, there's no success message. I wonder if for any new user, this will confuse them as to whether they changed the avatar or not.

I don't think any of those comments are blockers but if we tackle them the experience will be smooth IMHO. Anyway, the only blocker is updating the `CHANGELOG`.

podcasts/AccountViewController+TableView.swift
pinarol: Thank you! 🙇

pinarol:
I'll fwd these feedbacks to @wellyshen @aaronfc 

> 1. The email field has autocomplete and autocorrect on. I think we might want to disable those as it kinda keeps suggesting changes to email:

I agree, it would be better if those are disabled.

> 4. I felt an "avatar changed" confirmation is missing (either on the webpage or in the app). After cropping the image, there's no success message. I wonder if for any new user, this will confuse them as to whether they changed the avatar or not.

I think the confirmation is there but its position isn't right so it's not easily viewable to user. Seems like a bug.

issues created: https://github.a8c.com/Automattic/gravatar/issues/108240 https://github.a8c.com/Automattic/gravatar/issues/108241

pinarol:
> 3. Do we need to enter the email? We already have this information, in theory, we could be directly redirected to the code phase, without requiring the user to type any info. This will likely reduce friction.

Unfortunately, we need the user to enter their email for now. In Gravatar, users can define multiple avatars for multiple emails under one account. But they can log in to Gravatar only with their primary email. If the user is using one of those secondary emails in Pocket Casts, then pre-filling the field would be misleading because the login attempt would fail. But there's an ongoing project to update the login flow to allow logins by secondary emails. After that the pre-filling can also be an option I think.

pinarol:
> Not a big deal, but the icon feels a bit different from the rest to me. The border seems slightly thinner and the inside normally has a more subtle blue.

cc: @Luchadores this is a feedback about the new change avatar icon.

pinarol:
Icon updated:

aaronfc:
> Unfortunately, we need the user to enter their email for now. In Gravatar, users can define multiple avatars for multiple emails under one account. But they can log in to Gravatar only with their primary email. If the user is using one of those secondary emails in Pocket Casts, then pre-filling the field would be misleading because the login attempt would fail. But there's an ongoing project to update the login flow to allow logins by secondary emails. After that the pre-filling can also be an option I think.

Hey, sorry to jump in. But this is not accurate. You can (and you should) pass the `?email=` query parameter to the URL. The login process already handles secondary emails (just released) (p8Slzc-2NM-p2), and as soon as Quick Editor project (p8Slzc-3i1-p2) is done it will also redirect to the correct avatars page.

The correct URL, I believe, should be `https://gravatar.com/profile?is_quick_editor=true&email=xxxx&scope=avatars` (once the Quick Editor project is done). cc/ @wellyshen @jcheringer

pinarol:
> Hey, sorry to jump in. But this is not accurate. You can (and you should) pass the ?email= query parameter to the URL. The login process already handles secondary emails (just released) (p8Slzc-2NM-p2), and as soon as Quick Editor project (p8Slzc-3i1-p2) is done it will also redirect to the correct avatars page.

> The correct URL, I believe, should be https://gravatar.com/profile?is_quick_editor=true&email=xxxx&scope=avatars (once the Quick Editor project is done)

Okay just to avoid causing our Pocket Casts friends misunderstand the situation. Logging in with a secondary email flow was deployed just **yesterday**.

And the URL that enables us to pass the email is not completely ready yet, although it seems close to being ready. So I'll wait for it instead of opening a follow up PR which was the original plan. I'll communicate internally about this new URL and update the PR.

leandroalonso:
Ok! Just ping me when it's ready for another review @pinarol!

pinarol:
I think this is ready for another look. @leandroalonso

pinarol:
I tried triggering a prototype build but it seems like there's an unrelated error there. It shouldn't be a blocker for the merge I think.

podcasts/GravatarSafariViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    enum Destination {
+        case avatarUpdate(email: String)
+
+        var url: URL? {
+            switch self {
+            case .avatarUpdate(let email):
+                guard var components = URLComponents(string: "https://gravatar.com/profile") else { return nil }
+                components.percentEncodedQueryItems = [
+                    URLQueryItem(name: "is_quick_editor", value: "true"),
```
leandroalonso, line 14: Would it be possible to open the Gravatar site in a given language? I tested the app in another language but it opens in English (on the browser it opens in the correct language though).

david-gonzalez-a8c:
@Luchadores I suggest we use [this icon](https://www.figma.com/design/fs6qvDgcorsbCMUNyhGpd3uI/%F0%9F%9B%A0-Pocket-Casts-Design-System?node-id=2-73&t=NdUHln6v4Z19oH1C-1) in our [library](https://www.figma.com/design/fs6qvDgcorsbCMUNyhGpd3uI/%F0%9F%9B%A0-Pocket-Casts-Design-System?node-id=0-1), it might suit this new row already.

Thank you ✨

Luchadores:
@pinarol, @etoledom  - Could we update the icon with the one David provided the link to?

pinarol:
Sure thing! @david-gonzalez-a8c Which one do you mean exactly? I am afraid I can't see it through that link. Can you maybe share its screenshot?

leandroalonso (changes requested):
@pinarol I was talking to @danielebogo and I wonder if it would be possible to feature flag these changes?

We've been using feature flags for all new features so we can control them remotely if needed.

Luchadores:
@pinarol -> I believe he is talking about this one:
image

@david-gonzalez-a8c let us know if you refer to a different one ☝️

podcasts/GravatarSafariViewController.swift
pinarol: Do you mean the site opens in a different language than Pocket Casts?

leandroalonso: Yup!

pinarol: I checked this issue and there indeed is an inconsistency. But it seems like the in-app SFSafariViewController is the one who is acting weird and this is happening for other links inside the app as well. I inspected the network a bit, somehow SFSafariViewController sends the wrong `Accept-Language`, it is like it gets stuck with one of the previous languages it was opened with.

For example:

My device language: Turkish
Pocket Casts language: German
System Safari language: English

But the in-app Safari shows in Spanish 🫨 

I checked the network and it indeed sends the wrong language:

Screenshot 2024-07-19 at 12 38 57

Spanish was something I tried previously once, now the in-app Safari got stuck there. 

Gravatar also respects the `Accept-Language` header that's why we can observe the same thing with it. But the problem here is that the header value is just not set right by SFSafariViewController. 🤷‍♀️ 

<video src="https://github.com/user-attachments/assets/5c30f361-a73e-44fc-a1bd-bf09e26e73a9"/>

pinarol:
> I wonder if it would be possible to feature flag these changes?

yeah I think we can do it.

pinarol:
I am applying the new icon. Does it look ok @david-gonzalez-a8c ?




sharing the developer view also, the size differences are related with inner insets of the icons:

Screenshot 2024-07-19 at 15 25 43

Luchadores:
Hey Pinar, before David replies, that looks nice, but we don't have the different hues in the icon like the rest. Sync with @etoledom, I believe he solved that last time. 

Attached is a screenshot of the previous icon we used, but you can see the difference as it plays on the blue opacity color.

image

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/55">7.69</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

hamorillo:
> Hey Pinar, before David replies, that looks nice, but we don't have the different hues in the icon like the rest. Sync with @etoledom, I believe he solved that last time.

That's because the [icon from the library](https://www.figma.com/design/fs6qvDgcorsbCMUNyhGpd3uI/%F0%9F%9B%A0-Pocket-Casts-Design-System?node-id=2-73&t=NdUHln6v4Z19oH1C-1) doesn't have different opacities as other icons have ([for example](https://www.figma.com/design/fs6qvDgcorsbCMUNyhGpd3uI/%F0%9F%9B%A0-Pocket-Casts-Design-System?node-id=127-8&t=SKZtmRXGR3Ayymvd-4)). I've just updated the icon in [the Android PR](https://github.com/Automattic/pocket-casts-android/pull/2263#discussion_r1684343131), but let me know if you should use another one.

pinarol:
@hamorillo is right. For that effect to happen the icon should have different opacities in it. @david-gonzalez-a8c could you check the icon and share an updated version if you think it doesn't look right https://github.com/Automattic/pocket-casts-ios/pull/1770#issuecomment-2239034912 ?

pinarol:
Apart from the icon, this is ready for another look. @leandroalonso @danielebogo

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -170,6 +175,8 @@ public enum FeatureFlag: String, CaseIterable {
             "default_player_filter_callback_fix"
         case .upNextOnTabBar:
             "up_next_on_tab_bar"
+        case .gravatarChangeAvatar:
+            "gravatar_change_avatar"
```
pinarol: Let me know if this is the right thing to do to make the FF controllable remotely.

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -139,6 +142,8 @@ public enum FeatureFlag: String, CaseIterable {
             false
         case .transcripts:
             false
+        case .gravatarChangeAvatar:
+            true
```
pinarol, line 146: I made it true by default, let me know if you think it needs to be false.

Luchadores:
👋 I recreated that specific icon with transparency. Attached is an SVG, and PNGs (1x, 2x, 3x). Hope it helps 🤞 
@pinarol, @hamorillo. @david-gonzalez-a8c, hope it's the right icon :D 

ProfileIcon
[ProfileIcon.zip](https://github.com/user-attachments/files/16312634/ProfileIcon.zip)

pinarol:
>  I recreated that specific icon with transparency.

Okay I've used them. This is how it looks now:

podcasts/GravatarSafariViewController.swift
pinarol: This is my hypothesis about what's happening: The in-app Safari is a separate process and its lifespan is controlled by the OS. We can't kill and restart it when we want. And it seems like the in-app Safari process needs to be restarted to adapt to the Safari language settings. If I do `Device > Restart` on the simulator then the in-app Safari works as expected and uses the same language with system Safari.

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
danielebogo: Nit: In theory if the value doesn't change, you don't need to specify the lower snake case as it will be automatically converted below by `rawValue.lowerSnakeCased()`

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
danielebogo: @bjtitus correct me if I'm wrong, but in theory we can switch off a FF if the default value is `true` right?

danielebogo (approved):
Hey @pinarol I tested the feature and it worked for me. Also the avatar changed pretty fast after the vc has been dismissed. I left one comment and a question to @bjtitus about the FF.

pinarol:
> Also the avatar changed pretty fast after the vc has been dismissed

Yes that's the case now since we introduced a kind of cache buster parameter with this commit. https://github.com/Automattic/pocket-casts-ios/pull/1770/commits/04d8c08ddab09fa184051a9fc4020267d076f049.

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
bjtitus: Yes, that's right. This value is just used to set the initial value so that it's enabled or disabled prior to connecting to Firebase. We also tend to leave this value `false` while testing until we are mostly sure we want to release in whichever version this is going into.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 7.70
 -----
-
+- A new button is added under Account Settings: "Change Avatar". It opens the in-app Safari with Gravatar's Avatar Quick Editor. [#1770]
```
bjtitus: We changed our formatting a bit to make this into a markdown link for easier copying to p2. Could you change this to:
```suggestion
- A new button is added under Account Settings: "Change Avatar". It opens the in-app Safari with Gravatar's Avatar Quick Editor. [#1770](https://github.com/Automattic/pocket-casts-ios/pull/1770)
```

pinarol:
I think we are ready to merge. 🎉 Thank you all for your help and feedback. 🙇

pinarol:
We need 👍 from @leandroalonso as well.

leandroalonso (approved):

PR #1778, opened 2024-05-22
Signing back in: fixes issues with outdated podcast information

### LINKED ISSUE #791: Podcasts being removed from folders

leandroalonso:
Fixes https://github.com/Automattic/pocket-casts-ios/issues/791#issuecomment-2125086525

We recently had an issue where a user had all their podcasts removed from folders when signing back in on another device.

This happened because they had outdated information and the app marked it as unsynced.

### With `onlyMarkPodcastsUnsyncedForNewUsers` enabled

1. Go to Profile > Settings > Beta Features > enable `onlyMarkPodcastsUnsyncedForNewUsers`
1. Sign in to a Plus account
2. Create a folder with some podcasts in it and refresh
3. Log out
4. Go to Settings > Developer > Clear all folder information
5. Subscribe to a few new podcasts
6. Log back into the same account
7. ✅ Under "Podcasts" the folder should be visible and with the same podcasts from before
8. Go to Profile > Refresh Now
9. ✅  Login into the web player, and ensure the newly subscribed podcasts were correctly synced

### With `onlyMarkPodcastsUnsyncedForNewUsers` disabled

1. Go to Profile > Settings > Beta Features > disable `onlyMarkPodcastsUnsyncedForNewUsers`
1. Add a breakpoint in `SyncSigninViewController.swift:337`
1. Sign in to a Plus account
2. Create a folder with some podcasts in it and refresh
3. Log out
4. Go to Settings > Developer > Clear all folder information
6. Log back into the same account
7. ✅ Breakpoint should hit
8. ✅ Under "Podcasts" the podcasts will be out of their folders

## Checklist

### DIFF
--- a/podcasts/AuthenticationHelper.swift
+++ b/podcasts/AuthenticationHelper.swift
@@ -66,7 +66,8 @@ class AuthenticationHelper {
 
         // we've signed in, set all our existing podcasts to
         // be non synced if the user never logged in before
-        if ServerSettings.lastSyncTime == nil {
+        if (FeatureFlag.onlyMarkPodcastsUnsyncedForNewUsers.enabled && ServerSettings.lastSyncTime == nil)
+            || !FeatureFlag.onlyMarkPodcastsUnsyncedForNewUsers.enabled {
             DataManager.sharedManager.markAllPodcastsUnsynced()
         }
 
--- a/podcasts/AuthenticationHelper.swift
+++ b/podcasts/AuthenticationHelper.swift
@@ -65,7 +65,9 @@ class AuthenticationHelper {
         ServerSettings.refreshToken = response.refreshToken
 
         // we've signed in, set all our existing podcasts to be non synced
-        DataManager.sharedManager.markAllPodcastsUnsynced()
+        if ServerSettings.lastSyncTime == nil {
+            DataManager.sharedManager.markAllPodcastsUnsynced()
+        }
 
         SyncManager.syncReason = .login
         ServerSettings.clearLastSyncTime()
--- a/podcasts/AuthenticationHelper.swift
+++ b/podcasts/AuthenticationHelper.swift
@@ -64,7 +64,8 @@ class AuthenticationHelper {
         ServerSettings.syncingV2Token = response.token
         ServerSettings.refreshToken = response.refreshToken
 
-        // we've signed in, set all our existing podcasts to be non synced
+        // we've signed in, set all our existing podcasts to
+        // be non synced if the user never logged in before
         if ServerSettings.lastSyncTime == nil {
             DataManager.sharedManager.markAllPodcastsUnsynced()
         }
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/50">7.65</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/AuthenticationHelper.swift
```diff
@@ -65,7 +65,9 @@ class AuthenticationHelper {
         ServerSettings.refreshToken = response.refreshToken
 
         // we've signed in, set all our existing podcasts to be non synced
-        DataManager.sharedManager.markAllPodcastsUnsynced()
+        if ServerSettings.lastSyncTime == nil {
+            DataManager.sharedManager.markAllPodcastsUnsynced()
```
leandroalonso, line 71: @geekygecko Before merging this I'd like to double-check with you as I might be missing any user case.

The app will only mark all podcasts as unsynced if the user has never logged in to any account in this device.

geekygecko: If they have just signed in, the last sync should be nil already. 🤔
If they signed out for six months and subscribed to many podcasts, then signed in again, won't that mean those new podcasts won't appear on their other devices?

SergioEstevao (approved):
Code looks good and tested the scenario described and all worked correctly. 

I do wonder why do we need to do step 4, Clear all folder information, the end user will not be able to do that, shouldn't this work correctly even without doing that?

podcasts/AuthenticationHelper.swift
leandroalonso: > won't that mean those new podcasts won't appear on their other devices?

Based on my tests it syncs just fine (as per the PR description)

leandroalonso:
> I do wonder why do we need to do step 4, Clear all folder information, the end user will not be able to do that, shouldn't this work correctly even without doing that?

Very good question, @SergioEstevao. I should have been more explicit in the PR description.

if you test these same steps in `trunk` (including using the "Clear all folder information"), your podcasts will be out of their folders when you sign back.

This is to simulate the issue that happened that is reported in  https://github.com/Automattic/pocket-casts-ios/issues/791#issuecomment-2125086525

A user won't be able to do that, but in the past, we had a bug that caused this scenario. A user with an old device with this "wrong data" can end up with their podcasts out of their folders, which is exactly what happened for this user.

leandroalonso:
@SergioEstevao could you please review again? I added a feature flag so we can toggle this behavior in case it's needed.

SergioEstevao (approved):
All working great, and FF doing it's job!

PR #1780, opened 2024-05-24
Download Issues: Retry Cancelled Tasks

bjtitus:
This is a series of changes to better handle tasks which have been cancelled by the system and set the proper download `episodeStatus` and `downloadTaskId` even if an episode was orphaned from its download task somehow.

This also adds the `downloadFixes` feature flag so we can disable if something goes horribly wrong.

| Before | After |
| -- | -- |
| <video src="https://github.com/Automattic/pocket-casts-ios/assets/3250/9087a9de-652b-4ad4-974a-922df0456dcd"> | <video src="https://github.com/Automattic/pocket-casts-ios/assets/3250/cc598900-155f-4858-8173-54eaf29892d6"> |

## To test

* Begin downloads for a podcast (I've been using "Blame it on the Fame" which only has 4 episodes)
* Notice the downloads start
* Quickly kill the app before they complete
* Reopen the app
* Ensure downloads continue and that download progress appears
* Ensure that episodes complete downloading

## Checklist


Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
```diff
@@ -643,16 +643,6 @@ class EpisodeDataManager {
         save(fields: fields, values: values, dbQueue: dbQueue)
     }
 
-    func saveEpisode(downloadStatus: DownloadStatus, sizeInBytes: Int64, episode: Episode, dbQueue: FMDatabaseQueue) {
```
bjtitus, line 646: This is no longer needed because we will always provide a `downloadTaskID`.

podcasts/DownloadManager.swift
```diff
@@ -459,4 +470,21 @@ class DownloadManager: NSObject, FilePathProtocol {
         downloadTask?.taskDescription = taskId
         downloadTask?.resume()
     }
+
+    func resumeAllBackgrounded() {
+        let queuedEpisodes = DataManager.sharedManager.findEpisodesWhere(customWhere: "episodeStatus == ?", arguments: [DownloadStatus.queued.rawValue])
```
bjtitus, line 492: All episodes still queued should be retried since they may have been cancelled previously. `addToQueue` will check if there's already a task associated with this episode.

podcasts/DownloadManager.swift
```diff
@@ -287,7 +298,7 @@ class DownloadManager: NSObject, FilePathProtocol {
         let useCellularSession = (mobileDataAllowed || (!NetworkUtils.shared.isConnectedToWifi() && autoDownloadStatus != .autoDownloaded)) // allow cellular downloads if not on WiFi and not auto downloaded, because it means the user said yes to a confirmation prompt
 
         #if os(watchOS)
-            let sessionToUse = WKApplication.shared().applicationState == .background ? cellularBackgroundSession : cellularForegroundSession
+            let sessionToUse = await WKApplication.shared().applicationState == .background ? cellularBackgroundSession : cellularForegroundSession
```
bjtitus, line 295: Making this method async made this require await.

podcasts/DownloadManager+URLSessionDelegate.swift
```diff
@@ -36,19 +36,24 @@ extension DownloadManager: URLSessionDelegate, URLSessionDownloadDelegate {
             progressManager.updateProgressForEpisode(downloadingEpisode.uuid, totalBytesWritten: totalBytesWritten, totalBytesExpected: totalBytesExpectedToWrite)
         }
 
-        if !downloadingEpisode.downloading(), downloadingEpisode.downloadTaskId != nil {
+        // If our download status or downloadTaskId are incorrect, then we should update these
+        if !downloadingEpisode.downloading() || downloadingEpisode.downloadTaskId == nil {
```
bjtitus, line 40: I was finding that `downloadTaskId` was missing from some downloads, causing the download progress to fail. This didn't seem to be new with the changes so I think it makes sense to just set this value when we update the status to `downloading`.

podcasts/DownloadManager+URLSessionDelegate.swift
```diff
@@ -63,8 +68,16 @@ extension DownloadManager: URLSessionDelegate, URLSessionDownloadDelegate {
         switch error.code {
         case NSURLErrorCancelled:
             if !episode.downloadFailed() {
-                // already handled this error, since we failed the download ourselves
+                // we already handled this error, since we failed the download ourselves
+                let reason = error.userInfo[NSURLErrorBackgroundTaskCancelledReasonKey] as? Int
+                switch reason {
+                case NSURLErrorCancelledReasonUserForceQuitApplication, NSURLErrorCancelledReasonInsufficientSystemResources:
+                    DataManager.sharedManager.saveEpisode(downloadStatus: .queued, downloadTaskId: nil, episode: episode)
```
bjtitus, line 75: When the system cancels a download task, we should reset it to `queued` so we don't fail to download it when the app restarts or a background refresh re-triggers downloads.

podcasts/DownloadManager.swift
```diff
@@ -258,7 +262,14 @@ class DownloadManager: NSObject, FilePathProtocol {
         }
     }
 
-    private func performAddToQueue(episode: BaseEpisode, url: String, previousDownloadFailed: Bool, fireNotification: Bool, autoDownloadStatus: AutoDownloadStatus) {
+    private func performAddToQueue(episode: BaseEpisode, url: String, previousDownloadFailed: Bool, fireNotification: Bool, autoDownloadStatus: AutoDownloadStatus) async {
+
+        if let task = await existingTask(for: episode) {
```
SergioEstevao: A bit of an edge case but could tasks exists for the same episode, but downloading different URLs because the episode media changed? The Creator update the audio or the link was changed.

podcasts/DownloadManager.swift
```diff
@@ -459,4 +470,21 @@ class DownloadManager: NSObject, FilePathProtocol {
...
+            }
+        }
+    }
+
+    private func existingTask(for episode: BaseEpisode) async -> URLSessionTask? {
+        let cellTasks = await cellularBackgroundSession.allTasks
+        let wifiTasks = await wifiOnlyBackgroundSession.allTasks
+
+        let cellTask = cellTasks.first(where: { $0.taskDescription == episode.downloadTaskId })
+        return cellTask ?? wifiTasks.first(where: { $0.taskDescription == episode.downloadTaskId })
```
SergioEstevao: Can we tasks for the same episode coexist on both sessions? or it's one or the other?

podcasts/DownloadManager.swift
bjtitus: I considered this and thought some future changes to cancel old episode downloads could handle this. What would you think about cancelling any existing tasks which have a different URL in this spot?

SergioEstevao: If we find tasks that don't match the current episode url then we can cancel them.

podcasts/DownloadManager.swift
bjtitus: The main case I think would be if download settings were changed and an episode download was added to a different session (user allowed cellular but we had wifi tasks queued). Similar to above, maybe we could cancel tasks which are still added to the other sessions but I feel like I should test that a bit further just to be sure.

For now, I'll change this to only check for existing tasks in the session where the download task is about to be added.

podcasts/DownloadManager.swift
bjtitus: Since I changed to only check the current session from [your other comment](https://github.com/Automattic/pocket-casts-ios/pull/1780#discussion_r1613205988), I decided only to cancel tasks found in that session instead of checking all of them. I'm planning a set of changes to cancel tasks on refresh which don't match current episode URLs so that should handle the rest. This is really just a fallback since it's probably more likely that the episode uuid changes.

podcasts/DownloadManager.swift
```diff
@@ -287,11 +292,25 @@ class DownloadManager: NSObject, FilePathProtocol {
...
         #else
             let sessionToUse = useCellularSession ? cellularBackgroundSession : wifiOnlyBackgroundSession
         #endif
 
+        if let task = await sessionToUse.existingTask(for: episode) {
+            if task.originalRequest?.url == request.url {
+                if task.error == nil {
+                    // As long as we don't have an error, we'll skip starting a new download, otherwise we'll need the new task anyway
+                    // Before this change, we allowed any new download so we'd rather start out more restrictive
+                    FileLog.shared.addMessage("Skipped task for episode: \(episode)")
```
leandroalonso: Maybe add a "Download:" or "[Download]" before so it's easier to search for that?

```suggestion
                    FileLog.shared.addMessage("Download: skipped task for episode: \(episode)")
```

leandroalonso (approved):
✅ Tested the described steps and it works as described
✅ Tested with the flag disabled and download still work

SergioEstevao (approved):
Working great! :shipit:

PR #1785, opened 2024-05-28
Download Issues: Cancel URLSession tasks when deleting unused podcasts

bjtitus:
We clear out unused podcasts after a refresh but do not touch the scheduled URLSession tasks. This could lead to out of date tasks which are no longer needed.

All functional changes were added in commit: https://github.com/Automattic/pocket-casts-ios/pull/1785/commits/9709b82c05227e6152f40d7370aa5745b60f4ca2
* Any time we call `deletePodcastIfUnused`, we also cancel the task associated with any episodes related to that podcast
* This is an async operation now (since fetching the tasks is async), which makes some operations, like the podcast fetching in `PodcastViewController` async. This should be fine because this view controller already performs updating on the podcast asynchronously.

https://github.com/Automattic/pocket-casts-ios/pull/1785/commits/7a9445d1b9bebebcd90dfe5635274f9f2775a3e3 prepares `PodcastManager` and `DownloadManager` for unit testing by allowing a `DataManager` instance to be specified.

https://github.com/Automattic/pocket-casts-ios/pull/1785/commits/e5991fdd0bce90faf20353b095aea54594fdb13b adds a unit test to verify that download tasks are cancelled after calling `deletePodcastIfUnused`.

#### Specific Case

1. Install fresh app install
2. Subscribe to any podcast
3. Set breakpoints in `cancelTasks(for episodes)` to make sure tasks are being cancelled
4. Enable Network Link Conditioner with 3G Profile
5. Navigate to Acquired in Discover and download all episodes
6. Stop the app running in Xcode
7. Run the following script:

```bash
#!/bin/bash

# This command sets the podcast to a state where it has been synced and was added over a week ago
# This satisfies the "unused podcast" checks
UPDATE_PODCAST_COMMAND="UPDATE SJPodcast
SET 
    syncStatus = 1,
    addedDate = 1715375262
WHERE uuid = 'eed80bc0-599b-0133-cb25-0d11918ab357';"

sqlite3 "$(xcrun simctl get_app_container booted au.com.shiftyjelly.podcasts data)/Library/Application Support/Pocket Casts/podcast_newDB.sqlite3" "$UPDATE_PODCAST_COMMAND"
echo "Reset podcast"
```
8. Re-run the app
9. Quickly pull to refresh in the Podcasts list/grid
10. Note that tasks are cancelled

#### General Regressions

* Visit the Podcast page from the Podcast grid and verify everything works properly
* Visit the Podcast page from Discover and verify everything works properly
* Download a podcast episode and verify that it completes and plays back
* Stream a podcast episode and verify that playback works

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PocketCastsTests/Tests/End of Year/EndOfYearStoriesBuilderTests.swift
```diff
@@ -273,7 +273,7 @@ class EndOfYearStoriesBuilderTests: XCTestCase {
 }
 
 private class EpisodeMock: Episode {
-    override func parentPodcast() -> Podcast? {
+    override func parentPodcast(dataManager: DataManager) -> Podcast? {
```
leandroalonso, line 276: We added `dataManager` as an argument but it's not used inside the function. Did I miss anything?

PocketCastsTests/Tests/Utilities/PodcastManagerTests.swift
```diff
@@ -0,0 +1,92 @@
...
+        self.episode = episode
+        self.podcast = podcast
+    }
+
+    func testTaskCancellationForUnknownDeletion() async throws {
+        let podcastManager = PodcastManager(dataManager: dataManager, downloadManager: downloadManager)
+
+        // Verify the podcast and episode exist in the data manager after being added in `setUp`
+        XCTAssertEqual(dataManager.findPodcast(uuid: podcast.uuid, includeUnsubscribed: true), podcast)
+        XCTAssertEqual(dataManager.findEpisode(uuid: episode.uuid), episode)
```
leandroalonso, line 52: Isn't that kinda of a test to make sure the Mock Data Manager works? Do we need to test that here? Wouldn't the rest of the asserts fail in case this doesn't work?

leandroalonso (dismissed):
@bjtitus can we put this new logic behind the same feature flag we used in the other PR?

I know it all seems pretty straightforward and you added tests to ensure the behavior but we don't know what we don't know. 😇 

Ideally we should be able to restore the previous logic in case anything goes wrong.

PocketCastsTests/Tests/End of Year/EndOfYearStoriesBuilderTests.swift
bjtitus: I don't think it needs to be used here since it's a mock. It was required to override the [changed method in Episode](https://github.com/Automattic/pocket-casts-ios/pull/1785/files#diff-6e22a3f4b4630aa41269a0e49fa7dc254f04226db8056e7cef087237074128feR166).

bjtitus:
> can we put this new logic behind the same feature flag we used in the other PR

Absolutely! I did this locally but must have lost it in cherry picking somewhere. It's added in https://github.com/Automattic/pocket-casts-ios/pull/1785/commits/40a91ed4c0253ca34713e35396a21f7b6a17d2d5 but only covers the task cancellation. I _could_ see the async changes causing issues but I think I'd need to duplicate the methods to do that properly. I think it's sufficient to flag only the task cancellation bit if we test the Podcast view controller thoroughly, but what do you think?

PocketCastsTests/Tests/Utilities/PodcastManagerTests.swift
bjtitus: Yeah, it is, although this isn't the `DataManagerMock`, but a full separate test database instance.

I wondered if this is covering too much but when I was writing the test, any mismatch in the sample podcast or episode would cause the episode or podcast not to exist (failed database save, which is just logged to the console). In that case, the assertion to check for a running task would probably catch it but it didn't seem very clear.

I could also, or instead of these assertions, add a check that the first task is for the URL we expect, which would at least cover the case where some other tasks sneaks in.
```swift
XCTAssertEqual(task.originalRequest?.url?.absoluteString, episode.downloadUrl)
```

leandroalonso:
> I think it's sufficient to flag only the task cancellation bit if we test the Podcast view controller thoroughly, but what do you think?

Up to you!

One last question, would be possible to test the task cancelation? Like forcing a specific podcast to be deleted or something like that?

@SergioEstevao I'm sharing my 2 cents here and there but feel free to review it as you were the original reviewer.

SergioEstevao:
@bjtitus do we really need this to go in 7.65? Can't it go to trunk?

SergioEstevao (changes requested):
Great work in this feature, and replacing all the the use of direct singletons for instance classes.
Code wise is also looking good, just wanted to be sure that we want this on 7.65?

leandroalonso:
> Code wise is also looking good, just wanted to be sure that we want this on 7.65?

We're somewhat fine having that on 7.65 IMHO as we still have 1 and a half week of testing time. But from now on we should definitely target `trunk`.

bjtitus:
@leandroalonso I've updated with another test with a script to rewrite the database. I left some comments in the script to try and explain but it's essentially setting up a podcast which a user may have unsubscribed from or only ever visited in Discover. I could see an argument about whether these episodes should be deleted, but we should at least cancel any download tasks since they are _currently_ being deleted from the database and will be orphaned.

bjtitus:
@SergioEstevao + @leandroalonso I bumped this to 7.66 at this point since it's several days into the next release.

SergioEstevao (approved):
:shipit:

PR #1787, opened 2024-05-28
Implement parallel download and streaming of episode.

SergioEstevao:
Fixes #

Implement a method to allow download and streaming of the same episode content while it's being streamed

This is a first step on integration to test the performance and issues with piggybacking of downloads at the same time that the asset is streaming. This PR will start the download of the episodes when you press play and will switch to the downloaded asset when it's available.

The missing part is to integrate the resulting file with the rest of the DownloadManager actions. So for example if you start a download of an already streamed file you can benefit from the file and if you have an auto-download going on and you start streaming the auto-download is cancelled and replaced by the streamed file, etc..

## To test

 - Start the app on a **real device**
 - Enable the FF `cachePlayingEpisode`
 - Check that you have disable Auto-Downloads features so that they not override the streaming downloads
 - Start playing any audio episode of a podcast ( video podcasts are not supported for this method)
 - Check on the console for log messages that start with `Media Export Session`
 - Check if you see messages like `Media Export Session -> Finished exporting to:` or `Media Export Session -> Finished with error:`
 - While testing this check if performance on the player is good (no slowdowns or breaks in sound)
 - Test this with different podcasts with different effects.
 - Test this putting the app to the background while streaming

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/PlaybackManager.swift
```diff
@@ -1945,7 +1945,7 @@ class PlaybackManager: ServerPlaybackDelegate {
     // MARK: - Private helpers
 
     private func checkIfStreamBufferRequired(episode: BaseEpisode, effects: PlaybackEffects) {
-        let downloadEpisode = effects.trimSilence.isEnabled() || FeatureFlag.cachePlayingEpisode.enabled
+        let downloadEpisode = effects.trimSilence.isEnabled() && !FeatureFlag.cachePlayingEpisode.enabled
```
leandroalonso, line 1948: This needs to be OR. Otherwise, we will cause a regression: if the flag is off, users with trim silence enabled will never have this episode downloaded and trim silence won't work.

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
```diff
@@ -574,8 +574,12 @@ public class DataManager {
         episodeManager.saveFileType(episode: episode, fileType: fileType, dbQueue: dbQueue)
     }
 
-    public func saveEpisode(contentType: String, episode: Episode) {
-        episodeManager.saveContentType(episode: episode, contentType: contentType, dbQueue: dbQueue)
+    public func saveEpisode(contentType: String, episode: BaseEpisode) {
```
leandroalonso, line 577: I don't see that used anywhere. Can we delete this and the `UserEpisodeDataManager` one?

podcasts/DownloadManager.swift
```diff
@@ -231,6 +231,45 @@ class DownloadManager: NSObject, FilePathProtocol {
         }
     }
 
+    func downloadParallelToStream(of episode: BaseEpisode) -> AVPlayerItem? {
+        guard let playbackItem = PlaybackItem(episode: episode).createPlayerItem() else {
```
leandroalonso, line 235: When I ran the app, the flag was turned off. Then I turned on, paused, and played the episode, this was never hit.

This is not such a big deal but I wonder if we can be more dynamic? If the flag is on, we execute this. If it's not, then we don't.

podcasts/DownloadManager.swift
```diff
@@ -231,6 +231,45 @@ class DownloadManager: NSObject, FilePathProtocol {
...
+        guard FeatureFlag.cachePlayingEpisode.enabled,
+              !episode.videoPodcast(),
+              let urlAsset = playbackItem.asset as? AVURLAsset,
+              !urlAsset.url.isFileURL // only  start download if it's a remote file that we are playing
+        else {
+            return playbackItem
+        }
+        #if !os(watchOS)
+        Task {
+            //TODO: check for the following: download is not happening already because of auto download, cancel the auto download and use this file
```
leandroalonso: Could you please create an issue so we can keep track of these tasks?

I believe it's better than having a `TODO` here. :)

podcasts/DownloadManager.swift
```diff
@@ -231,6 +231,45 @@ class DownloadManager: NSObject, FilePathProtocol {
...
+            episode.autoDownloadStatus = AutoDownloadStatus.playerDownloadedForStreaming.rawValue
+            episode.contentType = UTType.mpeg4Audio.preferredMIMEType
+            DataManager.sharedManager.save(episode: episode)
+
+            NotificationCenter.postOnMainThread(notification: Constants.Notifications.episodeDownloadStatusChanged, object: episode.uuid)
+
+            let outputURL = URL(fileURLWithPath: streamingBufferPathForEpisode(episode), isDirectory: false)
+            //TODO: force contentType to M4A so downloads paths match on the remaing code
```
leandroalonso: Wasn't that done on line `250`?

podcasts/DownloadManager.swift
```diff
@@ -231,6 +231,45 @@ class DownloadManager: NSObject, FilePathProtocol {
...
+
+            downloadingEpisodesCache[episode.uuid] = episode
+            FileLog.shared.addMessage("Media Export Session -> starting exporting: \(episode.title ?? "")")
```
leandroalonso: [nit]

I think we can use the standard we use in other log messages, ie.:

```
DownloadManager export session: start exporting <episode uuid>
```

(episode UUID is easier for us to locate)

podcasts/MediaExporter.swift
```diff
@@ -0,0 +1,51 @@
...
+
+struct MediaExporter {
+
+    #if !os(watchOS)
+    static func exportMediaItem(_ item: AVPlayerItem, to outputURL: URL) async -> Bool {
+        let composition = AVMutableComposition()
+
+        guard let compositionAudioTrack = composition.addMutableTrack(withMediaType: .audio, preferredTrackID: CMPersistentTrackID(kCMPersistentTrackID_Invalid)),
+            let sourceAudioTrack = try? await item.asset.loadTracks(withMediaType: .audio).first else {
+            FileLog.shared.addMessage("Media Export Session -> Failed to create audio track")
```
leandroalonso: [nit]

Same comment as the other file

podcasts/MediaExporter.swift
```diff
@@ -0,0 +1,51 @@
...
+            return false
+        }
+        do {
+            try compositionAudioTrack.insertTimeRange(CMTimeRangeMake(start: .zero, duration: item.asset.duration), of: sourceAudioTrack, at: CMTime.zero)
+        } catch {
+            FileLog.shared.addMessage("Media Export Session -> Failed to create audio track: \(error)")
```
leandroalonso: [nit] Same

podcasts/MediaExporter.swift
```diff
@@ -0,0 +1,51 @@
...
+            return false
+        }
+
+        guard let exporter = AVAssetExportSession(asset: composition, presetName: AVAssetExportPresetAppleM4A) else {
+            FileLog.shared.addMessage("Media Export Session -> Failed to create export session")
```
leandroalonso: [nit] :D

podcasts/MediaExporter.swift
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        await exporter.export()
+
+        if let error = exporter.error {
+            FileLog.shared.addMessage("Media Export Session -> Finished with error: \(error)")
+            return false
+        }
+
+        FileLog.shared.addMessage("Media Export Session -> Finished exporting to: \(outputURL)")
```
leandroalonso: When looking at the logs I don't see much value in having the output URL there. I see the value when we're debugging though, so maybe you can replace that in favor of a print?

leandroalonso (changes requested):
@SergioEstevao tested and it worked as described. I left a few comments and there's also a potential regression with one of the changes.

podcasts/DownloadManager.swift
SergioEstevao: If you switch to a new episode, then the code to load the new episode will be hit and a new item will be created and the flag will take effect.

podcasts/DownloadManager.swift
SergioEstevao: You are correct this TODO comment  is no longer needed.

podcasts/DownloadManager.swift
SergioEstevao: You are right this is no longer needed.

podcasts/DownloadManager.swift
SergioEstevao: Will do.

podcasts/MediaExporter.swift
SergioEstevao: I will keep the log for us to know the export session ended successfully but keep the outputURL on a print for finding the file quickly if we need to exam it.

Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
SergioEstevao: It's being used on the `MetadataTask` class line 59. 
I just made the code more generic by using BaseEpisode and switching inside to the correct episode Manager.
I was also using it on `downloadParallelToStream` but then realized that I was already doing a global episode save and remove it on a later interaction.

podcasts/PlaybackManager.swift
SergioEstevao: Maybe I'm seeing this wrong but if the flag is off the condition above will say true ( there is a !) so the download will happen if trim is enabled.
If the flag is on we don't want to do this download because we will use the download done by the export session.

podcasts/DownloadManager.swift
SergioEstevao: Ticket here: https://github.com/Automattic/pocket-casts-ios/issues/1792

podcasts/PlaybackManager.swift
leandroalonso: I missed the `!`, shame on me! Makes sense, thanks for explaining it!

podcasts/DownloadManager.swift
leandroalonso: I was thinking of a more generic Issue, such as "Downloading streaming episodes with AVMutableComposition". Then we can add subtasks there, like this one.

podcasts/MediaExporter.swift
```diff
@@ -0,0 +1,52 @@
...
+
+struct MediaExporter {
+
+    #if !os(watchOS)
+    static func exportMediaItem(_ item: AVPlayerItem, to outputURL: URL) async -> Bool {
+        let composition = AVMutableComposition()
+
+        guard let compositionAudioTrack = composition.addMutableTrack(withMediaType: .audio, preferredTrackID: CMPersistentTrackID(kCMPersistentTrackID_Invalid)),
+            let sourceAudioTrack = try? await item.asset.loadTracks(withMediaType: .audio).first else {
+            FileLog.shared.addMessage("Media Export Session -> Failed to create audio track")
```
leandroalonso: This log hasn't been updated to the new pattern

podcasts/MediaExporter.swift
```diff
@@ -0,0 +1,52 @@
...
+
+        await exporter.export()
+
+        if let error = exporter.error {
+            FileLog.shared.addMessage("DownloadManager export session: finished with error -> \(error)")
+            return false
+        }
+
+        FileLog.shared.addMessage("DownloadManager export session: Finished exporting successfully")
+        print(outputURL)
```
leandroalonso: Do we want to leave that here after merging?

leandroalonso (changes requested):
✅ Tested with `cachePlayingEpisode` disabled and it worked as it used to be before
✅ Tested with `cachePlayingEpisode` enabled and the streaming is exported

However, when we save the stream buffer, I feel the change from `DefaultPlayer` to `EffectsPlayer` is more noticeable. To test this:

1. Make sure to enable trim silence
2. Disable the `cachePlayingEpisode` and play any episode
3. Keep an eye on the log to see when `DefaultPlayer` changes to `EffectsPlayer`
4. When that happens, the transition is almost unnoticeable
5. Enable `cachePlayingEpisode`
6. Play any other episode
7. Keep an eye on the log to see when `DefaultPlayer` changes to `EffectsPlayer`
8. Check how it's way more noticeable. Sometimes the audio pauses briefly, it seems it's not on the correct spot

Is curious though because there's no change in what happens AFTER it's downloaded. I'm just wondering if the exported audio is somehow a bit different?

Anyway I think it's good to take a look if we can have the same smooth transition because the idea is to enable that for everyone.

podcasts/MediaExporter.swift
SergioEstevao: Nope, I will remove it.

SergioEstevao:
>s curious though because there's no change in what happens AFTER it's downloaded. I'm just wondering if the exported audio is somehow a bit different?
> Anyway I think it's good to take a look if we can have the same smooth transition because the idea is to enable that for everyone.

This is curious in my tests I actually felt that the transition was smoother, but maybe this can be related to the original format the episode has. 

The Export Session is always trans-coding to `M4A` format,  this can have an impact on duration/sound volume/etc... do you think we should try to do  a cross-fade effect when transitioning?

podcasts/DownloadManager.swift
SergioEstevao: Like this: https://github.com/Automattic/pocket-casts-ios/issues/1795 ?

leandroalonso:
> The Export Session is always trans-coding to M4A format, this can have an impact on duration/sound volume/etc... do you think we should try to do a cross-fade effect when transitioning?

I'm testing with this podcast: https://pca.st/oubRql

Could try and see if it's the same for you?

Anyway, once we address the code changes here I think this PR is good to go, we can assess/fix this in another task.

SergioEstevao:
@leandroalonso I updated the code for your suggestions. 
Regarding the gap we can hear when switching from streaming to the downloaded file, it was already happening before but we will try to improve this on other PRs by doing smarter switching between players.

leandroalonso (approved):

PR #1789, opened 2024-05-28
Up Next: add history and the option to restore

leandroalonso:
One of the most frustrating things that can happen are destructive bugs such as the Up Next being wiped out, or random episodes removed.

As of today, when this happens it's pretty much done: the data is lost.

While we're still investigating and fixing sync issues, they're not gone, and many things that aren't necessarily bugs can happen. With that in mind, we now keep a history of the Up Next sync changes for **up to 2 weeks** so the user can: 1) check and 2) restore.

https://github.com/Automattic/pocket-casts-ios/assets/7040243/b2ca74e1-357c-4c8c-a1f2-835423596116

## To test

You'll need to login into the same account on two devices. I recommend using the web player and the iOS simulator.

1. On the non-iOS device, clear your Up Next
2. Go to the iOS device, Profile > Refresh Now
3. ✅ Up Next should be gone
4. Go to Profile > Settings > Up Next History
5. ✅ You should see an entry there
6. Tap on it
7. ✅ It should display the episodes that were on the Up Next before
8. Tap Restore and then tap Restore again
9. ✅ Your Up Next should be restored
10. Go to the other device and refresh
11. ✅ It should display your restored Up Next
12. On iOS, tap and hold the mini player then tap "Close and Clear Up Next"
13. Go to Profile > Settings > Up Next History
14. ✅ There should be an entry there from just before wiping the Up Next

## Technical discussion

The restore is done episode by episode, adding to the bottom of the Up Next. I also implemented a mechanism to completely replace the Up Next but then it would have been necessary to add multiple options and to explain that to the user. For the sake of simplicity, I just removed this option as it seems too confusing.

## Checklist


bjtitus (approved):
:shipit:
✅ Checked restoring from Up Next history item
✅ Verified history items on sync
✅ Verified history items after clearing

leandroalonso:
Ops, I've forgotten the `CHANGELOG`. Will do in the other PR

PR #1797, opened 2024-05-31
Playback skipping fix download sizes

SergioEstevao:
Fixes #

This change updates the streaming/download code to set the File Size when it exports correctly

## To test

- Start the app
- Ensure that the FF for `Cache Playing Episode` is active
- Start playing an episode that you never downloaded or played before
- Wait for a while until you see in the logs a message saying: `DownloadManager export session: Finished exporting successfully`
- Now open the episode detail
- Check if the episode size is show correctly
- Tap on the Download option where the size of the file show
- Check that the download is automatic because the file was already downloaded for streamming
- Check that size is still correct.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/FileManager+Helper.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+extension FileManager {
+
+    func fileSize(of url: URL) -> Int64? {
```
bjtitus, line 5: You can use `url.resourceValues(forKeys: [.fileSizeKey]).fileSize` and avoid the `FileManager` here. I _think_ this is preferable since the API has slightly better type safety and the [FileManager docs](https://developer.apple.com/documentation/foundation/filemanager) say to avoid the string APIs:

> When specifying the location of files, you can use either NSURL or NSString objects. The use of the [NSURL](https://developer.apple.com/documentation/foundation/nsurl) class is generally preferred for specifying file-system items because URLs can convert path information to a more efficient representation internally.

bjtitus (changes requested):
Everything looks good in testing but just left one small suggestion and had one question:

> Check that the download is automatic because the file was already downloaded for streamming

Would it make sense to just mark these as downloaded? When I'm using a "Downloaded" filter or something I'd like to see _anything_ I could possibly play, which would include these episodes, right?

SergioEstevao:
> Would it make sense to just mark these as downloaded? When I'm using a "Downloaded" filter or something I'd like to see anything I could possibly play, which would include these episodes, right?

I didn't do this, because I was keeping the previous behaviour for streamed episodes ( for example when trimming silence is active). The main difference is that because the user didn't explicitly press download these items will be clear/delete from the file system after a while. While if mark them downloaded they will stay in place until the user decides to delete them.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/52">7.66</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/FileManager+Helper.swift
SergioEstevao: Will update the code to use that method instead.

SergioEstevao:
@bjtitus ready for another look.

bjtitus (approved):
:shipit: Looks good!

PR #1798, opened 2024-05-31
Folders: add history and the option to restore

leandroalonso:
Last year we had a bug — due to Apple Watch sync issues — that was removing podcasts from their original folders. This bug has been fixed for a while, but recently we've seen two reports from people who experienced the same.

While this issue was also fixed, it's always bad when users contact us about this issue and we can't do anything.

Now we can! We'll store when podcasts are removed from folders as part of sync and the user can restore this if they want to.

https://github.com/Automattic/pocket-casts-ios/assets/7040243/37251d90-4676-47a4-bd8d-94f14c7c9fae

## To test

You'll need to login into the same account (a Plus one) on two devices. I recommend using the web player and the iOS simulator.

1. Create a folder and add podcasts to it
2. Go to Profile > Refresh Now
3. On web, remove all podcasts from this folder
4. Go back to the iOS app: Profile > Refresh Now
5. Confirm that all podcasts are out of their folders
6. Go to Profile > Settings > Folders History
7. ✅ You should see the history of this sync there
8. Tap on it
9. ✅ You should see a list with the podcasts removed from the folders
10. Tap restore, confirm
11. ✅ Check that podcasts are now back in their original folders
12. Go to web, refresh
13. ✅ Podcasts should be back in their original folders

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/FolderHistoryManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/FolderHistoryManager.swift
@@ -14,8 +14,9 @@ public class FolderHistoryManager {
             do {
                 db.beginTransaction()
 
+                let date = Date()
                 try podcastsAndFolders.forEach {
-                    try db.executeUpdate("INSERT INTO PodcastFoldersHistory VALUES ?, ?, ?", values: [$0.key, $0.value, Date()])
+                    try db.executeUpdate("INSERT INTO PodcastFoldersHistory VALUES (?, ?, ?)", values: [$0.key, $0.value, date])
                 }
                 try db.executeUpdate("DELETE FROM PodcastFoldersHistory WHERE date <= ?", values: [Date().addingTimeInterval(-periodOfSnapshot)])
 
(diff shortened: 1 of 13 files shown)


SergioEstevao:
I tested this feature in terms of UI and all the steps above are working correctly.

One suggestion for improvement will be that we should force a refresh of the folder thumbnail presented in the podcasts, for example after I restored my `Random` folder when I switched back to Podcasts view I still saw the folder has empty:

| 1 | 2 | 
| - | - | 
| Simulator Screenshot - iPhone 15 Pro - 2024-06-03 at 13 06 32 |  Simulator Screenshot - iPhone 15 Pro - 2024-06-03 at 13 12 56 | 

Of course when I entered the folder the podcasts were there but only I scrolled up down and forced the cell to refresh the folder thumbnail updated correctly.

SergioEstevao (approved):
Looking good in terms of code and general functionality just left a comment regarding a possible improvement in terms of folder refresh after restoring.

leandroalonso:
> Looking good in terms of code and general functionality just left a comment regarding a possible improvement in terms of folder refresh after restoring.

Good point! I'll look into it.

leandroalonso:
@SergioEstevao done in 96b2cd29c0947d2553a71ce70a9e2ad89c19bc39

Could you please test again?

SergioEstevao (approved):
Thanks for adding the folder refresh! :shipit:

PR #1801, opened 2024-06-03
Playback: only update podcast if episode playback fails

leandroalonso:
When playing an episode, we have a mechanism that given some conditions might lead to a bigger time between tapping play and the app playing. Eg.:

1. User taps play on a `The Daily` episode
2. Before playing, we check to see if the podcast needs to be updated
3. If it needs, we update the podcast (small podcasts updates fast, bigger ones slower)
4. Then we start the `AVPlayer` that will then start buffering the episode

We do that no matter whether the URL we have works or not.

This PR changes this to only update in case the playback fails. Which means it will only execute it if needed.

As this is a playback change, it's behind a feature flag that we can remote control.

## To test

1. Run a clean install of the app (we'll mess with the database)
6. Go to Discover, add different episodes from different podcasts to Up Next
7. Stop the app
8. Go to the terminal and execute the following command: `sqlite3 "$(xcrun simctl get_app_container booted au.com.shiftyjelly.podcasts data)/Library/Application Support/Pocket Casts/podcast_newDB.sqlite3" "UPDATE SJPodcast SET lastUpdatedAt = 'Fri, 29 May 2020 08:06:13 GMT'; UPDATE SJEpisode SET downloadUrl = 'https://pocketcasts.com/fail.mp3';"`
9. Re-run the app
10. Play any episode from Up Next
11. ✅ On the console you should see `PlaybackManager: URL failed to load, trying to update episode and playing again` and `PlaybackManager: Episode  updated, trying to play again.`
12. ✅ The episode should play
13. Try the same on a few more episodes, they all should play
14. Go to Profile > Beta Features > disable `whenPlayingOnlyUpdateEpisodeIfPlaybackFails`
15. Play any episode not played yet
16. ✅ It should play and you should not see the aforementioned logs on the console

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 7.66
 -----
 - Re-add Use contentType to set extension for downloaded files. [#1743]
+- Improve responsiveness when playing a podcast [#1801]
 
 7.65
 -----
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -63,6 +63,7 @@ public enum FeatureFlag: String, CaseIterable {
 
     /// Only update an episode if it fails playing
     /// If set to `false`, it will use the previous mechanism that always update
+    /// but can lead to a bigger time between tapping play and actually playing it
     case whenPlayingOnlyUpdateEpisodeIfPlaybackFails
 
     public var enabled: Bool {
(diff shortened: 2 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

SergioEstevao (approved):
Working great, you can notice the difference in terms of lag when playing new episodes.

I just thought of and edge case that I wanted to check with you, what will happen if the user sets an URL for the episode, but then updates to a new file but the old file is still around and available. Will it play the previous file?

leandroalonso:
> Will it play the previous file?

Yes, it will.

Whether this is a problem or not we'll need to test and see how it goes. :)

PR #1804, opened 2024-06-05
Upgrade release toolkit from version 9.4.0 to latest, 11.0.2

mokagio:
Follows up from https://github.com/Automattic/pocket-casts-ios/commit/7aa8af3719918bd74536e4f67240524d1b47a464

I followed the migration instruction [from the `release-toolkit` repo](https://github.com/wordpress-mobile/release-toolkit/blob/476f94f8cf02847cab18d88e59270c3b0372999f/MIGRATION.md#from-1000-to-1100)

## To test

Unfortunately, these can only be tested "for real" by running a release workflow. In the meantime, at least seeing CI green lets us know that the updated `Fastfile` parses correctly.

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary. – N.A., but **cc @Automattic/pocket-casts-ios for visibility. Let me know if something doesn't work as expected during the next code freeze or release finalization**


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/52">7.66</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

leandroalonso (approved):
I'll be the next release manager so I'll keep an eye on everything.

leandroalonso:
@mokagio It seems the prototype build is failing. However, my question is: can we make that optional? Either by running this step, adding a label, or something like that. I don't believe we need a prototype build for every PR. It's the opposite: the minority of PRs will need it.

We can save a lot of processing and networking by making it optional.

SergioEstevao:
> @mokagio It seems the prototype build is failing. However, my question is: can we make that optional? Either by running this step, adding a label, or something like that.

On Tumblr  I think we used to have a special PR label to trigger the process to create a build.

mokagio:
For reference, the Prototype Build was built fine, but failed to get uploaded to App Center. Looks like they were having some certs issue.

```
[04:02:08]: Error setting metadata: 400: <html>
<head><title>400 No required SSL certificate was sent</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<center>No required SSL certificate was sent</center>
<hr><center>Microsoft-Azure-Application-Gateway/v2</center>
</body>
</html>
```

From https://status.appcenter.ms/:

image

mokagio:
@leandroalonso @SergioEstevao 

> can we make that optional? Either by running this step, adding a label, or something like that. I don't believe we need a prototype build for every PR. It's the opposite: the minority of PRs will need it.

Sure thing. 

I'm always of two minds on the topic. On the one end, I don't like to "waste" resources building builds we won't use. On the other, those resources are relatively cheap and one could argue that having the build ready when one needs it is worth the cost over time.

Just to say, neither approach is wrong. As always in software, it's a matter of tradeoffs 😄 

The simplest thing I can think of is to have a button in the pipeline.

We could of course do a label like @SergioEstevao mentioned, but in that case we might also need to add a GitHub workflow similar to the one in [WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/.github/workflows/run-danger.yml#L3-L5) to trigger the pipeline when the label is added, as Buildkite does not subscribe to the `labeled` event.

Let me know what you'd prefer. Thanks!

PR #1819, opened 2024-06-11
AVExportSession - Update status of download correctly

### LINKED ISSUE #44: Playback: audio skipped
Initially reported p1650892587223279-slack-C02A333D8LQ

**What the listener said:**

> I was just listening to a 2hour podcast and at about 1hr10min it just skipped back to the beginning. Yet the progress bar still showed that the episode was at 1:10!?

It happened when they played the [Biblical Series: Genesis 1 - Chaos & Order episode](https://nodeweb.pocketcasts.com/admin/podcasts/770548/episodes/923869971).

**What has been done so far**

* Enabled Legacy Bluetooth Support
* Disabled Intelligent Playback Resumption
* Noticed this happens on Bluetooth and over phone speakers

SergioEstevao:
Fixes #44 

Update status of download correctly if user starts download or auto download was started running while playing.

## To test

- Start the app
- Ensure that you have the FF `cachePlayingEpisode` active

### Scenario 1 - Start streaming then start download
- Start playing an episode
- While it's playing tap on the download button on episode detail
- Check if the progress circle on episode detail starts showing progress detail for the download
- Check that the download finishes correctly and you can see the final episode size
- Check that playing continues correct

### Scenario 2 - Start downloading then start playing the episode
- Start download an episode, one with larger duration will make it easier to test
- While it's downloading tap on the play button on episode detail
- The progress circle should go back to 0% ( because we started a new download for the export session)
- Check if the progress circle on episode detail continues to show progress detail for the download
- Check that the download finishes correctly and you can see the final episode size
- Check that playing continues correctly

### Scenario 3 - Start downloading, the start playing, then cancel download

- Start download an episode, one with larger duration will make it easier to test
- While it's download tap on the play button on episode detail
- Check if the progress circle on episode detail continues to show progress detail for the download
- In the middle of the download stop the download on episode detail
- Check that playing continues correctly

### Scenario 4 - Start streaming on multiple episodes

- Start playing an episode.
- Start playing another episode before the first one ends
- Resume playing the original episode
- Check that all works correctly in terms of playing sound, and that you only see two `DownloadManager export session: start exporting ... ` in the logs.

### Scenario 5 - Auto-Downloads

- Enable Auto-downloads for items added to UpNext
- Add multiple items to UpNext
- Play one of those items
- See that the download continues ( it may go back in terms of progress, because we are starting a streaming download)
- Wait to see if the download completes
- Now tap play on  another item for the podcast
- Check that the player starts and at the same time download starts ( remember we have auto-downloads for all up next)
- Tap play on another item on the podcast
- Check the previous episode is queued and starts download again
- Check that new item is playing and downloading

### Scenario 6 - Auto-Downloads and AutoPlay

- Enable Auto-downloads for UpNext and Auto-Play episode
- Clean up your UpNext Queue
- Start playing an episode on podcast that has more episodes to play after
- Seek to the end of episode, less some seconds
- Wait for the episode to finish
- Check if the new episode starts and downloads goes correctly

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso:
> Scenario 2 - Start downloading then start playing the episode

@SergioEstevao for this scenario, the download is reset to 0% once I start playing it. Is that expected? Just double-checking cause it's not clear on the PRs description.

leandroalonso:
@SergioEstevao I'm facing a bug:

1. Start downloading an episode
2. Play it
3. Download will revert to 0%
4. In this case, it stays in 0%
5. Tap Download it
6. Based on the logs, a normal download starts (but you can also see the download indicator going briefly back to 0 and up again)

Video:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/99246749-ca74-4797-b213-e259d1801bd9

SergioEstevao:
> @SergioEstevao for this scenario, the download is reset to 0% once I start playing it. Is that expected? Just double-checking cause it's not clear on the PRs description.

I should have make it clearer on the description, but yes it's expected because we are cancelling the exiting download and replace it with the new one done by the export session, so the download restarts from zero.

leandroalonso:
@SergioEstevao got it. But take a look in this message: https://github.com/Automattic/pocket-casts-ios/pull/1819#issuecomment-2163869875

This one is a bug and I can reproduce it very often

leandroalonso (changes requested):
## One session at a time

Only one export session works at a time. I think that’s ok, the only issue I see is that the download keeps showing a “0%” progress, which feels like it’s not working. It would be better if we show “Queued”

To reproduce:
1. start playing an episode (so an export session starts)
2. Start playing another episode
3. Tap download
4. Until the first one finishes, the second one will never start

Another possible solution is:

1. If a user switches episodes
2. AND the previous episode wasn’t downloading (either because the user tapped download or auto-download)
3. Cancel the export session for the previous one

https://github.com/Automattic/pocket-casts-ios/assets/7040243/eb03913a-fb5f-4217-bb44-c59fdf56fd1f


## Download progress not reliable

1. Have auto download Up Next: enabled
2. Play multiple episodes
3. The download progress indicator shows progress, then shows nothing. Sometimes it gets messy between the episodes

https://github.com/Automattic/pocket-casts-ios/assets/7040243/4b196450-3c7f-4263-a945-87078120176c

## Redownloading

This one is quite bad. I have items in my Up Next already downloaded, but when switching between the episodes they start to download again. To reproduce:

1. Have auto download for Up Next enabled
2. Add items to your Up Next so they download
3. Wait for them all to finish it
4. Play one of them again
5. After a while, some already downloaded episodes will start downloading again (notice in the video the "David Sacks on his..." episode, it's downloaded, but then starts downloading again)

https://github.com/Automattic/pocket-casts-ios/assets/7040243/fb34b5df-8726-40d4-b4d0-641a31cdf2d1

SergioEstevao:
> The download progress indicator shows progress, then shows nothing. Sometimes it gets messy between the episodes

I've found out that when auto-download was enabled the regular download was running at the same time that the streaming download. I addressed that and I think it's now sorted out.

One thing that I found out in my tests, is that even before this changes on this PR the updating of progress for episodes has some issues.

One core issue is that the download updates are staggered at one update for second for all ongoing downloads. This means that some downloads that are faster than others, "starve" the update calls for other downloads.

Check the code in `updateProgressForEpisode` in `DonwloadProgressManager` I can address this in this PR but I think it's out of scope for this change.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

SergioEstevao:
> One core issue is that the download updates are staggered at one update for second for all ongoing downloads. This means that some downloads that are faster than others, "starve" the update calls for other downloads.

I've addressed the issue above by staggering the updates for each separated episode here: https://github.com/Automattic/pocket-casts-ios/pull/1819/commits/ff96cd6e35e26341ded4b56dc77d20add93fa307

leandroalonso:
> I can address this in this PR but I think it's out of scope for this change.

@SergioEstevao I agree with you! If it was like that before just leave as it was.

leandroalonso (dismissed):
edit: I'm running the wrong code. 🤦

leandroalonso (commented):
@SergioEstevao re-tested all the mentioned scenarios and the wall works like a charm now.

I only have one question about Downloading "Only on Wifi".

Edit: We download the episode anyways when playing. I wonder if that's the behavior we want.

If not we can tackle it on another PR exclusively focused on that.

SergioEstevao:
@leandroalonso and @bjtitus I updated the code to address the issues with Auto-play episodes.

Regarding the Download Only on Wifi at the moment if you start playing an episode, the system will download the episode to the local file.  My rational is to keep it like this, for us to better test if playback skipping is addressed. If the users feel that this is wrong we can disable the export when that option is active and the device is on celular.

bjtitus (approved):
:shipit: Looked great in my testing!

> One core issue is that the download updates are staggered at one update for second for all ongoing downloads. This means that some downloads that are faster than others, "starve" the update calls for other downloads.

I see what you mean and I think this should probably be fixed separately. It seems like we could either pass the updated episode as part of the notification or maybe even switch to a publisher.

PR #1821, opened 2024-06-11
Accelerate Effects

bjtitus:
This updates our APIs to use the [Accelerate framework](https://developer.apple.com/documentation/accelerate) in an effort to reduce battery usage during episode playback when the Trim Silence effect is enabled.

* Adds an `accelerateEffects` feature flag so we can test before/after and disable if needed.
* Adds a new implementation of the `calculateRms` and `calculateStereoRms` functions which use Accelerate's vDSP functions.
* Adds a new implementation of the `fadeAudio` function which use Accelerates `vDSP_vgen` and `vDSP_vmul` to scale the values from the audio buffer.

> [!NOTE]  
> While adding unit tests, I noticed that my simple example of fading from 0-1 showed drifting results in the old fade implementation. I believe there was an off-by-one error when using the length and fixed that here. It may not have been noticeable because the values typically were within 20% of the final 0 or 1 values for the fade.

These RMS and Fade calculations are run frequently while playing with Trim Silence enabled. In our Xcode Energy usage reports, RMS is reported as the second highest cause of high energy usage terminations.

CleanShot 2024-06-11 at 16 03 52@2x

The performance tests show about a 3x speed improvement and Accelerate should be more power efficient as well.

## To test

* Run with Trim Silence enabled and verify that audio sounds correct
* Disable the `accelerateEffects` feature flag and ensure audio still sounds correct with trim silence enabled

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Playback/Effects/AudioUtilsTests.swift
+++ b/PocketCastsTests/Tests/Playback/Effects/AudioUtilsTests.swift
@@ -35,7 +35,7 @@ final class AudioUtilsTests: XCTestCase {
         let samples: [Float32] = [-8, -4, -0, 2, 4, 8]
 
         var audioBuffer = AudioBuffer()
-        audioBuffer.mNumberChannels = 1 // Mono audio
+        audioBuffer.mNumberChannels = 2 // Stereo
         audioBuffer.mDataByteSize = UInt32(samples.count * MemoryLayout<Float32>.size)
         audioBuffer.mData = UnsafeMutableRawPointer.allocate(byteCount: Int(audioBuffer.mDataByteSize), alignment: MemoryLayout<Float32>.alignment)
 
--- a/PocketCastsTests/Tests/Playback/Effects/AudioUtilsTests.swift
+++ b/PocketCastsTests/Tests/Playback/Effects/AudioUtilsTests.swift
@@ -1,0 +1,172 @@
+import XCTest
+@testable import podcasts
+import AVFoundation
+import Accelerate
+
+final class AudioUtilsTests: XCTestCase {
+
+    var audioBuffer: AudioBuffer!
+
+    override func setUp() async throws {
+        audioBuffer = sampleBuffer()
+    }
+
+    override func tearDown() async throws {
+        cleanUpAudioBuffer(buffer: &audioBuffer)
+    }
+
+    private var bufferFormat: AVAudioFormat = {
+        var asbd = AudioStreamBasicDescription(
+            mSampleRate: 44100.0,
+            mFormatID: kAudioFormatLinearPCM,
+            mFormatFlags: kAudioFormatFlagIsFloat | kAudioFormatFlagIsPacked,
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

PocketCastsTests/Tests/Playback/Effects/AudioUtilsTests.swift
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+        return AVAudioFormat(streamDescription: &asbd)!
+    }()
+
+    private func sampleBuffer() -> AudioBuffer {
+        let samples: [Float32] = [-8, -4, -0, 2, 4, 8]
+
+        var audioBuffer = AudioBuffer()
+        audioBuffer.mNumberChannels = 1 // Mono audio
```
leandroalonso: When using `EffectsPlayer` we always convert mono to stereo, as per: https://github.com/Automattic/pocket-casts-ios/pull/213

Shall we update the tests here?

podcasts/AudioUtils.swift
```diff
@@ -1,34 +1,116 @@
...
 import Foundation
+import Accelerate
+import PocketCastsUtils
 
 class AudioUtils {
     private static let bufferLength = UInt32(Constants.Audio.defaultFrameSize)
     private static let bufferByteSize = Float32(MemoryLayout<Float32>.size)
 
     class func fadeAudio(_ audio: BufferedAudio, fadeOut: Bool, channelCount: UInt32) {
+        if FeatureFlag.accelerateEffects.enabled {
```
leandroalonso, line 11: I don't think this is a problem (at least it wasn't in my tests): changing from one method to the other doesn't matter because the result will be the same right?

I'm just thinking about an occasion in which we switch the feature flag while someone is playing. Based on my real device tests, this doesn't matter, but I want to double-check with you.

leandroalonso (approved):
@bjtitus thanks for working on that, this is nice! Left just some minor comments but everything works as expected.

podcasts/AudioUtils.swift
bjtitus: > changing from one method to the other doesn't matter because the result will be the same right

Yes, this shouldn't matter since we are expecting the values to be the same.

I did a similar test by toggling the flag during playback and it does switch methods but I wasn't able to notice any change. Probably to be expected since the values should be the same.

PocketCastsTests/Tests/Playback/Effects/AudioUtilsTests.swift
bjtitus: It doesn't seem to make a difference to the test results. I'll set it to 2 channels just to make it better match the handling from within `EffectsPlayer`.

PR #1833, opened 2024-06-14
EffectsPlayer: remove different setup mechanisms

leandroalonso:
This undo the changes from #1055 keeping the setup number 2 which solved the issue (that was later updated on #1132).

Ps.: I won't remove Firebase flag because it will make the apps revert to the number 1 setup, which we don't want.

## To test

1. Comment (or remove) `EffectsPlayer.swift` lines `144` to `158` (by doing this the engine won't start thus the playback will fail)
1. Download some episodes or add Files and **enable** trim silence

Then:

2. Play a downloaded episode or local file with trim silence enabled
3. ✅ The play shouldn't work
4. Check the logs, you should see three `Loading <episode title>`, and for the first two you should see something like `EffectsPlayer: failed to start playback`

## Checklist

### DIFF
--- a/podcasts/Constants.swift
+++ b/podcasts/Constants.swift
@@ -275,9 +275,6 @@ struct Constants {
         static let endOfYearRequireAccount = "end_of_year_require_account"
         static let endOfYearRequireAccountDefault: Bool = true
 
-        static let effectsPlayerStrategy = "effects_player_strategy"
-        static let effectsPlayerStrategyDefault: Int = 1
-
         static let patronCloudStorageGB = "patron_custom_storage_limit_gb"
         static let patronCloudStorageGBDefault = 100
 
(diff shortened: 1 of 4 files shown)


SergioEstevao:
When testing the app following the steps in the PR description I didn't see the app crash but just showing the following on the screen when I tap play:

https://github.com/Automattic/pocket-casts-ios/assets/651601/6219ff35-be0b-47c0-b723-264ac97678b7



Is that what you mean by crash? Or something else going on?

Logs:
```
Using EffectsPlayer
EffectsPlayer: converting mono to stereo
EffectsPlayer: engine reported not running, calling playbackDidFail
```

leandroalonso:
@SergioEstevao oh you're right. I forgot about #1132, in which we changed the mechanism not to retry. I've updated the testing steps.

SergioEstevao (approved):
:shipit:

PR #1836, opened 2024-06-18
Xcode 16.0 Beta 1: update project

leandroalonso:
Updates the project so it runs on Xcode 16.0 Beta 1.

## To test

1. 🟢 CI
2. ✅ Run the project using Xcode 16.0 Beta 1

## Checklist

### DIFF
--- a/podcasts/MainTabBarController.swift
+++ b/podcasts/MainTabBarController.swift
@@ -8,13 +8,13 @@ import PocketCastsUtils
 class MainTabBarController: UITabBarController, NavigationProtocol {
     enum Tab { case podcasts, filter, discover, profile, upNext }
 
-    var tabs = [Tab]()
+    var pcTabs = [Tab]()
 
     let playPauseCommand = UIKeyCommand(title: L10n.keycommandPlayPause, action: #selector(handlePlayPauseKey), input: " ", modifierFlags: [])
 
     private lazy var endOfYear = EndOfYear()
 
-    private lazy var profileTabBarItem = UITabBarItem(title: L10n.profile, image: UIImage(named: "profile_tab"), tag: tabs.firstIndex(of: .profile) ?? -1)
+    private lazy var profileTabBarItem = UITabBarItem(title: L10n.profile, image: UIImage(named: "profile_tab"), tag: pcTabs.firstIndex(of: .profile) ?? -1)
 
 
     /// The viewDidAppear can trigger more than once per lifecycle, setting this flag on the first did appear prevents use from prompting more than once per lifecycle. But still wait until the tab bar has appeared to do so.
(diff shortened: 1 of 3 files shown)


podcasts/MainTabBarController.swift
```diff
@@ -8,13 +8,13 @@ import PocketCastsUtils
 class MainTabBarController: UITabBarController, NavigationProtocol {
     enum Tab { case podcasts, filter, discover, profile, upNext }
 
-    var tabs = [Tab]()
+    var pcTabs = [Tab]()
```
leandroalonso, line 11: I had to do this because of:

Screenshot 2024-06-18 at 12 16 40

bjtitus (approved):
✅ Compiled with Xcode 16 + Xcode 15
✅ Double checked VoiceOver on the tabs and it seems alright

rviljoen:
@leandroalonso @bjtitus Were you able to run the XCode 16 builds on a real device? I upgraded to the iOS and MacOS public betas today. I can build the project and run it in the iOS 18 simulator without issues, but when I run it on my iPhone 15 Pro with iOS 18 then I get multiple crashes across the UI, similar to this:
`Fatal error: Attempted to read an unowned reference but object 0x3029a78c0 already deallocated`

Running the latest TestFlight build on my iPhone works without issues..

leandroalonso:
Hi @rviljoen! I haven't tested yet on a real device with iOS 18. I'll try it and see

Does it happen immediately or it depend?

rviljoen:
@leandroalonso Thanks! Happens very soon after starting playback. If I just open the app and leave it, then it does not crash but as soon as I expand the mini-player it will crash. At the time of the crash it seems to be requesting show info via this closure:

https://github.com/Automattic/pocket-casts-ios/blob/076e03b5d7c26e4af2535a2fd0e7efe1a32f35ba/podcasts/Episode%20Info%20Coordinator/ShowInfoCoordinator.swift#L115-L124

leandroalonso:
If you change that to:

```swift
let task = Task<Episode.Metadata?, Error> { [weak self] in 
     do { 
         let data = try await dataRetriever.loadEpisodeDataFromCache(for: podcastUuid, episodeUuid: episodeUuid) 
         requestingShowInfo[episodeUuid] = nil 
         return await getShowInfo(for: data?.data(using: .utf8)) 
     } catch { 
         requestingShowInfo[episodeUuid] = nil 
         throw error 
     } 
 } 
```

Does it fixes it?

rviljoen:
@leandroalonso Thanks for the direction. I could not reproduce the problem this morning, which drove me crazy 😄 I then looked at the Beta features and found that the crash only happens with `newShowNotesEndpoint` enabled. Changing `unowned self` to `weak self` does seem to resolve it for now, but I'll keep testing. I did not add the guard because the return requires an `Episode.Metadata` object and if I returned `nil` I would get this error:

> Actor-isolated property 'requestingShowInfo' can not be mutated from a non-isolated context

Issue #1848, opened 2024-06-21
Transcripts - Phase 1


SergioEstevao:
```[tasklist]
### Tasks
- [x] Database data structure
- [x] Network code to retrieve the transcript files and store them locally https://github.com/Automattic/pocket-casts-ios/pull/1955
- [x] Parse data in Transcript file and implement in memory data structure
- [x] Implement UI component to show transcript synced with position of AVPlayer
- [x] Search on single transcript https://github.com/Automattic/pocket-casts-ios/pull/1982
- [x] Fix the logic to check if an episode has a transcript (sometimes when opening the player it doesn't update) https://github.com/Automattic/pocket-casts-ios/pull/1931
- [x] Ensure it works while offline https://github.com/Automattic/pocket-casts-ios/pull/1955
- [x] Format HTML, leave only text (strip tags) #1941
- [x] Check if transcript is empty and show appropriated error message #1944
- [x] Disable scroll on sync until we get a way to to resync out of sync scripts.
- [x] Accessibility - https://github.com/Automattic/pocket-casts-ios/pull/1961
- [x] Implement analytics https://github.com/Automattic/pocket-casts-ios/pull/1994
- [x] Fallback to other transcripts types provided in feed, if main preference is not working ( empty, corrupt, etc..) https://github.com/Automattic/pocket-casts-ios/pull/2022
```

```[tasklist]
### UI
- [x] Add initial transition in the player #1899
- [x] Add the already created VC to the container in the player https://github.com/Automattic/pocket-casts-ios/pull/1909
- [x] Style the Transcript font https://github.com/Automattic/pocket-casts-ios/pull/1961
- [x] Add search controls https://github.com/Automattic/pocket-casts-ios/pull/1953
- [x] Update shelf icon and shelf action https://github.com/Automattic/pocket-casts-ios/pull/2004
- [x] Display error messages in styled way and with translated text https://github.com/Automattic/pocket-casts-ios/pull/2007
- [x] Add speakers info https://github.com/Automattic/pocket-casts-ios/pull/2012
```

Out of scope:

- Add Bookmark action (given there's no syncing in this version)
- Ensure screen doesn't lock while transcripts is being shown (same as above)

rviljoen:
[...]

PR #1852, opened 2024-06-24
Transcripts add network parse code

SergioEstevao:
Refs #1848 

Adds the network code to parse transcript information on episode. This only ensures that the transcript metadata is parsed.
This also ensure that the transcript info is load when episodes are load to play and cache/loaded when the other episode notes are loaded.

## To test

- Start the app
- Ensure  that you have the `transcripts` and `useNewShowNotesEndpoint` FF enabled
- Add a breakpoint on ShowInfoCoordinator line 59
- Open an Episode of the Podcast `Cautionary Tales with Tim Hartford` 
- Check if the break point is hit and you can do in console `po metadata.transcripts` and see the transcripts info.

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -162,6 +162,8 @@ class PlaybackManager: ServerPlaybackDelegate {
             chapterManager.clearChapterInfo()
         }
 
+        loadTranscripts()
+
         if saveCurrentEpisode && currentEpisode() != nil && !switchingToDifferentUpNextEpisode {
             recordPlaybackPosition(sendToServerImmediately: false, fireNotifications: false)
         }
@@ -2079,6 +2081,16 @@ private extension PlaybackManager {
 
         lastSeekTime = Date()
         isBack ? skipBack() : skipForward()
+    }
+
+    // MARK: - Transcripts
+    private func loadTranscripts() {
+        guard FeatureFlag.transcripts.enabled, let episode = currentEpisode(), let podcast = currentPodcast else {
+            return
+        }
+        Task.init {
+            let _ = try? await ShowInfoCoordinator.shared.loadTranscripts(podcastUuid: podcast.uuid, episodeUuid: episode.uuid)
+        }
     }
 }
 
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -2088,6 +2088,7 @@ private extension PlaybackManager {
         guard FeatureFlag.transcripts.enabled, let episode = currentEpisode(), let podcast = currentPodcast else {
             return
         }
+
         Task.init {
             let _ = try? await ShowInfoCoordinator.shared.loadTranscripts(podcastUuid: podcast.uuid, episodeUuid: episode.uuid)
         }
--- a/podcasts/ShowNotesUpdater.swift
+++ b/podcasts/ShowNotesUpdater.swift
@@ -4,10 +4,14 @@ import PocketCastsUtils
 
 class ShowNotesUpdater {
     class func updateShowNotesInBackground(podcastUuid: String, episodeUuid: String) {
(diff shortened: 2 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/ShowNotesUpdater.swift
```diff
@@ -4,10 +4,14 @@ import PocketCastsUtils
 
 class ShowNotesUpdater {
     class func updateShowNotesInBackground(podcastUuid: String, episodeUuid: String) {
-        if FeatureFlag.newShowNotesEndpoint.enabled {
+        if FeatureFlag.newShowNotesEndpoint.enabled || FeatureFlag.transcripts.enabled {
             Task {
                 // Load the show notes and any available chapters
                 _ = try? await ShowInfoCoordinator.shared.loadChapters(podcastUuid: podcastUuid, episodeUuid: episodeUuid)
+
+                if FeatureFlag.transcripts.enabled {
```
SergioEstevao, line 12: I know that the above will also load the transcripts metadata but just want be sure that this is called here even if the above is removed.

podcasts/ShowNotesUpdater.swift
```diff
@@ -4,10 +4,14 @@ import PocketCastsUtils
 
 class ShowNotesUpdater {
     class func updateShowNotesInBackground(podcastUuid: String, episodeUuid: String) {
-        if FeatureFlag.newShowNotesEndpoint.enabled {
+        if FeatureFlag.newShowNotesEndpoint.enabled || FeatureFlag.transcripts.enabled {
```
leandroalonso: I believe we don't need the `|| FeatureFlag.transcripts.enabled` do we?

Regardless of the transcripts, we want to continue caching chapters/show notes, so it can be the same as it always has.

I feel like this might lead to a regression when removing `newShowNotesEndpoint`. If we leave `transcripts` there cache for downloaded episodes will stop working.

podcasts/PlaybackManager.swift
```diff
@@ -2080,6 +2082,16 @@ private extension PlaybackManager {
         lastSeekTime = Date()
         isBack ? skipBack() : skipForward()
     }
+
+    // MARK: - Transcripts
+    private func loadTranscripts() {
+        guard FeatureFlag.transcripts.enabled, let episode = currentEpisode(), let podcast = currentPodcast else {
+            return
+        }
```
leandroalonso, line 2090: [nit]

```suggestion
        }

```

leandroalonso (approved):
@SergioEstevao left a minor comment.

It would be good to point on the PR description that in order to test we need to enable `transcripts` and `newShowNotesEndpoint`

podcasts/ShowNotesUpdater.swift
SergioEstevao: Ok, will remove this then.

SergioEstevao: Better yet, should I separate the transcripts loading completely  from the `newShowNotesEndpoint`, do we need to link transcripts  to the new endpoint? 
I see the endpoint is the one that provides the data for transcripts but was the flag just done because of the way we update notes or to control the use of the endpoint?

leandroalonso: Not really. But I plan to remove this flag anyway, so I think we can leave this for another task. Wdyt?

SergioEstevao: Ok, just updated the code [here](https://github.com/Automattic/pocket-casts-ios/pull/1852/commits/1b351697bf0bf1c1a88ec60d6626a7afddec0a61) and the PR description to reflect the changes.

PR #1853, opened 2024-06-24
Transcripts add temporary player item to see transcripts

SergioEstevao:
Refs #1848 

This PR adds a temporary tab view on the player to show a transcripts if available.

## To test

- Start the app
- Ensure that you have the `transcripts` and `newShowNotesEndpoint` FF enabled
- Open an episode with Transcripts ( Cautionary Tales podcast episode have translations)
- Play the episode
- Open the Full Screen Player
- Open the tab for Transcripts
- See if you can see the transcript and scroll around.
- Check that when the episode finishes and a new episode is loaded the transcripts updates correctly.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/PlayerTabsView.swift
```diff
@@ -21,6 +22,8 @@ enum PlayerTabs: Int {
             return L10n.chapters
         case .bookmarks:
             return L10n.bookmarks
+        case .transcripts:
+            return "Transcripts"
```
SergioEstevao: I'm not localising the strings on this PR because these are temporary until we have final designs.

leandroalonso: I believe we can start doing so, unless is a very specific copy? "Transcripts" I'm sure we will have in the final version somehow.

Anyway, not a strong feeling.

leandroalonso (approved):
Tested with the flag enabled and disabled and it works as expected.

@SergioEstevao one logic we might want to add is to display the tab only if a transcript is ready. This is a logic we will need anyway, independently of design: if there are transcripts we will show a tab/card/button/something, if not, we won't show.

However, I don't feel this needs to be tackled now, just something for us to keep in mind (and this is kinda done in the transcripts VC).

SergioEstevao:
> However, I don't feel this needs to be tackled now, just something for us to keep in mind (and this is kinda done in the transcripts VC).

I will add this on the follow up PR

PR #1896, opened 2024-07-09
Reimagine Sharing: Add scrollable timeline view

bjtitus:
| 📘 Part of: https://github.com/Automattic/pocket-casts-ios/issues/1910 |
|:---:|

Adds the timeline view for clip sharing from the Reimagine Sharing project.

https://github.com/Automattic/pocket-casts-ios/assets/3250/d7902514-2999-427b-b0b3-17d48e3b540e

### Code Changes

* `MediaTrimBar` contains the play button and the timeline with associated styled elements
* `MediaTrimView` contains the scrollable timeline portion which will include Trim handles in the future
* `ScrollableScrollView` is a wrapper which handles scrolling operations via the `scrollTo` modifier and set of `invisibleTickMarks`. There are some newer APIs we could use in the future but this approach works for now.
* `AudioWaveformView` renders the lines for our "pseudo waveform". This may be replaced with an actual waveform in the future but is a static stylized version for now.
* `PCUnevenRoundedRectangle` is a fallback for rounding corners in iOS 15 similar to `UnevenRoundedRectangle` in iOS 16+.

## To test

> [!NOTE]

* Open the Now Playing view in the player
* Tap the share button in the upper right
* Tap the "Clip" option
* Verify that the Clip sharing page appears
* Ensure that timeline is scrolled to the correct position
* Verify that the timeline behaves as expected by scrolling back and forth and changing playback position from the episode and re-entering

A few missing items which I'll submit separately:
* Zoom gesture handling - the building blocks are there in `AudioWaveformView ` for hiding and showing lines
* Playback following
* Play/Pause button functionality

The tick marks may not match the designs exactly as they will be recalculated depending on scale (which may adjust with zooming). I'll be cleaning this up in a future update.

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Extensions/ComparableExtension.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Extensions/ComparableExtension.swift
@@ -9,3 +9,13 @@ public extension Comparable {
         self > limits.lowerBound && self < limits.upperBound ? self : self.clamped(to: limits)
     }
 }
+
+public extension Comparable {
+    func clamped(to limits: ClosedRange<Self>) -> Self {
+        min(max(self, limits.lowerBound), limits.upperBound)
+    }
+
+    func betweenOrClamped(to limits: ClosedRange<Self>) -> Self {
+        self > limits.lowerBound && self < limits.upperBound ? self : self.clamped(to: limits)
+    }
+}
--- a/podcasts/Common SwiftUI/RoundedCorner.swift
+++ b/podcasts/Common SwiftUI/RoundedCorner.swift
@@ -1,0 +1,70 @@
+import SwiftUI
+
+/// Used as a fallback for iOS 15 when `UnevenRoundedRectangle` is necessary
+struct PCUnevenRoundedRectangle: Shape {
+    let topLeadingRadius: CGFloat
+    let bottomLeadingRadius: CGFloat
+    let bottomTrailingRadius: CGFloat
+    let topTrailingRadius: CGFloat
+
+    init(topLeadingRadius: CGFloat,
+         bottomLeadingRadius: CGFloat,
+         bottomTrailingRadius: CGFloat,
+         topTrailingRadius: CGFloat) {
+        self.topLeadingRadius = topLeadingRadius
+        self.bottomLeadingRadius = bottomLeadingRadius
+        self.bottomTrailingRadius = bottomTrailingRadius
+        self.topTrailingRadius = topTrailingRadius
+    }
+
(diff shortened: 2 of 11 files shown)


Modules/Utils/Sources/PocketCastsUtils/Extensions/ComparableExtension.swift
```diff
@@ -9,3 +9,13 @@ public extension Comparable {
         self > limits.lowerBound && self < limits.upperBound ? self : self.clamped(to: limits)
     }
 }
+
+public extension Comparable {
+    func clamped(to limits: ClosedRange<Self>) -> Self {
```
bjtitus, line 14: These are duplicates of the `Range` versions for `ClosedRange`. Unfortunately, they don't share a suitable protocol so I couldn't make these generic.

podcasts/Common SwiftUI/RoundedCorner.swift
```diff
@@ -0,0 +1,70 @@
+import SwiftUI
+
+/// Used as a fallback for iOS 15 when `UnevenRoundedRectangle` is necessary
```
leandroalonso, line 3: 🥲

podcasts/Sharing/AudioWaveformView.swift
```diff
@@ -0,0 +1,112 @@
+import SwiftUI
+
+struct AudioWaveformView: View {
+    let baseLineWidth: CGFloat = 2
+    let baseLineSpacing: CGFloat = 2
+    var scale: CGFloat
+    var width: CGFloat
```
leandroalonso: [nit] Do these need to be exposed?

podcasts/Sharing/Clip/PlayButton.swift
```diff
@@ -0,0 +1,19 @@
+import SwiftUI
+
+struct PlayButton: View {
```
leandroalonso: Maybe rename it to `TrimPlayButton`?

I can see confusion when searching for the current `PlayPauseButton` in the full player.

leandroalonso (approved):
Tested on real deice and on multiple devices. On iPad there's a small issue, the bottom padding is missing:

Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-07-11 at 11 07 43

It's not a blocker though and can be addressed in the future.

bjtitus:
@leandroalonso Good catch! This was an issue on all of the new sharing views. I added a change (https://github.com/Automattic/pocket-casts-ios/pull/1896/commits/c6f89188bc501474b1c158700f928b1d22badc75) to pad both top and bottom on iPad.

PR #1899, opened 2024-07-10
Transcripts: add player transition to Transcripts

leandroalonso:
Part of #1848

This adds a transition in the player, allowing transcripts to be displayed while the play controls are still present.

Notice that we haven't defined yet the point of entry to the feature.

https://github.com/Automattic/pocket-casts-ios/assets/7040243/fe7ea98a-98f3-4978-8207-a661a6f39d53

Important: the transition here is not strictly the same as the [one here](https://pocketcastsp2.wordpress.com/2024/04/29/transcripts-i1/#a-icon-toggle). There are still tweaks to be done, but I've spent quite some time on this task for now. I believe the current state allows the work to proceed and then we can refine transitions once the bulk part of the work is done.

Important pt 2: During scoping, we discussed migration of the player to SwiftUI but that will be a complex task and a project on its own.

## To test

Change `playPauseTapped` in `NowPlayingPlayerItemViewController.swift` to:

```swift
    @IBAction func playPauseTapped(_ sender: Any) {
        displayTranscript.toggle()
    }
```

1. Run the app
2. If the miniplayer is visible, tap it. If not, play any episode
3. Tap the play/pause button on the full player
4. ✅ Notice how the player controls and the progress bar goes to the bottom and the tabs disappear
5. Test on small iPhone, big iPhone and iPad

## Checklist

### DIFF
--- a/podcasts/NowPlayingPlayerItemViewController.swift
+++ b/podcasts/NowPlayingPlayerItemViewController.swift
@@ -352,6 +352,10 @@ class NowPlayingPlayerItemViewController: PlayerItemViewController {
     }
 
     private func showTranscript() {
+        skipBackBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+        skipFwdBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+        playPauseBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+
         transcriptContainerView.layer.opacity = 0
         UIView.animate(withDuration: 0.35, animations: { [weak self] in
             guard let self else { return }
@@ -379,12 +383,20 @@ class NowPlayingPlayerItemViewController: PlayerItemViewController {
 
             // Ask parent VC to hide tabs
             playerContainer?.hideTabsAndLockScrollView()
-        }, completion: { _ in
+        }, completion: { [weak self] _ in
+            guard let self else { return }
 
+            playPauseBtn.finishedTransition()
+            skipBackBtn.finishedTransition()
+            skipFwdBtn.finishedTransition()
         })
     }
 
     private func hideTranscript() {
+        skipBackBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+        skipFwdBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+        playPauseBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+
         UIView.animate(withDuration: 0.35, animations: { [weak self] in
             guard let self else { return }
 
@@ -414,6 +426,10 @@ class NowPlayingPlayerItemViewController: PlayerItemViewController {
             guard let self else { return }
 
             transcriptContainerView.isHidden = true
+
+            playPauseBtn.finishedTransition()
+            skipBackBtn.finishedTransition()
+            skipFwdBtn.finishedTransition()
         })
     }
 }
--- a/podcasts/NowPlayingPlayerItemViewController.swift
+++ b/podcasts/NowPlayingPlayerItemViewController.swift
@@ -369,8 +369,8 @@ class NowPlayingPlayerItemViewController: PlayerItemViewController {
             fillView.isHidden = false
 
             // Change skip back and forward size
-            skipBackBtn.change(width: 32, height: 32, fontSize: 10)
-            skipFwdBtn.change(width: 32, height: 32, fontSize: 10)
+            skipBackBtn.changeSize(to: .small)
+            skipFwdBtn.changeSize(to: .small)
             skipBackBtn.layoutIfNeeded()
             skipFwdBtn.layoutIfNeeded()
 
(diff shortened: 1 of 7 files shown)


podcasts/NowPlayingPlayerItemViewController.swift
```diff
@@ -330,4 +346,86 @@ class NowPlayingPlayerItemViewController: PlayerItemViewController {
...
         present(navController, animated: true, completion: nil)
     }
+
+    private func showTranscript() {
+        skipBackBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+        skipFwdBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+        playPauseBtn.prepareForAnimateTransition(withBackground: view.backgroundColor)
+
+        transcriptContainerView.layer.opacity = 0
+        UIView.animate(withDuration: 0.35, animations: { [weak self] in
```
bjtitus: 🔧 I feel like the sequence of events here is very similar in the two animations. Would it make sense to combine them into a single animation and set the values based on an `isShowing` property? Is that overthinking it? I guess I'm more worried that something would be missing from one side of the animation so this would make it less likely.

```swift
UIView.animate(withDuration: 0.35, animations: { [weak self] in
	
	shelfBg.isHidden = !isShowing
	shelfBg.layer.opacity = isShowing ? 1 : 0
	
	transcriptContainerView.isHidden = isShowing
	transcriptContainerView.layer.opacity = isShowing ? 0 : 1
	
	bottomControlsStackView.distribution = isShowing ? .equalSpacing : .fill
	bottomControlsStackView.spacing = isShowing ? 30 : 10
	
	fillView.isHidden = isShowing
	
	let skipButtonSize: SkipButtonSize = isShowing ? .large : .small
	skipBackBtn.changeSize(to: skipButtonSize)
	skipFwdBtn.changeSize(to: skipButtonSize)
	skipBackBtn.layoutIfNeeded()
	skipFwdBtn.layoutIfNeeded()
	
	if isShowing {
		playerContainer?.showTabsAndUnlockScrollView()
	} else {
		playerContainer?.hideTabsAndLockScrollView()
	}
}, completion: { [weak self] _ in
	
	if isShowing {
		transcriptContainerView.isHidden = true
	}
	
	playPauseBtn.finishedTransition()
	skipBackBtn.finishedTransition()
	skipFwdBtn.finishedTransition()
})
```

podcasts/SkipButton.swift
```diff
@@ -98,4 +105,56 @@ class SkipButton: UIButton {
...
+            snapshot.centerYAnchor.constraint(equalTo: lottieView.centerYAnchor)
+        ])
+
+        animationView.clipsToBounds = true
+    }
+
+    func finishedTransition() {
+        guard let lottieView = subviews.first else { return }
+
+
```
bjtitus: I think there's an extra space here
```suggestion

```

podcasts/PlayPauseButton.swift
```diff
@@ -45,4 +48,24 @@ class PlayPauseButton: BasePlayPauseButton {
             animation.heightAnchor.constraint(equalTo: circleView.heightAnchor, multiplier: 0.48)
         ])
     }
+
+    // When using UIVIew.animate LottieAnimationView doesn't play nice with it
+    // Here we snapshot the view to provide a smooth animation
+    func prepareForAnimateTransition(withBackground: UIColor?) {
```
bjtitus: ❓ I don't think the `withBackground` color is used here, can it be removed?

podcasts/PlayerContainerViewController.swift
```diff
@@ -267,6 +271,22 @@ class PlayerContainerViewController: SimpleNotificationsViewController, PlayerTa
...
+        topSpaceToSafeArea.isActive = true
+        view.layoutIfNeeded()
+    }
+
+    func showTabsAndUnlockScrollView() {
+        mainScrollView.isScrollEnabled = true
+        topSpaceToHeader.isActive = true
+        topSpaceToSafeArea.isActive = false
+        view.layoutIfNeeded()
+    }
```
bjtitus: 🤔 Would it make sense to combine these or is that less clear?

```suggestion
    func updateTabsAndUnlockScrollView(isEnabled: Bool) {
            mainScrollView.isScrollEnabled = isEnabled
            topSpaceToHeader.isActive = isEnabled
            topSpaceToSafeArea.isActive = !isEnabled
            view.layoutIfNeeded()
    }
```

bjtitus (changes requested):
👏 Looks really good in testing! 💅

I left just a few pieces of code feedback.

podcasts/NowPlayingPlayerItemViewController.swift
leandroalonso: Done in 3328b3f34b1aa9d76be6a34ae9e4de666bed1dfd

podcasts/PlayPauseButton.swift
leandroalonso: Done in f1553062043db5be7db1c89f49a4dc411b17990d

podcasts/PlayerContainerViewController.swift
leandroalonso: Done in debf21cc39ffaf78438d7e377ef658ebccbcbb6a

leandroalonso:
@bjtitus thanks for the suggestions! Ready for another look!

bjtitus (approved):
:shipit: Looks good! 🙇

Issue #1795, opened 2024-05-30
Downloading streaming episodes with AVExportSession


SergioEstevao:
```[tasklist]
### Tasks
- [x] Implement downloading of audio of episode, using the same data/stream used by the player, #1787
- [x] Integrate exporting/downloading code with the Downloading system. #1792
- [ ] https://github.com/Automattic/pocket-casts-ios/pull/1819
```

PR #1904, opened 2024-07-11
Determine download content type from URL extension

bjtitus:
| 📘 Part of: https://github.com/Automattic/pocket-casts-ios/issues/1795 |
|:---:|

Fixes a streaming audio glitch when caching the episode using `AVExportSession`

## To test

* Navigate to an episode like [this one](https://pca.st/xn1iqlip)
* Play the episode
* Verify that no skipping happens within the first 10 seconds (other than the minor jump from stream to download)

## Checklist

### DIFF
--- a/podcasts/DownloadManager.swift
+++ b/podcasts/DownloadManager.swift
@@ -281,7 +281,7 @@ class DownloadManager: NSObject, FilePathProtocol {
             } else {
                 episode.autoDownloadStatus = Settings.downloadUpNextEpisodes() ? AutoDownloadStatus.autoDownloaded.rawValue :  AutoDownloadStatus.playerDownloadedForStreaming.rawValue
             }
-            episode.contentType = UTType.mpeg4Audio.preferredMIMEType
+            episode.contentType = UTType(episode.fileExtension())?.preferredMIMEType
             let downloadTaskUUID = episode.uuid
             downloadingEpisodesCache[downloadTaskUUID] = episode
             downloadAndStreamEpisodes.insert(downloadTaskUUID)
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/DownloadManager.swift
```diff
@@ -301,7 +301,7 @@ class DownloadManager: NSObject, FilePathProtocol {
             downloadAndStreamEpisodes.remove(downloadTaskUUID)
 
             if exportCompleted, let episode = dataManager.findBaseEpisode(uuid: episode.uuid) {
-                if let contentType = UTType.mpeg4Audio.preferredMIMEType {
+                if let contentType = episode.contentType {
```
bjtitus, line 304: This may be overwritten by something according to https://github.com/Automattic/pocket-casts-ios/commit/37f1cd9. Probably need to test against that. Instead, I could store the `preferredMIMEType`  from above?

danielebogo (commented):
I made a test using the episode you suggested and I didn't have any problems. I played the episode for few minutes and it was smooth. I also changed the network connection to see if I can stress it, but it still worked perfectly

SergioEstevao:
@bjtitus was this the fix that we tested and found out that using this basically disabled the AVExportSEssion?

bjtitus:
@SergioEstevao Yes, this ends up causing the AVExportSession to fail so it's bypassing the export process altogether and just streaming.

bjtitus:
I'm going to close this since we know it isn't really a solution to the problem. To summarize, the `AVAssetExportSession` in `MediaExporter` will fail because it only supports `m4a` audio and doesn't seem to be able to transcode from anything other than `mpeg4` from what we could tell.

In testing, we also discovered that _delaying_ the export session would delay the artifacts in the downloaded file which may indicate the problem lies with the simultaneous stream and export tasks rather than the content type.

PR #1909, opened 2024-07-12
Transcripts: show transcripts through a shelf button

leandroalonso:
| 📘 Part of: #1848 | 
|:---:|

This PR:

1. Add a shelf icon so transcripts can be accessed (and remove the tab)
2. Show the transcript in the player view
3. Add a close button to dismiss the transcript

https://github.com/user-attachments/assets/bfbd2e24-b4ee-4143-b016-dd5331b0ad65

### Flag disabled

1. With the `transcripts` flag **disabled**
2. Open the player
3. Tap "..." on the shelf (below the player buttons)
4. Scroll down
5. ✅ There should be no mention of Transcripts

### Flag enabled

1. Enable the `transcripts` flag
2. Open the player
3. ✅ Transcripts should be the 4th button, if it's not tap "..." and move it to one of the first 4 positions
4. Play an episode from the podcast Cautionary Tales
5. ✅ The transcript button should be enabled
6. Tap it
7. ✅ Transcript should show
8. Swipe down the transcript scrollbar
9. ✅ The whole player should disappear
10. Open the player again and tap the transcript icon
11. Tap the close button ("X" at the top-left corner)
12. ✅ Transcript should disappear

#### Episode without transcript

1. Enable the `transcripts` flag
2. Play an episode without transcript (eg.: The Daily)
3. ✅ Transcripts button should be disabled
4. Tap it
5. Nothing will happen

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/Enumerations.swift
```diff
@@ -399,15 +408,17 @@ extension PlayerAction: AnalyticsDescribable {
...
     }
 
     func canBePerformedOn(episode: BaseEpisode) -> Bool {
         switch self {
         case .starEpisode, .shareEpisode:
             return episode is Episode
-        case .addBookmark:
-            return isAvailable
+        case .transcript:
+            return FeatureFlag.transcripts.enabled
```
SergioEstevao, line 421: Shouldn't we check that if the episode has transcripts available?

leandroalonso: Sorry, I missed adding this testing step.

We don't need it, if a transcript is not available the button will still be there but disabled.

podcasts/TranscriptGradientView.swift
```diff
@@ -0,0 +1,52 @@
+import UIKit
+
+class TranscriptGradientView: UIView {
```
SergioEstevao: Nitpick Suggestion: Could this be a generic gradient view? to be used on other places?

podcasts/Enumerations.swift
SergioEstevao: Thanks for clarifying! I see that is disabled when none is available!

SergioEstevao (approved):
Thanks for implementing this! Excellent work on the animation.
Just left some suggestion about the gradient view but could be a follow up PR if justified.
I tested the following scenarios on iPhone and iPad and all working:
- FF disabled
- FF enabled
- FF enabled and no podcast available.

podcasts/TranscriptGradientView.swift
leandroalonso: Done!

I also went ahead and fix a bug that wasn't updating the gradient colors when the episode changed.

SergioEstevao: Thanks! will have a look again. Was the bug when the player went to the next episode when the previous episode finished?

leandroalonso: Yep

SergioEstevao (approved):
Grandients looking good, and updating when switching to a new episode! Thanks for taking the suggestion.
:shipit:

PR #1928, opened 2024-07-23
Show Notes: force the request to be made if an episode is missing

leandroalonso:
We have a few complaints about the episode notes not showing for some users. After looking at it, I believe this is what happens:

1. The show notes for a given episode are requested
2. The app can't find it in the cache, so it requests
3. However we were using `reloadRevalidatingCacheData`. If our `etag` or `lastModified` matches the server (which is weird for me but it seems it's happening), the request won't be made and the cached data will be returned again

This creates a scenario where the show notes will never update unless the `etag` or `lastModified` changes server-wise.

In order to fix it, I changed the policy to `reloadIgnoringLocalAndRemoteCacheData`. This shouldn't be a problem given we check the cache ourselves before calling the method that does it.

I also went ahead and added logs so we can have more info.

## To test

1. Run the app
2. Open any episode from a podcast you haven't interacted with before
3. ✅ `Show Info: requesting info for podcast X` should be logged
4. Open any other episode
5. ✅ `Show Info: returning cached data for episode X` should be logged

## Checklist

### DIFF
--- a/podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
+++ b/podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
@@ -71,38 +71,6 @@ actor ShowInfoCoordinator: ShowInfoCoordinating {
         try await requestShowInfo(podcastUuid: podcastUuid, episodeUuid: episodeUuid)
     }
 
-    func loadRawMetadata(
-        podcastUuid: String,
-        episodeUuid: String
-    ) async throws -> String? {
-        try await requestRawMetadata(podcastUuid: podcastUuid, episodeUuid: episodeUuid)
-    }
-
-    @discardableResult
-    func requestRawMetadata(
-        podcastUuid: String,
-        episodeUuid: String
-    ) async throws -> String? {
-        if let task = requestingRawMetadata[episodeUuid] {
-            return try await task.value
-        }
-
-        let task = Task<String?, Error> { [unowned self] in
-            do {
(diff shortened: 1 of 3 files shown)


podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
```diff
@@ -71,38 +71,6 @@ actor ShowInfoCoordinator: ShowInfoCoordinating {
         try await requestShowInfo(podcastUuid: podcastUuid, episodeUuid: episodeUuid)
     }
 
-    func loadRawMetadata(
```
leandroalonso, line 74: This is not used anywhere, so I deleted it to avoid confusion.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/55">7.69 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
danielebogo: Nice catch! It's probably some leftovers from the first implementation 🤔 ?

danielebogo (approved):
It worked as described!

PR #1947, opened 2024-07-26
Reimagine Sharing: Clip playback

bjtitus:
| 📘 Part of: #1910 |
|:---:|

Adds clip playback via a new `ClipPlaybackManager` which uses a separate `AVPlayer` instance.

https://github.com/user-attachments/assets/ceabce5f-2162-484e-8944-b214dff11c63

### Known Issues

* The playback head disappears behind the trailing trim handle
* The "Audio waveform" doesn't look right at the zoom level
* The playback position indicator is not as smooth as it might be
* Trim handles are slightly sluggish during playback, especially on older devices

## To test

> [!NOTE]

### Playback

* Play an episode
* Navigate to the player and then the "Share" button in the action bar
* Tap "Clip"
* Verify that tapping the Play button plays back the audio for the clip
* Tapping Pause should stop audio playback
* Moving the playhead should change the current play time when ending the gesture
* Moving the Trim Handles should change the start/stop point of playback
* Upon reaching the end, the clip should seek back to the beginning and pause

### Sharing 

* Tap the "Clip" button
* Tap "Copy Link"
* Verify that the link which was created mirrors the clip you expected to create. This clip should play in the web page.
* Try testing with changing the range of the clip in the editor

### Dismiss Clip Editor

* Navigate back to the Player
* Ensure that the episode playback position remains where it was when you began sharing the clip

### Episode Detail

* Navigate to an Episode Detail page for an episode which isn't playing but which you have played a bit of
* Tap the share button in the upper right
* Tap "Clip"
* Verify that the clip editor page comes up
* Verify that playback works but doesn't affect the currently playing episode

## Checklist

### DIFF
--- a/podcasts/Sharing/Clip/ClipPlaybackManager.swift
+++ b/podcasts/Sharing/Clip/ClipPlaybackManager.swift
@@ -11,6 +11,9 @@ class ClipPlaybackManager: ObservableObject {
     @Published var currentTime: TimeInterval?
     @Published var duration: TimeInterval = 0
 
+    private let normalPlaybackManager = PlaybackManager.shared
+    private let downloadManager = DownloadManager.shared
+
     private var avPlayer: AVPlayer?
     private var timeObserverToken: Any?
     private var cancellables = Set<AnyCancellable>()
@@ -23,9 +26,9 @@ class ClipPlaybackManager: ObservableObject {
             avPlayer = nil
         }
 
-        PlaybackManager.shared.pause()
+        normalPlaybackManager.pause()
 
-        guard let playerItem = DownloadManager.shared.downloadParallelToStream(of: episode) else {
+        guard let playerItem = downloadManager.downloadParallelToStream(of: episode) else {
             return
         }
 
@@ -37,7 +40,7 @@ class ClipPlaybackManager: ObservableObject {
 
         let playbackCMTime = CMTime(seconds: playbackTime, preferredTimescale: .audio)
 
-        PlaybackManager.shared.activateAudioSession(completion: { [weak self] activated in
+        normalPlaybackManager.activateAudioSession(completion: { [weak self] activated in
             self?.avPlayer?.seek(to: playbackCMTime)
             self?.avPlayer?.play()
         })
--- a/podcasts/Sharing/Clip/ClipPlaybackManager.swift
+++ b/podcasts/Sharing/Clip/ClipPlaybackManager.swift
@@ -1,0 +1,123 @@
+import AVFoundation
+import PocketCastsDataModel
+import Combine
+import SwiftUI
+
+class ClipPlaybackManager: ObservableObject {
+
+    static var shared = ClipPlaybackManager()
+
+    @Published var isPlaying: Bool = false
+    @Published var currentTime: TimeInterval?
+    @Published var duration: TimeInterval = 0
+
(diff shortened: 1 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/56">7.70</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/Sharing/Clip/ClipPlaybackManager.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    @ObservedObject var clipTime: ClipTime = ClipTime(start: 0, end: 0)
+
+    func play(episode: BaseEpisode, clipTime: ObservedObject<ClipTime>) {
+        if avPlayer != nil {
+            stop()
+            avPlayer = nil
+        }
+
+        PlaybackManager.shared.pause()
```
leandroalonso: Shall we inject this?

leandroalonso (changes requested):
@bjtitus tested and it works as described, there are a few things worth mentioning:

1. When tapping play and closing the screen, playback don't stop. We need to ensure that leaving the screen stops the playback;
2. The playhead can be out of bounds with the trimmer: is that expected? Taking as an example the Photos app: the playhead is always contained inside the trimmer bounds.
3. I wonder if we should follow the Photos behavior: when moving the trimmer, any playback stops and the playhead disappears. After the trimming is done, the playhead re-appears at the beginning of the trimmed part. This should also fix the "Trim handles are slightly sluggish during playback, especially on older devices" known issue
4. When moving the playhead for a streaming episode it moves back to the previous position while the seek is being performed (it would be better to stay at the moved position)

None of those are blockers except (1)

bjtitus:
**1**

> When tapping play and closing the screen, playback don't stop

✅ Fixed

**2**
> The playhead can be out of bounds with the trimmer

Yep, I have code for this but just haven't pushed it yet. I'll do that as a separate PR.

**3**
> I wonder if we should follow the Photos behavior: when moving the trimmer, any playback stops and the playhead disappears.

This is an interesting idea. I'll take a look at this alongside 2. It certainly would help a bit with performance, which I think is mostly a bunch of SwiftUI updates triggered by the moving playhead. Would this make it harder to line up the exact clip position, though?

**4**
> When moving the playhead for a streaming episode it moves back to the previous position while the seek is being performed

Are you talking about the quick jump? I was looking a bit and I think I will have to delay the update while seek is happening.

https://github.com/user-attachments/assets/1e5fa956-64d9-4eaf-b90e-6a8b82055a84

bjtitus:
@leandroalonso I added the seek ignoring logic for number 4 in https://github.com/Automattic/pocket-casts-ios/pull/1947/commits/601a27fa230d88ef3124b69a4f372b78dbcc7500

leandroalonso (approved):
✅ Tested closing the screen while playback is ongoing
✅ Tested moving the playhead while streaming

PR #1955, opened 2024-07-29
Implement data retriever with cache

SergioEstevao:
| 📘 Part of: #1848  | 
|:---:|

Implement transcripts local storage.

The cache is implemented using an URL cache with back of local storage path.

Some implementation considerations:

- We are using an ephemeral session so no other HTTP cache will be used
- We only fill the cache if there is a proper 200 response. 
- If we get a 200 we cache the response should we handle scenarios were a transcript is updated later on with new text? Should eject transcripts based on time or any HTTP headers?
- I'm only caching the data object returned from HTTP, in terms of use from my tests looks is good enough. We can try to cache the full parsed TranscriptModel object later on.

## To test

 Start the app
- Ensure that you have the `transcripts` and `newShowNotesEndpoint` FF enabled

### Test regular online scenario

- Start playing an episode with transcript supports. Ex:  One of this list: https://lists.pocketcasts.com/b852a088-ccee-4e4b-a3c5-4bb7fd700a02
- Open the full screen player
- Tap on the transcript shelf button 
- Check if the transcripts shows correctly
- Close the transcripts view
- Open it again
- Check that no loading indicator is view or if show goes away quickly
- Start playing another episode with transcripts
- Tap on transcript shelf button. You should see the loading indicator
- Close transcript
- Open the transcript again and see if it's quickly load, no loading view
- Start playing the previous episode that should be already cached
- Tap on transcript shelf
- Check if the the transcript is show quickly

### Test offline scenario

- Go offline
- For the previous podcasts you opened, open the episode again and see if you can see the transcript

### Test auto-downloads

- Enable auto-downloads for UpNext Queue
- Add some episodes with Transcript to UpNext Queue
- Wait for the downloads to happen
- Go offline
- Open one of the episode donwloaded
- Check that you can see the transcript

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso (changes requested):
@SergioEstevao when downloading an episode, Transcript is not available while offline. This is the default behavior for chapters, show notes and episode artwork.

> If we get a 200 we cache the response should we handle scenarios were a transcript is updated later on with new text? Should eject transcripts based on time or any HTTP headers?

Like the previous show notes, you can return whatever you have in cache and do a new request if online. However, we'll need to change the method and use [AsyncStream](https://developer.apple.com/documentation/swift/asyncstream) I believe.

SergioEstevao:
@leandroalonso so the idea is when the user downloads an episode ( or auto-downloads) we should also fetch the transcript?

leandroalonso:
@SergioEstevao yep!

SergioEstevao:
@leandroalonso ready for another look I added mechanisms to cache the transcript whenever the episode is downloaded and also a way to update/refresh the cache.

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
```diff
@@ -59,7 +64,10 @@ actor ShowInfoCoordinator: ShowInfoCoordinating {
         guard let transcripts = metadata?.transcripts else {
             return []
         }
-
+        if let transcript = Episode.Metadata.TranscriptFormat.bestTranscript(from: transcripts),
```
leandroalonso: This doesn't look good 😬

What do you think about moving that to `transcriptDataRetriever`? Or is there any reason why it is on the episode?

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
```diff
@@ -59,7 +64,10 @@ actor ShowInfoCoordinator: ShowInfoCoordinating {
...
+           let url = URL(string: transcript.url) {
+           let _ = try? await transcriptDataRetriever.loadTranscript(url: url)
```
leandroalonso: I think this fundamentally changes something:

1. This method didn't download the transcript before. So it was used in `checkTranscriptAvailability()` to just check if there was, or wasn't, a transcript available ¶
2. However, we did change that now. It means that we will download transcripts for all episodes, no matter whether the user wants to check them or not

This is not wrong per se. But, do we want it? Do we want to load transcripts even if, maybe, some users will never actually open them?

I also wonder if we should change this method to return the loaded transcript.

Modules/Server/Sources/PocketCastsServer/Public/Helpers/URLRequest+RefreshIfNeeded.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public extension URLRequest {
+
+    mutating func setRefreshHeadersUsing(cachedResponse: CachedURLResponse) {
```
leandroalonso: Maybe rename it to `setEtagAndLastModifiedHeaders` to be explicit about what it does?

leandroalonso (changes requested):
✅ Tested downloading an episode and playing while offline and transcripts are shown

I left some comments about some changes in behavior in methods and whether is something we want or not.

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
SergioEstevao: Good suggestion, let me have a look into this.

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
SergioEstevao: My understanding is that we wanted to have the transcript file the moment we downloaded an episode, If I understood correctly the codebase this was the place to be used. 
Do you only get shown notes and artwork info if the user opens the episode screen?
Another advantage of this is the transcript is just ready when the user taps on it.

Modules/Server/Sources/PocketCastsServer/Public/Helpers/URLRequest+RefreshIfNeeded.swift
SergioEstevao: Ok

SergioEstevao: Done here: https://github.com/Automattic/pocket-casts-ios/pull/1955/commits/1cffb0c49b6567f3d7b266d6c767bd281b406b87

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
SergioEstevao: Done here: https://github.com/Automattic/pocket-casts-ios/pull/1955/commits/a14b2ce979d57dd37026dc552cc1554d1ac8f47b

SergioEstevao:
@leandroalonso updated the code, ready for another look.

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
SergioEstevao: Done here: https://github.com/Automattic/pocket-casts-ios/pull/1955/commits/c941afbb889fbe628af01062f9b8c525acc4a1c0

Modules/Server/Sources/PocketCastsServer/Public/Helpers/URLRequest+RefreshIfNeeded.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public extension URLRequest {
+
+    mutating func setEtagAndLastModifiedHeaders(cachedResponse: CachedURLResponse) {
+        if let etag = cachedResponse.response.etag {
```
leandroalonso, line 6: While testing different podcasts I found an issue that is not directly related to this PR but it will affect this:

http://www.stephen-gurnett.org/unlikely-objects/2017/2/1/fylkgu09vo4g3zhofzyv9gucz64alt

Basically, when testing some podcasts (like "If Books Could Kill") cache never worked, even though there's a `etag` there. The issue is that it's written `Etag` not `ETag` as we expect.

I believe we should implement this fix so we don't run in cases like this.

Not in this PR though.

podcasts/TranscriptsDataRetriever.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    private func loadTranscriptFromServer(_ url: URL, previousResponse: CachedURLResponse? = nil) async throws -> String? {
+        if let task = dataRequestMap[url] {
+            return try await String(data: task.value, encoding: .utf8)
+        }
+        FileLog.shared.addMessage("Transcripts Data Retriever: requesting transcript data \(url)")
+
+        let task = Task<Data, Error> {
+            do {
+                var request = URLRequest(url: url, cachePolicy: .reloadIgnoringLocalAndRemoteCacheData)
```
leandroalonso: By using `reloadIgnoringLocalAndRemoteCacheData` this request sends a `cache-control: no-cache`. In my testing this doesn't make any difference but I think we should not send it, so maybe `reloadRevalidatingCacheData` might be a better choice here.

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Episode.swift
```diff
@@ -206,6 +206,43 @@ public class Episode: NSObject, BaseEpisode {
             public let url: String
             public let type: String
             public let language: String?
+
+            public var transcriptFormat: TranscriptFormat? {
+                return TranscriptFormat(rawValue: self.type)
+            }
+        }
+
+        public enum TranscriptFormat: String {
```
leandroalonso: What do you think about moving this outside of `Episode`?

I feel like `Episode` and `Transcript` are not **directly** related and we're adding a lot of knowledge about Transcripts inside an `Episode` (like file extension, best transcript, etc).

Maybe `TranscriptManager` is a better place?

leandroalonso (commented):
@SergioEstevao working as described!

I tested many podcasts and found some issues related to how we read headers. I think it would be good to change but not on this PR.

One last thing is that I believe we're now coupling `Episode` and `Transcript`. Left a few comments, let me know what do you think!

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
leandroalonso: >Do you only get shown notes and artwork info if the user opens the episode screen?

Before, yes. Show notes would only display if downloading or checking them. Artwork info is also only cached (the image itself) if downloading or playing the episode.

This has changed though because the same endpoint that returns the show notes also return artwork, transcript, etc, etc.

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Episode.swift
SergioEstevao: Are you suggesting to move `TranscriptFormat` only right?

leandroalonso: Yep!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/56">7.70</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/DataModel/Sources/PocketCastsDataModel/Public/Model/Episode.swift
SergioEstevao: Done here: https://github.com/Automattic/pocket-casts-ios/pull/1955/commits/fcf6f156be0a05eca84e1a1526b05e64c43e3977

Decided to move it to a separate file.

leandroalonso:
@SergioEstevao any thoughts about the other 2 comments? (the headers and the `cachePolicy`)

podcasts/Episode Info Coordinator/ShowInfoCoordinator.swift
SergioEstevao: Ok now we only download transcripts metadata when opening the episode, but if we download the episode we will also cache the transcript locally so it can be opened even if offline.

Modules/Server/Sources/PocketCastsServer/Public/Helpers/URLRequest+RefreshIfNeeded.swift
SergioEstevao: So the plan is another PR is to try to fetch the elements by using multiple possible variations of the tag? I wonder if we could copy the dictionary and lowercase all the keys?

podcasts/TranscriptsDataRetriever.swift
SergioEstevao: I think because I'm using the ephemeral configuration for the session, it does not make a difference but I can switch to the default.

SergioEstevao:
> any thoughts about the other 2 comments? (the headers and the cachePolicy)

@leandroalonso Sorry I missed those two, I just answered them, tell me what do you think.

podcasts/TranscriptsDataRetriever.swift
SergioEstevao: Changed here: https://github.com/Automattic/pocket-casts-ios/pull/1955/commits/a919037a2e0d26277b1ff416f0a3c48e47ffc2f2

leandroalonso (approved):

PR #1969, opened 2024-07-31
Implement ExPlat

### LINKED ISSUE #1987: Integrate A/A test trough ExPlat

danielebogo:
| 📘 Part of: #1986
|:---:|

Fixes #1987 

This PR integrates ExPlat and runs an AA test to double check our implementation works as expected.

## To test

- CI must be 🟢 
- Check `p1722974042562459/1722961411.090999-slack` to set your bookmarklet
- Make sure your A8C proxy is set correctly and is running
- You must login with your A8C account
- Set a breakpoint into `ExPlatService` line `78` -> `let decoder = JSONDecoder()`
- Run the app
- When the app stops at the breakpoint
- Check if response request URL: it should be `https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0/assignments/pocketcasts?_locale=en&experiment_names=pocketcasts_paywall_ios_aa_test&anon_id=YOUR_ID`
- Print the JSON result with `po String(data: (try? JSONSerialization.data(withJSONObject: (try? JSONSerialization.jsonObject(with: data))!, options: [.prettyPrinted, .sortedKeys]))!, encoding: .utf8)!`, checking if the body contains this:
```
"{\n  \"debug\" : {\n    \"backend_aa_result\" : \"request not sampled\"\n  },\n  \"ttl\" : 60,\n  \"variations\" : {\n    \"pocketcasts_paywall_ios_aa_test\" : null\n  }\n}"
```

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/Analytics/Adapters/Tracks/TracksAdapter.swift
```diff
@@ -111,12 +117,22 @@ private extension TracksAdapter {
...
+        if let userId = ServerSettings.userId {
+            tracksService.switchToAuthenticatedUser(withUsername: nil, userID: userId, skipAliasEventCreation: false)
+        } else {
             tracksService.switchToAnonymousUser(withAnonymousID: anonymousUUID)
-            return
         }
+        reloadABTest()
+    }
+
+    func reloadExPlat() {
```
danielebogo, line 128: We need to reload ExPlat after the TrackService is created, as we need to force setting up the `platform`.

SergioEstevao (approved):
After some going around to setup the environment I was able to test it correctly!

:shipit:

howarding:
Thanks for setting up the code change for the A/A test! I don't have front-end coding experience so I might not be able to give correct review on this PR. I think @necronet or @samacart should have more experience on reviewing the coding changes for experiments.

podcasts/ABTest/ABTestProvider.swift
```diff
@@ -0,0 +1,27 @@
...
+
+    func start() async {
+        await withCheckedContinuation { continuation in
+            let experiments = ABTest.allCases.map { $0.rawValue }
+            if experiments.isEmpty {
+                return continuation.resume()
+            }
+
+            ExPlat.shared.register(experiments: experiments)
+            ExPlat.shared.refresh {
```
aaronyan, line 24: It seems that assignments are batched all at once, am I understanding that correctly? This requires A/B tests to use exposure events to reduce data collection noise.

Is there an option to call for the assignment of a single experiment?

podcasts/ABTest/ABTestProvider.swift
```diff
@@ -0,0 +1,27 @@
+import AutomatticTracks
+
+struct ABTestProvider: ABTestProviding {
+    static let shared = ABTestProvider()
+
+    func variation(for abTest: ABTest) -> Variation {
+        ExPlat.shared.experiment(abTest.rawValue) ?? .control
```
aaronyan, line 12: For `control` variants, `GET /assignments` [endpoint](https://app.swaggerhub.com/apis/yanir/experiments/0.1.0#/assignment/get_assignments__platform_) returns `null`. Glad to see this is checking for that and managing the `.control` variant.

Keep in mind that both users that assigned `control` and users that are not eligible for the experiment both are assigned `null` variations. We don't differentiate, so please do not use the `variation` value in future Tracks event  props.

It may be worth adding a comment here so others know that users assigned to `control` and users ineligible for the experiment are given the assignment `null` which corresponds to them experiencing the `control` variation.

aaronyan (commented):
Thanks for the implementation! I'll let others more experienced with iOS code to provide official approvals, but I've left some notes and thoughts.

The `ExPlat` implementation in Automattic-Tracks-iOS can be found [here](https://github.com/Automattic/Automattic-Tracks-iOS/blob/8cf6461cf4dfb0f0f48524ddd6376e5ba1a553b1/Sources/Experiments/ExPlat.swift#L8) and is a helpful resource when reviewing this!

podcasts/ABTest/ABTestProvider.swift
necronet: Yes this is batching indeed all the experiment available in `ABTest` enum at once.

But I don't think the `Explat` tracks library has a way to do single experiment, they leverage the fact that we can send multiple experiments (comma separated) and get the assignment for each one. As I can see in the [test code](https://github.com/Automattic/Automattic-Tracks-iOS/blob/8cf6461cf4dfb0f0f48524ddd6376e5ba1a553b1/Tests/Tests/ABTesting/ExPlatServiceTests.swift#L11) this is how they intended to do it.

podcasts/ABTest/ABTest.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+enum ABTest: String, CaseIterable {
+    case pocketcastsPaywallAATest = "pocketcasts_paywall_ios_aa_test"
```
necronet, line 4: Make sense! we may also have an enum for ExposureEven based on the [experiment design](https://experiments.a8c.com/experiments/21898-pocketcasts-paywall-ios-aa-test/overview)

```enum ExposureEvent: String, CaseIterable {
    case exposureEventA = "pcios_plus_promotion_shown"
}```

File: podcasts/Analytics/Adapters/Tracks/TracksAdapter.swift [hunk 5]
```diff
@@ -111,12 +117,22 @@ private extension TracksAdapter {
...
+        abTestProvider.reloadExPlat(platform: tracksService.platform, oAuthToken: nil, userAgent: nil, anonId: anonymousUUID)
```
necronet, line 129: oAuthToken nil, and also userAgent why ?

necronet (changes requested):
Overall this is a very solid changes to add Explat into PC ios great work, couple of details here and there!

podcasts/ABTest/ABTest.swift
aaronyan: What did you have in mind for having an exposure event as an enum here?

We want to make sure for all experiments, we do this in the following order:
- Assign A/A test assignment
- Call exposure event specified in experiment profile

For the A/A test, we just need to make sure that we fire the Tracks event `pcios_plus_promotion_shown` for both users assigned `control` and `treatment`.

danielebogo: @aaronyan we use `enum` as it's the best way to represent it for a few reasons:
• Conform to `CaseIterable` protocol, which allows us to easily iterate trough all cases and read the raw value for each case (which is in this case the original test name)
• Can not be instantiated
• Easy to filter while iterating
• Easy to use while checking what's the variant for a specific case: see `func variation(for abTest: ABTest) -> Variation`
• Consistency with WP, Woo and DayOne as all platforms use the same approach

podcasts/Analytics/Adapters/Tracks/TracksAdapter.swift
danielebogo: > oAuthToken nil, and also userAgent why ?

If I'm not wrong these 2 parameters are not needed. Besides:

• `oAuthToken` is the `wpComToken ` and usually set the `Bearer` token while reloading `Tracks` with these 2 methods:
```
- (void)switchToAuthenticatedUserWithUsername:(NSString *)username userID:(NSString *)userID wpComToken:(NSString *)token skipAliasEventCreation:(BOOL)skipEvent;
- (void)switchToAuthenticatedUserWithUsername:(NSString *)username userID:(NSString *)userID anonymousID:(NSString *)anonymousID wpComToken:(NSString *)token skipAliasEventCreation:(BOOL)skipEvent;
```  
These 2 specific methods are used by WP iOS and Woo iOS only (which makes sense as those are tight with WP), while DayOne doesn't and uses the same method we use to reload Tracks for a logged in user (which doesn't reload ExPlat underneath)
```
- (void)switchToAuthenticatedUserWithUsername:(NSString *)username userID:(NSString *)userID skipAliasEventCreation:(BOOL)skipEvent;
```

So both DayOne and PocketCasts use the `abTestProvider.reloadExPlat(platform: tracksService.platform, oAuthToken: nil, userAgent: nil, anonId: anonymousUUID)` to only set the platform name, the constructor of the `TracksService` initialize ExPlat with the default `wpiOS` platform name.
Day One: `ExPlat.configure(platform: "dayone", oAuthToken: nil, userAgent: nil, anonId: DOSharedSettings.default.tracksUUID)`

• The user agent  is a private property of the `TracksService` instance. I set it to nil as DayOne did.

CC @aaronyan

podcasts/ABTest/ABTestProvider.swift
leandroalonso: > We don't differentiate, so please do not use the variation value in future Tracks event props.

If I got it right it means that we should **never** send `variation` as a property in an event, right? If we want to set the property what would be the correct way of doing it?

leandroalonso (approved):
Tested and I got the assignment.

However, I believe the "manual assignment" (with the bookmarklet) will never work as our account is not tied to WP.com. So when we make the request there's no `oAuthToken`, thus the backend doesn't know which user it is.

podcasts/Analytics/Adapters/Tracks/TracksAdapter.swift
danielebogo: Just [an addition](https://github.com/Automattic/pocket-casts-ios/pull/1969#pullrequestreview-2233319609) about what @leandroalonso mentioned below regarding the token

podcasts/ABTest/ABTestProvider.swift
aaronyan: The way you are setting it is the correct way. Just be careful with interpreting those Tracks events, as a12s may think that this is the variation assigned vs the variation experienced.

podcasts/Analytics/Adapters/Tracks/TracksAdapter.swift
aaronyan: > So both DayOne and PocketCasts use the abTestProvider.reloadExPlat(platform: tracksService.platform, oAuthToken: nil, userAgent: nil, anonId: anonymousUUID) to only set the platform name

Neat, good to know!

podcasts/ABTest/ABTest.swift
necronet: Yes, @aaronyan agree, and the exposure event do come after the assignment phase, just wanted to point out if we're gonna have a similar structure (or if there is one in place) for it!

necronet (approved):
LGTM

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/57">7.71</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

Issue #23, opened 2022-07-11
Watch: migrate the Watch app to SwiftUI


pocketcasts:
## Phase 1 - Isolated screens
- [x] EffectsInterfaceController
- [x] MenuController (**Edit** This item is TBD see below on the comments)
- [x] FilesOrderInterfaceController
- [x] PodcastEpisodeOrderController
- [x] PodcastsOrderInterfaceController

## Phase 2 - Episode List Prep
- [x] Reference each other and should be merged together or account for the navigation
    - [x] EpisodeController
    - [x] PodcastEpisodeListController 

## Phase 3 - Episode Lists and Inter-dependencies
- [x] PodcastsInterfaceController
- [x] DownloadInterfaceController
- [x] FilterInterfaceController
- [x] FilesInterfaceController
- ~[ ] Reference each other and should be merged together or account for the navigation~
    - [x] UpNextInterfaceController
    - [x] NowPlayingInterfaceController

## Phase 4 - Remaining views (Should be done sequentially) 
- [x] FiltersInterfaceController - https://github.com/Automattic/pocket-casts-ios/pull/1524
- [x] InterfaceController
    - [x] handle state restoration 
- [x] SourceInterfaceController #1800

## Phase 5 - Review unknown elements
- [x] Static Notification Interface Controller Scene #1800
- [x] NotificationController (Empty view) #1800
- [x] Clean up any remaining WatchOS 7 migration warnings #1800
- [ ] Ensure `Screenshot Automation Watch` still runs

SergioEstevao:
Should this project also include migrating the `.xcproj` file to have a single Watch OS project and also get rid of the Storyboards files and the startup cycle of the app?

It looks this was changed on XCode 14: https://stackoverflow.com/questions/52687924/ios-watchos-what-is-the-difference-between-watchkit-app-and-watchkit-app-exten

PR #1989, opened 2024-08-06
Disable SPM caching in CI

mokagio:
This is a temporary workaround to address the SPM-resolution-related build failure in the Prototype Build step, example https://buildkite.com/automattic/pocket-casts-ios/builds/7634#019113ca-d4bd-4838-b93a-aae0e4a0528f.

The long term solution will be one of the two options proposed in https://github.com/Automattic/pocket-casts-ios/issues/1990

While touching the CI scripts to implement the workaround, I also:

- Centralized shared setup steps in a dedicated script
- Removed an unnecessary `install_cocoapods` call
- Upgraded the CI toolkit
- Explicitly set the `agents: queue: mac` node in a few steps, to enable us to remove the implicit `agents: queue: mac` setting from the pipeline configuration at the Buildkite level, in turn letting us use a standard upload command which supports shared pipeline vars.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:
<details><summary>Pocket Casts Prototype Build📲 You can test the changes from this Pull Request in <b>Pocket Casts Prototype Build</b> by scanning the QR code below to install the corresponding build.</summary>
<table>
<tr>
  <td><b>App Name</b></td><td>Pocket Casts Prototype Build Pocket Casts Prototype Build</td>
</tr>
<tr><td><b>Build Number</b></td><td>pr1989-2fd53e3</td></tr>
<tr><td><b>Version</b></td><td>7.70</td></tr>
<tr><td><b>Bundle ID</b></td><td>au.com.shiftyjelly.podcasts.prototype</td></tr>
<tr><td><b>Commit</b></td><td>2fd53e3a847593e519d2a16ef86ff51dc40af696</td></tr>
<tr><td><b>App Center Build</b></td><td><a href='https://install.appcenter.ms/orgs/automattic/apps/Pocket-Casts-Prototype-Builds/releases/23'>Pocket Casts - Prototype Builds #23</a></td></tr>
</table>
<em>Automatticians: You can use our internal self-serve MC tool to give yourself access to App Center if needed.</em>
</details>

.buildkite/commands/build.sh
```diff
@@ -1,17 +1,6 @@
 #!/bin/bash -u
 
-echo "--- :swift: Installing Swift Package Manager Dependencies"
-install_swiftpm_dependencies
-
-echo "--- Setup Ruby tooling"
-
-export LANG=en_US.UTF-8
-export LC_ALL=en_US.UTF-8
```
spencertransier, line 9: Do these need to be added to the `shared_setup.sh` script?

.buildkite/commands/shared_setup.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "--- :ruby: Setting up Ruby tools"
+install_gems
+
+echo "--- :cocoapods: Setting up Pods"
+install_cocoapods
+
+echo "--- :swift: Installing Swift Package Manager Dependencies"
+# Disabled until the xcodeproj will be in a folder other than the project root
+#
+# install_swiftpm_dependencies
```
AliSoftware: Nitpick because I found it potentially a bit confusing what was disabled vs what to revert once the fix will be ready

```suggestion
# Temporarily disable `install_swiftpm_dependencies` and calling `xcodebuild -resolvePackageDependencies`
# command directly instead until we move the `.xcodeproj` to a folder other than the project root to workaround
# an issue with `xcodebuild` not using the right `Package.resolved` file (`*.xcworkspace/*` vs `*.xcodeproj/*`)
#
# install_swiftpm_dependencies
```

.buildkite/commands/build.sh
AliSoftware: Good question. I wonder why we needed to export those… and if it's still needed though 🤔 

I know that in general, those special env vars sometimes have impacts on some third-party toolds relying on the locale and that it's thus a good practice to set those to `en_US.UTF-8` to fix various issues with those… But I have no idea:
1. which of the tools we use might have required for us to add those in the past (would have been nice to add a comment back then to leave breadcrumbs 😓
2. and thus if it's still necessary (maybe we don't use the tool that had an issue with this anymore, or maybe the CI agents we run our `build.sh` command on now has those env vars already set to `en_US.UTF-8` at the system level?)

AliSoftware (approved):
LGTM, with only nitpicks.

Approving to unblock, providing that we don't forget to follow-up with https://github.com/Automattic/pocket-casts-ios/issues/1990 or any alternative long-term solution for it.

.buildkite/commands/build.sh
mokagio: They were legacy settings [coming from the previous automation setup](https://github.com/Automattic/pocket-casts-ios-private/commit/cffc29815892339c79d6d9b69947d52fe04d9420). Unfortunately I don't have additional context on the previous automation. 

Those exports triggered some old memories from other projects though, so if I had to guess I'd say they are related to old CocoaPods requirements, e.g. https://github.com/CocoaPods/CocoaPods/issues/6333

PR #2000, opened 2024-08-07
Transcripts fix crash on empty search

### LINKED ISSUE #1999: Transcripts Search crashes when pressing up and down on a empty result
### Description

Transcripts Search crashes when pressing up and down on a empty result

1. Enable Transcripts FF
2. Start playing an episode with transcripts. Ex: any Cautionary Tales episode
3. Open the full screen player
4. Tap on transcripts icon
5. Tap on Search
6. Search for a word non-existing on the transcript
7. Tap on up or down
8. See crash

_No response_

iPad Mini iOS 17.5

SergioEstevao:
| 📘 Part of: #1848  
|:---:|

Fixes #1999 

I updated the code with a `guard` to prevent the crash detected when accessing empty search results, and I also added code to disable the up-down buttons when no search results are available.

## To test

1. Start the app
2. Ensure that you have the `transcripts` and `newShowNotesEndpoint` FF enabled
3. Start playing an episode with transcript supports. Ex:  One of this [list](https://lists.pocketcasts.com/b852a088-ccee-4e4b-a3c5-4bb7fd700a02)
4. Open the full screen player
5. Tap on transcripts icon
6. Tap on Search
7. Search for a word non-existing on the transcript
8. Check that the up-down buttons are disabled
9. Write now a word that exists in the transcript
10. See that the up-down buttons are enabled again

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/TranscriptsViewController.swift
```diff
@@ -541,6 +542,9 @@ extension TranscriptsViewController: TranscriptSearchAccessoryViewDelegate {
     }
 
     private func processMatch() {
+        guard !searchIndicesResult.isEmpty, currentSearchIndex >= 0, currentSearchIndex < searchIndicesResult.count else {
+            return
+        }
```
leandroalonso: I don't believe we need al those checks, we can do it like this:


```suggestion
        if searchIndicesResult.isEmpty {
            return
        }

```

Also not using a `guard` negation — we've been avoid doing those.

podcasts/TranscriptSearchAccessoryView.swift
```diff
@@ -63,6 +63,11 @@ class TranscriptSearchAccessoryView: UIInputView {
         contentInsets: .init(top: 8, leading: 8, bottom: 8, trailing: 8)
     )
 
+    func enableUpDownButtons(_ enable: Bool) {
```
leandroalonso, line 66: It was intentional to not do this because the search on other parts of iOS doesn't handle disabling the buttons.

I don't think it's bad having though.

However, when I first open the search the buttons are enabled, they should be disabled if we want this behavior to be correct.

leandroalonso (changes requested):

podcasts/TranscriptSearchAccessoryView.swift
SergioEstevao: Updated here: [68a71f3](https://github.com/Automattic/pocket-casts-ios/pull/2000/commits/68a71f30977360d288e50254655404e124231845)

leandroalonso:
@SergioEstevao since I renamed `TranscriptsViewController` to `TranscriptViewController` there's a merge issue here, I believe we should fix that before I do a final review

SergioEstevao:
@leandroalonso done here: https://github.com/Automattic/pocket-casts-ios/pull/2000/commits/35c457d3a12f9895154d8d9f2b0e7df60276c8f2

leandroalonso:
@SergioEstevao it looks like we have more merge conflicts due to my PR, sorry 😬

pachlava:
While I was testing something else, on a different branch, I noticed that searching for emojis also results in crash. Not sure if it's the same crash as here, since technically emojis search is not an empty search.

pachlava:
>Not sure if it's the same crash

Oh, I now realize it's the same kind of crash. Initially, I understood current crash as searching for empty string, whereas this meant navigating across empty search results. Never mind my comments here.

SergioEstevao:
@leandroalonso conflicts sorted out.

leandroalonso (approved):

PR #1982, opened 2024-08-02
Transcripts: search logic

leandroalonso:
| 📘 Part of: #1848 | 
|:---:|

| iPhone | iPad |
| ------- | ---- |
| Simulator Screenshot - iPhone 15 - 2024-08-02 at 15 24 31 | Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-08-02 at 15 24 57 |

This adds the search logic inside a single transcript. I implemented a version of the [Knuth–Morris–Pratt algorithm](https://en.wikipedia.org/wiki/Knuth–Morris–Pratt_algorithm) to be as performant as possible.

## To test

Before testing, ensure **transcripts** feature flag is enabled. It's also better to test on a real device. If you want to test on the simulator, use the simulator keyboard, not your Mac's.

### Testing UI componentes

1. Run the app
2. Play any episode with transcripts, [there's a list here](https://lists.pocketcasts.com/b852a088-ccee-4e4b-a3c5-4bb7fd700a02)
3. Tap the Transcript button and then tap "Search"
4. Scroll down to the end of the transcript
5. ✅ The scroll view should end when the keyboard starts
6. Tap "Done"
7. ✅ The scroll view should get back to its original position
8. Test the same on iPad, but when the keyboard is appearing toggle between portrait/landscape and ensure the scroll view is adjusted accordingly

### Testing search

1. Run the app
2. Play any episode with transcript, [there's a list here](https://lists.pocketcasts.com/b852a088-ccee-4e4b-a3c5-4bb7fd700a02)
3. Tap the transcript button and then tap "Search"
4. Enter any word
5. ✅ The app scrolls to the first results of your search
6. ✅ The input accessory should display "1 of X" where X is the number of results
7. ✅ The first result should have a white highlight color and black text color
8. ✅ The rest of the results should have a white background color with opacity
9. Tap the chevron down or up to toggle between results
10. ✅ The app should scroll to the next result updating which one is highlighted
11. ✅ When you reach the last one and tap the chevron down, it should go to the first
12. ✅ When showing the first result and you tap the chevron up, it should go to the last
13. ✅ The label should update indicating which result you see, eg.: "3 of 10"
14. Tap "search" on the keyboard
15. ✅ The keyboard should disappear, but the results are still highlighter and the input accessory is at the bottom of the screen
16. ✅ Ensure you can still use the chevrons up and down
17. Tap "Done"
18. ✅ All highlights are removed
19. Tap "Search" again and search for something
20. Tap "Done"
[...]

PR #2014, opened 2024-08-08
Transcripts: block scroll when dismissing keyboard

leandroalonso:
| 📘 Part of: #1848 |
|:---:|

As reported by @pachlava [here](https://github.com/Automattic/pocket-casts-ios/pull/1982#issuecomment-2274295556) we have an issue where tapping outside the keyboard when the search is active activates the scroll on the transcript:

https://github.com/user-attachments/assets/57df34e0-a361-48cb-b870-2500f5075fa6

After some investigation, I was able to pinpoint that the issue is with the `TranscriptViewController.inputAcessoryView`. If we set it to `nil` this doesn't happen. However, it prevent it from appearing when you tap "Search".

I wasn't able to find a proper solution, so what I ended up doing was to block the scrollview from scrolling during keyboard hiding.

This causes a small glitch though:

https://github.com/user-attachments/assets/25202b8d-0db2-4b47-b1b5-bef7225c37aa

I still think this is better than having the issue which normally scrolls far away from where the user was.

----

I did some extra work on this and was able to found a way to fix the issue without the glitch.
From my investigation the issue was happening when we changed the bottom inset of the text view when the keyboard animated in/out.
My solution saves the scroll position before the keyboard is show/hide and keeps it in place while the animation is on going.

Here is a video:

https://github.com/user-attachments/assets/d87f23f9-1a32-4e77-9607-7ae4581718c8

## To test

1. Ensure **transcripts** is enabled
2. Play any episode with transcript
3. Open the transcript
4. Tap "Search"
5. Tap outside the keyboard
6. ✅ The transcript should not scroll to another position

## Checklist

### DIFF
--- a/podcasts/TranscriptViewController.swift
+++ b/podcasts/TranscriptViewController.swift
@@ -432,7 +432,6 @@ class TranscriptViewController: PlayerItemViewController {
         addCustomObserver(Constants.Notifications.playbackTrackChanged, selector: #selector(update))
         NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillShow(_:)), name: UIResponder.keyboardWillShowNotification, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillHide(_:)), name: UIResponder.keyboardWillHideNotification, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(keyboardDidHide(_:)), name: UIResponder.keyboardDidHideNotification, object: nil)
         //We disabled the method bellow until we find a way to resync/shift transcript positions
         //addCustomObserver(Constants.Notifications.playbackProgress, selector: #selector(updateTranscriptPosition))
     }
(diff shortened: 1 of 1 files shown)


bjtitus (changes requested):
I think I'm still seeing this behavior when I open search and tap the transcripts.

Here's a recording from the simulator:

https://github.com/user-attachments/assets/793736af-17ce-42f2-b3ce-41122ff01b01

And a device:

https://github.com/user-attachments/assets/eb08b177-b959-41b2-abe7-59126458505a

SergioEstevao:
@bjtitus thanks for testing, I confirm your testing I will give it a try to fix the issue.

SergioEstevao:
@leandroalonso and @bjtitus I did some changes to the code and I believe the issue is now sorted but please give it some tests.

Changes here: https://github.com/Automattic/pocket-casts-ios/pull/2014/commits/7e485235533dd118aaa6386a91ead1055f4749c9

leandroalonso:
@SergioEstevao on it!

leandroalonso:
@SergioEstevao amazing! This indeed fixes the issue. I cannot approve my PR though, so if you prefer feel free to close this one and reopen another one then I can approve it.

SergioEstevao:
@leandroalonso we can wait for @bjtitus to test it again and when he approves we merge it.

leandroalonso:
@SergioEstevao ok! Can you update the PR description with the final solution?

bjtitus (approved):
:shipit: Looks good in my testing 👍

PR #2018, opened 2024-08-10
Reimagine Sharing: Export video

bjtitus:
| 📘 Part of: #1910 |
|:---:|

Adds Video export to clip sharing. Note that the asset and loading progress are not final. This is mostly adding the video asset writing code.

https://github.com/user-attachments/assets/57e8a42e-5b54-4bae-a09e-8d38072d190f

#### Video Export process
* Exports a 5 second clip of animated kidney shapes - this roughly follows the design specs but could still use tweaking in the future
* Loops this 5 second clip repeatedly over the length of the clip - this was done to speed up rendering as rendering each frame is quite slow
* Adds the audio from the episode to the clip

#### Progress
Adds a progress bar to verify export progress which reports each frame rendered. This isn't the final design but the current version is mainly meant for debugging to verify parts of the process.

### Code Changes

There are a lot of additions here but you can check separately or mostly ignore the helpers added in:

* `UnsafeTransfer.swift` - Removes compiler warnings about unsafe access to `CVPixelBuffer`
* `UnsafeWait.swift` - Allows synchronous blocking of the async operation in `unsafeRequestMediaDataWhenReady`
* `View+CVPixelBuffer.swift` - Generates `CVPixelBuffer` rendering to its context instead of converting to an image first and create a pixel buffer from that, reducing memory usage during export

The major additions are in
* `VideoExporter` - to export the video file
* `AnimatedShareImageView` - to wrap the `ShareImageView` with progress updates

## To test

> [!NOTE]

* Play an episode
* Tap the "Share" button in the action shelf
* Tap the "Clip" button
* Set the clip length by adjusting the trim handles
* Tap the "Clip" button
* Verify that export progress appears
* Verify that export completes
* Save audio to Photos or Files or share to an app
* Verify that the resulting file includes the clipped audio & subtle animation

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/57">7.71</a>. This milestone is due in less than 2 days.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/Sharing/Clip/VideoExporter.swift
```diff
@@ -0,0 +1,203 @@
...
+            FileLog.shared.addMessage("VideoExporter Ended: \(start.timeIntervalSinceNow)")
+        } onCancel: {
+            progress.cancel()
+        }
+    }
+
+    //MARK: - Private Methods
+
+    // Step 1 of video export
+    private static func exportInitialVideo<Content: AnimatableContent>(of view: Content, to outputURL: URL, with parameters: Parameters, frameCount: Int, progress: Progress) async throws {
```
bjtitus, line 66: This just creates the AVAsset related objects for exporting and then kicks off the `writeFrames` process.

podcasts/Sharing/Clip/VideoExporter.swift
```diff
@@ -0,0 +1,203 @@
...
+                              fps: parameters.fps,
+                              videoWriterInput: videoWriterInput,
+                              videoWriter: videoWriter,
+                              adaptor: adaptor,
+                              progress: progress,
+                              frameCount: frameCount)
+    }
+
+    // Part of Step 1
+    private static func writeFrames<Content: AnimatableContent>(of view: Content, size: CGSize, fps: Int, videoWriterInput: AVAssetWriterInput, videoWriter: AVAssetWriter, adaptor: AVAssetWriterInputPixelBufferAdaptor, progress: Progress, frameCount: Int) async throws {
```
bjtitus, line 91: This writes the actual frames to the AVAsset objects

podcasts/Sharing/Clip/VideoExporter.swift
```diff
@@ -0,0 +1,203 @@
...
+        let counter = Counter()
+        try await videoWriterInput.unsafeRequestMediaDataWhenReady { continuation in
```
bjtitus, line 93: The `requestMediaDataWhenReady` API isn't great for async/await since the block can be called repeatedly but the block needs to run synchronously producing frames. At the same time, it needs to end when the frames are done, so needs some context of an external frame count. I've added a "counter" actor to track the frame count and `unsafeWait` to synchronously block on the async `pixelBuffer` call. Seemed to work well in testing and there shouldn't be many of these running simultaneously.

I investigated `AsyncStream` but it won't work for this sync style case where backpressure is needed for the `requestMediaDataWhenReady`. I took a look at [`AsyncChannel`](https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncAlgorithms/AsyncAlgorithms.docc/Guides/Channel.md) provided by the Swift async algorithms but wasn't able to make it work here.

I might revisit later in the week but this solution works for now and reads relatively cleanly IMO.

podcasts/Sharing/Clip/VideoExporter.swift
```diff
@@ -0,0 +1,203 @@
...
+
+        guard let compositionVideoTrack = composition.addMutableTrack(withMediaType: .video, preferredTrackID: kCMPersistentTrackID_Invalid),
+              let sourceVideoTrack = try? await asset.loadTracks(withMediaType: .video).first else {
+            throw ExportError.failedToCreateCompositionTrack
+        }
+
+        let sourceTimeRange = try await asset.load(.duration)
+        var currentTime: CMTime = .zero
+
+        while currentTime < CMTime(seconds: parameters.duration, preferredTimescale: 600) {
```
bjtitus, line 139: Loops the video track over the full duration. Since our animation is a looping background change, we can just reuse those frames instead of generating frames over the whole duration, dramatically speeds up export especially on long clips. It rarely took more than 45 seconds to export in my testing regardless of clip length.

podcasts/Sharing/Clip/VideoExporter.swift
```diff
@@ -0,0 +1,203 @@
...
+            try compositionVideoTrack.insertTimeRange(CMTimeRange(start: .zero, duration: sourceTimeRange),
+                                                      of: sourceVideoTrack,
+                                                      at: currentTime)
+            currentTime = CMTimeAdd(currentTime, sourceTimeRange)
+        }
+
+        if let audioTrack = try await parameters.audioPlayerItem.asset.loadTracks(withMediaType: .audio).first {
```
bjtitus: Adds the audio track if one is available

podcasts/Sharing/Clip/VideoExporter.swift
```diff
@@ -0,0 +1,203 @@
...
+
+        await exportSession.export()
+
+        guard exportSession.status == .completed else {
+            throw ExportError.exportFailed(exportSession.error)
+        }
+    }
+
+    // Part of Step 2 of video export to add the audio track
+    private static func add(audioTrack: AVAssetTrack, at timeRange: CMTimeRange, to composition: AVMutableComposition) throws {
```
bjtitus, line 171: Just a helper which throws a `failedToAddAudioTrack` error if `addMutableTrack` fails for some reason

podcasts/Sharing/ShareImageView.swift
```diff
@@ -24,46 +24,74 @@ enum ShareImageStyle: CaseIterable {
             return "small"
         }
     }
+
+    var videoSize: CGSize {
```
bjtitus, line 28: These are now hard coded based on the designs and these sizes are used on export. These are scaled down for preview when creating the clip.

podcasts/Sharing/ShareImageView.swift
```diff
@@ -24,46 +24,74 @@ enum ShareImageStyle: CaseIterable {
...
     let info: ShareImageInfo
     let style: ShareImageStyle
 
+    @Binding var angle: Double
+
     var body: some View {
         ZStack {
             LinearGradient(gradient: info.gradient, startPoint: .top, endPoint: .bottom)
+            let rotationFactor = sin(Angle(degrees: angle).radians)
+            KidneyShape()
```
bjtitus, line 51: These are likely not the final animations but get us relatively close to the original designs.

podcasts/Sharing/ShareImageView.swift
```diff
@@ -24,46 +24,74 @@ enum ShareImageStyle: CaseIterable {
...
+                .opacity(0.15 + 0.5 * abs(rotationFactor))
+                .rotationEffect(.degrees(-180 * rotationFactor))
+                .blendMode(.softLight)
             Color.black.opacity(0.2)
             switch style {
             case .large:
                 VStack(spacing: 32) {
                     image()
-                        .frame(width: 200, height: 200)
+                        .aspectRatio(1, contentMode: .fit)
```
bjtitus, line 69: Since sizes are now defined, I preferred to define an aspect ratio here.

bjtitus:
@SergioEstevao I was hoping to get this under the 500 line count but I just don't think I can do it without removing too much functionality. I've stripped out iOS 15 support for now and tried to leave a bunch of comments to describe what's going on. There aren't too many files here, mostly just the export logic and some helpers to deal with async changes.

I'm happy to walk through it with you if needed, just let me know!

podcasts/Sharing/ShareImageView.swift
SergioEstevao: Could the aspect ratio value be connected to the enum: .large, .medium, ... and be return from there?

podcasts/Sharing/AnimatedShareImageView.swift
```diff
@@ -0,0 +1,35 @@
...
+                }
+                .scaleEffect(CGSize(width: 2.0, height: 2.0))
+        }
+    }
+
+    func update(for progress: Double) {
+        animationProgress.progress = progress
+    }
+
+    func calculateAngle(progress: Float) -> Float {
```
SergioEstevao: This could be marked private

podcasts/Sharing/Clip/VideoExporter.swift
SergioEstevao: Could this GH comment be part of of the function/code comment for the method?

podcasts/Sharing/Clip/VideoExporter.swift
```diff
@@ -0,0 +1,203 @@
...
+import AVFoundation
+import UIKit
+import PocketCastsUtils
+
+protocol AnimatableContent: View {
+    func update(for progress: Double)
+}
+
+@available(iOS 16, *)
+class VideoExporter {
```
SergioEstevao: Does it make sense for VideoExporter to be a class? from what I can see all it's methods are static. Maybe it can be an enum if you just using for to be a namespace?

podcasts/Sharing/SharingModal.swift
```diff
@@ -162,6 +162,24 @@ extension SharingModal.Option {
         return imageInfo
     }
 
+    func itemProvider(style: ShareImageStyle) -> NSItemProvider {
+        switch self {
+        case .clipShare(_, _, _, let progress):
+            if let fileURL = progress.fileURL {
```
SergioEstevao: Maybe use a `guard `here instead of an  `if`

SergioEstevao:
I realize that this is not related to this changes but when testing on the iPad there is some issues in layout:

| 1 | 2 | 
| - | - | 
| Simulator Screenshot - iPad mini (6th generation) - 2024-08-12 at 11 48 46 |  Simulator Screenshot - iPad mini (6th generation) - 2024-08-12 at 11 46 16 |

And a crash when you tap on the More (...) button in landscape mode ( the code is not expecting a pop up I believe)

SergioEstevao (changes requested):
Great work on the video exporter. 

I left some nitpick comments and more important some layout issues and a crash on the iPad.

podcasts/Sharing/SharingModal.swift
bjtitus: The only issue is I have to add `return` in each block:

Personally that doesn't bother me too much, though.

```suggestion
    func itemProvider(style: ShareImageStyle) -> NSItemProvider {
        switch self {
        case .clipShare(_, _, _, let progress):
            guard let fileURL = progress.fileURL else {
                return ShareImageView(info: imageInfo, style: style, angle: .constant(0)).itemProvider()
            }
            if #available(iOS 16.0, *) {
                return NSItemProvider(contentsOf: fileURL, contentType: UTType(filenameExtension: fileURL.pathExtension) ?? .mpeg4Movie)
            } else {
                return NSItemProvider(contentsOf: fileURL)!
            }
        default:
            return ShareImageView(info: imageInfo, style: style, angle: .constant(0)).itemProvider()
        }
    }
```

bjtitus:
@SergioEstevao Ready for review again!

I switched to `ShareLink`, which also required adopting the new [Transferable](https://developer.apple.com/documentation/coretransferable/transferable#) API instead of `NSItemProvider`. I'll submit code for iOS 15 separately to fix the `sourceView` issue directly since it's not covered in this PR anyway and a bit of a hack, so `ShareLink` will be the better implementation going forward.

SergioEstevao:
@bjtitus when I try to share the video to X or Mail I get this exception:

```
Thread 40: "value for key 'NS.objects' was of unexpected class 'NSValue' (0x1f7434588) [/System/Library/Frameworks/Foundation.framework].\nAllowed classes are:\n {(\n    \"'NSString' (0x1f74346c8) [/System/Library/Frameworks/Foundation.framework]\",\n    \"'NSArray' (0x1f742cde0) [/System/Library/Frameworks/CoreFoundation.framework]\",\n    \"'NSDate' (0x1f742ca70) [/System/Library/Frameworks/CoreFoundation.framework]\",\n    \"'NSNumber' (0x1f74346f0) [/System/Library/Frameworks/Foundation.framework]\",\n    \"'NSOrderedSet' (0x1f742c9a8) [/System/Library/Frameworks/CoreFoundation.framework]\",\n    \"'NSSet' (0x1f742cc28) [/System/Library/Frameworks/CoreFoundation.framework]\",\n    \"'NSDictionary' (0x1f742cd68) [/System/Library/Frameworks/CoreFoundation.framework]\",\n    \"'NSNull' (0x1f742c7c8) [/System/Library/Frameworks/CoreFoundation.framework]\",\n    \"'NSData' (0x1f742cc00) [/System/Library/Frameworks/CoreFoundation.framework]\"\n)}"
```

Expected?

I also noticed that when trying to share to Tumblr or WordPress we only get a link to a web page instead of the video.

SergioEstevao (changes requested):
@bjtitus see this comment: https://github.com/Automattic/pocket-casts-ios/pull/2018#issuecomment-2285923670

bjtitus:
I'm having a lot of trouble with the `ShareLink` and `Transferable` APIs right now, especially with third party apps. I think many of them may not be expecting these share types. I've added multiple items for each content type, which seems to fix Mail.app but I'm still not seeing video in X, Tumblr, or any of our apps. I'll circle back to this after this PR, though, since it's not really part of the video export.

SergioEstevao:
@bjtitus I think you are missing a`try` in this commit https://github.com/Automattic/pocket-casts-ios/pull/2018/commits/8ba393275fa39802077849e23586ac2a7eb148fd

SergioEstevao:
@bjtitus I ran the tests again and now it looks that the audio is not being recorded to the resulting video clip... Are you also seeing that?

bjtitus:
@SergioEstevao Sorry, I used the video clip asset for the audio track. It's fixed now.

SergioEstevao (approved):
:shipit:

PR #2041, opened 2024-08-15
Skip Back: do not allow skipping forward when skipping back

### LINKED ISSUE #1950: Rapidly Tapping Skip Back Button Results in App Skipping Forward
### Description

Rapidly and repeatedly tapping the "skip back" button to go back actually results in the app skipping forward. This was reported in 8529645-zd-a8c.

Slack discussion: p1721757368412189-slack-C02A333D8LQ

I'm unable to reproduce.

The user confirmed they have "trim silence" set to "medium", the playback speed is set to 1x, and the volume boost is not activated.

1. Open the full screen player.
2. Tap the play button.
3. Tap the "skip back" button in rapid succession.

https://github.com/user-attachments/assets/e8842555-7b8b-4111-808c-4a0cbcf14b8c

iPhone14,2, iOS 17.5.1, Pocket Casts 7.68

leandroalonso:
Fixes https://github.com/Automattic/pocket-casts-ios/issues/1950

Adds a mechanism to ensure that rapidly tapping Skip Back do not result in skipping forward.

This doesn't actually fixes the issue but prevents it from happening.

## To test

I highly recommend testing on a real device. In order to reproduce the issue and test this PR you need:

1. To download an episode
2. To enable trim silence
3. Then, play the episode

Then:

1. Open the full screen player.
4. Tap the play button.
5. Tap the "skip back" button in rapid succession.
6. ✅ The audio should skip back, not forward
7. Skip forward
8. ✅ It should work
9. Skip forward, then back
10. ✅ It should work

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -298,7 +298,7 @@ class PlaybackManager: ServerPlaybackDelegate {
 
         let currPos = currentTime()
         let backTime = max(currPos - amount, 0)
-        seekTo(time: backTime)
+        seekTo(time: backTime, seekHint: .back)
     }
 
     func skipForward() {
@@ -401,8 +401,8 @@ class PlaybackManager: ServerPlaybackDelegate {
         seekingTo != PlaybackManager.notSeeking
     }
 
-    func seekTo(time: TimeInterval, startPlaybackAfterSeek: Bool = false) {
-        seekTo(time: time, syncChanges: SyncManager.isUserLoggedIn(), startPlaybackAfterSeek: startPlaybackAfterSeek)
+    func seekTo(time: TimeInterval, startPlaybackAfterSeek: Bool = false, seekHint: SeekHint? = nil) {
+        seekTo(time: time, syncChanges: SyncManager.isUserLoggedIn(), startPlaybackAfterSeek: startPlaybackAfterSeek, seekHint: seekHint)
     }
 
     func seekToFromSync(time: TimeInterval, syncChanges: Bool, startPlaybackAfterSeek: Bool) {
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/57">7.71</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/PlaybackManager.swift
```diff
@@ -401,18 +401,44 @@ class PlaybackManager: ServerPlaybackDelegate {
...
+
+    func seekTo(time: TimeInterval, syncChanges: Bool, startPlaybackAfterSeek: Bool = false, seekHint: SeekHint? = nil) {
         guard let playingEpisode = currentEpisode() else { return } // nothing to actually seek
 
+        // When using EffectsPlayer we have an issue in which rapidly tapping
+        // skip back results (sometimes) in skipping forward
+        // Here we handle that to avoid this issue
+        // See https://github.com/Automattic/pocket-casts-ios/issues/1950
+        if seekHint == .back {
+            if let previousSeekTime, time > previousSeekTime {
```
SergioEstevao: Why not use the currentTime value that we have bellow? And avoid the use of this variable and the deboucer?

leandroalonso: Because it will never work.

`currentTime` is the wrong value being given. If we compare it to it, `time > currentTime` will never succeed. You can try adding and testing it, the bug will be there.

Not sure if the usage of debouncer is clear, if not I can add a comment clarifying it. But we have the debouncer to fix issues with skip forward or other seeks after rapidly tapping skipping back happens. If we don't have it, we might get stuck into a state where seek will never work.

SergioEstevao: Just wanted to clarify if the `currentTime` returned is correct or not.
So our current understanding is that value sometimes is not correct/or not in sync with the previous seeks

Regarding the debouncer my understanding is that we want to clear the previousSeektime, so it's only used when you are doing multiple skips back in a quick interval. Right?

podcasts/PlaybackManager.swift
```diff
@@ -401,18 +401,44 @@ class PlaybackManager: ServerPlaybackDelegate {
...
+    private let debouncer = Debounce(delay: 1.second)
+
+    func seekTo(time: TimeInterval, syncChanges: Bool, startPlaybackAfterSeek: Bool = false, seekHint: SeekHint? = nil) {
         guard let playingEpisode = currentEpisode() else { return } // nothing to actually seek
 
+        // When using EffectsPlayer we have an issue in which rapidly tapping
+        // skip back results (sometimes) in skipping forward
+        // Here we handle that to avoid this issue
+        // See https://github.com/Automattic/pocket-casts-ios/issues/1950
+        if seekHint == .back {
```
SergioEstevao: Nit: Could we refactor this to a separate method something like `isValidSeek` ?

SergioEstevao (approved):
I tested this on simulator and real device and it looks that it avoid the bug of moving forward.
The only thing that I notice it that sometime the seekPosition get's "stucks: in a specific position ( I see that is because our fail guard kicks in) but it then starts moving again when the times resync.
All in all I think it's better than our current state so I think we should move forward with this solution.

podcasts/PlaybackManager.swift
leandroalonso: Done in 601a19609556bbf1461e9e4f5913728e2711bc5f

PR #2054, opened 2024-08-16
watchOS 11: make play/pause button a primary action

leandroalonso:
While building the project for Xcode 16 I went ahead and implemented the primary action for the play/pause button which should allow to play/pause by double-tapping the thumb and index finger together.

Before merging that we need Xcode 16.0 on the CI though, so I'll leave as a draft till the support is added.

## To test

1. Run the app on a device connected to a Apple Watch Series 9 and Apple Watch Ultra 2 with watchOS 11
2. Open the playing screen on the watch
3. Double-tap using the thumb and index finger
4. ✅ Episode should play/pause

## Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -6,7 +6,7 @@ common_params:
   # Common environment values to use with the `env` key.
   - &common_env
     # If you update the image to a newer Xcode version, don't forget to also update the badge in the README.md file accordingly for consistency
-    IMAGE_ID: xcode-16.0-beta5
+    IMAGE_ID: xcode-15.4
   # Common agents values to use with the `agents` key.
   - &common_agents
     queue: mac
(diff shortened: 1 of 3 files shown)


SergioEstevao:
Just tested this on my updated Series 9 Watch and the action is working correctly.

Keep in mind that the action only works when the app is open and we have the Playing Now screen in the screen.

If the apps is background or in another screen the Play/Pause action will not be triggered.

Another thing to keep in mind is that by default when you double tap, the system will show and fade a shape around the control to indicate the selection. By default the shape is a rectangle that encloses the play/pause button.

We can change the shape used: a triangle for the play action and a rectangle for the pause action but maybe not worth the extra work.

leandroalonso:
> Another thing to keep in mind is that by default when you double tap, the system will show and fade a shape around the control to indicate the selection. By default the shape is a rectangle that encloses the play/pause button.

@SergioEstevao can you share video recording of that please?

SergioEstevao:
@leandroalonso here is a video of the action:


https://github.com/user-attachments/assets/6f4ef89a-abdd-43f2-b85a-e035e2e0fd13

SergioEstevao (approved):
Let's merge this PR and we will the clip shape on a separate PR.

leandroalonso:
@SergioEstevao TIL about watch mirroring.

PR #2065, opened 2024-08-19
[Tooling][CI] Modernize Buildkite pipelines syntax to use shared values

AliSoftware:
This improves the syntax we use in `.buildkite/*.yml` pipelines to rely on the `shared-pipeline-vars` file, which exports common values as env variables before the YAML files are evaluated by the `buildkite-agent`.

This will make the pipelines easier to read and maintain—and especially updating the Xcode version used by CI.

> [!NOTE]
> This is in preparation of facilitating the update of the Xcode image used by this project, to soon adopt the new Xcode 16.0b5 image (via https://github.com/Automattic/pocket-casts-ios/pull/2066) that we made available in our CI (ref: paaHJt-72d-p2)

> [!IMPORTANT]
> - [x] This PR requires https://github.com/Automattic/buildkite-ci/pull/496 to be deployed first

## To Test

If CI is green (which won't happen until we land https://github.com/Automattic/buildkite-ci/pull/496 and retry the build) we should be good to go.

### DIFF
--- a/.buildkite/shared-pipeline-vars
+++ b/.buildkite/shared-pipeline-vars
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+# This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
+# to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
+
+# 🎗️ If you update the image to a newer Xcode version, don't forget to also update the badge in the README.md file accordingly for consistency
+export IMAGE_ID="xcode-15.4"
+
+export CI_TOOLKIT="automattic/a8c-ci-toolkit#3.5.0"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,22 +1,17 @@
-# Nodes with values to reuse in the pipeline.
-common_params:
-  # Common plugin settings to use with the `plugins` key.
-  - &common_plugins
-    - automattic/a8c-ci-toolkit#3.5.0
-  # Common environment values to use with the `env` key.
-  - &common_env
-    # If you update the image to a newer Xcode version, don't forget to also update the badge in the README.md file accordingly for consistency
-    IMAGE_ID: xcode-15.4
-  # Common agents values to use with the `agents` key.
-  - &common_agents
-    queue: mac
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+# Variables used in this pipeline are defined in `shared-pipeline-vars`, which is `source`'d before calling `buidkite-agent pipeline upload`
+
+env:
+  IMAGE_ID: $IMAGE_ID
 
 steps:
   - label: Build and Test
     command: .buildkite/commands/build.sh
-    agents: *common_agents
-    env: *common_env
-    plugins: *common_plugins
+    agents:
+      queue: mac
+    plugins: [$CI_TOOLKIT]
 
   - group: "Linters"
     steps:
@@ -40,9 +35,9 @@ steps:
 
       - label: ":sleuth_or_spy: Lint Localized Strings Format"
         command: lint_localized_strings_format
-        agents: *common_agents
-        plugins: *common_plugins
-        env: *common_env
+        agents:
+          queue: mac
+        plugins: [$CI_TOOLKIT]
 
   - block: Deploy Prototype Build
     prompt: Share a Prototype Build via App Center?
@@ -59,9 +54,9 @@ steps:
         key: build_prototype
         depends_on: prototype_triggered
         command: .buildkite/commands/prototype-build.sh
-        agents: *common_agents
-        env: *common_env
-        plugins: *common_plugins
+        agents:
+          queue: mac
+        plugins: [$CI_TOOLKIT]
         # The folder name is are configured in Fastlane
         artifact_paths:
           - artifacts/*.ipa
@@ -73,9 +68,9 @@ steps:
       - label: Prototype Build - Upload
         depends_on: build_prototype
         command: .buildkite/commands/prototype-upload.sh
-        agents: *common_agents
-        env: *common_env
-        plugins: *common_plugins
+        agents:
+          queue: mac
+        plugins: [$CI_TOOLKIT]
         notify:
           - github_commit_status:
               context: Prototype Build - Upload
--- a/.buildkite/release-builds.yml
+++ b/.buildkite/release-builds.yml
@@ -1,13 +1,11 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+# Variables used in this pipeline are defined in `shared-pipeline-vars`, which is `source`'d before calling `buidkite-agent pipeline upload`
 # This pipeline is meant to be run via the Buildkite API, and is only used for release builds
 
-# Nodes with values to reuse in the pipeline.
-common_params:
-  # Common plugin settings to use with the `plugins` key.
-  - &common_plugins
-    - automattic/a8c-ci-toolkit#2.15.1
-  # Common environment values to use with the `env` key.
-  - &common_env
-    IMAGE_ID: xcode-15.4
+env:
+  IMAGE_ID: $IMAGE_ID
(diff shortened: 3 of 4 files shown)


.buildkite/shared-pipeline-vars
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
+# to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
+
+# If you update the image to a newer Xcode version, don't forget to also update the badge in the README.md file accordingly for consistency
+export IMAGE_ID="xcode-15.4"
```
leandroalonso, line 7: @AliSoftware If I'm not mistaken, in the current configuration there are two places to set the Xcode image, so it is possible to set a beta Xcode for non-release builds.

While this creates a bit of code friction, it allows us to have betas using a beta Xcode and releases to be submitted to Apple using a stable Xcode version as they require.

Is the same possible here? Or does this affect all builds — including releases?

AliSoftware: Oh, I didn't know this was potentially intentional to have two different versions of Xcode for the different pipelines.

The only change I made in this PR is replacing the existing values with those constants. Before this PR the `IMAGE_ID` used in both `pipeline.yml` and `release-builds.yml` were already both `xcode-15.4` and the same, so that PR won't change that.

If you want to be able to change them at different pace, we could consider using two different constants — e.g. declare separate `DEFAULT_IMAGE_ID` and `RELEASE_IMAGE_ID` values to be exported in this `shared-pipeline-vars` here, and then use `IMAGE_ID: $DEFAULT_IMAGE_ID` on `pipeline.yml` to be used on PR builds, and `IMAGE_ID: $RELEASE_IMAGE_ID` on `release-builds.yml` to be used for beta and final release builds sent to TestFlight and App Store.

That way when you'll be using a stable version including for PR builds, you can just set `export DEFAULT_IMAGE_ID=$RELEASE_IMAGE_ID` to use the same in both cases, while when you're experimenting with betas you can set `DEFAULT_IMAGE_ID=xcode-16.0-beta5` to be used for PR builds, all while keeping `RELEASE_IMAGE_ID=xcode-15.4` pointing to a stable version for builds sent to TestFlight/App Store.

wdyt?

leandroalonso: In fact, we would like to have the final build use a stable Xcode version (so we can submit it for review) and all the rest to use the beta one (including TestFlight). This helps us in assessing issues because the TestFlight user base is quite active so we can fix them asap.

However, this is a bit confusing, happens just this time of the year and we can change the image manually ourselves if we want to. So I wonder if we should or should not reflect this on the code. 🤷

AliSoftware: In the current setup, we can have the VM images used to be different for `pipeline.yml` (the pipeline used for PR builds and builds made when new commits land in  `trunk`) vs `release-builds.yml` (the pipeline used to generate builds sent to TestFlight—both for beta or not)

But note that, even before this PR, this has always been the case that both betas (from `new_beta_release` lane) and final build (from `finalize_release` lane) use the same `release-builds.yml` pipeline, and thus always used the same Image and Xcode version on CI when those are built.
So I'm not sure how you were handling this even before the change from this PR 😅  because the Xcode version and VM image used by CI has always been the same for both beta and final builds, even if it could be different from the one used for building and running tests on PRs and `trunk`.

---

If you think that could be valuable, we can totally brainstorm how we could rewrite the pipelines to make it possible to use different VM images / Xcode versions for intermediate betas vs final builds, as opposed to only use the beta for PR builds. But that would be a new feature that wasn't supported even by the current setup before this PR 😉 

My initial thoughts about how to enable such a setup is to make use of the fact that [the `BETA_RELEASE` env var](https://github.com/Automattic/pocket-casts-ios/blob/trunk/fastlane/Fastfile#L971) is [set to `true` when the `release-builds.yml` pipeline is triggered for a beta](https://github.com/Automattic/pocket-casts-ios/blob/trunk/fastlane/Fastfile#L406), while it is [set to `false` when the same `release-builds.yml` pipeline is triggered for a final build](https://github.com/Automattic/pocket-casts-ios/blob/trunk/fastlane/Fastfile#L414).

So I guess something like this in the `shared_pipeline-vars` would cover it?
```bash
export IMAGE_ID=xcode-15.4
if [[ "$PIPELINE" == "pipeline.yml" || "$BETA_RELEASE" == "true" ]];
  # Use the beta only for PRs, `trunk` and beta builds
  export IMAGE_ID=xcode-16.0-beta5
fi
```
Which, btw, is a logic that we wouldn't be able to implement in the `release-builds.yml` directly in the current state, and would be an approach only possible thanks to this migration to use this new `shared-pipeline-vars` (which is in practice just a bash script that is`source`'d before the YML is evaluated by Buildkite, allowing us such additional flexibility)

AliSoftware: @leandroalonso I've updated https://github.com/Automattic/pocket-casts-ios/pull/2066 to have the logic mentioned in my above comment; let's continue the discussion about this approach in that other PR (since it's related to updating the image to the Xcode beta so feels a more relevant place to discuss strategies around it)

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/shared-pipeline-vars
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
+# to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
+
+# 🎗️ If you update the image to a newer Xcode version, don't forget to also update the badge in the README.md file accordingly for consistency
```
iangmaia, line 6: While at it, what about adopting a `.xcode-version` file?

AliSoftware: yeah, I thought about it too.

But then I thought:
 - The `.xcode-version` file doesn't exist yet, that means it's not yet used by any tooling (no use of e.g. [the `xcversion` action](https://docs.fastlane.tools/actions/xcversion/) to enforce the right Xcode version to be `xcode-select`'d before fastlane can continue). So introducing it would only serve as having a conventional place to declare the expected Xcode version for humans (developers, contributors) to look and check, but not enforced by any of our tooling (yet) in this repo
 - But on the other hand, there was already a source of truth for developers and contributors to look at to know which version of Xcode they were supposed to use to work on this project: the `README.md` has a dedicated badge to document exactly that.

So that made me think that introducing the `.xcode-version` file would mean that we'd suddenly have 2 sources of truth instead of one… 🤔 

That being said, once could argue that we already have 2 sources of truth with `shared-pipeline-vars` vs the `README.md` anyway, so moving to still 2 sources of truth but those now being `.xcode-version` (that the `shared-pipelines-vars` would read and derive the `IMAGE_ID` from) and `README.md`… would not change the status quo that much 😅 

But again… given the fact that [the team see to want to be able to use different Xcode versions (stable for final releases, Xcode beta for PRs & betas) during the period where there are Xcode betas around](https://github.com/Automattic/pocket-casts-ios/pull/2065#discussion_r1722243780), it's not clear what the `.xcode-version` file would be expected to contain in such case (even if those cases only happens during a short periods when Apple releases new Xcode betas, not all year round).

In particular, [the conditional logic implemented in #2066](https://github.com/Automattic/pocket-casts-ios/pull/2066/files) allowing to pick a different Xcode version depending on final release build or not… would not really work well with a single source of truth that would be the `.xcode-version` file…

So that's why in the end I decided not to introduce it after all 😅

leandroalonso (approved):

.buildkite/shared-pipeline-vars
iangmaia: > The .xcode-version file doesn't exist yet, that means it's not yet used by any tooling

Perhaps we could use this as the main factor to not introduce it now 👍
But with some convention + additional logic, I guess we could work around the remaining points in the future.

iangmaia (approved):

PR #2066, opened 2024-08-19
[Tooling][CI] Update Xcode image to 16.0b5

AliSoftware:
This updates the VM image used by CI from the one using Xcode `15.4` to Xcode `16.0-beta-5`.

Ref: pdnsEh-1Oa-p2

> [!IMPORTANT]
> This PR is based on top of https://github.com/Automattic/pocket-casts-ios/pull/2065 and depends on it, so before we can hope for CI to go green on this PR, we need to:
> - [x] deploy https://github.com/Automattic/buildkite-ci/pull/496
> - [ ] then ideally land https://github.com/Automattic/pocket-casts-ios/pull/2065 first before this one (to keep the diffs separate)

### Note on potential ongoing WIP for features requiring Xcode 16.0+ APIs

This chain of PRs (https://github.com/Automattic/pocket-casts-ios/pull/2065 + this one) are all build on top of `trunk`.
@leandroalonso this means that if you already cut a branch locally to start working on the features that require the new API early, you'll have to rebase your feature branch on `trunk` once the whole thing has landed to benefit from the new image on CI.

## To test

 - CI is green
 - The CI logs show that it used the Xcode 16.0b5 VM image and Xcode version to do the build

### DIFF
--- a/.buildkite/shared-pipeline-vars
+++ b/.buildkite/shared-pipeline-vars
@@ -6,4 +6,10 @@
 # 🎗️ If you update the image to a newer Xcode version, don't forget to also update the badge in the README.md file accordingly for consistency
 export IMAGE_ID="xcode-15.4"
 
+# Uncomment this to experiment with an Xcode beta only on PRs and TestFlight beta builds
+# (while still using the latest stable Xcode declared above for final Release Builds)
+if [[ "$PIPELINE" == "pipeline.yml" || "$BETA_RELEASE" == "true" ]]; then
+  export IMAGE_ID="xcode-16.0-beta5"
+fi
+
 export CI_TOOLKIT="automattic/a8c-ci-toolkit#3.5.0"


leandroalonso (changes requested):
CI is failing. 😬

AliSoftware:
> CI is failing. 😬

… which is expected, as indicated by the "Important" note in the PR description 😛 

> This PR is based on top of https://github.com/Automattic/pocket-casts-ios/pull/2065 and depends on it, so we need to deploy https://github.com/Automattic/buildkite-ci/pull/496 then land https://github.com/Automattic/pocket-casts-ios/pull/2065 before we can hope for CI to go green on this PR and land it

AliSoftware:
@leandroalonso so, following our discussion from https://github.com/Automattic/pocket-casts-ios/pull/2065#discussion_r1722243780

> In fact, we would like to have the final build use a stable Xcode version (so we can submit it for review) and all the rest to use the beta one (including TestFlight). This helps us in assessing issues because the TestFlight user base is quite active so we can fix them asap.

I've updated the logic in this PR to make the use of the `xcode-16.0-beta5` image now be conditional and only if we're building a PR (`PIPELINE` = `pipeline.yml`) or building a beta (`release-builds.yml` with `BETA_RELEASE` set to `true`), while it'd still uses the `xcode-15.4` image for release builds.

---

That being said, I'm still wondering if that's the right approach, because by going with this logic that means that if a change on a PR makes use of new APIs only available in the Xcode 16.0b5 SDK… since the PRs and the betas will build with the Xcode beta that will pass and all be green on CI… but then you'll only notice the failure on the day you do the final release (due to it being the only one still using `15.4` and not `16.0b5`), not before. Which might not be ideal, as one doesn't want to rush a fix for adding some missing `#if available` last minute on a Friday / submission day instead of catching it earlier and keep the zen on submission day 😛 

wdyt?

.buildkite/shared-pipeline-vars
```diff
@@ -6,4 +6,10 @@
 # 🎗️ If you update the image to a newer Xcode version, don't forget to also update the badge in the README.md file accordingly for consistency
 export IMAGE_ID="xcode-15.4"
 
+# Uncomment this to experiment with an Xcode beta only on PRs and TestFlight beta builds
```
AliSoftware, line 9: FYI: I wrote this intentionally as "Uncomment this" while in practice it's currently already uncommented, because my idea was for y'all to just comment the 3 `if…fi` lines below back once you'll be done with experimenting the Xcode 16 beta.

In other words, basically once the final Xcode 16.0 will be out, you'd update the `export IMAGE_ID="xcode-16.0"` from line 7 above, then comment the `if…fi` lines below to remove them but still keep the syntax around for next time.
Then, if some months later in the future after you migrated all builds to the stable `16.0`, you wanted to start experimenting again with the next Xcode beta (e.g. say Xcode 16.1b3), you'd just have to un-comment those lines below again and update the `export IMAGE_ID="xcode-16.1-beta3"` on line 12… instead of having to remember (or dig in git history) how to do such conditional when experimenting with betas.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso:
@AliSoftware 

> wdyt?

Yeah, forget about that. I think we will be trying to solve a problem that has multiple layers and as you mentioned in the other PR:

> So I'm not sure how you were handling this even before the change from this PR 😅

I think I handled this manually and got confused with the two `XCODE_IMAGE` declarations.

Sorry about all the additional back-and-forth.

leandroalonso (approved):

iangmaia (approved):

AliSoftware:
> Yeah, forget about that. I think we will be trying to solve a problem that has multiple layers and as you mentioned in the other PR:
> 
> > So I'm not sure how you were handling this even before the change from this PR 😅
> 
> I think I handled this manually and got confused with the two `XCODE_IMAGE` declarations.
> 
> Sorry about all the additional back-and-forth.

Ok. Which means that, while I see you ended up approving this PR (while it's using that previously discussed conditional logic rn), now that we've decided that it's not the right choice after all, we'll need to adjust the implementation of this PR before landing it.
I'll return the PR back to Draft state in the meantime then, to avoid an accidental merge of it too soon.

---
 
So, given all that… what should we end up doing after all to move this forward? I can imagine two options:

### 1️⃣  Option 1

 - `xcode-16.0-beta5` for `pipeline.yml` — i.e. Buildkite builds run on `trunk` and PRs and which are running the Unit Tests etc. so you can start working on various PRs that would use the beta
 - `xcode-15.4` for `release-builds.yml` — i.e. for Buildkite builds uploading a build to TestFlight / ASC, both if that's for a beta or if that's for the final build

This means that:
 - your beta testers will still get builds done with the latest stable Xcode (I'm actually not 100% sure that it's possible to submit a build compiled with a beta version of Xcode to TestFlight—and have it pass Apple review even if only for public beta distribution—to begin with tbh?)
 - if you accidentally start using an API only available in Xcode 16.0b5 without properly protecting it with `#if available`, your PRs will still go green and you'll only notice it during the next beta build (which is better than noticing it during the final release, but still not ideal)


### 2️⃣  Option 2

 - Keep an unconditional `IMAGE_ID="xcode-15.4"` in `trunk` (used for all YAML pipelines, so for `trunk`, PR cut from `trunk`, betas, final release
 - Start cutting a long-running branch named `xcode-16` from `trunk`, change `IMAGE_ID="xcode-16.0-beta5` in that branch, and then each dev who is working on a WIP feature requiring the latest APIs would cut a feature branch from that `xcode-16` branch, and make their PR target that `xcode-16` branch. Then you'd finally only merge the `xcode-16` branch into `trunk` once the final Xcode 16.0 stable version is released by Apple (after updating `IMAGE_ID="xcode-16.0"` one last time before this final merge)

@leandroalonso which of those two options do you prefer we implement in this PR 2066 then?

.buildkite/shared-pipeline-vars
```diff
@@ -6,4 +6,10 @@
...
+# (while still using the latest stable Xcode declared above for final Release Builds)
+if [[ "$PIPELINE" == "pipeline.yml" || "$BETA_RELEASE" == "true" ]]; then
```
AliSoftware, line 11: I just realized that this logic didn't work anyway (i.e. [this CI build associated with building this PR](https://buildkite.com/automattic/pocket-casts-ios/builds/7884#019170a4-b84b-489d-9f47-c59765467e27) still used `xcode-15.4`), because the `PIPELINE` env var is not defined to any value in the case of builds triggered from commits to `trunk` or PR branches (we do rely on `pipeline.yml` being the default if `PIPELINE` env var is not defined).
So I should have written `[[ "${PIPELINE:-pipeline.yml}" == "pipeline.yml" … ]]` for that test to work properly in call cases instead.

(We're going to change this whole logic anyway [based on ongoing discussions in this PR](https://github.com/Automattic/pocket-casts-ios/pull/2066#issuecomment-2299364262), but wanted to mention in case we end up still wanting to use a similar logic in our final implementation)

leandroalonso:
> which of those two options do you prefer we implement in this PR 2066 then?

Option 2. Let's keep it simple!

AliSoftware:
> > which of those two options do you prefer we implement in this PR 2066 then?
> 
> Option 2. Let's keep it simple!

Ok, works for me 👍 

Given I initially named the branch of this PR `tooling/xcode-16.0b5`, which now feels a bit too specific for a long-running branch of option 2:
 - I think we should then close this PR and delete my `tooling/xcode-16.0b5` branch
 - And then @leandroalonso I'll let you cut a new `xcode-16.0` branch (or whatever you want to call it) from `trunk` (or maybe naming it `xcode-16` if you think you'll keep it for 16.1 too—as Apple has apparently already started release Xcode 16.1 betas already?), push a commit on it to update just the `IMAGE_ID="xcode-16.0-beta5"` on line 7 of `shared-pipeline-vars`, and communicate with your team that any WIP features intending to use Xcode 16 APIs should cut from and target this new branch

Works for you?

leandroalonso:
> Works for you?

Yup, absolutely! Thanks for working on that!

PR #2068, opened 2024-08-20
Fix for iPad crash on xcode 16 and iOS 18

### LINKED ISSUE #2063: iOS 18 -> TabController new TabBar design makes app crash
| 1 | 2 | 
| - | - |
| Simulator Screenshot - iPad mini (6th generation) - 2024-08-13 at 13 50 55 | Simulator Screenshot - iPad mini (6th generation) - 2024-08-13 at 13 50 52 |

When running the app using Xcode 16 beta 5 and on iOS18 iPad simulator de app crashes on startup.

This is because the new design of the tab bar on iOS 18 on `.regular` size modes uses a new floating design that is not linked to the current view.

On startup we are trying to constraint the mini-player to the top of the old tabbar, because it's no longer on the same hierarchy makes the app to crash.

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes #2063

This PR avoid the crash resulting from the iPad Tab bar new design on iOS18 when using a TabViewController.

The way we are avoiding the crash is to force the TabBarController  to believe is always in the compact mode and keep using the regular tab bar on the bottom.

I tried to adapt to the new design but there are several issues with our current design of the app:

- Our mini-player is positioned above the tab bar ( this was reason for the crash because on the new design the tabbar is not in the same hierarchy as the mini-player)
- Our animations for the mini-player assume the tabbar is on the bottom
- The new tab bar design only allows to show 4 items on portrait mode and we currently 5 items so it show a chevron to display more options

## To test

- Using Xcode 16 beta 5 and an iOS 18 iPad device start the app
- Ensure that the app does not crash at startup
- Rotate the device and use the multiple split-view modes and check the app is working correctly
- Test on iOS 18 iPhone device to see if the tab bar shows correctly
- Repeat the tests on XCode 15.4 and iOS 17

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/MainTabBarController.swift
```diff
@@ -137,9 +139,24 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
         Settings.shouldShowInitialOnboardingFlow = false
     }
 
+    private func fixTarBarTraitCollectionOnIpadForiOS18() {
+        if #available(iOS 18.0, *),
```
leandroalonso, line 143: I tested and it all works as expected.

However, if we merge this one we'll run into the following issue:

1. This code requires Xcode 16 to build
2. However, we cannot submit final builds to Apple using beta Xcode
3. Which means that we will have to change both the `IMAGE_ID` and comment this code

I've been researching a way to do that by using the `$(XCODE_VERSION_MAJOR)` without any luck so far though.

leandroalonso (changes requested):

podcasts/MainTabBarController.swift
SergioEstevao: I was able to run and build this code on Xcode 15. All the calls made are already present on iOS 17, the only thing that changed in iOS 18 was the way iOS implements the tab bar so only on iOS 18 we need to do these calls.
I think I can revert the CI configurations changes and will still get the CI green.

leandroalonso: Oh ok!

Changing the CI should be enough then.

But I think we should wait to merge until we decide if we want to merge that on the frozen branch or `trunk`.

SergioEstevao:
Closing in favour of: https://github.com/Automattic/pocket-casts-ios/pull/2077

PR #2071, opened 2024-08-20
Xcode 16: add flag to allow building the app to Xcode 16 and Xcode 15

leandroalonso:
This is a solution for a temporary problem:

1. We would like to release betas using Xcode 16 beta
2. However, we need to submit final builds to Apple using Xcode 15

This poses the challenge of how to handle building the app in both versions.

~In this PR I added a Swift Compiler Custom Flag that allows us to detect if we're building on Xcode 16 or not.~

I changed the solution to use a built-in compiler directive as [suggested here](https://github.com/Automattic/pocket-casts-ios/pull/2071#discussion_r1724685336).

## To test

1. ✅ Build the app with Xcode 15
2. ✅ Build the app with Xcode 16 beta

## Checklist

### DIFF
--- a/Pocket Casts Watch App/NowPlayingControls.swift
+++ b/Pocket Casts Watch App/NowPlayingControls.swift
@@ -131,7 +131,7 @@ struct NowPlayingView_Previews: PreviewProvider {
 
 struct HandGestureShortcutPrimaryAction: ViewModifier {
     public func body(content: Content) -> some View {
-        #if XCODE1600
+        #if compiler(>=6.0)
         if #available(watchOS 11.0, *) {
             content
                 .handGestureShortcut(.primaryAction)
--- a/Pocket Casts Watch App/NowPlayingControls.swift
+++ b/Pocket Casts Watch App/NowPlayingControls.swift
@@ -131,11 +131,15 @@ struct NowPlayingView_Previews: PreviewProvider {
 
 struct HandGestureShortcutPrimaryAction: ViewModifier {
     public func body(content: Content) -> some View {
+        #if XCODE1600
         if #available(watchOS 11.0, *) {
             content
                 .handGestureShortcut(.primaryAction)
         } else {
             content
         }
+        #else
+            content
+        #endif
     }
 }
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/57">7.71 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

Pocket Casts Watch App/NowPlayingControls.swift
```diff
@@ -131,11 +131,15 @@ struct NowPlayingView_Previews: PreviewProvider {
 
 struct HandGestureShortcutPrimaryAction: ViewModifier {
     public func body(content: Content) -> some View {
+        #if XCODE1600
```
AliSoftware: That's a nice trick, but that also means that once you'll migrate to Xcode 16.1 betas for example, this condition will be false (as `XCODE_MAJOR_VERSION` will be `1601` and `XCODE1600` will now be undefined). Sure, you probably expect to migrate to the stable Xcode 16.0 once it's out before moving to Xcode 16.1 betas, and this `#if` conditional is only intended to be temporary and will be removed once migration to stable Xcode 16 can be done, but still easy to miss one of those when cleaning up, so…

I'm wondering if it wouldn't instead be possible/enough to use [the built-in compiler control directives](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Compiler-Control-Statements) like `#if compiler(>=6.0)` instead.
 - On one hand that would allow this condition to stay true as we migrate to higher versions of Xcode even if you forget to clean up those `#if` after the migration is over
 - On the other hand, semantically speaking what one would really want to test is the version of the SDK used, not the version of the compiler… but since those are updated in tandem when they update Xcode versions… 🤷

leandroalonso: @AliSoftware `XCODE_MAJOR_VERSION` will return `XCODE1600` for 16.1. On 15.4 it returns `XCODE1500`, because it's the major version.

Screenshot 2024-08-21 at 10 50 40

That being said, I think your solution requires fewer changes and it's more elegant so I'll change to it! Thanks for sharing that!

AliSoftware: Ah interesting. I wonder why they return 1600 and not 16 then 🤷

danielebogo (approved):
Built both with Xcode 15 and 16 and it worked!

PR #2072, opened 2024-08-21
Update podcast list when episode is archived or marked as played from player

### LINKED ISSUE #1976: Podcast grid not updated when archiving or marking episode as played from full screen player
### Description

When marking an episode as played or archived from the fullscreen player, the podcast grid is not updated with the correct badge count and sort order. When navigating away from the grid and then back again, the UI is updated (see attached video).

I had a look at the UI code and `PodcastListViewController` does listen for changes to playback, which should trigger a grid refresh:
https://github.com/Automattic/pocket-casts-ios/blob/f4c02d8e0493912d41947f12cb4b6f236c42b7bd/podcasts/PodcastListViewController.swift#L149

I debugged this and found that the event is indeed fired and triggers a refresh, but at that time the podcast episode update was still in progress by `DataManager`. I think `PodcastListViewController` should also listen for the notifications that are fired when the episode play or archive status has been changed:

https://github.com/Automattic/pocket-casts-ios/blob/f4c02d8e0493912d41947f12cb4b6f236c42b7bd/podcasts/Constants.swift#L51-L53

1. Go to Podcast grid view
2. Play episode and open fullscreen player
3. Mark episode as played
4. Minimise player
5. Notice that the badges are not updated and the podcast order is not changed
6. Navigate away from and then back to the grid
7. Notice that the grid is now correct

https://github.com/user-attachments/assets/3d5daea0-a1ed-449d-801c-ea1bf1cb0106

iPhone 14 Pro, iOS 18 beta 4, PocketCasts Trunk Build

rviljoen:
Fixes #1976 

Updated `PodcastListViewController` to listen for episode state changes initiated by the user from the player screen (mark as played or archived). Previously the podcast list would remain out of date until you would tab away from the screen and then back (see video in issue).

## To test

1. Go to Podcast grid view
2. Play episode and open fullscreen player
3. Mark episode as played
4. Minimise player
5. Notice that the badges are not updated and the podcast order is not changed
6. Navigate away from and then back to the grid
7. Notice that the grid is now correct
8. Apply the change in this PR
9. Re-do steps above. Grid should now be correct without the need to navigate away and back.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 7.72
 -----
+- Update podcast list when episode is archived or marked as played from player [#1976](https://github.com/Automattic/pocket-casts-ios/issues/1976)
 
 
 7.71
--- a/podcasts/PodcastListViewController.swift
+++ b/podcasts/PodcastListViewController.swift
@@ -148,6 +148,8 @@ class PodcastListViewController: PCViewController, UIGestureRecognizerDelegate, 
         addCustomObserver(ServerNotifications.syncCompleted, selector: #selector(refreshGridItems))
         addCustomObserver(Constants.Notifications.playbackTrackChanged, selector: #selector(refreshGridItems))
         addCustomObserver(Constants.Notifications.playbackEnded, selector: #selector(refreshGridItems))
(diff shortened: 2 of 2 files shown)


leandroalonso (approved):
Tested before and after and it works as described! Thanks for fixing it @rviljoen!

PR #2073, opened 2024-08-21
Fix playback when switching from iPhone speaker to other audio outputs with trim silence enabled

### LINKED ISSUE #2049: Trim silence breaks playback when switching from iPhone speaker to other audio outputs
### Discussed in https://github.com/Automattic/pocket-casts-ios/discussions/1538

<div type='discussions-op-text'>

<sup>Originally posted by **ajkessel** March 19, 2024</sup>
I've experienced an issue since first installing Pocket Casts at the start of this year: if trim silence is enabled and audio is playing from the iPhone speaker, switching to another audio output (Bluetooth earbuds, speaker, or CarPlay) causes playback to stop. The lock screen shows the player in "play" rather than "pause" mode, but the player is not playing, and tapping play/pause has no effect. The only way to resume playing is to foreground the app and hit play/pause a couple of times.

I've attempted to work through this issue with Pocket Casts support but hit a dead-end.

Wondering if anyone here has ideas for further troubleshooting/narrowing the issue? Apparently it's not widespread, but I can't think of anything that would be unusual about my own setup. I've tried purging and reinstalling the app to no avail. 

FWIW, this is on an iPhone Pro Max 15 now running iOS 17.4. The behavior occurs with various external audio devices: Jabra Elite 75t earbuds, a Sangean H202 speaker, and Hyundai CarPlay.</div>

## Steps to reproduce

1. Play podcast on iPhone speaker
2. Enable Trim Silence
3. Switch to BT headphones (non-Apple)
4. Switch off headphones. Playback should stop
5. Press Play. Playback should start on speaker.
6. Turn on headphones. They will connect to iPhone and playback will stop, instead of continuing on headphones. Player shows pause button, i.e. it thinks it is still playing, but the progress bar is stopped.
7.  You have to press pause and then play again to fix playback.

The above steps were performed in the app, but also happens outside of the app (i.e. when the app is minimized), as initially reported by the user.

rviljoen:
Fixes #2049 

See original issue for issue description and root cause. Since iOS 17 and WatchOS 10, we receive a notification when the audio route changes due to disconnect. This PR sets the state for `interruptionInProgress` correctly based on the reason of the interruption.

This is complex part of the player state, so I'm submitting a draft PR first for review and discussion. Open to other ideas for resolution.

## To test

1. Play podcast on iPhone speaker
2. Enable Trim Silence
3. Switch to BT headphones (non-Apple)
4. Switch off headphones. Playback should stop
5. Press Play. Playback should start on speaker.
6. Turn on headphones. They will connect to iPhone and playback will stop, instead of continuing on headphones. Player shows pause button, i.e. it thinks it is still playing, but the progress bar is stopped.
7. Apply the PR and repeat steps above. Playback should continue when switching from speaker to headphone and player state should be correct.
8. Test other audio interruptions, e.g. activating Siri, to ensure that the app still handles those interruptions correctly (i.e. pauses playback and resumes when interruption is done).

## Checklist


rviljoen:
@leandroalonso @SergioEstevao Would appreciate your review and thoughts on this issue and potential fix.

SergioEstevao:
Thanks for this fix @rviljoen! I will try to give it a test today.

SergioEstevao:
I tested this on the iPhone Xs with iOS 17 and Apple Watch Series 9 with WatchOS 18 with Sony MX-1004 headphone and solves the issue!

@leandroalonso or @danielebogo  do you have time to double check this?

leandroalonso:
Will do it now

leandroalonso:
@rviljoen all good from my side!

Just want to suggest one small thing:

1. Given this is related to Playback, I think it would be good to put this change behind a feature flag
2. Why? If something goes wrong after the app is released, we can turn off this change

When it comes to playback, there are a lot of exceptions, different devices, and situations that even a good beta testing user base doesn't help cover, so staying on the cautious side avoids issues in the future. :) 

Let me know what do you think!

rviljoen:
@leandroalonso 100% agree. I'd sleep much better knowing the FF safety net is in place 😄 I'll add it.

rviljoen:
@leandroalonso Ready for another look. I assume you'll take care of setting the remote flag on the server?

leandroalonso:
> I assume you'll take care of setting the remote flag on the server?

Yup!

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -88,6 +88,9 @@ public enum FeatureFlag: String, CaseIterable {
     /// Enable the new Upgrade Experiments
     case upgradeExperiment
 
+    ///
```
leandroalonso: Should we add a comment about it?

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -185,6 +190,8 @@ public enum FeatureFlag: String, CaseIterable {
             "default_player_filter_callback_fix"
         case .upNextOnTabBar:
             "up_next_on_tab_bar"
+        case .ignoreRouteDisconnectedInterruption:
```
leandroalonso: This one you don't need to set it actually, we need to update this documentation. This is generated automatically (see `default`)

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
rviljoen: Doh! I started the comment but dit not finish. There now.

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
rviljoen: Got it. Backed out this change.

leandroalonso:
@rviljoen just one conflict left on the `CHANGELOG.md`

[We configured this to be merged using union](https://github.com/Automattic/pocket-casts-ios/blob/trunk/.gitattributes) but it's not working. :(

rviljoen:
@leandroalonso Good to go!

SergioEstevao (approved):
All looking good, thanks for adding the FF!

PR #2080, opened 2024-08-22
[Paywall Experiments] Features Carousel variant

### LINKED ISSUE #2061: Feature Variation: Create new Carousel based on features

danielebogo:
| 📘 Part of: #1986
|:---:|

Fixes #2061

This PR introduces the Features Carousel Variant

https://github.com/user-attachments/assets/0bc007a5-a721-4cfc-8a39-332385ce2f17

## To test

> [!NOTE]
> Make sure to enable the `upgradeExperiment` feature flag

- CI must be 🟢 
- Use a non-plus account 
- Go to Profile -> Settings and tap on `Pocket Casts Plus`
- Check the new carousel like the video above

## Checklist


leandroalonso (changes requested):
@danielebogo from testing the feature and code side, all good. I also tested with dynamic font size and it works as expected.

There are a few design issues though and one concern I have.

### Concern: the amount of images

We're adding a bunch of new images. All of them add up 1,4 MB on the project. Of course, for App Store, they will be inside the IPA, and Apple will do its own processing but I still think it's a lot for a screen that users won't see that often.

Wouldn't it be possible to add just the first image on the project and load the rest from a URL?

This is not a blocker though, but I feel we should discuss to see if there's any clever way of doing it.

## Wear OS image

On the Apple Watch part, we're showing the Wear OS image + this weird watch shape. I commented in the Figma but this one I feel we should definitely change because it's quite bad to show another platform here. 😅

## Spacing

In the cards the title/content is quite close, as opposed to the designs. See:

| App | Design |
| ---- | ------- |
| Screenshot 2024-08-22 at 17 14 45 | Screenshot 2024-08-22 at 17 15 22 |

Also the subtitle color is gray.

danielebogo:
@leandroalonso thanks for the feedback.
I fixed the text color and the spacing. I also optimized the images for now, reducing the whole folder to less than 700kb. Would you mind give it another look while waiting for some other feedback on the WearOS image and url loading?

leandroalonso (approved):
> Would you mind give it another look while waiting for some other feedback on the WearOS image and url loading?

Donezo! This is not a blocker for merging this PR.

PR #2084, opened 2024-08-22
Dismiss search if already active when double tap is performed on tab bar

### LINKED ISSUE #1652: Search: tapping Discover/Podcasts with the search open should dismiss it
### Description

When tapping Podcast or Discover tab bar item with the search open, the keyboard is shown again for a new search instead of dismissing the search.

1. Tap Discover or Podcasts at the tab bar
2. Tap it again so the search starts
3. Dismiss the keyboard or search for something
4. Tap the tab bar item again
5. Search restart and the keyboard is shown again instead of dismissing search

_No response_

_No response_

rviljoen:
Fixes #1652 

Search: tapping Discover/Podcasts with the search open should dismiss it

## To test

1. Tap Discover or Podcasts on the tab bar
2. Tap it again so the search starts
3. Dismiss the keyboard or search for something
4. Tap the tab bar item again
5. Search should be dismissed at this point. Old behaviour was for search to restart and the keyboard to be shown again, which should not happen

## Checklist

### DIFF
--- a/podcasts/DiscoverViewController.swift
+++ b/podcasts/DiscoverViewController.swift
@@ -95,7 +95,13 @@ class DiscoverViewController: PCViewController {
         if mainScrollView.contentOffset.y.rounded(.down) > defaultOffset.rounded(.down) {
             mainScrollView.setContentOffset(CGPoint(x: 0, y: defaultOffset), animated: true)
         } else {
-            searchController.searchTextField.becomeFirstResponder()
+            // When double-tapping on tab bar, dismiss the search if already active
+            // else give focus to the search field
+            if searchController.cancelButtonShowing {
+                searchController.cancelTapped(self)
+            } else {
+                searchController.searchTextField.becomeFirstResponder()
+            }
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 7.72
 -----
 - Update podcast list when episode is archived or marked as played from player [#1976](https://github.com/Automattic/pocket-casts-ios/issues/1976)
-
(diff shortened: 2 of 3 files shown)


podcasts/DiscoverViewController.swift
```diff
@@ -95,7 +95,13 @@ class DiscoverViewController: PCViewController {
         if mainScrollView.contentOffset.y.rounded(.down) > defaultOffset.rounded(.down) {
             mainScrollView.setContentOffset(CGPoint(x: 0, y: defaultOffset), animated: true)
         } else {
-            searchController.searchTextField.becomeFirstResponder()
+            // When double-tapping on tab bar, dismiss the search if already active
+            // else give focus to the search field
+            if searchController.cancelButtonShowing {
+                searchController.cancelTapped(self)
```
rviljoen, line 101: - Feels a bit hacky to simulate a button tap, but not sure if it is work duplicating the code in that method or refactoring it.
- Also, should I pass `self` or `nil` or not important? More of a style question I guess.

rviljoen:
@leandroalonso Since you logged this issue, I would appreciate your eyes on the proposed fix. Also see my comment on the change for some guidance, please.

podcasts/DiscoverViewController.swift
leandroalonso: I think it's fine to pass `self`.

While it seems hack I think it's a ok trade-off for a gesture I assume not many people use that much. I learned of that quite recently by the way.

leandroalonso (approved):

leandroalonso:
@rviljoen all good on my testing, I think your solution is ready to go. :)

PR #2093, opened 2024-08-23
Transcripts: Reset scroll position when changing transcript episode

SergioEstevao:
| 📘 Part of: #1848  |  
|:---:|

Fixes #2092 

This PR ensures that a new transcript is loaded we reset the transcript position to the beginning. 

## To test

 - Start the app
 - Ensure that you have `transcripts` and `newShowNotesEndpoint` FF enabled
 - Play an episode that as transcripts support. EX: "Cautionary Tales"
 - Add another episode with transcripts to be played in your UpNext Queue
 - Open the full screen player
 - Tap on transcripts option
 - Scroll around in the transcripts to the end of it
 - Await for the episode to get to the end and the new episode starts
 - Check that position is reset to the start.

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #2094, opened 2024-08-26
Transcripts: Reset scroll position when changing transcript episode

### LINKED ISSUE #2092: Transcripts: transcript position is not reset when changing episode
Steps to reproduce:

- Open the app
- Start playing an episode with a transcript
- Add to the UpNext queue another episode with a transcript
- Open the full screen player
- Open the transcript view
- Scroll down
- Close transcript view
- Wait for episode to finish and the next one start
- Open transcript view
- See that the transcript is on previous position instead of the start of the transcript

leandroalonso:
| 📘 Part of: #1848  |
|:---:|

Fixes #2092 

This is the same as #2093, I just changed the solution slightly to be more simple.

This PR ensures that a new transcript is loaded we reset the transcript position to the beginning. 

## To test

 - Start the app
 - Ensure that you have `transcripts` and `newShowNotesEndpoint` FF enabled
 - Play an episode that as transcripts support. EX: "Cautionary Tales"
 - Add another episode with transcripts to be played in your UpNext Queue
 - Open the full screen player
 - Tap on transcripts option
 - Scroll around in the transcripts to the end of it
 - Await for the episode to get to the end and the new episode starts
 - Check that position is reset to the start.

## Checklist


leandroalonso:
This solution doesn't work for a specific case when loading a transcript for the first time — instead of switching between episodes.

PR #2117, opened 2024-09-03
Update BookmarkRow.swift

yinuoyang01:
Fix accessibility issues of the play button inside episode row


leandroalonso:
Hi @yinuoyang01! Thanks for this change!

Would you mind adding some testing steps to the PR description? Thank you so much!

PR #2146, opened 2024-09-05
Fix #2139: Add export caching + improved file names

### LINKED ISSUE #2139: Cache exported asset upon creation
The asset is exported each time a sharing option is selected. We should cache this option, at least while the sharing view is visible.

The Facebook export is different than the System Sheet export so this may complicate things a little bit.

bjtitus:
| 📘 Part of: #1910 |
|:---:|

Fixes #2139

* Caches clip exports so they don't need to be recreated for a new share destination.
* Adds improved shared file name to `<Podcast> - <Episode> - <Start>-<End>` like `This American Life - 754 Spark Bird - 34.727-60.532`

## To test

* Play an episode
* Tap the "Share" icon
* Tap "Cip"
* Create a clip
* Tap the "More" button
* Share the asset
* Dismiss the system share sheet
* Tap the "More" button
* Verify that the clip is not re-exported and the cached version is used instead
* Verify that the file name is of the format `<Podcast> - <Episode> - <Start>-<End>`

## Checklist

### DIFF
--- a/podcasts/Sharing/SharingModal.swift
+++ b/podcasts/Sharing/SharingModal.swift
@@ -177,13 +177,13 @@ extension SharingModal.Option {
     }
 
     @MainActor
-    func shareData(style: ShareImageStyle, destination: ShareDestination, progress: Binding<Float?>) async throws -> [Any] {
+    func shareData(style: ShareImageStyle, destination: ShareDestination, clipUUID: String, progress: Binding<Float?>) async throws -> [Any] {
         let url = URL(string: shareURL) as NSURL?
 
         let media: Any?
         switch self {
         case .clipShare(let episode, let clipTime, _):
-            media = try await mediaData(imageInfo: imageInfo, style: style, episode: episode, clipTime: clipTime, destination: destination, progress: progress)
+            media = try await mediaData(imageInfo: imageInfo, style: style, episode: episode, clipTime: clipTime, destination: destination, clipUUID: clipUUID, progress: progress)
         default:
             media = ShareImageView(info: imageInfo, style: style, angle: .constant(0)).frame(width: style.previewSize.width, height: style.previewSize.height).snapshot(scale: 3)
         }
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/58">7.72 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/Sharing/SharingModal.swift
```diff
@@ -255,3 +281,17 @@ fileprivate extension Podcast {
         return L10n.paidPodcastReleaseFrequencyFormat(frequency)
     }
 }
+
+fileprivate extension String {
```
danielebogo: Do you think this can be re-used in other areas? Just thinking if we should move this to a generic extension, like `String+RegEx`.

podcasts/Sharing/SharingModal.swift
```diff
@@ -206,20 +206,46 @@ extension SharingModal.Option {
...
+            )
+        }
 
         progress.wrappedValue = nil
 
         if destination == .instagram {
             return try? Data(contentsOf: fileURL) // For some reason, I couldn't get this to work with just a URL
         } else {
-            return fileURL as NSURL // Third party apps need URLs and won't accept Data
+            func format(time: TimeInterval) -> String {
```
danielebogo: Nit: Can this one be a new extension for TimeInterval? Like the one Sentry has.

danielebogo (approved):
Tested and it worked as described.
I left few nits other than that it looks good.
Just a question:
I tested it on the sim and I got this log when the sheet opened:
```
error fetching file provider domain for URL:file:///SIM_PATH/tmp/Accidental%20Tech%20Podcast%20-%20A%20Glimpse%20of%20a%20Better%20World%20-%201230.962-1290.962.mp4 : (null)
```
also on real device
```
error fetching item for URL:file:///DEVICE_PATH/tmp/The%20Wirecutter%20Show%20-%20How%20to%20Clean%20The%20Air%20You%20Breathe%20-%201193.250-1205.292.mp4 : (null)
```
Is it something we should look at?

podcasts/Sharing/SharingModal.swift
bjtitus: This was really just intended to not repeat the code in this function. I've added it as a `fileprivate` extension here so it's a bit cleaner.

bjtitus:
> I tested it on the sim and I got this log when the sheet opened:

I'm not sure if this log has any impact. I tried stripping the file name back to something very basic and the log persists so I don't think it's some problem with finding the file extension. I wasn't able to see a problem when sharing to other apps but I'll keep an eye on it.

PR #2182, opened 2024-09-20
Add Mimetype lib and helper

### LINKED ISSUE #1933: Corrupt files because HTTP contentType and real file type do not match
This zendesk tickets reports that episode of the following podcasts are corrupted when downloaded:

- https://play.pocketcasts.com/podcasts/f77f2320-60c8-0139-33f3-0acc26574db2
- https://play.pocketcasts.com/podcasts/e8a4ad80-09e9-013c-f5ae-0acc26574db2

I debugged the issue and found out the following:
- When downloading the episode files the server is returning contentType as `video/mp4`
- The actual file being downloaded is an `audio/mp3` file
- Because we use the [contentType to check for the real file type](https://github.com/Automattic/pocket-casts-ios/blob/05d568c55ca40c3065be06f75c4890c76310b3f9/Modules/Server/Sources/PocketCastsServer/Private/API%20Tasks/MetadataTask.swift#L51) we then use that type to check if it's a video file or not and also set the extension of the local file using that contentType. This causes the system to fail when opening file.

In order to fix this we will need to change our code to actual look at the mime type written inside the file.

One possibility is to do something like this after we donwload the file:

```
import Foundation

func mimeType(forFileAtPath path: String) -> String {
    let fileExtension = (path as NSString).pathExtension
    // the code bellow should be changed to only read the bytes we need from the file instead of loading everything into memory
    guard let fileData = try? Data(contentsOf: URL(fileURLWithPath: path)) else {
        return "application/octet-stream"
    }
    
    let bytes = [UInt8](fileData)
    
    var fileHeader = [UInt8](repeating: 0, count: 12)
    for i in 0..<min(fileHeader.count, bytes.count) {
        fileHeader[i] = bytes[i]
    }
    
    let headerString = fileHeader.map { String(format: "%02x", $0) }.joined()
    
    let mimeTypes: [String: String] = [
        "25504446": "application/pdf",
        "504b0304": "application/zip",
        "47494638": "image/gif",
        "ffd8ff": "image/jpeg",
        "89504e47": "image/png"
        // Add more signatures as needed
    ]
    
    for (signature, mimeType) in mimeTypes {
        if headerString.hasPrefix(signature) {
            return mimeType
        }
    }
    
    return "application/octet-stream"
}

```

danielebogo:
Fixes #1933

This PR fixes the issue when the HTTP content type doesn't match the file type.
To fix it we read the first part of the file and match the correct mime type using `Swime`.
If the file and HTTP  types don't match, we update the episode metadata and store the file with the correct type and extension.

## To test

> [!NOTE]
> Make sure you have FF `useMimetypePackage` enabled

- CI must be 🟢 
- Try to download some episodes from:
https://play.pocketcasts.com/podcasts/f77f2320-60c8-0139-33f3-0acc26574db2
https://play.pocketcasts.com/podcasts/e8a4ad80-09e9-013c-f5ae-0acc26574db2
- After the download you should play it normally
- You can also try with https://rts-vod-dd.akamaized.net/ww/15153718/d3f6e99d-7bdb-37eb-96e1-aaa67fa0a582.mp4

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
@@ -12,4 +12,3 @@ public struct MimetypeHelper {
         }
     }
 }
-
--- a/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
@@ -5,7 +5,7 @@ public struct MimetypeHelper {
     // The max number of bytes required to cover all the mimetypes
     // needed by `Swime`.
     private static let maxFileSize: Int = 4500
-    
+
     public static func contetType(for url: URL) -> String? {
         do {
             guard let fileSize = try url.resourceValues(forKeys: [.fileSizeKey]).fileSize,
--- a/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
@@ -1,0 +1,15 @@
+import Swime
+import Foundation
+
+public struct MimetypeHelper {
+    public static func contetType(for url: URL) -> String? {
+        do {
+            let data = try Data(contentsOf: url)
+            let mimeType = Swime.mimeType(data: data)
+            return mimeType?.mime ?? "application/octet-stream"
+        } catch {
+            return nil
+        }
+    }
+}
+
--- a/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
@@ -4,13 +4,13 @@ import Foundation
 public struct MimetypeHelper {
     // The max number of bytes required to cover all the mimetypes
     // needed by `Swime`.
-    private static let maxFileSize: Int = 4500
+    private static let maxBytesNeeded: Int = 4500
 
     public static func contetType(for url: URL) -> String? {
         do {
             guard let fileSize = try url.resourceValues(forKeys: [.fileSizeKey]).fileSize,
                   let fileHandle = FileHandle(forReadingAtPath: url.relativePath),
-                  let data = try fileHandle.read(upToCount: min(fileSize, Self.maxFileSize)) else {
+                  let data = try fileHandle.read(upToCount: min(fileSize, Self.maxBytesNeeded)) else {
                 return nil
             }
             let mimeType = Swime.mimeType(data: data)
--- a/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
@@ -4,8 +4,13 @@ import Foundation
 public struct MimetypeHelper {
     public static func contetType(for url: URL) -> String? {
         do {
-            let data = try Data(contentsOf: url)
+            guard let fileSize = try url.resourceValues(forKeys: [.fileSizeKey]).fileSize,
+                  let fileHandle = FileHandle(forReadingAtPath: url.relativePath),
+                  let data = try fileHandle.read(upToCount: min(fileSize, 4500)) else {
+                return nil
+            }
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><code>Modules/Server/Package.swift</code> was changed without updating its corresponding <code>Package.resolved</code>. Please resolve the Swift packages in Xcode.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/59">7.73 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
```diff
@@ -0,0 +1,15 @@
+import Swime
+import Foundation
+
+public struct MimetypeHelper {
+    public static func contetType(for url: URL) -> String? {
+        do {
+            let data = try Data(contentsOf: url)
```
SergioEstevao: This call will pull all the file data into memory... Is there a way to just get the first bytes?

danielebogo: Yes in theory I can use the FileHandle to get the data for n amount of bytes. I need to make some change to the package anyway, I will try there

Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
```diff
@@ -0,0 +1,19 @@
+import Swime
+import Foundation
+
+public struct MimetypeHelper {
+    public static func contetType(for url: URL) -> String? {
+        do {
+            guard let fileSize = try url.resourceValues(forKeys: [.fileSizeKey]).fileSize,
+                  let fileHandle = FileHandle(forReadingAtPath: url.relativePath),
+                  let data = try fileHandle.read(upToCount: min(fileSize, 4500)) else {
```
danielebogo: @SergioEstevao 
here we use 4.5kb to generate the data and read the correct type below

SergioEstevao: Do we need that much bytes? maybe define the value as a constant just to have a clear idea of why this amount.

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -188,6 +191,8 @@ public enum FeatureFlag: String, CaseIterable {
             false
         case .playerIsReadyToPlay:
             true
+        case .useMimetypePackage:
+            false
```
SergioEstevao: Do you plan to just activate using the remote flag later on?

SergioEstevao (changes requested):
I think it will be safer to also put a check on the FF on the `MetadataTask` class in line 51 , to avoid racing conditions between updates to `contentType`

SergioEstevao:
This is working great, tested with all the podcasts we detected issues and all working great now. Just did some nitpick comments and suggestion to improve it.

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
danielebogo: Yes but I might enable this now

Modules/Server/Sources/PocketCastsServer/Private/MimetypeHelper.swift
danielebogo: > Do we need that much bytes?

Yes as it's the max to cover all the types

danielebogo:
> I think it will be safer to also put a check on the FF on the MetadataTask class in line 51 , to avoid racing conditions between updates to contentType

@SergioEstevao `MetadataTask` didn't change as the metadata is set before having the file in the temp folder and it happens before the `DownloadManager` returns the downloaded location.

SergioEstevao (approved):
:shipit:

Issue #2104, opened 2024-08-28
Discover Collection View Refactor


bjtitus:
### Discover Collection View Refactor Project

Project Thread: pdeCcb-6GQ-p2

The `DiscoverViewController` uses a scroll view with manual layout. We should refactor this to use a `UICollectionView` so that we can better track visibility of sections and make the code easier to manage.

danielebogo:
Version `7.75` has now entered code-freeze, so the milestone of this PR has been updated to `7.76`.

SergioEstevao:
Version `7.76` has now entered code-freeze, so the milestone of this PR has been updated to `7.77`.

danielebogo:
Version `7.77` has now entered code-freeze, so the milestone of this PR has been updated to `7.78`.

PR #2238, opened 2024-10-03
Discover Collection View: Call appearance methods on cells

bjtitus:
| 📘 Part of: #2104 |
|:---:|

Modifies the Discover collection view cells to call `viewWillAppear`, `viewDidAppear`, `viewWillDisappear`, and `viewDidDisappear` (indirectly via `beginAppearanceTransition` and `endAppearanceTransition`) when the cell is shown or hidden according to UICollectionView delegate methods. This behavior reuses our existing code for each Discover view controller which loads content, triggers analytics events, and handles some timer additions (for the Featured Summary cell).

## To test

1. Enable `tracksLogging`
2. Navigate to Discover
3. Verify that the `discover_list_impression` event is called when a cell is _just_ about to be shown.
4. Ensure that cells are correctly loaded
5. Navigate to a category
6. Ensure that the `discover_featured_page_changed` event is not being logged while in a category
7. Close the category
8. Ensure that the `discover_featured_page_changed` event is logged again

## Checklist

### DIFF
--- a/podcasts/DiscoverCollectionViewController+Search.swift
+++ b/podcasts/DiscoverCollectionViewController+Search.swift
@@ -33,6 +33,27 @@ extension DiscoverCollectionViewController: UICollectionViewDelegate {
 
         searchController.parentScrollViewDidScroll(scrollView)
     }
+
+    func collectionView(_ collectionView: UICollectionView, willDisplay cell: UICollectionViewCell, forItemAt indexPath: IndexPath) {
+        let item = dataSource.itemIdentifier(for: indexPath)
+
+        switch item {
+        case .item(let item):
+            let viewController = (cell.contentConfiguration as? UIViewControllerContentConfiguration)?.viewController as? DiscoverSummaryProtocol & UIViewController
+            viewController?.populateFrom(item: item.item, region: item.region, category: item.selectedCategory)
+            viewController?.beginAppearanceTransition(true, animated: false)
+            viewController?.endAppearanceTransition()
+        default:
+            ()
+        }
+    }
+
+    func collectionView(_ collectionView: UICollectionView, didEndDisplaying cell: UICollectionViewCell, forItemAt indexPath: IndexPath) {
+
+        let viewController = (cell.contentConfiguration as? UIViewControllerContentConfiguration)?.viewController
+        viewController?.beginAppearanceTransition(false, animated: false)
+        viewController?.endAppearanceTransition()
+    }
 }
 
 extension DiscoverCollectionViewController: PCSearchBarDelegate {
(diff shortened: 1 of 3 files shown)


podcasts/DiscoverCollectionViewController+Search.swift
```diff
@@ -33,6 +33,27 @@ extension DiscoverCollectionViewController: UICollectionViewDelegate {
...
+        let item = dataSource.itemIdentifier(for: indexPath)
+
+        switch item {
+        case .item(let item):
+            let viewController = (cell.contentConfiguration as? UIViewControllerContentConfiguration)?.viewController as? DiscoverSummaryProtocol & UIViewController
+            viewController?.populateFrom(item: item.item, region: item.region, category: item.selectedCategory)
+            viewController?.beginAppearanceTransition(true, animated: false)
+            viewController?.endAppearanceTransition()
+        default:
+            ()
```
danielebogo, line 47: Can we just `break` here?

danielebogo (commented):
Hey Brandon, all good with events tracked.

Before approve this, I have few questions:

Do we track the same event when a cell appears and disappear?
```
🔵 Tracked: discover_list_impression ["list_id": "nasa"]
🔵 Tracked: discover_list_impression ["list_id": "myths-and-more"]
🔵 Tracked: discover_list_impression ["list_id": "nasa"]
```

When I select a category I see `discover_featured_page_changed` being tracked immediately. Is it correct?

(Not about this PR)
About the layout warning, I think one of the breaking constraint is the title:
Simulator Screenshot - iPhone 16 - 2024-10-07 at 11 30 40

bjtitus:
> Do we track the same event when a cell appears and disappear?

Nope! This was a bug. I missed that the child appearance methods are called by the system so I've disabled them in https://github.com/Automattic/pocket-casts-ios/pull/2238/commits/4d1eebbe35e7810051755f7f208d093e0f06a9ab for `DiscoverCollectionViewController`. Before, it was calling this method whenever the view controller's view was added, now, from what I can tell, it's only logged on appearance when I call from the collection view delegate. I think this is relatively safe since Discover is a tab anyway.

bjtitus:
> When I select a category I see discover_featured_page_changed being tracked immediately. Is it correct?

Hmmm. This shouldn't be happening. Could you add a recording of what you're seeing? Here's what it looks like for me. I wonder if the timer is just happening to fire when you switch?

https://github.com/user-attachments/assets/e0f50b48-71d4-4fad-b9f5-2bc34ee3a083

podcasts/DiscoverCollectionViewController+Search.swift
danielebogo: @bjtitus just a heads up about this

danielebogo:
@bjtitus I don't see the event anymore. Maybe was just a coincidence when tapping on the category

https://github.com/user-attachments/assets/2d4ba870-9e4d-4695-bb8c-6146274f8fa8

danielebogo (approved):
Approving this as the events now are tracked correctly.
There's still one comment left but not a blocker.

PR #2265, opened 2024-10-14
[Tooling] Add script to spell-check `en.lproj` strings

AliSoftware:
This is a proof-of-concept to spell-check the strings from `en.lproj` to find typos.

This idea comes from a discussion with @danielebogo in Slack DM

## To test

 - Install `aspell` using `brew install aspell` (note that it _might_ be already installed on your Mac by another brew dependency)
 - ~Run `scripts/spell-check-en-strings.rb`~ Run `bundle exec fastlane spellcheck_strings`

<details><summary> 🖼️ This is the output when run on my machine</summary>

_Old output from initial implementation. See [below comment](https://github.com/Automattic/pocket-casts-ios/pull/2265#issuecomment-2417218201) for the new rendering after I migrated the code to be a lane_
image
</details>

## What's Next

This is only a proof of concept / draft for now. As a follow-up after this lands, we probably want to:

 - Fix the current typos, and potentially add more acceptable exceptions to the `KNOWN_WORDS` lists if needed
 - Make this run on CI _(Note: this will require to run on a macOS agent because it uses `plutil`[^1])_

[^1]: We could arguably re-implement `plutil` in Swift for Linux if we wanted to, so we could run that script on a Linux machine instead of Mac; but that's out of scope of this PR / spike

## Checklist

- [x] ~I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary~. N/A
- [x] ~I have considered adding unit tests for my changes.~ N/A
- [ ] ~I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~ N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

scripts/spell-check-en-strings.rb
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'json'
+require 'open3'
+
+STRINGS_FILE = File.join('podcasts', 'en.lproj', 'Localizable.strings')
+KNOWN_WORDS = %w[
```
AliSoftware: Note that according to its man page, `aspell` is supposed to be able to take a `--personal={file}` parameter to provide a personal dictionary of known words. That being said, I tried to use it but it still reported the words I listed in my text file as typos in the output of the command. I probably missed something, and this could be worth looking a bit more into `aspell` to figure this out, instead of handling the logic ourselves in the ruby script? 🤷

danielebogo:
It works as I have the same result! I'm wondering if our default file is UK or US 🤔

AliSoftware:
I've added a couple of `KNOWN_WORDS` to remove some false positives (`url`, `rss`, `Heyooo`, …), and also extracted the language to use for spellcheck in a `CHECK_LANG` constant.

> I'm wondering if our default file is UK or US 🤔

Now running the command with:
 - `CHECK_LANG = 'en-US'` => 23 typos found
 - `CHECK_LANG = 'en-UK'` => 10 typos found
🤔

danielebogo (approved):
It works well! Thanks for adding this!

AliSoftware:
@danielebogo what about next steps now (see PR description)?

We can probably merge this PR as is, to at least have the script available to be run manually.
But before we do a follow-up PR to make it run on CI, we'll need to:
 - Fix the existing typos, so that a CI check won't be red when the check starts being implemented
 - Finalize the proper setup (add `KNOWN_WORDS` if appropriate, decide on the appropriate `CHECK_LANG` locale to use…)

And then we can add a step in the pipeline to make this run on every PR on CI…
 - Which can be easy if all we do is add a step in the `pipeline.yml` to run `brew install aspell && scripts/spell-check-en-strings.rb` on every PR…
 - … but does it really make sense to run it on every PR (which would actually require to `generate_strings_file_from_code` first, to extract any new string introduced in code into the `.strings` file first so those newly added strings would be spell-checked during that PR)?!
    - On one end that would allow to detect typos in the PR that introduce them…
    - But OTOH, not only would this not match the timing of when we generate the `.strings` files from code (which we do during `code_freeze`) but this will make it spellcheck all strings over and over on every commit of every PR…
 - Or would it make more sense to just run that script during code-freeze, i.e. at the time we run `generate_strings_file_from_code` to generate the `.strings` file from code (aka `genstrings`)?
 - @iangmaia also floated the idea of using Danger to only run this on strings added via a PR… but given the `.strings` file is generated during code-freeze, a PR that changes the `value` used in a `NSLocalizedString(…)` will not be as easy to be detected by Danger compared to a condition like "if the .strings file changed"… and also, I'm not sure how it would work to be able to only spell-check the changed strings, as opposed to everything…

---

Personally I vote for:
 - First maybe convert this script into a lane in the `Fastfile` instead of a `scripts/*.rb` file
 - Then merge this PR as is, so that the spell-check can be invoked manually
 - Then you take care of deciding what locale `en.lproj` is supposed to be (`en-US` vs `en-GB`), and of fixing the current typos that the script found
 - Then we can discuss the CI integration (in particular, in which pipeline to run the check—I'm thinking only during code-freeze ie when `.strings` file is generated, even if that means typos from `NSLocalizedString` will only be reported during code-freeze not during the PR that introduced them…)

danielebogo:
> First maybe convert this script into a lane in the Fastfile instead of a scripts/*.rb file
Then merge this PR as is, so that the spell-check can be invoked manually
Then you take care of deciding what locale en.lproj is supposed to be (en-US vs en-UK`), and of fixing the current typos that the script found
Then we can discuss the CI integration (in particular, in which pipeline to run the check—I'm thinking only during code-freeze ie when .strings file is generated, even if that means typos from NSLocalizedString will only be reported during code-freeze not during the PR that introduced them…)

@AliSoftware 100% agree with this!

AliSoftware:
@danielebogo I've updated the PR to implement this as a `lane` in the `Fastfile` rather than the script.

This can now be invoked using `bundle exec fastlane spellcheck_strings`.

<details><summary>🖼️ This is how it looks like on my machine</summary>

</details>


## Next steps

 - Merge this PR into `trunk`
 - @danielebogo to fix the 25 typos found in a separate PR at some point.
    - Up to you to decide if you want to keep `spellcheck_locale = 'en-US'` or change it to `en-GB` if you acknowledge that your `en.lproj` is expected to be in `en-GB` (while this is unusual imho, as `en.lproj` is usually `en-US` by convention?)
    - Up to you to add words to the `SPELLCHECK_KNOWN_WORDS` array (line 110 of `Fastfile`) for any exceptions you'd deem reasonable to keep (instead of fixing the corresponding typo the tool found)
 - Discuss CI integration during `code_freeze` separately later (probably lower priority)

<details><summary>Some preliminary brainstorm wrt CI integration</summary>

 - Do we want the `code_freeze` lane to fail if it detect typos right after `generate_strings_file_for_glotpress`?
    - The problem with that is that if it fails and gives you an opportunity to fix those, the `release/*` branch would have already been created, so you can't re-run `code_freeze` lane after fixing the typos.
    - But maybe we should split `code_freeze` into `start_code_freeze` and `complete_code_freeze` to give you the necessary pause to fix typos if any arise?
 - Another option would be to run the check on CI during the beta build pipeline on CI, so that even if there are typos, it doesn't interrupt the `code_freeze` lane and process.
     - The issue with that is that if there are typos you need to fix, you'd have to make a new beta (to re-spellcheck your fixes + backmerge them into `trunk`) and ensure the updated strings are re-send to translators in time…

</details>

danielebogo (approved):
Tested `bundle exec fastlane spellcheck_strings` and it works perfectly!
Thanks for adding this!

AliSoftware:
Side note that if you decide that UK English should be the default after all, you'd need to use `en-GB` (and [not `en-UK` like I initially thought and tried](https://github.com/Automattic/pocket-casts-ios/pull/2265#issuecomment-2411340326)), as `en-GB` is the official ISO639 name for that locale.

After comparing the 2 with the correct locale for `en-GB` this time, I got:

 - `en-US` => Finds 25 typos, including some instances of `cancelled` and `colour` (expected to be `canceled` and `color`, respectively)
 - `en-GB` => Finds 34 typos, including multiple instances of `favorite`, `canceled`, `color` (expected to be `favourite`, `cancelled`, `colour`, respectively)

So it seems that ultimately your `en.lproj` has more words that use the en-US spelling (less typos) than words that use the en-GB spelling (more typos), which suggests to keep `spellcheck_locale = 'en-US'` and just fix the `colour`=>`color` / `cancelled` => `canceled` en-GB spellings with their `en-US` spelling instead.

PR #2303, opened 2024-10-18
Update `release-toolkit` to latest version

AliSoftware:
This updates the `release-toolkit` from `11.x` to the latest version `12.2.1`, in particular to get the fix from https://github.com/wordpress-mobile/release-toolkit/pull/601

This should solve the problem that @SergioEstevao and @danielebogo encountered in the recent sprints when running `new_beta_release` from their machine while an old `merge/release-x.y.into-trunk` intermediate branch was still existing on their local working copy from doing a previous backmerge earlier in their sprint.

## To test

Not really testable outside doing normal release management operations for real.
I guess we'll see during the next `code_freeze` and `new_beta_release` if that works as expected.

## Checklist

- [x] ~I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.~ _This is an internal tooling change_
- [x] ~I have considered adding unit tests for my changes.~ _This is an internal tooling change_
- [x] ~I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.~ N/A


Gemfile
```diff
@@ -9,7 +9,7 @@ gem 'danger-dangermattic', '~> 1.0'
 gem 'fastlane', '~> 2.216'
 gem 'fastlane-plugin-appcenter', '~> 2.1'
 gem 'fastlane-plugin-sentry', '~> 1.14'
-gem 'fastlane-plugin-wpmreleasetoolkit', '~> 11.0', '>= 11.0.2'
+gem 'fastlane-plugin-wpmreleasetoolkit', '~> 12.2'
```
AliSoftware, line 12: I have checked [our Migration Guide from 11.x to 12.0](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md#from-11x-to-1200) and didn't see any changes listed there that would apply to what's used in PCiOS's `Fastfile`. But feel free to double-check to be sure I didn't overlooked something!

iangmaia: Ah, just noticed something the `CHANGELOG.md` mentions but `MIGRATION.md` doesn't: `ios_current_branch_is_hotfix` has been removed, so `finalize_release` will fail.

I had the same problem recently on WPiOS and [had to add a similar function](https://github.com/wordpress-mobile/WordPress-iOS/pull/23674/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R129).

AliSoftware: Ah, nice catch. ~We should update the `MIGRATION.md` so that it's not missed by migrations of other apps!~ [EDIT] nvm, it's already listed in the `CHANGELOG.md` alongside all other removed methods after all, it's just that I missed it the first time around 🙃

iangmaia: ~Oh, there's also `create_release`, now it should be `create_github_release`.~ NVM, the action isn't really being called 😅 it's just a `create_release` flag.

AliSoftware: > `ios_current_branch_is_hotfix` has been removed, so finalize_release will fail.

This should now be fixed in 264cf8e. Thanks for linking to your WPiOS implementation for inspiration!

> ~Oh, there's also `create_release`, now it should be `create_github_release`.~ NVM, the action isn't really being called 😅 it's just a `create_release` flag.

Yeah, almost made the same mistake when I searched the Fastfile for the various methods in `MIGRATION.md` 😅

fastlane/Fastfile
```diff
@@ -14,9 +14,6 @@ APP_STORE_METADATA_FOLDER = File.join(FASTLANE_FOLDER, 'metadata')
 ARTIFACTS_FOLDER = File.join(PROJECT_ROOT_FOLDER, 'artifacts')
 SECRETS_FOLDER = File.join(Dir.home, '.configure', 'pocketcasts-ios', 'secrets')
 VERSION_XCCONFIG_PATH = File.join(PROJECT_ROOT_FOLDER, 'config', 'Version.xcconfig')
-# Unfortunately, ios_current_branch_is_hotfix relies on this ENV var under the hood.
-# We can't get rid of it just yet.
-ENV['PUBLIC_CONFIG_FILE'] = VERSION_XCCONFIG_PATH
```
AliSoftware, line 19: A quick search of `PUBLIC_CONFIG_FILE` in the source code of `release-toolkit` confirms that this ENV var is not referenced by any other action from `release-toolkit` anymore.

A quick search of `PUBLIC_CONFIG_FILE` in this repo also confirms that this ENV var doesn't seem to be used by the `Fastfile` directly either.

So I'm pretty sure it's safe to remove now.

iangmaia (approved):

PR #2370, opened 2024-10-31
[Up Next Shuffle] Enable the Feature for Plus and Patron only

danielebogo:
| 📘 Part of: [Up Next Shuffle](https://github.com/orgs/Automattic/projects/1179/views/1)
|:---:|

This PR adds:
• The feature available for Plus and Patron user only
• Hides the shuffle button and clear button if the queue is empty
• Update Changelog

| Plus Icon | Default Icon | Empty Screen |
| -------- | ------- | ------- |
| RocketSim_Screenshot_iPhone_16_6 1_2024-10-31_14 37 54 | RocketSim_Screenshot_iPhone_16_6 1_2024-10-31_14 38 19 | RocketSim_Screenshot_iPhone_16_6 1_2024-10-31_14 38 36 |

## To test

- CI must be 🟢 
- Run the app with the StoreKit scheme or run this PR on a real device with a sandbox account
- Run a fresh install and log in with a non plus/patron user
- Add few episodes in the queue
- Go to Up Next 
- Confirm the Shuffle button gold is present
- Clear the queue
- Confirm both shuffle and clear button disappear from the view
- Add few episodes again
- Go back to Up Next
- Tap shuffle gold
- Confirm the paywall is presented
- Purchase a plan
- When the paywall is dismissed confirm the Shuffle icon is the gray one
- Confirm tapping the shuffle doesn't promtp the paywall and enable/disable the feature

## Checklist


SergioEstevao:
Very edge case but when testing this when switching from a Patron account ( where I had Shuffle active) to a free account the shuffle option stayed active on the free active, and I couldn't disable it with out going trough the paywall.

I think the issue here, is that the UpNext that is part of the tab, only updates on startup, so you probably need to refresh it on logout notification.

It will also be good to reset the shuffle setting when logging out of an account so it goes to the disable state.

SergioEstevao:
If you open the the app, without any account installed, then create an account but without signing up for Plus or Patron the shuffle button stays inactive until I restart the app.

SergioEstevao (changes requested):
See my comments about logging out from an account that has Patron/Plus

danielebogo:
Thansk @SergioEstevao ! The PR should be ready for another look.

> If you open the the app, without any account installed, then create an account but without signing up for Plus or Patron the shuffle button stays inactive until I restart the app.

I think this is because of the FF being enabled after the view did load. I realized that this may affect the old button too

> It will also be good to reset the shuffle setting when logging out of an account so it goes to the disable state.

I added the check of the user login state

podcasts/UpNextViewController.swift
```diff
@@ -221,10 +219,47 @@ class UpNextViewController: UIViewController, UIGestureRecognizerDelegate {
     }
 
     @objc private func themeDidChange() {
-        let unselected = UIImage(named: "shuffle")?.withTintColor(AppTheme.colorForStyle(.primaryIcon02, themeOverride: themeOverride), renderingMode: .alwaysOriginal)
-        let selected = UIImage(named: "shuffle-enabled")?.withTintColor(AppTheme.colorForStyle(.primaryIcon01, themeOverride: themeOverride), renderingMode: .alwaysOriginal)
-        shuffleButton.setImage(unselected, for: .normal)
-        shuffleButton.setImage(selected, for: .selected)
+        if !SubscriptionHelper.hasActiveSubscription() || !SyncManager.isUserLoggedIn() {
+            shuffleButton.setImage(UIImage(named: "shuffle-plus"), for: .normal)
```
SergioEstevao, line 223: I believe you also nee to change the `selected` status to the `shuffle-plus` image, or else if it was on the selected state when a subscription was active it will swill stay like that.
Related to this we also should disable shuffle if no subscription is active, because at the moment it will stay active if you had if before with a subscription.

SergioEstevao (changes requested):
While the action of the shuffle updates correctly when disabling a subscription the look of the button and actual shuffle state still need to be changed.

podcasts/UpNextViewController.swift
danielebogo: Good catch @SergioEstevao ! Ready for another look!

SergioEstevao (approved):
All working correctly now! :shipit:

PR #2533, opened 2024-12-10
[Tooling] Do some `Fastfile` cleanup

AliSoftware:
Just some cleanup and maintenance in preparation for some more upcoming automation work on the `Fastfile` soon

- Remove extra instructions from Slack message—because those instructions are now in the MC scenario and we prefer to only have one source of truth
- Remove `after_confirming_push` helper—because this will become obsolete once we migrate to the lanes being run by CI
- Clean up and standardize YARD doc for lanes

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1183,21 +1183,17 @@ inal copy, and try again.")
     message_root = lambda { |tag, display_name|
       ":announcement: <#{GITHUB_URL}/releases/tag/#{tag}|*#{display_name}*>"
     }
-    appstoreconnect_url = 'https://appstoreconnect.apple.com/apps/414834813'
-    testflight_submit_link = "<#{appstoreconnect_url}/testflight/ios|App Store Connect>"
-    appstore_submit_link = "<#{appstoreconnect_url}/appstore/ios/version/deliverable|App Store Connect>"
-    merge_pr_link = "<#{GITHUB_URL}/pulls?q=is%3Apr+is%3Aopen+#{build_number}+into|merge the PR>"
 
     if is_beta
       if (build_number_split[3] || '0') == '0'
-        "#{message_root.call(build_number, version)} code freeze is completed.\nPlease submit #{build_number} for testers on #{testflight_submit_link} and #{merge_pr_link} (<https://wp.me/PdeCcb-1ju|need help?>)"
+        "#{message_root.call(build_number, version)} code freeze is completed."
       else
-        "#{message_root.call(build_number, build_number)} beta has been submitted to Apple.\nPlease distribute #{build_number} to testers on #{testflight_submit_link} and #{merge_pr_link} (<https://wp.me/PdeCcb-1ku|need help?>)"
+        "#{message_root.call(build_number, build_number)} beta has been submitted to Apple."
       end
     elsif (build_number_split[2] || '0').to_i.positive?
-      "#{message_root.call(version, version)} hotfix has been uploaded to Apple.\nPlease submit it for review on #{appstore_submit_link} and #{merge_pr_link}."
+      "#{message_root.call(version, version)} hotfix has been uploaded to Apple."
     else
-      "#{message_root.call(version, version)} final build has been uploaded to Apple.\nPlease submit it for review on #{appstore_submit_link} and #{merge_pr_link}."
+      "#{message_root.call(version, version)} final build has been uploaded to Apple."
     end
   end
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -513,40 +513,14 @@ platform :ios do
     trigger_buildkite_release_build(branch: branch_to_build, beta: false)
   end
 
-  # Builds the Pocket Casts app and uploads it to TestFlight, for beta-testing or final release
+  # Builds the app binary for distribution to App Store Connect
   #
-  # @param beta_release [Boolean] If true, the GitHub release will be marked as being a pre-release
-  # @param skip_prechecks [Boolean] If true, don't run the prechecks and ios_build_preflight (default: false)
-  # @param skip_confirm [Boolean] If true, avoids any interactive prompt (default: false)
-  # @param create_release [Boolean] If true, creates a GitHub Release draft after the upload, with zipped xcarchive as artefact
+  # Once the app binary is built by this lane, you might want to call:
+  #  - `upload_app_store_connect_build_to_testflight` to upload the binary to TestFlight
+  #  - `symbols_upload` to upload dSYMs to Sentry
+  #  - `create_release_on_github(beta_release: …)` to create the GitHub Release
   #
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -1105,9 +1117,13 @@ inal copy, and try again.")
     )
   end
 
-  # - Loop though all PRs which were still open and were assigned the milestones of the code-frozen version,
-  #   and move them to the next milestone + leaving a PR comment about it
+  # Update the milestone of still-open PRs
+  #
```
dangermattic, line 1121: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-ios/blob/7cbbdbbc913b3a3c47ffe9e3209e81a16981863c/fastlane/Fastfile#L1121"></span>Layout/TrailingWhitespace: Trailing whitespace detected.</td>
    </tr>
  </tbody>
</table>

fastlane/Fastfile
```diff
@@ -306,15 +314,15 @@ platform :ios do
...
-      create_backmerge_pr
+    push_to_git_remote(tags: false)
 
-      # We cannot use the `copy_branch_protection` action here yet, as it would create a branch protection rule with the same
-      # restrictions we have in `trunk`, and the release managers wouldn't be able to push due to permissions.
-      # This should be changed only when we have PCiOS releases done on CI, when the CI bot is the one running `git push`.
-      set_branch_protection(repository: GHHELPER_REPO, branch: release_branch_name)
-    end
+    trigger_beta_build(branch_to_build: release_branch_name)
+    create_backmerge_pr
```
leandroalonso, line 320: @AliSoftware I left a question on Slack about this one.

I know this is going in the direction of running everything to a machine but:

1. How it is now, it will open the back merge PR even if the beta build fails for any reason. (while before it would only open the PR if the build succeeded)
2. [PRs are not created](https://github.com/Automattic/pocket-casts-ios/pulls?q=is%3Apr+author%3Apocketcasts+is%3Aclosed) by our pocketcasts user anymore
3. Since the PR is created by whoever ran the command, someone else from the team needs to approve. This is a small friction but the process was super lean before. Most of the time the release manager would approve it.

I believe moving it to a machine will reinstate the `pocketcasts` user as the creator of those PRs but creating it even after the build failed seems a bit wrong.

AliSoftware: Yes the fact that the PR is not created by the bot will be fixed once I make those lanes be run by CI (but I got delayed a bit on that work, thought I'd have happened sooner, sorry)

As for creating the backmerge PR before the build has finished, I did this change intentionally (via https://github.com/Automattic/pocket-casts-ios/commit/b3fa8e7fc8a1522dfdc18915257b29f0bcca6171) back then.

My rationale for it was to:
 - Create the backmerge PR ASAP, so that the Release Manager would not have to wait for the beta or release build to finish building before the backmerge PR is created only at the end of the build… which would then itself trigger its own set of CI checks (the classical ones for any PR, like unit tests etc), and thus require you to wait even more for that CI build to finish too before you can merge the backmerge PR.
 - Instead, now the 2 CI builds (build for the beta/release and CI checks for the backmerge PR) are run in parallel, leading you to less wait time and ability to review and approve the backmerge PR sooner[^1]
 - In theory at least, I wouldn't expect the beta build to fail to build, because everything that lands in the `release/*` branch did that via a dedicated PR already, so whatever is in `release/*` and being built by the beta build in CI is supposed to have already been proven to build successfully when it was validated in the PR landing the beta fix(es).
 - The only exception I could think of for this if the CI build for the beta fail not due to a compilation error but due to e.g. an ASC API error, or some issue when uploading the build to TestFlight. But those should be rare and in most cases either you'd not need to add any new commit to fix it (e.g. just a retry or similar), or you'd retrigger a `new_beta_release` after having pushed a commit to fix the issue, and that will close the previous backmerge PR and create a new one.

[^1]: All of this setup was assuming that we'd have moved to the lanes being run by CI sooner, and thus that those backmerge PRs would be created by the bot that runs the CI, allowing the Release Manager to approve and merge them themselves. This is still the plan, it just won't happen until January do to unforeseen delays on my project for this. But of course I understand that the fact that currently that's not yet the case and the PR is currently authored by the Release Manager instead of the bot kinda throws a spanner in the works of this reasoning until it will happen.

---

That being said, I'm OK if you prefer that we revert to do the backmerge PR only at the end of the CI build, even if this means the process will be back to be sequential and will require you to wait longer before you can start reviewing the PR.
But I just wanted to highlight my initial rationale in case it helped making sense of the initial plans 😉 

_PS: Also if we end up deciding to revert to doing the backmerge PR only after the build has finished, I'll probably do that change in a follow-up PR rather than this one, to keep the diff easier to review and on topic._

leandroalonso: > I'll probably do that change in a follow-up PR rather than this one, to keep the diff easier to review and on topic.

Oh, absolutely! Don't worry!

I think we're aligned in terms of speeding up the waiting time. 🤓 My main concern with the PR being open by a member it's that the release manager itself cannot approve. So if they want to move quickly that might require pinging someone else. (That's what happened today, by the way)

> Create the backmerge PR ASAP, so that the Release Manager would not have to wait for the beta or release build to finish building before the backmerge PR is created only at the end of the build… which would then itself trigger its own set of CI checks (the classical ones for any PR, like unit tests etc), and thus require you to wait even more for that CI build to finish too before you can merge the backmerge PR.

Regarding that, I just want to mention that my workflow wasn't affected by the CI time because I would just approve and enable auto-merge. Then I'd just move on.

I think your rationale is good, the only issue is that one of us opening the PR requires someone else to approve, so a lot of time it will require a ping to someone else.

> The only exception I could think of for this if the CI build for the beta fail not due to a compilation error but due to e.g. an ASC API error, or some issue when uploading the build to TestFlight. But those should be rare 

That happened in the past. 😅 And I believe more recently we had the Sentry part failing.

Anyway, I think my only request would be to open the PR with the `pocketcasts` user, the rest is fine. But if we're doing that by January then it's fine to wait.

A temporary solution might be to use the Github API token from the `.mobile-secrets` to open the PRs, but I'll leave the decision to you.

Thanks for taking the time to explain!

fastlane/Fastfile
```diff
@@ -306,15 +314,15 @@ platform :ios do
...
+
+    # We cannot use the `copy_branch_protection` action here yet, as it would create a branch protection rule with the same
+    # restrictions we have in `trunk`, and the release managers wouldn't be able to push due to permissions.
+    # This should be changed only when we have PCiOS releases done on CI, when the CI bot is the one running `git push`.
```
iangmaia, line 324: 👍

iangmaia (approved):
Nice improvements and clean-up 🧹

PR #2534, opened 2024-12-10
Add flagged logout avoidance from background task

### LINKED ISSUE #2529: App gets logged out randomly
### Description

We've had a few users report that they are randomly getting logged out. Internal thread: p1729595145505109-slack-C02A333D8LQ

Something in common we've been noticing is that these users are using Google Sign In. 

We've also seen cases where this happens when a VPN is used

@danielebogo was able to replicate, but it is not something we have found easy to replicate on demand. 

n/a

- [x] I have searched for existing bug reports.

_No response_

bjtitus:
Fixes #2529, an issue reported by users where they were logged out coming back to the app.

From the logs, we log users out during the background refresh task, potentially because data is not accessible due to encryption. This change only logs the user out if the app is currently foregrounded. One downside to this might be increased requests which result in 401s. However, I think it's a better user experience to avoid logout unless the app is foregrounded when we can be more sure stored data is available for access.

An alternative could be to pass through a value indicating that a particular request came from the background refresh task and only log out in that case.

This change is flagged with the `avoid_logout_in_background` flag in case we need to disable it.

## To test

* Apply this patch ([ForcedLogoutRefresh.patch](https://github.com/user-attachments/files/18131573/ForcedLogoutRefresh.patch)) to either this branch or `trunk`
* Add a breakpoint at `AppDelegate:110` in `didFinishLaunching` after `scheduleNextBackgroundRefresh()` is called.
* Build and run the app
* When the breakpoint is hit, background the app
* Execute this command in the debug console to trigger the scheduled background refresh task:
```
e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"au.com.shiftyjelly.podcasts.Refresh"]
```
* Before this change, you should have been logged out
* After this change, you should see `TokenHelper: Skipped logout in background due to error` logged

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Private/TokenHelper.swift
@@ -85,17 +85,21 @@ class TokenHelper {
             ServerSettings.setRefreshToken(refreshedRefreshToken)
         }
         else {
-            if ServerConfig.avoidLogoutOnError {
-                // if the user doesn't have an email and password or SSO token, they aren't going to be able to acquire a sync token
-                switch error as? APIError {
-                case APIError.TOKEN_DEAUTH?, APIError.PERMISSION_DENIED?:
-                    tokenCleanUp()
-                default:
-                    // Do nothing so the user is not disrupted in the case of non-auth errors
-                    FileLog.shared.addMessage("TokenHelper: Unable to acquire token but avoided logout due to error: \(String(describing: error))")
-                }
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso (approved):

PR #2539, opened 2024-12-12
Protobuf: ensure protoc and swift-protobuf is up-to-date

leandroalonso:
While updating our protobuf files I noticed that a lot of changes were being reverted. After talking to the team I realized my libraries were completely outdated.

This PR ensures protoc and swift-protobuf are updated before updating the files.

## To test

1. Run `make update_proto` pointing to the correct pocket-casts-sync-api path in your machine
2. Check that `Sendable` stuff is not removed
3. Run `protoc --version`
4. Output should be `libprotoc 29.1`
5. Run brew upgrade swift-protobuf
6. Output should have `Warning: swift-protobuf 1.28.2_1 already installed`

## Checklist

### DIFF
--- a/scripts/update_proto.sh
+++ b/scripts/update_proto.sh
@@ -13,5 +13,6 @@ then
     exit 1
 fi
 
+brew upgrade protobuf
 protoc --swift_out=./Modules/Server/Sources/PocketCastsServer/Private/Protobuffer --proto_path=$API_BASE_FOLDER/ $API_BASE_FOLDER/api.proto
 protoc --swift_out=./Modules/Server/Sources/PocketCastsServer/Private/Protobuffer --proto_path=$API_BASE_FOLDER/ $API_BASE_FOLDER/files.proto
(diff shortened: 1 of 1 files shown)


danielebogo (approved):
Just did the test and it works... no Sendable removed and I got the same warning for the swift protobuf

scripts/update_proto.sh
```diff
@@ -13,5 +13,8 @@ then
     exit 1
 fi
 
+brew upgrade protobuf
+brew upgrade swift-protobuf
```
bjtitus, line 17: Could we check that `brew` is available before running these? Maybe:

```suggestion
if command -v brew >/dev/null 2>&1; then
    echo "Updating protobuf and swift-protobuf with homebrew"
    brew upgrade protobuf
    brew upgrade swift-protobuf
else
    echo "Homebrew is not installed, check protobuf and swift-protobuf versions manually"
fi
```

bjtitus (changes requested):
Sorry, I think I missed this before merge @leandroalonso

PR #2541, opened 2024-12-12
Stats: apply API changes to prevent reset

leandroalonso:
We had a few instances of users' complaining that their Stats reset or reduced. This PR applies the latest changes deployed on Staging ensuring:

1. That the backend never "reduces" (or reset) any Stat;
2. That we pull back the stats in case we need

## To test

1. First, change the `Build Configuration` to `Staging` and run the app
2. Login to a staging account
3. Ensure this account has some Stats. If it doesn't, [manipulate these lines](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Modules/Server/Sources/PocketCastsServer/Public/StatsManager.swift#L23-L29) and trigger a refresh (just add numbers there)
4. Then, remove all the `updateQueue.sync` block from the `StatsManager` `init` method. This will simulate having negative stats
5. Go to Stats
6. ✅ Reduced Stats should appear but it should update to the correct one
7. Re-run the app
8. Trigger a refresh from Profile
9. Go to Stats
10. ✅ Stats should be correct

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/StatsManager.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/StatsManager.swift
@@ -30,7 +30,7 @@ public class StatsManager {
     }
 
     func updateStatsIfNeeded(savedDynamicSpeed: TimeInterval, savedVariableSpeed: TimeInterval, totalListenedTo: TimeInterval, totalSkipped: TimeInterval, savedAutoSkipping: TimeInterval) {
-        let minimumStatsChangeToUpdate = 100
+        let minimumStatsChangeToUpdate: TimeInterval = 100
 
         updateQueue.sync {
             if savedDynamicSpeed - self.savedDynamicSpeed > minimumStatsChangeToUpdate {
--- a/Modules/Server/Sources/PocketCastsServer/Public/StatsManager.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/StatsManager.swift
@@ -30,28 +30,30 @@ public class StatsManager {
     }
 
     func updateStatsIfNeeded(savedDynamicSpeed: TimeInterval, savedVariableSpeed: TimeInterval, totalListenedTo: TimeInterval, totalSkipped: TimeInterval, savedAutoSkipping: TimeInterval) {
+        let minimumStatsChangeToUpdate = 100
+
         updateQueue.sync {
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

Modules/Server/Sources/PocketCastsServer/Public/StatsManager.swift
```diff
@@ -29,6 +29,39 @@ public class StatsManager {
         }
     }
 
+    func updateStatsIfNeeded(savedDynamicSpeed: TimeInterval, savedVariableSpeed: TimeInterval, totalListenedTo: TimeInterval, totalSkipped: TimeInterval, savedAutoSkipping: TimeInterval) {
+        let minimumStatsChangeToUpdate = 100
```
leandroalonso: I'm debating whether this should be `100` or something bigger or smaller.

Maybe if the local stat is just smaller than remote we should update?

Modules/Server/Sources/PocketCastsServer/Public/StatsManager.swift
```diff
@@ -29,6 +29,39 @@ public class StatsManager {
         }
     }
 
+    func updateStatsIfNeeded(savedDynamicSpeed: TimeInterval, savedVariableSpeed: TimeInterval, totalListenedTo: TimeInterval, totalSkipped: TimeInterval, savedAutoSkipping: TimeInterval) {
+        let minimumStatsChangeToUpdate: TimeInterval = 100
+
+        updateQueue.sync {
+            if savedDynamicSpeed - self.savedDynamicSpeed > minimumStatsChangeToUpdate {
```
SergioEstevao, line 36: Nit: Do you want to save only if the difference is positive right? No need to use `abs` ?

SergioEstevao (approved):
Followed the test instructions and the stats recovered to the correct values.
Just left a small nit that you may want to consider, but feel free to ignore if does not make sense for the scenario.
:shipit:

leandroalonso:
@srjakes are the server changes deployed?

srjakes:
> are the server changes deployed?

The server changes have been deployed.

PR #2555, opened 2024-12-16
Add a configuration to allow the display of the Toast above the mini-player

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes # 

| 1 | 2 |
| - | - |
| Simulator Screenshot - iPhone 16 Pro - 2024-12-16 at 14 20 59 | Simulator Screenshot - iPhone 16 Pro - 2024-12-16 at 14 20 49 |

## To test

- Start the app with a user with a subscription
- Start playing an episode so the mini-player is visible
- Go to UpNext tab
- Activate UpNext Shuffle
- See if the Toast shows above the Mini-Player
- Tap on UpNext icon the mini-player
- Activate UpNext Shuffle
- - See if the Toast shows at the regular position

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/Common SwiftUI/Toast/ToastView.swift
```diff
@@ -109,10 +109,10 @@ struct ToastView<Style: ToastTheme>: View {
         .shadow(color: .black.opacity(0.3), radius: 10)
 
         // Animates the toast in from the bottom of the screen
+        .miniPlayerSafeAreaInset(multiplier: viewModel.aboveMiniPlayer ? 1.5 : 0)
```
leandroalonso, line 112: We'll probably forget this if we change the mini-player and the toast will probably appear incorrectly again. Also, given this is a property I see one of us repeating the same issue 

I wonder if there's any way we can:

1. Detect mini player is visible;
2. Set this value based on the mini player height/position

However, such a solution still couple this to the mini-player and might require some additional hack code. 🤷

SergioEstevao: The modifier actual ensures that the miniPlayer is taking in account the issue we have is the tabbar, while on the other places the standard safeInset ensure thats the bottom already takes in account the tab bar, on the toast because we are on another Window the tab bar is not taken in account.
Do you see a way to detect the tab bar presence and size from another window?

danielebogo: @leandroalonso @SergioEstevao this solution it's similar to what Android team has implemented. In our codebase we have other classes, like the `InsetAdjuster`, that takes the same argument to adjust the padding.

However, if we want to automatically adjust it without injecting any parameters, we can use the `NavigationManager` (with some adjustments) as source to get the miniplayer and tabbar access.

bjtitus: I think it should be possible to adjust using the `MainTabBarController` at the `ToastWindow` level but I wasn't quite able to get the values right:

```
if windowScene.keyWindow?.rootViewController?.presentedViewController == nil {
    let tabBarController = (windowScene.keyWindow?.rootViewController as? MainTabBarController)
    if let tabBarController, tabBarController.tabBar.isHidden == false {
        let tabBarHeight = tabBarController.tabBar.bounds.height
        controller.additionalSafeAreaInsets = UIEdgeInsets(top: 0, left: 0, bottom: tabBarHeight, right: 0)
    }
}
```

I was hoping the `keyWindow.safeAreaInset` would include the tab bar height but it does not seem to right now and adjusting that would probably mess with a bunch of things.

Would this make sense as an alternative? It's still somewhat messy to pull out that `MainTabBarController` and make sure there's nothing presented.

danielebogo: > Would this make sense as an alternative? It's still somewhat messy to pull out that MainTabBarController and make sure there's nothing presented.

@bjtitus  The `NavigationManager` singleton has this information. `AppDelegate` and the `MainTabBarController` rely on this. If we want to have an auto position of the Toast I think we can use this as source.
Otherwise my suggestion is to use Sergio solution and align with Android

SergioEstevao: @danielebogo and @leandroalonso I updated the code so that instead of hard-coding the miniplayer inset attribute on toast, I added a inset provider parameter that allows the customization of the inset depending of the screen. 
By default it will use the safeAreaInset.bottom but customizations can be made to for example take in account the mini-player.
Give it a look and tell me what you think.

SergioEstevao: @bjtitus ^

leandroalonso (changes requested):
@SergioEstevao when displaying the toast above the mini player I'm not able to swipe it down to dismiss.

SergioEstevao:
> when displaying the toast above the mini player I'm not able to swipe it down to dismiss.

@leandroalonso great catch, I updated the code to support dismiss on that scenario. Can you give it another look?

leandroalonso (dismissed):
@SergioEstevao the first the Toast animation happens the background seems to move in a different place (or speed) than the text. After the first appearance, it works just fine. Check:


https://github.com/user-attachments/assets/dc3f33b2-2e9d-4b2f-95d9-c5b0b050c4e6

I don't remember seeing that before. 🤔 

This is meant to be a simple change, so I think that if the newest code changes are introducing more complexity/issues maybe we should just stick with the first one. But it's up to you!

danielebogo:
Version `7.80` has now entered code-freeze, so the milestone of this PR has been updated to `7.81`.

SergioEstevao:
@leandroalonso just reverted my latest changes, and went back to the original changes.

Can you please give it another look?

danielebogo (approved):

PR #2556, opened 2024-12-16
Download copy error logging

bjtitus:
* Moves non-throwing code out of the `do`/`catch` block so it's clear which code is throwing
* Adds error logging when we fail to copy the downloaded file. This is the point at which we show an error message saying "have you run out of space", which may not be accurate depending on the error.

## To test

* Ensure episode downloads still work
* In `StorageManager.copyItem(at: URL, to: URL)` you can add the following: `throw CocoaError(.fileNoSuchFile)` and ensure this error is logged when downloading:

```
DownloadManager: Failed to copy downloaded file from location: <path> to destination: <path>  error: Error Domain=NSCocoaErrorDomain Code=4 "The file doesn’t exist."
```

## Checklist


SergioEstevao (approved):
:shipit:

Issue #2550, opened 2024-12-13
App Clip Target


bjtitus:
Add an App Clip target with all necessary code to run a basic playback page.

bjtitus:
This has been completed with new target.

Issue #2549, opened 2024-12-13
Project: App Clip


bjtitus:
| **Project Thread** |
|-|
| pdeCcb-7H8-p2 |

[Github Project](https://github.com/orgs/Automattic/projects/1219/)

bjtitus:
| **Project Thread** |
|-|
| pdeCcb-7H8-p2 |

SergioEstevao:
Version `7.81` has now entered code-freeze, so the milestone of this PR has been updated to `7.82`.

pocketcasts:
Version `7.82` has now entered code-freeze, so the milestone of this PR has been updated to `7.83`.

pocketcasts:
Version `7.83` has now entered code-freeze, so the milestone of this PR has been updated to `7.84`.

pocketcasts:
Version `7.84` has now entered code-freeze, so the milestone of this PR has been updated to `7.85`.

mokagio:
🎉

PR #2566, opened 2024-12-18
App Clip: Add target and player implementation

bjtitus:
| 📘 Part of: #2549 |
|:---:|

Implements #2550

* Adds the App Clip target
* Adds necessary files and ifdefs for Now Playing functionality
* Adds SKOverlay for upgrade prompting

| | | |
|-|-|-|
| Simulator Screenshot - iPhone 16 - 2024-12-17 at 21 13 23 | Simulator Screenshot - iPhone 16 - 2024-12-17 at 21 13 18 | Simulator Screenshot - iPhone 16 - 2024-12-17 at 21 16 32 |

## Review

This is a large PR and I apologize for not being able to break it up smaller. The VAST majority of the changes are in the `project.pbxproj` file to add various files to the App Clip target and many more are from the Asset Catalog JSON files.

If you filter to only the `swift` files, using the "File filter" in Github's UI it should be easier to parse through. All of the modified existing swift files are `ifdefs` for App Clip support. The new App Clip files are just for adding the Now Playing screen, handling the user activity, and showing the `SKOverlay`.

## To test

* Build and run the "Pocket Casts App Clip" target
* Ensure the Hidden Brain episode loads and plays
* Test playback, sleep timer, and air play functionality
* Ensure that the SKOverlay upgrade banner appears after 10 seconds
* Leave the app and check the App Library at the end of your Home Screens. You should see a "Webclips" group which includes the Pocket Casts logo with dashes representing the App Clip. Tapping it will bring you back into the App Clip.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Pocket Casts App ClipUITests/Pocket_Casts_App_ClipUITests.swift
```diff
@@ -0,0 +1,36 @@
+import XCTest
+
+final class Pocket_Casts_App_ClipUITests: XCTestCase {
```
SergioEstevao: Is it worth to keep this default template tests?

SergioEstevao:
Should we have the option to choose global a local effects on the App Clip? If nothing of those are saved for future use maybe it's worth to simplify and not show the top segment bar?

Simulator Screenshot - iPhone 16 Pro - 2024-12-18 at 11 12 19

podcasts/VideoViewController.swift
```diff
@@ -90,7 +90,11 @@ class VideoViewController: SimpleNotificationsViewController, AVPictureInPicture
     @IBOutlet var pipButton: UIButton!
 
     @IBOutlet var airplayButton: UIButton!
+    #if APPCLIP
+    @IBOutlet var castButton: UIButton!
```
SergioEstevao, line 95: Does it make sense to keep the cast button on the App clip? Will it work?

SergioEstevao (changes requested):
App Clip is working correctly.

I'm assuming this is just a step into getting it to work and as a follow up we will have a way to test it when opening a web page?

One thing that I noticed is that when you activate a sleep timer we don't see an animation on the sleep timer button? I'm assuming it's because of the animation library but can we add a new icon that reflects the active status of the sleep timer? maybe just tint with an highlight?

Those aside I also left some nit comments on the code.

Pocket Casts App ClipUITests/Pocket_Casts_App_ClipUITests.swift
bjtitus: I was planning to add unit and UI tests later on, although I had quite a bit of trouble getting the invocation URL to work. I'll remove them for now.

bjtitus:
> I'm assuming this is just a step into getting it to work and as a follow up we will have a way to test it when opening a web page?

Yes, but unfortunately it is only testable in TestFlight or the App Store so we will need to include it in a test version. However, it's only really invoked when certain HTML is added to the web page so we could try only on staging first.

> when you activate a sleep timer we don't see an animation on the sleep timer button

Woops, this is a bug. I've fixed it in this commit.

> Should we have the option to choose global a local effects on the App Clip? If nothing of those are saved for future use maybe it's worth to simplify and not show the top segment bar?

I'll look into this a bit more. I made a separate issue https://github.com/Automattic/pocket-casts-ios/issues/2577 and I'll tackle that in a separate PR.

podcasts/VideoViewController.swift
bjtitus: This needs to remain because the xib sets it and it will crash otherwise. It isn't used in the App Clip, though.

You can test the video playback by changing the Invocation URL to `https://pca.st/q319mb8y` in the Run Arguments of the App Clip target:
CleanShot 2024-12-19 at 16 39 39@2x

SergioEstevao (approved):
Sleep timer is now working and double checked the cast button and it's also working.
:shipit:

PR #2582, opened 2024-12-20
Disable private feed sharing

bjtitus:
Disables sharing of private feeds with a Toast message shown when trying to share.

The core feature is flagged under `disable_private_feed_sharing` but I left the database changes unflagged.

https://github.com/user-attachments/assets/ea3ecd31-e997-49c3-aa71-5db3929743d5

## To test

1. Try sharing a podcast and make sure it works
2. Try sharing a private podcast (try this URL: `https://pca.st/private/707a8bf0-9abf-013c-74fe-027201e7e97f`)
3. Ensure this message appears in a toast: "Sharing is not available for private podcasts"

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #2587, opened 2024-12-26
Move podcastShared analytics events

bjtitus:
This handles the early returns added in #2582 so we don't log a `podcastShared` event when private podcasts are shared.

### Private Podcast

* Enable the `tracksLogging` feature flag

* Open the Podcast detail screen for a private podcast
* Tap "Share" and verify that only the following event is logged with no `podcast_shared` event:
```
🔵 Tracked: podcast_screen_share_tapped
```

* Open the episode detail page
* Verify that no event is logged when tapping share

* Play an episode from this podcast
* Open the player
* Tap the "Share" action button in the shelf
* Verify that no `podcast_shared` event is logged

### Public Podcast

* For a public podcast, do the following:
* Open the Podcast detail page
* Tap Share and see the following logged:
```
🔵 Tracked: podcast_shared ["type": "podcast", "source": "podcast_screen"]
```

* Open the Episode detail page
* Tap Share and select the options and you should see the following logged with the "type" swapped out as above:
```
🔵 Tracked: podcast_shared ["source": "episode_detail", "type": "episode"]
```

* Open the player
* Tap the "Share" action button in the shelf
* Select one of the options from the menu and you should see the following logged for each corresponding option:
```
🔵 Tracked: podcast_shared ["type": "episode", "source": "player"]
🔵 Tracked: podcast_shared ["type": "podcast", "source": "player"]
🔵 Tracked: podcast_shared ["type": "current_position", "source": "player"]
🔵 Tracked: podcast_shared ["type": "clip", "source": "player"]
```

## Checklist

### DIFF
--- a/podcasts/NowPlayingPlayerItemViewController+Shelf.swift
+++ b/podcasts/NowPlayingPlayerItemViewController+Shelf.swift
@@ -499,11 +499,7 @@ extension NowPlayingPlayerItemViewController: NowPlayingActionsDelegate {
     }
 
     private func shareEpisode(source: UIView, episode: Episode, fromTime: TimeInterval) {
-        guard let buttonSuperview = source.superview else { return }
-
-        let type = fromTime == 0 ? "episode" : "current_position"
-
-        Analytics.track(.podcastShared, properties: ["type": type, "source": "player"])
+        guard let _ = source.superview else { return }
 
         if fromTime == 0 {
             SharingModal.show(option: .episode(episode), from: analyticsSource, in: self)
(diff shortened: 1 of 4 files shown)


danielebogo (approved):
Hey @bjtitus ,
everything seems working ok as you described. However, for a private podcast, from the fullscreen player, when I tap the share button in the shelf, I see:
`🔵 Tracked: player_shelf_action_tapped ["from": "overflow_menu", "action": "share_episode"]`

Is it expected?

bjtitus:
I think we should keep both the `player_shelf_action_tapped` and `podcast_screen_share_tapped` events since the button is being tapped. Android is doing the same so we shouldn't be out of sync there. In my mind, the button is still being tapped. We _could_ add a property that indicates the button is disabled.

PR #2598, opened 2025-01-03
Add entry for App Clip in list of targets + DRY code sign config

mokagio:
Builds on top of https://github.com/Automattic/pocket-casts-ios/pull/2591 with the changes I used to update the provisioning profiles for the App Clip after adding the app group capability and test the builds locally.

I thought that the entry for the app clip target in `Fastfile` was there already because there were provision profiles in the store generated for it via Fastlane. I must have forgotten to commit it at the time I worked on it...

See also pdnsEh-1Xy-p2

---

When I tried to archive an App Store build with these changes, I got the following failure: "error: Cycle inside podcasts; building could produce unreliable results." It happened via Fastlane build as well as Xcode.

image

```
error: Cycle inside podcasts; building could produce unreliable results.
Cycle details:
→ That command depends on command in Target 'podcasts': script phase “[CP] Embed Pods Frameworks”
○ That command depends on command in Target 'podcasts': script phase “Modify Plist”
○ Target 'podcasts' has process command with output '/Users/gio/Developer/a8c/pcios/DerivedData/podcasts/Build/Intermediates.noindex/ArchiveIntermediates/pocketcasts/InstallationBuildProductsLocation/Applications/podcasts.app/Info.plist'
○ Target 'podcasts' has copy command from '/Users/gio/Developer/a8c/pcios/DerivedData/podcasts/Build/Intermediates.noindex/ArchiveIntermediates/pocketcasts/BuildProductsPath/Release-iphoneos/Pocket Casts App Clip.app' to '/Users/gio/Developer/a8c/pcios/DerivedData/podcasts/Build/Intermediates.noindex/ArchiveIntermediates/pocketcasts/BuildProductsPath/Release-iphoneos/podcasts.app/AppClips/Pocket Casts App Clip.app'


Raw dependency cycle trace:

target:  ->

node: <all> ->

command: <all> ->

node: /Users/gio/Developer/a8c/pcios/DerivedData/podcasts/Build/Intermediates.noindex/ArchiveIntermediates/pocketcasts/InstallationBuildProductsLocation/Applications/podcasts.app/_CodeSignature ->

command: P0:target-podcasts-1831d32372b889e9ea6d6e495914e180a889c15545cabe47fe671f7f3b81e86e-:Release:CodeSign /Users/gio/Developer/a8c/pcios/DerivedData/podcasts/Build/Intermediates.noindex/ArchiveIntermediates/pocketcasts/InstallationBuildProductsLocation/Applications/podcasts.app ->

[...]

Issue #43, opened 2022-07-12
Siri Suggestions: artwork not appearing on lock screen while playing with Siri Suggestions


pocketcasts:
A user reported that the artwork on the playback control is not appearing on the lock screen. According to the screenshot, it looks like the playback was initiated from Siri Suggestions.

> Thumbnails aren't loading in iOS's lock screen media control. This is with the latest iOS and Pocket Casts versions (non-beta):

On iOS 15.2 and app version 7.19.1.

<img src="https://user-images.githubusercontent.com/40906847/148329162-20d26c73-950f-44d8-9234-0dba03f0668e.png" width="300">

thabotswana:
Another report in 5704082-zen

>1. Remove everything from the queue from the pocket casts
>2. Start playing any episode
>3. Lock the screen
>4. While on the lock screen, scroll to the end of the episode using
>widget
>
>As soon as episode stops playing, “now playing” widget will show
suggestions and you can observe the issue.


https://user-images.githubusercontent.com/35628116/201613621-f543a2ef-e9f1-456d-84d8-ae8cd9362d2d.mp4

rviljoen:
I have tracked the issue down to this section of code:

https://github.com/Automattic/pocket-casts-ios/blob/6f51ce58b5150231e7bdb7b130b0fcec95f8d261/podcasts/SiriShortcutsManager.swift#L174-L176

It [seems like you](https://stackoverflow.com/questions/69292991/does-creating-an-inimage-from-a-url-work) have to init ```INImage``` with image data instead of an URL. 

The following hack resolved the problem:
```Swift
let data = try? Data(contentsOf: podcastUrl)
let podcastArtwork = INImage(imageData: data!)
```

However, doing a synchronise load of the image on the main thread triggers a warning. I think this could be a quick fix for someone with more knowledge of Swift concurrency and the Pocket Casts codebase. 

Below you can see the issue in the App Store build (red icon) and top right is fixed (black icon) using the code above.

![Image](https://user-images.githubusercontent.com/25402814/203647465-b9a45bbd-160c-480c-aa98-49e5e23a0609.png)

PR #2617, opened 2025-01-07
Update project and automation to build with App Clip

### LINKED ISSUE #2601: Exclude App Clip from Prototype build
Enterprise builds don't support App Clips and we use this for our Prototype build.

I think we should be able to exclude the App Clip from the Prototype build but I'm not 100% sure how to accomplish this. The "Embed App Clip" build phase includes a "Code Sign On Copy" field which might help with this.

mokagio:
📘 Part of: #2549   

Fixes https://github.com/Automattic/pocket-casts-ios/issues/2601

As discussed in the issue, this PR makes it possible to produce a prototype build by dynamically removing the App Clip target from the main app target at build time. 

This happens only in Fastlane, however; it is no longer possible to archive the Prototype Build from Xcode. Given that automation via Fastlane should be the primary and mostly used way to generate such builds, the compromise seems acceptable. I did this because it felt cleaner to modify the project file from a script that mostly runs in CI, rather than as part of the build in Xcode.

We _could_ add a build phase run script that check the build configuration and modifies the project if we're archiving a Prototype Build. But, again, given we shouldn't be doing that from Xcode often, or at all, the current implementation seems better to me.

## To test

I unblocked the Prototype Build for this PR [in CI](https://buildkite.com/automattic/pocket-casts-ios/builds/9302#019441f0-93b9-4ad2-b7f2-7def502adcb5) to verify that it works and that the generated app works fine.

I also run `bundle exec fastlane build_app_store_connect set_up_code_signing:false` locally and ensured it passed to make sure that the changes still result in a build that can be uploaded to the App Store. As an aside, that was very useful because it revealed various additional settings to update.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

config/PocketCasts.debug.xcconfig
```diff
@@ -1,5 +1,5 @@
 #include "PocketCasts.base.xcconfig"
 
-PRODUCT_BUNDLE_IDENTIFIER = $PRODUCT_BUNDLE_IDENTIFIER_ROOT
+PRODUCT_BUNDLE_IDENTIFIER = $(PRODUCT_BUNDLE_IDENTIFIER_ROOT)
```
mokagio, line 3: Reading this way prevents Fastlane from picking up `PRODUCT_BUNDLE_IDENTIFIER_ROOT` as one of the bundle identifiers to map a provisioning profile to.

podcasts.xcodeproj/project.pbxproj
```diff
@@ -8550,9 +8550,9 @@
 				BD5231EA1E3EDA6A000F0D38 /* Embed App Extensions */,
 				BDD3018A1EFB9356004A9972 /* Embed Watch Content */,
 				40BF0B3F241F55FE003E3ABD /* Embed Frameworks */,
-				C716FF45296F2BE3006B787D /* Modify Plist */,
-				CD7B4456FFD51509B59727FB /* [CP] Embed Pods Frameworks */,
 				9AC09DDA2D2D4103003EBA96 /* Embed App Clips */,
+				CD7B4456FFD51509B59727FB /* [CP] Embed Pods Frameworks */,
+				C716FF45296F2BE3006B787D /* Modify Plist */,
```
mokagio, line 8555: This is what addressed the cycle that I reported in https://github.com/Automattic/pocket-casts-ios/pull/2598

podcasts.xcodeproj/project.pbxproj
```diff
@@ -12994,14 +12990,12 @@
...
-				MARKETING_VERSION = 1.0;
 				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
 				MTL_FAST_MATH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = au.com.shiftyjelly.podcasts.Clip;
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				PROVISIONING_PROFILE_SPECIFIER = "";
-				"PROVISIONING_PROFILE_SPECIFIER[sdk=iphoneos*]" = "Pocket Casts App Clip Development";
-				SKIP_INSTALL = NO;
+				PROVISIONING_PROFILE_SPECIFIER = "Pocket Casts App Clip Development";
+				SKIP_INSTALL = YES;
```
mokagio, line 12998: With `SKIP_INSTALL` set to `NO` for the App Clip target, Xcode is able to archive and export the app, but `xcodebuild` via Fastlane fails with:

```
error: exportArchive exportOptionsPlist error for key "method" expected one {} but found app-store
```

I don't know if this change will have downstream effects in the distribution capability for the App Clip.

For reference, from https://developer.apple.com/documentation/xcode/build-settings-reference

> SKIP_INSTALL: If enabled, don’t install built products even if deployment locations are active.

pocketcasts:
<details><summary>Pocket Casts Prototype Build📲 You can test the changes from this Pull Request in <b>Pocket Casts Prototype Build</b> by scanning the QR code below to install the corresponding build.</summary>
<table>
<tr>
  <td><b>App Name</b></td><td>Pocket Casts Prototype Build Pocket Casts Prototype Build</td>
</tr>
<tr><td><b>Build Number</b></td><td>pr2617-9098dee</td></tr>
<tr><td><b>Version</b></td><td>7.80</td></tr>
<tr><td><b>Bundle ID</b></td><td>au.com.shiftyjelly.podcasts.prototype</td></tr>
<tr><td><b>Commit</b></td><td>9098dee4a20fa825e43945bb16e9772e351447cd</td></tr>
<tr><td><b>App Center Build</b></td><td><a href='https://install.appcenter.ms/orgs/automattic/apps/Pocket-Casts-Prototype-Builds/releases/43'>Pocket Casts - Prototype Builds #43</a></td></tr>
</table>
<em>Automatticians: You can use our internal self-serve MC tool to give yourself access to App Center if needed.</em>
</details>

AliSoftware:
> We could add a build phase run script that check the build configuration and modifies the project if we're archiving a Prototype Build. But, again, given we shouldn't be doing that from Xcode often, or at all, the current implementation seems better to me.

Actually I don't think that idea would be possible to implement anyway, because Xcode prevents dynamically modifying the targets that are currently being built. It will detect that files have been changed in the middle of the build and will abort the build.
But I agree that's not worth it anyway and only supporting using Fastlane to build Prototype Builds is totally enough for our use case.

config/PocketCasts.debug.xcconfig
AliSoftware: TIL

podcasts.xcodeproj/project.pbxproj
AliSoftware: Nice find!

podcasts.xcodeproj/project.pbxproj
```diff
@@ -12755,7 +12755,7 @@
 				CLANG_WARN_SUSPICIOUS_MOVE = YES;
 				CLANG_WARN_UNREACHABLE_CODE = YES;
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
-				CODE_SIGN_IDENTITY = "Apple Distribution: Automattic, Inc. (PZYM8XX95Q)";
+				CODE_SIGN_IDENTITY = "Apple Distribution";
```
AliSoftware, line 12758: 👍

podcasts.xcodeproj/project.pbxproj
```diff
@@ -12968,13 +12968,9 @@
 				CLANG_WARN_OBJC_ROOT_CLASS = YES_ERROR;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_ENTITLEMENTS = "Pocket Casts App Clip/Pocket_Casts_App_Clip.entitlements";
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				CODE_SIGN_STYLE = Manual;
-				CURRENT_PROJECT_VERSION = 1;
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				DEVELOPMENT_ASSET_PATHS = "\"Pocket Casts App Clip/Preview Content\"";
-				DEVELOPMENT_TEAM = "";
-				"DEVELOPMENT_TEAM[sdk=iphoneos*]" = PZYM8XX95Q;
```
AliSoftware, line 12977: 👍

podcasts.xcodeproj/project.pbxproj
AliSoftware: Nah I'm pretty sure `SKIP_INSTALL=NO` is for things like when you build a macOS (GUI or CommandLine) app and want the built app to be installed in `/Applications` / `/usr/local/bin` or something like that? So shouldn't have any incidence in this case 👍

fastlane/Fastfile
```diff
@@ -1420,4 +1425,43 @@ platform :ios do
...
+    UI.message("Removing #{app_clip_file_reference.path} from #{embed_app_clips_phase.display_name}...")
+    UI.verbose("Paths to embed count before deletion #{embed_app_clips_phase.files_references.count}.")
+    embed_app_clips_phase.remove_file_reference(app_clip_file_reference)
+    UI.verbose("Paths to embed count after deletion #{embed_app_clips_phase.files_references.count}.")
+
+    UI.message('Saving the project file....')
+    project.save
+
+    UI.success('Dependency successfuly removed.')
+  end
```
AliSoftware, line 1466: See the improvement suggestions I made for this script in https://github.com/Automattic/pocket-casts-ios/commit/a89eeea1dbd420e6c04601ea3075fd34da647183#commitcomment-151007623 before you opened this PR 😅 

```ruby
native_targets = project.targets.select { |t| t.is_a?(Xcodeproj::Project::Object::PBXNativeTarget) }
app_targets = native_targets.select { |t| t.product_type == 'com.apple.product-type.application' }
app_clip_targets = native_targets.select { |t| t.product_type == 'com.apple.product-type.application.on-demand-install-capable' }

app_targets.each do |app|
  # Delete any dependency of the app target that is an app_clip target
  app.dependencies.delete_if { |dep| app_clip_targets.include?(dep.target) }
  # Delete the "Embed App Clips" build phase
  app.build_phases.delete_if { |b| b.display_name == 'Embed App Clips' }
end
```

This is not really a blocker (as long as your code works…). But I think my suggestions are still worth it given that it would (1) simplify the code a bit (just using `delete_if` instead of having the same be done with multiple lines of Ruby) and (2) be more flexible and generic (avoid hardcoding the target's names, automatically handle potential future targets needed same cleanup…)

AliSoftware (commented):
LGTM, but I have some [suggestions](https://github.com/Automattic/pocket-casts-ios/commit/a89eeea1dbd420e6c04601ea3075fd34da647183#commitcomment-151007623) to simplify the `remove_app_clip_dependency!` and make it more resilient if you're up for it.

fastlane/Fastfile
mokagio: Thanks @AliSoftware ! That's definitely simpler and easier to read.

I came to the implementation in this PR iteratively, sort of using all those logs to figure out what step to take next. I do agree it's a lot the way it is now... At the same time, I like how it would fail if the project is not set up with an App Clip, to signify that something has changed in the setup the automation expects and the user should act accordingly.

I'm going to merge as is just to unblock future builds, then follow up with a PR that will try to take the best of both implementations. Thanks again!

PR #2714, opened 2025-01-30
[Podcast Feed Reload] Pull Down to Reload feed

### LINKED ISSUE #2705: Add Pull down to reload

danielebogo:
| 📘 Part of: #2703 
|:---:|

Fixes #2705 

This PR adds the pull down to reload to call the view model and invoke the API. The component used for the pull down is called `CustomRefreshControl`, which extends the system `UIRefreshControl`. The instance is set to the table view property `refreshControl` and it's not manually added as subview.

This because the `PCRefreshControl` does some computations to the scroll view edge inset which interferes with the view controller and its custom navigation bar and search controller.

https://github.com/user-attachments/assets/cc8b032d-2e49-4dfe-a490-2ee2ba1140ff

## To test

- Make sure you have the FF `podcastFeedUpdate` enabled 
- Open a podcast detail
- Pull down to reload
- Even here the view model simulates the response
- Confirm the UI is correct and the label correctly updates

## Checklist

### DIFF
--- a/podcasts/CustomRefreshControl.swift
+++ b/podcasts/CustomRefreshControl.swift
@@ -1,0 +1,196 @@
+import Foundation
+
+class CustomRefreshControl: UIRefreshControl {
+    var perform: (() -> Void)?
+
+    private var refreshInnerImage = UIImageView()
+    private var refreshOuterImage = UIImageView()
+    private let innerStartingAngle = -90 as CGFloat
+    private let innerEndingAngle = 90 as CGFloat
+    private var innerRotationAngle = 0 as CGFloat
+    private var outerRotationAngle = 0 as CGFloat
+    private var pullDownAmountForRefresh = 170 as CGFloat
+    private var refreshLabel = UILabel()
+    private var isAnimating = false
+    private var didTriggerHaptic = true
+
+    override init() {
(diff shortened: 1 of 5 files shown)


podcasts/CustomRefreshControl.swift
```diff
@@ -0,0 +1,203 @@
...
+
+        refreshOuterImage.image = UIImage(named: "refresh_outer")
+        addSubview(refreshOuterImage)
+
+        addTarget(self, action: #selector(beginRefreshing), for: .valueChanged)
+    }
+
+    private func setupLayout() {
+        refreshLabel.translatesAutoresizingMaskIntoConstraints = false
+        refreshLabel.leadingAnchor.constraint(equalTo: leadingAnchor).isActive = true
```
SergioEstevao, line 69: Don't you prefer to do NSLayoutConstraint.activate([...]) instead of one by one?

SergioEstevao (approved):
All working great and looking good!

Just left a small nit regarding code layout, feel free to ignore.

podcasts/CustomRefreshControl.swift
danielebogo: I do but I kept the same as the other component

PR #2731, opened 2025-02-06
[App Clips] Add Build Phases

danielebogo:
This PR re-enable the Build Phases and bump the App Clips min deployment target to iOS 17.
If Apple will reject it again I will revert this change again.

## To test

- CI must be 🟢 

## Checklist

### DIFF
--- a/podcasts.xcodeproj/project.pbxproj
+++ b/podcasts.xcodeproj/project.pbxproj
@@ -680,6 +680,7 @@
 		9A4BB1F32D4A3F2A00B68D94 /* CustomRefreshControl.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A4BB1F22D4A3F2A00B68D94 /* CustomRefreshControl.swift */; };
 		9A509DE02D14606100E8CEB1 /* CancelSubscriptionOfferSuccessView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A509DDF2D14606100E8CEB1 /* CancelSubscriptionOfferSuccessView.swift */; };
 		9AA3B6312D27F0B600A0E30E /* CancelSubscriptionPlansViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9AA3B6302D27F0B600A0E30E /* CancelSubscriptionPlansViewModel.swift */; };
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/77">7.82 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

AliSoftware (approved):
Was the `Agrume` framework that you removed the most likely culprit for the size of the AppClip binary being too large?

danielebogo:
@AliSoftware not really but it's not important. I tried to export the size report with both Agrume and Lottie removed and the compressed size was 11.1MB, which is still not enough to stay below 10MB. Using the min deployment target set to iOS 17 for the app clips only should bring the limit to 50MB. Let's see what Apple will say 🤞

AliSoftware:
Ah, I see, TIL that they increased the limit in higher deployment targets. Sneaky 🙂 

But now I'm wondering if the fact that [the `IPHONEOS_DEPLOYMENT_TARGET` of all the Pods targets except the Watch App](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Podfile#L44-L49) being forced [to `15.0`](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Podfile#L9) by the `Podfile` won't hinder that bump you made on the AppClip's deployment target?

I hope not, because I think that if the AppClip's own deployment target is `17.0` then even if it depends on pods frameworks that have their deployment target set to a lower value (`15.0`) the AppClip's whole min deployment target should still be `17.0`… but then that also makes me wonder why the script in the `Podfile` makes an exception for the `target.name != 'Pocket Casts Watch App'` (which I wouldn't expect to be the name of a pod target in the **Pods** Project to begin with btw?)… 🤔

danielebogo:
It shouldn't be a problem as I believe Xcode would have complained about it. Besides the classes in the App Clips bundle shouldn't use Cocoapods. Cocoapods is not used for App Clips so I believe that Xcode would have triggered some errors while building it?

PR #2544, opened 2024-12-13
Fix highPassFilterRenderCallback callback

danielebogo:
Ref. #2522

This PR fixes the `highPassFilterRenderCallback` which was still using the `Unmanaged<DefaultPlayer>` and not the proxy.

## To test

- CI must be 🟢
- Clear your UpNext Queue and your downloaded episodes
- Add at least 4 episodes into the queue you never listened to
- Run this branch and play an episode with the volume boost on and trim silence as global settings
- Confirm the app won't crash
- Switch to a new episode
- Confirm the app won't crash
- Switch to a new episode and drag almost at the end to make it finishing and switching to a new episode
- Confirm the app won't crash
- Try to play and change episodes while playing
- When the system decides to finalize the tap you will see 2 logs in this order:

1. The log from the finalize closure: `[AudioProcessingTapProxy] Finalize tap: <MTAudioProcessingTap 0x6000030513b0> Retain count 3 Created with i/f/p/u/t callbacks = {0x10aa14448/0x10aa14600/0x10aa149b4/0x10aa14d08/0x10aa15330} flags = 0x1`
2. The log from the proxy deinit: `[AudioProcessingTapProxy] Deinit proxy`

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #2744, opened 2025-02-11
Default Player: Fix retain self

danielebogo:
Related #2544

We still see crashes due to the player not being release correctly. I believe the reason is that even if the `inputProcRefCon` is a reference to the proxy, correctly extracted while calling both blocks `peakLimiterRenderCallback` and `highPassFilterRenderCallback` , the 2 functions (in which the blocks are injected) refers to `self` and not to the proxy -> default player

## To test

I'm not able to reproduce the crash but to test it we should stress the player by playing episodes, switching between episodes, using the effect player as well etc...

## Checklist


leandroalonso:
Shall this change be under a feature flag just for safety?

danielebogo:
@leandroalonso it's part of the `useDefaultPlayerTapCookie` FF

leandroalonso:
Oh that's right, sorry! 🤦

SergioEstevao (approved):
:shipit:

PR #2754, opened 2025-02-14
[Tooling] Fix `finalize_release` and the `release/*` branch checkout

AliSoftware:
# Why?

During release process we need the lanes invoked from the `release-pipelines/*.yml` and that will require to do a `git push` to the release branch to actually `git fetch` + `git checkout` those release branches explicitly, otherwise the `git push` will fail (e.g. unknown `refspec`)

Example of a failure of such case: https://buildkite.com/automattic/pocket-casts-ios/builds/9594#019504fc-44dd-433a-a9fe-39cf81f67510

# How

 - We still need to checkout the `release/*` branch before the compilation in `release-build.sh`, because we want to ensure that we include the latest commit that was just pushed by the `new_beta_release` / `finalize_release` lanes from previous job, instead of building the commit that the CI build was initially created on
 - But we **also** need to checkout the `release/*` in `release-pipelines/*.yml` before calling any of the lanes that would do a version bump and push the commit, otherwise it won't try to do the commit on the branch but on the detached HEAD.

# Links

Internal ref: p1739547154659529/1739543498.075849-slack-CC7L49W13

### DIFF
--- a/.buildkite/release-pipelines/finalize-hotfix-release.yml
+++ b/.buildkite/release-pipelines/finalize-hotfix-release.yml
@@ -18,6 +18,9 @@ steps:
       echo '--- 🔐 Access Secrets'
       bundle exec fastlane run configure_apply
 
+      echo '--- :git: Checkout Release Branch'
+      .buildkite/commands/checkout-release-branch.sh "${RELEASE_VERSION}"
+
       echo '--- 🚀 Finalize Hotfix Release'
       bundle exec fastlane finalize_hotfix_release skip_confirm:true
     plugins: [$CI_TOOLKIT]
--- a/.buildkite/release-pipelines/finalize-hotfix-release.yml
+++ b/.buildkite/release-pipelines/finalize-hotfix-release.yml
@@ -18,7 +18,6 @@ steps:
       echo '--- 🔐 Access Secrets'
       bundle exec fastlane run configure_apply
 
-      echo '--- :git: Checkout Release Branch'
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/77">7.82 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

danielebogo (approved):

danielebogo (approved):

.buildkite/release-pipelines/finalize-hotfix-release.yml
```diff
@@ -18,6 +18,8 @@ steps:
       echo '--- 🔐 Access Secrets'
       bundle exec fastlane run configure_apply
 
+      .buildkite/commands/checkout-release-branch.sh "${RELEASE_VERSION}"
```
iangmaia: Shouldn't the checkout happen before `install_gems`, just to be 100% consistent we're going to install the exact gems that are in the branch? (this also applies to the other pipelines ofc).

AliSoftware: Good catch! Indeed.

PR #2776, opened 2025-02-24
[Suggested folders] Show suggested folder podcast list

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes # 

This PR adds the new podcast list view for each folder. 
**The data being displayed is still mock data.**

It also fixes two bugs found on a previous review:
 - Scroll indicator overlapping content on the folder and podcast cells
 - Create Custom button was too small

| 1 | 2 |
| - | - |
| Simulator Screenshot - iPhone 16 Pro Max - 2025-02-24 at 15 16 30 |  Simulator Screenshot - iPhone 16 Pro Max - 2025-02-24 at 15 15 49 |

## To test

- Start the app and login with a Plus or Patron user
- Ensure that you have the Feature Flag enabled. Profile -> Settings -> Beta Features -> SuggestedFolders
- Go to Podcasts tab
- Tap on the Create Folder icon on the top
- Check if the suggested folder screen shows and you see a grid view like the above with the folder
- Scroll around in the folder to see if all is correct.
- Tap on one of the folders
- Check if the content is correct and the podcast images are shown.

## Checklist

### DIFF
--- a/podcasts/SuggestedFoldersView.swift
+++ b/podcasts/SuggestedFoldersView.swift
@@ -46,6 +46,14 @@ struct SuggestedFoldersView: View {
             Text(L10n.suggestedFoldersDescription)
                 .textStyle(SecondaryText())
             foldersView
+                .padding(.horizontal, -20)
+                // hack to allow the scroll indicators to be visible without overlapping the content
+                .safeAreaInset(edge: .trailing) {
+                    EmptyView().frame(width: 20)
+                }
+                .safeAreaInset(edge: .leading) {
+                    EmptyView().frame(width: 20)
+                }
             Button {
                 Analytics.track(.suggestedFoldersModalUseTheseFoldersTapped, properties: [:])
                 dismissAction(nil)
--- a/podcasts/SuggestedFoldersView.swift
+++ b/podcasts/SuggestedFoldersView.swift
@@ -21,7 +21,7 @@ struct SuggestedFoldersView: View {
     var dismissAction: (String?) -> Void
 
     var body: some View {
-        NavigationView {
+        NavigationContainer {
             mainBody
                 .toolbar {
                     ToolbarItem(placement: .navigationBarLeading) {
--- a/podcasts/SuggestedFoldersView.swift
+++ b/podcasts/SuggestedFoldersView.swift
@@ -14,6 +14,11 @@ class SuggestedFoldersModel: ObservableObject {
 }
 
 struct SuggestedFoldersView: View {
+
+    enum Constants {
+        static var margin: CGFloat = 20
+    }
+
     @EnvironmentObject var theme: Theme
     @State private var createFolderActive = false
     @ObservedObject var model: SuggestedFoldersModel = SuggestedFoldersModel()
@@ -46,9 +51,9 @@ struct SuggestedFoldersView: View {
             Text(L10n.suggestedFoldersDescription)
                 .textStyle(SecondaryText())
             foldersView
-                .padding(.horizontal, -20)
+                .padding(.horizontal, -Constants.margin)
                 // hack to allow the scroll indicator to be visible without overlapping the content
-                .customHorizontalMargin(margin: 20)
+                .customHorizontalMargin(margin: Constants.margin)
             Button {
                 Analytics.track(.suggestedFoldersModalUseTheseFoldersTapped, properties: [:])
                 dismissAction(nil)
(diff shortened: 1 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

joshheald (commented):
Behaviour noticed while testing:

1. The images seem to be based on test data or placeholders, not my subscribed podcasts. Two of the test data images were pocketcast logos. (See video)
2. It's very easy to accidentally tap on a folder while scrolling. This may be an artifact of using the simulator, but I've never noticed it in other apps.
3. Scrolling is not smooth, but again, simulator may be to blame. (See video)


https://github.com/user-attachments/assets/63d2123d-58fd-4a08-b9c7-ce4de4e43f54


https://github.com/user-attachments/assets/c189c95d-8bbd-4372-bf37-36d3306a6bb7

I'm reviewing the code now.

podcasts/SuggestedFoldersView.swift
```diff
@@ -46,6 +46,9 @@ struct SuggestedFoldersView: View {
             Text(L10n.suggestedFoldersDescription)
                 .textStyle(SecondaryText())
             foldersView
+                .padding(.horizontal, -20)
```
joshheald: Is this padding related to the horizontal margin? If so, it might be worth defining it in a constant, especially if it's worth using the same number in the `SuggestedFolderPodcastView` - what do you think?

SergioEstevao:
> 1. The images seem to be based on test data or placeholders, not my subscribed podcasts. Two of the test data images were pocketcast logos. (See video)

This is expected, we are still using mock data, while the endpoint is being finalized

> 2. It's very easy to accidentally tap on a folder while scrolling. This may be an artifact of using the simulator, but I've never noticed it in other apps.

Will investigate that one

> 3 Scrolling is not smooth, but again, simulator may be to blame. (See video)

It looks for some reason you are unable to fetch the image for two of the podcasts, I will investigate if we have some loop there that is causing multiple requests.

podcasts/SwiftUI/NavigationContainer.swift
```diff
@@ -0,0 +1,21 @@
+import SwiftUI
+
+struct NavigationContainer<Content: View>: View {
```
joshheald, line 3: This is nice, it's so easy to forget to use `NavigationStack` instead of `NavigationView`. Is it something you plan to apply wider in the codebase?

joshheald (approved):
Works well, though note the previous comment with behavioural findings...

I tested on an iPad Simulator, and an iPhone 16 Simulator. No scroll indicators visible in the folder contents view, presumably because it only showed me 5 podcasts which isn't enough to require scrolling. Button and scroll indicators worked fine on the list view.

A couple of minor questions/suggestions inline 😊

joshheald:
> > 2. It's very easy to accidentally tap on a folder while scrolling. This may be an artifact of using the simulator, but I've never noticed it in other apps.
> 
> Will investigate that one 

For me, accidental taps happened at the start of a scroll. I can't see anything custom or unusual in the code which would be causing it, so it's probably just me – tbh, I usually test on a device rather than a simulator.

> It looks for some reason you are unable to fetch the image for two of the podcasts, I will investigate if we have some loop there that is causing multiple requests.

This happened on both simulators for me, so it wasn't a one-off.

podcasts/SwiftUI/NavigationContainer.swift
SergioEstevao: Yes, we have been seen situations that we want to use the newer NavigationStack but been holding up because of our min. IOS requirement, I saw this is an opportunity to finally starting migrating away.

SergioEstevao: I also needed it because of some animation bugs on iOS 18 with NavigationView.

SergioEstevao:
> For me, accidental taps happened at the start of a scroll. I can't see anything custom or unusual in the code which would be causing it, so it's probably just me – tbh, I usually test on a device rather than a simulator.

Tested this on the device and it's not an issue there it looks a simulator only issue.

> This happened on both simulators for me, so it wasn't a one-off.

We debugged this together and strangely enough the mock uuids that I used look they are not the same for Josh so that is why they are not showing.

PR #2820, opened 2025-03-07
GRDB: add GRDB abstraction

leandroalonso:
| 📘 Part of: #2773 | 
|:---:|

Add GRDB abstraction.

This should be merged with #2840 because this PR alones has some issues.

#### With `grdb` feature flag enable

Go to `FeatureFlag.swift` and ensure `grdb` case returns `true`.

##### Clean install

1. Do a clean install of the app
2. ✅ Check that the database schema is correct

##### Migration

1. Install an old version of the app (`7.75` is recommended, you can do so by `git checkout 7.75`)
2. Now go to this branch and run it
3. ✅ Ensure database was migrated corrected and that there are no warnings

#### With `grdb` feature flag disabled

Go to `FeatureFlag.swift` and ensure `grdb` case returns `false`.

##### Clean install

1. Do a clean install of the app
4. ✅ Check that the database schema is correct

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

Issue #2773, opened 2025-02-21
GRDB migration


leandroalonso:
## Goal

We would like to migrate from FMDB to GRDB to take advantage of concurrent reads and improve the app's performance. We want this migration to not lead to duplicate code and to be manageable through a remote feature flag.

## Tasks

- [x] Create abstractions so we can toggle between frameworks https://github.com/Automattic/pocket-casts-ios/pull/2774
- [x] Start using the FMDB abstraction https://github.com/Automattic/pocket-casts-ios/pull/2778
- [x] Implement the GRDB abstraction https://github.com/Automattic/pocket-casts-ios/pull/2820
- [x] Handle how Dates are stored between FMDB and GRDB #2840 
- [x] Check all usage of double-quotes and replace by a single quote #2840 
- [x] Double-check usage of `enum` to make sure GRDB can use them, [see this comment](https://github.com/Automattic/pocket-casts-ios/pull/2840#issuecomment-2792907135)
- [x] Start testing the GRDB abstraction #3009

## Testing in TestFlight

- [x] Log errors to Sentry https://github.com/Automattic/pocket-casts-ios/pull/3192
- [ ] ~Backup database~
- [x] Ensure GRDB and FMDB initialization are the same https://github.com/Automattic/pocket-casts-ios/pull/3194
- [x] Change feature flag name https://github.com/Automattic/pocket-casts-ios/pull/3279
- [x] Test GRDB in TestFlight https://github.com/Automattic/pocket-casts-ios/pull/3279
- [ ] Release to all users https://github.com/Automattic/pocket-casts-ios/pull/3334

## Concurrency reads

- [x] Create a new feature flag for this
- [x] Change the abstraction to have methods for read/write https://github.com/Automattic/pocket-casts-ios/pull/3040
- [x] `shm` and `wal` should not be protected file
- [x] Change the code to make use of these new methods
- [x] Test with GRDB
- [x] Test in TestFlight
- [x] Release to all users

leandroalonso:
## Dates: FMDB vs GRDB

GRDB and FMDB handle `Date` types differently. See below:

| FMDB | GRDB |
| ------ | ------ |
| <img width="1149" alt="Image" src="https://github.com/user-attachments/assets/813b0b89-c870-433d-9734-695e98ff484d" /> | <img width="1149" alt="Image" src="https://github.com/user-attachments/assets/6afbce1d-beef-4062-a5e4-4f02756af89f" /> |

[...]

PR #2840, opened 2025-03-12
GRDB: ensure FMDB and GRDB compatibility

leandroalonso:
| 📘 Part of: #2773 |
|:---:|

While doing the process to migrate from FMDB to GRDB we want something simple: that we can alternate between FMDB and GRDB as needed and that this doesn't cause any issue.

#### Issue 1: Dates

While working on #2820, I discovered that FMDB and GRDB handle dates differently. FMDB saves it using `timeIntervalSince1970` while GRDB uses "the format "YYYY-MM-DD HH:MM:SS.SSS" in the UTC time zone. It is precise to the millisecond." (from their docs).

GRDB is much better, however — at this stage, we want compatibility between them. So this PR forces GRDB to also save using `timeIntervalSince1970`.

#### Issue 2: usage of double quotes

There was a query using double quotes. This should not be done and causes an error that was caugh by a unit test, you can see the [change here](https://github.com/Automattic/pocket-casts-ios/pull/2840/commits/227b0d83f1615af9160c8d7327357091ad68c502).

More context about [this issue here](https://github.com/Automattic/pocket-casts-ios/issues/2773#issuecomment-2721222352).

#### More issues?

We might have more issues. The only way to know is by using GRDB. So my proposal is to enable it first for our development environment (using `#if DEBUG`).

## Changes

* Extract `Episode` and `Podcast` creation from a dictionary to its own method (so it can be used in unit tests)
* Update the old code where the extracted code lived to call the new methods
* Change `GRDBDatabase.executeUpdate` to parse the given values and change `Date` to `TimeInterval`
* Add unit tests that compare that the same podcast/episode saved from GRDB and FMDB produces the same object
* Add unit tests to ensure that podcasts/episodes saved with GRDB and read from FMDB produces the same results
* Add unit tests to ensure that podcasts/episodes saved with FMDB and read from GRDB procudes the same results
* Changed all unit tests to run with the enabled framework. If GRDB is disabled it uses FMDB, otherwise it uses GRDB
* Changed a SQL Query that was using double quotes and failing to execute when running using GRDB

## Discussion

I think adding unit tests for `Podcast` and `Episode` are enough because it coverts all the data types used by other entities (such as Bookmark, Folder, etc).

### From FMDB to GRDB

1. Run `trunk`
2. Checkout this branch
3. Run the app again
4. ✅ Check the console, it should have no database warnings
5. Check the podcasts and episodes, they should display correctly

### From GRDB to FMDB

1. Delete the app
2. Run this branch
3. Log in to an account with items in Up Next, followed podcasts, etc.
4. Change the `grdb` feature flag to `false`
5. Re-run the app
6. ✅ Check the console, it should have no database warnings
7. ✅ Check the podcasts, episodes, Up Next, it all should behave correctly

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -258,7 +258,7 @@ public enum FeatureFlag: String, CaseIterable {
         case .podcastNewformAppsFlyer:
             false
         case .grdb:
-            false
+            true
```
leandroalonso: **This should not be merged with the flag enabled.**

My proposal is that we change this to:

```suggestion
            #if DEBUG
            true
            #else
            false
            #endif
```

Modules/DataModel/Sources/PocketCastsDataModel/Private/DB/GRDB/GRDBDatabase.swift
```diff
@@ -17,7 +18,12 @@ class GRDBDatabase: PCDatabase {
     }
 
     func executeUpdate(_ sql: String, values: [Any]?) throws {
-        try database.execute(sql: sql, arguments: StatementArguments(values != nil ? values! : [])!)
+        // We want GRDB to save `Date` the same way FMDB does: using `timeIntervalSince1970`
+        // In terms of raw performance changing that on the model layer would be much better.
+        // However, that's a large change that we want to avoid.
+        let filteredValues = values?.map { ($0 as? Date)?.timeIntervalSince1970 ?? $0 }
+
+        try database.execute(sql: sql, arguments: StatementArguments(filteredValues != nil ? filteredValues! : [])!)
```
SergioEstevao, line 26: Do we want to keep the force unwrapping for StatementArguments? 
Is this for us to detect any arguments that is not `DatabaseValueConvertible` ?

SergioEstevao:
When running the unit tests locally with the `grdb` FF active I'm seeing the following exceptions:

`
PocketCastsDataModel/GRDBQueue.swift:13: Fatal error: 'try!' expression unexpectedly raised an error: SQLite error 21: out of memory - while executing `BEGIN IMMEDIATE TRANSACTION`
`

Expected?

leandroalonso:
@SergioEstevao definetely not.

I've checked what was going on and there were multiple issues:

1. I've inverted the correct `DataManager` initialization. That is fixed in 7a0208c79328a1a686a4957c8b346eedf44b8ddd
2. By running the tests using GRDB — and forcing unwrapping — I discovered an issue with the download tests. Those were leaking while other tests were running, this is fixed in 55d33cb089a1e46b6e8067116638b5443568914c
3. The `DataManager` initialization for GRDB was wrong when the `grdb` flag was disabled. This was causing the database connection to be closed due to the Feature Flag being `false`. Fixed on 7a0208c79328a1a686a4957c8b346eedf44b8ddd

Tests should run correctly now, with `grdb` flag on/off.

Modules/DataModel/Sources/PocketCastsDataModel/Private/DB/GRDB/GRDBDatabase.swift
leandroalonso: I would keep this way for now.

You can see a good example described in this comment: https://github.com/Automattic/pocket-casts-ios/pull/2840#issuecomment-2787429208

tl;dr: forcing unwrapping for GRDB made me discover underlying issues on our unit tests.

SergioEstevao:
Tests are now running correctly!
However when running in GRDB mode, I got an exception when playing around with bookmarks.
After I dig at it a bit I found the exception was resulting from this line:
`
let result = await dbQueue.executeUpdate(query, values: [SyncStatus.synced])
`
on `BookmarkDataManager`

From what I could found the issue is that we are not calling rawValue on the enum when sending to the database and then the `StatementArguments` does not know how to convert it properly and return nil.

Changing it to this:
`
let result = await dbQueue.executeUpdate(query, values: [SyncStatus.synced.rawValue])
`
solves the issue, but we will need to keep an eye on the database call other places were we are using enums directly instead of using rawValue

leandroalonso:
@SergioEstevao That's the good thing about force unwraps while testing. :)

We should make `SyncStatus` to conform to `DatabaseConvertible`, I added this on #2840.

This PR is already quite big, and we will still find a few things here and here.

SergioEstevao (approved):
Now that unit tests are working correctly let's merge this one, and handle the other issues we found on follow up PRs.
:shipit:

danielebogo (approved):
🎉 🎉 🎉 🎉 🎉

PR #2886, opened 2025-03-24
[Suggested folders] Don't show suggested folders if podcasts didn't change since last application

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes # 

This PR adds a mechanism to check if the podcast available changed since the last time suggested folders were applied

## To test

1. Start the app using a paid account
2. Got to the podcasts tab
3. Tap on the folders button
4. Check if suggestions are show and applied them, tap on Replace Folders
5. Check if results were applied
6. Tap on Folders again
7. Check if manual folder creation flow is show
8. Create a new folder
9. Unfollow some of the podcasts
10. Tap on Folders again
11. Check if suggested folders flow is show again

## Checklist

### DIFF
--- a/podcasts/FoldersCoordinator.swift
+++ b/podcasts/FoldersCoordinator.swift
@@ -42,7 +42,7 @@ class FoldersCoordinator: NSObject {
     }
 
     func startFolderCreationFlow(from vc: UIViewController) {
-        if FeatureFlag.suggestedFolders.enabled, suggestedFoldersModel.loadingState == .loaded {
+        if FeatureFlag.suggestedFolders.enabled, suggestedFoldersModel.loadingState == .loaded, didPodcastsChanged() {
             suggestedFolderCreationFlow(from: vc, source: .podcastsList)
         } else {
             manualFolderCreationFlow(from: vc)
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/80">7.85 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/FoldersCoordinator.swift
```diff
@@ -150,6 +151,18 @@ class FoldersCoordinator: NSObject {
         NotificationCenter.postOnMainThread(notification: ServerNotifications.podcastsRefreshed, object: nil)
     }
 
+    private func saveLastUuidsUsed() {
+        let uuids = dataManager.allPodcastsOrderedByAddedDate().map { $0.uuid }.sorted()
+        let md5 = String(uuids.joined(separator: "")).md5
```
danielebogo, line 156: Can we avoid to repeat these 2?

danielebogo: Silly question but does this mean we fetch all podcasts each time? Would the last podcast added be enough? I'm worried about users with huge database.

danielebogo (approved):
@SergioEstevao tested following the flow and it's working as expected.
I left 2 questions though if you can check it before merging 🙏

podcasts/FoldersCoordinator.swift
SergioEstevao: There is a cache in the data manager, so we are not hitting the database and not using more memory than it's already being used.

SergioEstevao: Refactored the code to be in a single method here: https://github.com/Automattic/pocket-casts-ios/pull/2886/commits/5b52730eb835fa6672a7865ddc42e3e3b9432347

Using the last podcast will not be enough because user can remove older podcast that they followed and that will make the test to be incorrect.

danielebogo: Thanks for addressing that! Im wondering if there's a better way to now if that column in that table is modified 🤔

PR #2914, opened 2025-03-31
[WatchApp] Revert audio session to be always spokenAudio

### LINKED ISSUE #2245: Watch app should pause play during announcements from other apps
### Description

Other Watch-based podcast apps pause (or can be configured to pause) when another app interrupts with an announcement, for example, a fitness app milestone or a GPS app instruction. With Pocketcasts, the third-party app speaks over the podcast, making it difficult to hear both. I believe it would be a better UX to have Pocketcasts pause when another app is speaking, or at least have the option to enable that behavior.

_No response_

### Did you search for existing list?

- [x] I have searched for existing issues.

danielebogo:
Fixes #2245 

This PR revert an old change and set the audio session property to `spokenAudio`. This should let the Watch app to pause/play during announcements from other apps

## To test

- CI must be 🟢 
- Needs to be tested on an Apple Watch
- Play an episode on the watch
- Trigger an announcement from another app (like Google Maps)
- The play should pause

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 7.86
 -----
 - New podcast page design [#2825](https://github.com/Automattic/pocket-casts-ios/pull/2825)
+- Fix Watch app should pause play during announcements from other apps [#2245](https://github.com/Automattic/pocket-casts-ios/issues/2245)
 
 7.85
 -----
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
On my tests I couldn't notice difference between the two modes. With that being said I will aprove it to see if it solves the user reports, because looking at the documentation this looks to be correct mode to use.

SergioEstevao:
@danielebogo do you want to target 7.86 and add an entry to the Changelog?

PR #2960, opened 2025-04-10
[Fix] Avoid to assign wrong frame position

danielebogo:
Fixes sentry-5543778228

This fix should avoid setting a wrong frame position which causes a crash

## To test

- Run the app
- Use the effect player
- Try to play different episodes
- The if statement should enter if you play with an episode already started
- Check the logs to see if it works
- The app shouldn't crash

## Checklist

### DIFF
--- a/podcasts/AudioReadTask.swift
+++ b/podcasts/AudioReadTask.swift
@@ -47,10 +47,10 @@ class AudioReadTask {
         if playPositionHint > 0 {
             currentFramePosition = framePositionForTime(playPositionHint).framePosition
             if currentFramePosition < audioFile.length {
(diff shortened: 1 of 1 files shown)


bjtitus (approved):
This branch seemed to work well when switching episodes with the effects player active.

I was having trouble reproducing this on branches outside of this PR. Is there anything else I should try just to confirm this is fixing it?

danielebogo:
> Is there anything else I should try just to confirm this is fixing it?

Not really as we don't have breadcrumbs saying what episodes cause this. We can keep track of it in the next sprint testing 7.87 and collect more logs

PR #3005, opened 2025-04-17
Watch Log crash fix

bjtitus:
Fixes a crash with Watch Log fetching from the background.

`WatchManager` may be getting deallocated when this thread is called and we want to skip these calls in that case.

### Normal Functionality

- Build and run the iOS and Apple Watch app. A simulator will work.
- Perform several actions on the watch seek back and forward in Now Playing, play pause, and change the Up Next queue.
- Quit the watchOS app
- Navigate to Profile > Help & Feedback > ⋯ > Export Database
- Make sure the exported file contains a watchos-logs.txt log with recent log lines.

### The Crash

- Apply [this change](https://github.com/user-attachments/files/19797716/mutable-watchmanager.patch)
 to make the WatchManager singleton mutable:
```swift
    static var shared: WatchManager! = WatchManager()
```
- Add a breakpoint which executes: `expr WatchManager.shared = nil` as a debugger command. You can also check "Automatically continue" if you don't want to skip.
CleanShot 2025-04-17 at 10 20 20@2x
- You should not see any crash after this change

## Checklist

### DIFF
--- a/podcasts/WatchManager+Send.swift
+++ b/podcasts/WatchManager+Send.swift
@@ -23,14 +23,17 @@ extension WatchManager {
             return
         }
 
+        // Hold a local reference so we don't potentially run into a deallocated `self` when the below blocks are run.
+        let cachedLog = self.cachedLog
+
         // since we don't know how long it takes for a send message to timeout, wait only 10 seconds for a watch response before giving up here
         var haveCalledCompletion = false
-        logFileRequestTimedAction.startTimer(for: 5.seconds) { [weak self] in
+        logFileRequestTimedAction.startTimer(for: 5.seconds) { [cachedLog] in
             if haveCalledCompletion { return }
 
             haveCalledCompletion = true
 
-            completion(self?.cachedLog)
+            completion(cachedLog)
         }
 
         // if we get here then it's likely we'll be able to ask the watch for a log file, so let's try
@@ -47,14 +50,14 @@ extension WatchManager {
                 }
                 completion(logContents)
             } else {
-                completion(self?.cachedLog)
+                completion(cachedLog)
             }
 
-        }) { [weak self] _ in
+        }) { _ in
             if haveCalledCompletion { return }
             haveCalledCompletion = true
 
-            completion(self?.cachedLog)
+            completion(cachedLog)
         }
     }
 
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/84">7.87 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/WatchManager+Send.swift
```diff
@@ -23,14 +23,17 @@ extension WatchManager {
             return
         }
 
+        // Hold a local reference so we don't potentially run into a deallocated `self` when the below blocks are run.
+        let cachedLog = self.cachedLog
+
         // since we don't know how long it takes for a send message to timeout, wait only 10 seconds for a watch response before giving up here
         var haveCalledCompletion = false
-        logFileRequestTimedAction.startTimer(for: 5.seconds) { [weak self] in
+        logFileRequestTimedAction.startTimer(for: 5.seconds) { [cachedLog] in
```
bjtitus, line 31: Keep a local copy in case self has been deallocated. It shouldn't be changing in the interim anyway and its a cached copy so isn't expected to be the most up to date.

podcasts/WatchManager.swift
```diff
@@ -399,10 +399,11 @@ class WatchManager: NSObject, WCSessionDelegate {
             return
         }
         DispatchQueue.global(qos: .background).async { [weak self] in
-            self?.sendStateToWatch()
+            guard let self else { return }
```
bjtitus, line 402: This was the main cause of the crash. `sendStateToWatch` was already effectively checking the optional value but the lower value was not and I think this caused a crash while `WatchManager` is being deallocated.

danielebogo (approved):
It seems to work as described. Let's try if this actually fix it.

SergioEstevao (commented):
I also tested this using my real watch and phone and all worked correctly.
No crash was detected.

PR #3007, opened 2025-04-17
GRDB: fix crash on `BookmarkDataManager`

leandroalonso:
| 📘 Part of: #2773 |
|:---:|

As mentioned by @SergioEstevao [here](https://github.com/Automattic/pocket-casts-ios/pull/2840#issuecomment-2792907135) there was a crash when dealing with Bookmarks.

## About the fix

There was two ways to fix that:

1. By using `rawValue` (as did in this PR)
2. By making `SyncStatus` to conform to `DatabaseValueConvertible`

I did (1) because:

1. This seems to be the only place where an enum was used directly 
2. By not using `DatabaseValueConvertible` we keep the `Enums` decoupled of the database framework. 

## To test

Enable `grdb` feature flag in `FeatureFlag.swift`

2. Login to an account with Bookmarks
3. ✅ The login should succeed

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/BookmarkDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/Bookmarks/BookmarkDataManager.swift
@@ -158,7 +158,7 @@ public struct BookmarkDataManager {
         SET \(Column.syncStatus) = ?
         """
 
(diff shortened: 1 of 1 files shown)


danielebogo (approved):
@leandroalonso :shipit: 
Logged in with the FF on and everything worked!

Issue #3016, opened 2025-04-18
[Task] Hide Recommendations when empty


bjtitus:
Recommendations can be empty but we need to check the `source` to see if this is the case. There are a couple cases where we want to hide the sections:
- The `source` endpoint returns anything but a 200
- There are fewer than 3 items in the section

bjtitus:
We no longer need this as the Discover sections are already handled and the tab is always shown and we have fallback categories. We have [empty state and loading handling for network failures](https://github.com/Automattic/pocket-casts-ios/pull/3097).

Issue #2989, opened 2025-04-15
[Epic] Recommendations


bjtitus:
Adding personalized podcast recommendations

### Useful Links

[Project Thread](https://pocketcastsp2.wordpress.com/2025/01/26/project-thread-recommendations/)
[Client Tasks](https://pocketcastsp2.wordpress.com/2025/02/27/client-tasks-for-the-pocket-casts-recommender/)
[Designs](https://pocketcastsp2.wordpress.com/2024/11/19/improving-podcast-recommendations-i1/)
[Android Project](https://href.li/?https://github.com/orgs/Automattic/projects/1292/views/1)

BKrnk1337:
I really like this update and the new function. However, I still have an improvement request that I expressed a few months ago in the forum and certainly a year or two ago here on GitHub. Please give us the opportunity to personalise the Discover page more. There are a lot of sections there that, although I have set my correct country, don't interest me at all and make the page very confusing.

There is always something like "Get your Facts straight" which seems like a categories that you want to promote but non of the podcast there are interesting for me. But it takes me scrolling and a lot of wasted space. Furthermore the "network highlights" which are although not interesting at all. As well the "stuff picks" and the "guest curator". Your stuff seems anonymous as I don't know how recommend what and I actually don't want anonymous recommendations. The guest I don't know and can't relate to the recommendations. 

So in the end a looot of wasted screen space. The important parts are to cluster of the screen (popular in <country>)

Finally: As a paying user of your app, I don’t actually expect to see sponsored posts at this point - only users who use your app for free should see them.


![image](https://github.com/user-attachments/assets/841f7e80-8874-4e65-abc4-ee6182a52689)
![image](https://github.com/user-attachments/assets/929484bc-2a78-40c6-8f18-59d6242290db)
![image](https://github.com/user-attachments/assets/6d2b01ce-430b-46c5-80f7-5b3fdeb24122)

PR #3022, opened 2025-04-21
[Recommendations] Authentication failure handling

bjtitus:
| 📘 Part of: #2989 |
|:---:|

Part of #3016

Hides recommendations cells when the API returns anything but 200.

We expect to see:
* `404` if the response is missing
* `401` if authentication failed for some reason (although we shouldn't even make the call since we know whether the user is authenticated)

Here's an example showing only one of the filled out sections while others remain hidden:

https://github.com/user-attachments/assets/6c52c75b-a487-4480-926b-6aab90a45a3a

## To test

* Log in to an account with some listening history
* Enable the `recommendations` feature flag & restart the app
* Navigate to Discover
* You may need to force reload discover in the Developer menu
* ✅ Verify that you see one of the Recommendations sections
* Other sections should be missing and nothing should be completely empty
* If you have trouble verifying which sections should appear, use Proxyman to verify that the `api.pocketcasts.com/recommendations/<user, social, podcast>` API calls are returning a 200 for at least one. If not, reach out to me and I can try to help.

On many accounts I only see one of these endpoints successfully returning:

CleanShot 2025-04-21 at 11 03 15@2x

## Checklist

### DIFF
--- a/Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
+++ b/Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
@@ -10,6 +10,11 @@ public class DiscoverServerHandler {
 
     public static let shared = DiscoverServerHandler()
 
+    private let tokenHelper = {
+        let connection = URLConnection(handler: URLSession.shared)
+        return TokenHelper(urlConnection: connection)
+    }()
+
     public private(set) lazy var discoveryCache: URLCache = {
         let cache = URLCache(memoryCapacity: 1024 * 1024, diskCapacity: 5 * 1024 * 1024, diskPath: "discovery")
         return cache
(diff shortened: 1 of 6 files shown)


Modules/Server/Sources/PocketCastsServer/Public/Discover/DiscoverServerHandler.swift
```diff
@@ -101,55 +106,101 @@ public class DiscoverServerHandler {
...
-        let connection = URLConnection(handler: URLSession.shared)
-        return TokenHelper(urlConnection: connection)
-    }()
+    /// A method to check whether the response from the source URL authenticated successfully.
+    /// - Parameters:
+    ///   - item: An item which is `authenticated == true`
+    /// - Returns: Whether or not the authentication succeeded
+    public func checkSourceAuthentication(for item: DiscoverItem) async -> Bool {
+        // If there's no source URL, consider authentication failed. This shouldn't happen.
+        guard item.authenticated == true, let source = item.source else {
```
danielebogo: Nit: instead of checking against the optional `authenticated` I'm wondering if `DiscoverItem` can have a computed property `isAuthenticated` that unwraps it or fallback to false

danielebogo (changes requested):
Hey @bjtitus good looks good. I left just a nit but not a blocker.
However I had problems while testing this:
### Infinite Loading
I'm not sure if it's because I reloaded the Discover from the Developer panel, but after that I had an infinite loading. It never disappeared and it was repeated in few sections. Restarting the app it disappeared.


### Crash
While scrolling to the bottom, I got a crash. I can always replicated it
| 1 | 2 |
| -------- | ------- |
| Screenshot 2025-04-22 at 11 33 35 | Screenshot 2025-04-22 at 11 33 48 |

The error:
```
podcasts/DiscoverCellType.swift:92: Fatal error: Unknown Discover Item: podcast_list large_list_with_podcast
Unknown Discover Item: podcast_list large_list_with_podcast
```
I also see this warning before the crash:
```
Warning: You are setting a new content configuration to a cell that has an existing content configuration, but the existing content view does not support the new configuration. This means the existing content view must be replaced with a new content view created from the new configuration, instead of updating the existing content view directly, which is expensive. Use separate registrations or reuse identifiers for different types of cells to avoid this. Make a symbolic breakpoint at UIContentConfigurationAlertForReplacedContentView to catch this in the debugger.
Cell: <UICollectionViewCell: 0x10453c8d0; frame = (0 2461.67; 393 100); layer = <CALayer: 0x600000338220>>;
Existing content configuration: UIHostingConfiguration<ModifiedContent<LoadingView, _EnvironmentKeyWritingModifier<Optional<Theme>>>, EmptyView>(rootView: SwiftUI.ModifiedContent<podcasts.LoadingView, SwiftUI._EnvironmentKeyWritingModifier<Swift.Optional<podcasts.Theme>>>(content: podcasts.LoadingView(_theme: SwiftUI.EnvironmentObject<podcasts.Theme>(_store: nil, _seed: 0)), modifier: SwiftUI._EnvironmentKeyWritingModifier<Swift.Optional<podcasts.Theme>>(keyPath: \EnvironmentValues.<computed 0x00000001d410a480 (Optional<Theme>)>, value: Optional(podcasts.Theme))), backgroundView: SwiftUI.EmptyView(), storage: SwiftUI.(unknown context at $1d3c90730).UIHostingConfigurationStorage(wantsBackground: false, margins: SwiftUI.OptionalEdgeInsets(top: nil, leading: nil, bottom: nil, trailing: nil), _minSize: (width: nil, height: nil), createsUIInteractions: true, disablesAnimatedSizeInvalidation: false, lastState: Optional(<UICellConfigurationState: 0x600000d6cd50; traitCollection = <UITraitCollection: 0x104649ef0; UserInterfaceIdiom = Phone, DisplayScale = 3, DisplayGamut = P3, HorizontalSizeClass = Compact, VerticalSizeClass = Regular, UserInterfaceStyle = Light, UserInterfaceLayoutDirection = LTR, ForceTouchCapability = Unavailable, PreferredContentSizeCategory = L, AccessibilityContrast = Normal, UserInterfaceLevel = Base, ImageDynamicRange = 0, SceneCaptureState = 0>>), wantsPlatformItemList: false, delegate: nil));
New content configuration: UIHostingConfiguration<EmptyView, EmptyView>(rootView: SwiftUI.EmptyView(), backgroundView: SwiftUI.EmptyView(), storage: SwiftUI.(unknown context at $1d3c90730).UIHostingConfigurationStorage(wantsBackground: false, margins: SwiftUI.OptionalEdgeInsets(top: nil, leading: nil, bottom: nil, trailing: nil), _minSize: (width: nil, height: nil), createsUIInteractions: true, disablesAnimatedSizeInvalidation: false, lastState: Optional(<UICellConfigurationState: 0x600000d6c5d0; traitCollection = <UITraitCollection: 0x103a1c410; UserInterfaceIdiom = Phone, DisplayScale = 3, DisplayGamut = P3, HorizontalSizeClass = Compact, VerticalSizeClass = Regular, UserInterfaceStyle = Light, UserInterfaceLayoutDirection = LTR, ForceTouchCapability = Unavailable, PreferredContentSizeCategory = L, AccessibilityContrast = Normal, UserInterfaceLevel = Base, SceneCaptureState = 0, ImageDynamicRange = 0>>), wantsPlatformItemList: false, delegate: nil))
```

bjtitus:
Thanks @danielebogo! I've added the cell type for `large_list_with_podcast`, this wasn't in production when I first submitted. I also made that change to add an `isAuthenticated` computed property.

I don't think the `UIContentConfiguration` replacement is new and I think I should address it separately. We'll probably need separate cell registrations for each non-item type.

danielebogo (approved):
Thanks for addressing those! No crashes anymore! :shipit:

PR #3040, opened 2025-04-25
GRDB: add `write` and `read` methods

leandroalonso:
| 📘 Part of: #2773 | 
|:---:|

To unlock the power of concurrent reads, we need to use GRDB `write` and `read` methods. This is a first step in this direction.

Since there are many places that this needs to be changed, I'll raise one PR per data manager — this makes it easier to be reviewed.

## To test

1. ✅ Run the app form scratch and ensure the app works just fine
2. Subscribe to a few podcasts
3. Change `grdb` in `FeatureFlag.swift` to `false`
4. Re-run the app
5. ✅ Everything should be there

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
@@ -4,22 +4,12 @@ import PocketCastsUtils
 
 class DatabaseHelper {
     class func setup(queue: PCDBQueue) {
-        var startingSchemaVersion: Int32 = 0
-
-        queue.read { db in
+        queue.write { db in
             do {
                 try db.executeQuery("PRAGMA busy_timeout = 10000", values: nil).close()
 
-                let rs = try db.executeQuery("PRAGMA user_version", values: nil)
-                if rs.next() { startingSchemaVersion = rs.int(forColumnIndex: 0) }
-                rs.close()
-            } catch {
-                FileLog.shared.addMessage("Failed to setup database \(db.lastErrorCode()): \(db.lastErrorMessage()) actual error: \(error)")
-            }
-        }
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
```diff
@@ -4,22 +4,12 @@ import PocketCastsUtils
 
 class DatabaseHelper {
     class func setup(queue: PCDBQueue) {
-        var startingSchemaVersion: Int32 = 0
-
-        queue.inDatabase { db in
+        queue.write { db in
```
leandroalonso, line 7: This call was split in two in one of the GRDB PRs as [explained here](https://github.com/Automattic/pocket-casts-ios/pull/2820/files#r1985476408).

This method is back to being a single block, which is what it should be. I had to add a method so that the crash that was happening before doesn't happen anymore.

With this change we're back at how it was before + working correctly with GRDB.

SergioEstevao:
Non related to the changes above but noticed this quotes issues when testing this PR on the `ServerPodcastManager+Update.swift` in the `cleanupDeletedEpisodes` method.

`let serverUuids = serverEpisodes.compactMap { $0["uuid"] as? String }.map { "\"\($0)\"" }`

SergioEstevao (approved):
Working correctly.

Installed from scratch, smoke test it and them changed the FF back and ran again.

All good! :shipit:

leandroalonso:
About the Sentry error, this code hasn't even make to TestFlight yet.

PR #3159, opened 2025-05-19
Fix Up Next queue replacement by specifying episode IDs

bjtitus:
| 📘 Part of: #24 |
|:---:|

We have received several user reports that old episodes are coming back into their queue. I analyzed several server logs and found cases where _we_ are sending those episodes as part of a "replace" command. This command sends those episodes right back to us, which _is_ expected since we're replacing the list with those new episodes.

I think there could be some race conditions in the way we handle deleting episodes around our sync API call. The `saveReplaceIfRequired` method fetches all episodes in the Up Next Queue and that queue is emptied by deleting all episodes just before, however, this process is not really synchronous and I think we may be ending up with a set of older episodes from the queue. It seems safer to me to specify the episode we want to replace and use this to empty the queue.

## To test

I don't have a full end to end test when running the app. Instead, I added a unit test which reproduces the issue.

* Change the feature flag value in the `testOverrideAllEpisodesWith_shouldNotIncludeStaleEpisodesInReplace` method to see the original issue.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

SergioEstevao (approved):
Working correctly on my tests.

PR #3192, opened 2025-05-27
GRDB: remove `try!` and log errors in Sentry

leandroalonso:
| 📘 Part of: #2773 | 
|:---:|

Removes force unwraps and log errors to Sentry.

This basically does what FMDB does now, which is silently failing instead of crashing the app.

Ps.: we talked about `Breadcrumbs` but what (I believe) we want is to log the error even if it doesn't crash the app.

## Changes

1. Created an `ErrorLogger` protocol, since Sentry is not available on `DataModel` so we can inject it
2. Expose the `CrashLoggingAdapter` so we can send events to Sentry
3. Create `SentryLogger` conforming to `ErrorLogger`
4. Inject `SentryLogger` into `GRDBQueue`
5. Log to Sentry in case any issue is triggered

## To test

**Thoroughly** review the code, ensuring that write database operations use `write` and read operations use `read`.

1. In `GRDBQueue.swift` create an error enum (or just copy the code below), such as:
```swift
enum Test: Error {
    case fail
}
```
2. Then inside the `do` block in `inDatabase`, add:
```swift
UserDefaults.standard.set(true, forKey: "force-crash-logging")
                throw Test.fail
```
3. Run the app
4. Go to Sentry and filter for the latest events
5. ✅ You should see your event there

### `GRDBqueue.swift` code

To make it easier:

```swift
import GRDB
import Foundation

enum Test: Error {
    case fail
}

class GRDBQueue: PCDBQueue {
    private let dbPool: DatabasePool
    private let logger: ErrorLogger?

    init(dbPool: DatabasePool, logger: ErrorLogger? = nil) {
        self.dbPool = dbPool
        self.logger = logger
    }

    func inDatabase(_ block: (any PCDatabase) -> Void) {
        withoutActuallyEscaping(block) { block in
            do {
                UserDefaults.standard.set(true, forKey: "force-crash-logging")
                throw Test.fail
                try dbPool.write { db in
                    let dbWrapper = GRDBDatabase(database: db)
                    block(dbWrapper)
                }
            } catch {
                logger?.log(error: error, context: [:])
            }
        }
    }
```

## Checklist

### DIFF
--- a/podcasts/Analytics/Adapters/Crash Logging/CrashLoggingAdapter.swift
+++ b/podcasts/Analytics/Adapters/Crash Logging/CrashLoggingAdapter.swift
@@ -4,8 +4,11 @@ import AutomatticRemoteLogging
 class CrashLoggingAdapter: AnalyticsAdapter {
     let crashLogging: CrashLogging?
 
+    static var sharedManager: CrashLoggingAdapter?
+
     init() {
         self.crashLogging = try? CrashLogging(dataProvider: CrashLoggingDataProvider()).start()
+        Self.sharedManager = self
     }
 
     func track(name: String, properties: [AnyHashable: Any]?) { }
--- a/podcasts/AppDelegate.swift
+++ b/podcasts/AppDelegate.swift
@@ -1,4 +1,5 @@
 import BackgroundTasks
+import AutomatticRemoteLogging
 import Firebase
 import FirebasePerformance
 import Foundation
@@ -7,7 +8,6 @@ import PocketCastsServer
 import PocketCastsUtils
 import Combine
 import FacebookCore
-import Sentry
 
 class AppDelegate: UIResponder, UIApplicationDelegate {
     private static let initialRefreshDelay = 2.seconds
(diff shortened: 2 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/Analytics/Adapters/Crash Logging/CrashLoggingAdapter.swift
```diff
@@ -4,8 +4,11 @@ import AutomatticRemoteLogging
 class CrashLoggingAdapter: AnalyticsAdapter {
     let crashLogging: CrashLogging?
 
+    static var sharedManager: CrashLoggingAdapter?
```
leandroalonso, line 7: Not proud of that.

However, I don't want to refactor this as it would be it's own task and comes with another complexities.

Since this is mean to be temporary I believe it is a good trade-off for the time being.

podcasts/AppDelegate.swift
```diff
@@ -419,3 +422,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         backgroundSignOutListener = BackgroundSignOutListener(presentingViewController: SceneHelper.rootViewController())
     }
 }
+
+struct SentryLogger: ErrorLogger {
+    func log(error: Error, context: [String: String]?) {
+    #if os(iOS)
+    CrashLoggingAdapter.sharedManager?.crashLogging?.logError(error, tags: context ?? [:], level: .warning)
```
leandroalonso, line 429: Should we use `.warning`? Or something else?

SergioEstevao: I think warning should be ok for now

SergioEstevao (approved):
:shipit:

bjtitus (approved):
:shipit: 

✅ I tested this for a while yesterday and everything seems good.
✅ Code looks good

PR #3216, opened 2025-05-30
GRDB: add TestFlight feature flag

leandroalonso:
| 📘 Part of: #2773 | 
|:---:|

This PR adds the remote feature flag value for enabling GRDB for TestFlight users **only**.

We will do a rollout, meaning not all users will have it enabled. We want to test that in a controlled way so we ensure everything goes smoothly.

## To test

1. [Comment this line](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/AppDelegate.swift#L316) so the feature flag code runs in `DEBUG`
2. Do a clean install of the app
3. ✅ Search for `Overriding grdb to true`. This log might be or might not be there. Try a few times, each time deleting and running a clean install of the app. This way you can also test that the rollout is working as intended.
4. When you see the log, see that right below there's a `Failed to set remote feature flag grdb: cannotBeOverridden` because this is not TestFlight
5. [Change this](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Modules/Utils/Sources/PocketCastsUtils/Feature%20Flags/FeatureFlag.swift#L369) to `true`
6. Delete the app, run it again until you see the `Overriding grdb to true`
7. ✅ This time there won't be a `Failed to set remote feature flag grdb: cannotBeOverridden` though
8. Re-run the app
9. ✅ You should see `[DataManager] Initialized using GRDB`

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -355,7 +355,7 @@ public enum FeatureFlag: String, CaseIterable {
             // Never remove this line or change to "grdb" as it will enable it
             // in previous versions in which the work was in progress.
             // tl;dr: this flag can never be "grdb"
-            "grdb_testflight"
+            "grdb_pr_test"
         default:
             rawValue.lowerSnakeCased()
         }
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -351,6 +351,11 @@ public enum FeatureFlag: String, CaseIterable {
             "use_podcast_html_description"
         case .podcastViewChanges:
             "podcast_view_changes_2025"
+        case .grdb:
+            // Never remove this line or change to "grdb" as it will enable it
+            // in previous versions in which the work was in progress.
+            // tl;dr: this flag can never be "grdb"
+            "grdb_testflight"
         default:
             rawValue.lowerSnakeCased()
         }
@@ -363,7 +368,13 @@ extension FeatureFlag: OverrideableFlag {
     }
 
     public var canOverride: Bool {
-        true
+        switch self {
+        // GRDB can only change in TestFlight versions
+        case .grdb:
+            Self.isTestFlight
+        default:
+            true
+        }
     }
 
     private static let isTestFlight = Bundle.main.appStoreReceiptURL?.lastPathComponent == "sandboxReceipt"
--- a/podcasts/AppDelegate.swift
+++ b/podcasts/AppDelegate.swift
@@ -334,6 +334,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 let remoteValue = RemoteConfig.remoteConfig().configValue(forKey: remoteKey)
                 if remoteValue.source == .remote {
                     do {
+                        print("Overriding \(flag) to \(remoteValue.boolValue)")
                         try FeatureFlagOverrideStore().override(flag, withValue: remoteValue.boolValue)
                     } catch {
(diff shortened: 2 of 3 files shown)


Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -351,6 +351,11 @@ public enum FeatureFlag: String, CaseIterable {
             "use_podcast_html_description"
         case .podcastViewChanges:
             "podcast_view_changes_2025"
+        case .grdb:
+            // Never remove this line or change to "grdb" as it will enable it
+            // in previous versions in which the work was in progress.
+            // tl;dr: this flag can never be "grdb"
+            "grdb_pr_test"
```
leandroalonso, line 358: This needs to change before we merge the PR, of course.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso:
@Automattic/pocket-casts-ios since this one is very important it would be good if y'all can give it a try.

podcasts/AppDelegate.swift
```diff
@@ -334,6 +334,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 let remoteValue = RemoteConfig.remoteConfig().configValue(forKey: remoteKey)
                 if remoteValue.source == .remote {
                     do {
+                        print("Overriding \(flag) to \(remoteValue.boolValue)")
```
danielebogo, line 337: Nit: you can use `FileLog.shared.console(".......")`, also wrapping it into `#DEBUG`

danielebogo (approved):
Thanks @leandroalonso !
I run few tests and tried to replicate al the scenarios you described.
I wasn't able to replicate the first step, when getting the FF from the config with `canOverride` to true. But I discovered that was me with a real device and a sandbox user logged in.

I left a nit about the `print`, other than that it's good.

podcasts/AppDelegate.swift
leandroalonso: Oh, this is here for testing purposes and I was going to remove before merging.

leandroalonso:
@danielebogo 

> I wasn't able to replicate the first step, when getting the FF from the config with canOverride to true. But I discovered that was me with a real device and a sandbox user logged in.

Would you mind giving more context about that? Weren't you able to see `Overriding grdb to true`?

SergioEstevao:
> Would you mind giving more context about that? Weren't you able to see Overriding grdb to true?

I'm having the same issue, I'm unable to see that log, I've tried at least 10 installs and still no luck ( device and simulator)

danielebogo:
> Would you mind giving more context about that? Weren't you able to see Overriding grdb to true?
@leandroalonso I saw `Overriding grdb to true` at my first tentative but I didn't see `Failed to set remote feature flag grdb: cannotBeOverridden`. Then I tried it again but it took me a while to see again `Overriding grdb to true`.

Each time I got `Overriding grdb to true`, I never saw the failed log. So I checked the `cannotBeOverridden` and I saw we check for the `sandboxReceipt`.

All my tests were done on a real device with a sandbox user setup, so the `isTestflight` is always `true` for me.

leandroalonso:
@danielebogo @SergioEstevao I'll put that on hold for now given the upcoming Firebase change that will allow us to have feature fags values only for TestFlight.

leandroalonso:
Closing this one for now in favor of an upcoming change to enable the flag only for TestFlight users without affecting non-TestFlight users.

SergioEstevao:
For the record any built done when using a Debug configuration in the simulator or real device will think it's a Testflight build.
See here: https://stackoverflow.com/questions/26081543/how-to-tell-at-runtime-whether-an-ios-app-is-running-through-a-testflight-beta-i

PR #3279, opened 2025-06-19
GRDB: add new feature flag to enable for 20% of users in TestFlight

leandroalonso:
| 📘 Part of: #2773 | 
|:---:|

Add the new `grdb_testflight` feature flag. In Firebase, this is configured to:

1. Always return `false` if it's a non-TestFlight build
2. `true` for 20% of TestFlight users
3. `false` for the rest

### Functionality

It's quite tricky to test this one as it involves the backend settings the value.

1. [Comment this line ](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/AppDelegate.swift#L316)
2. [Change](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/FirebaseManager.swift#L5)`2.hour` to `1.second`
3. [Change](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/FirebaseManager.swift#L25) `isTestflight` to `false`
4. Run the app multiple times (always doing a clean install)
5. ✅ You should **always** see a `Overriding grdb to false`
6. [Change](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/FirebaseManager.swift#L25) `isTestflight` to `true`
7. Run the app multiple times
8. ✅ Eventually you should get `Overriding grdb to true` (the number of times until it returns `true` might vary, so you might need to do the clean reinstall for a few times)

### Flag

[Please review the rollout conditions](https://console.firebase.google.com/u/0/project/singular-vector-91401/config/env/firebase/rollout/rollout_9?app=1:910146533958:ios:cf844de2510aeece&dateRange=24h).

## Checklist

### DIFF
--- a/podcasts/AppDelegate.swift
+++ b/podcasts/AppDelegate.swift
@@ -334,6 +334,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 let remoteValue = RemoteConfig.remoteConfig().configValue(forKey: remoteKey)
                 if remoteValue.source == .remote {
                     do {
+                        print("Overriding \(flag) to \(remoteValue.boolValue)")
                         try FeatureFlagOverrideStore().override(flag, withValue: remoteValue.boolValue)
                     } catch {
                         FileLog.shared.addMessage("Failed to set remote feature flag \(flag): \(error)")
--- a/podcasts/AppDelegate.swift
+++ b/podcasts/AppDelegate.swift
@@ -334,7 +334,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 let remoteValue = RemoteConfig.remoteConfig().configValue(forKey: remoteKey)
                 if remoteValue.source == .remote {
                     do {
-                        print("Overriding \(flag) to \(remoteValue.boolValue)")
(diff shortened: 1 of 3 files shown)


pocketcasts:
Version `7.92` has now entered code-freeze, so the milestone of this PR has been updated to `7.93`.

podcasts/AppDelegate.swift
```diff
@@ -334,6 +334,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 let remoteValue = RemoteConfig.remoteConfig().configValue(forKey: remoteKey)
                 if remoteValue.source == .remote {
                     do {
+                        print("Overriding \(flag) to \(remoteValue.boolValue)")
```
danielebogo: Nit: `FileLog.shared.console`

danielebogo (commented):
@leandroalonso I did few tests but I never get the `true` value

danielebogo (approved):

podcasts/AppDelegate.swift
leandroalonso: This was added just for testing purposes. Removed in 1ee35597c724206c96612c02928bc3f928011134

PR #3293, opened 2025-06-25
Fix the force reload of Feature Flag from developer panel

danielebogo:
While testing the Libro.fm FF remotely enabled, I discovered the force reload from the Dev panel didn't work:
- Xcode complained about the reload in background
- The force reload wasn't starting because of the debug environment 

This PR adds:
- Fix the background issue
- Add a force reload argument
- Add a debug log when the Firebase configuration overrides the local values

## To test

- Run this branch
- Confirm you don't see any any `Override \(flag): \(remoteValue.boolValue)`
- Open the dev panel and tap on `Force Reload Feature Flags`
- Confirm you see the override logs
- Stop the build
- Replace [this line](https://github.com/Automattic/pocket-casts-ios/blob/71d863874cb8e590ff902d91eaf73603d528f629/Modules/Utils/Sources/PocketCastsUtils/General/BuildEnvironment.swift#L24) with `.appStore`
- Start the build
- Confirm you see the override logs

## Checklist

### DIFF
--- a/podcasts/AppDelegate.swift
+++ b/podcasts/AppDelegate.swift
@@ -312,8 +312,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
     }
 
-    func updateRemoteFeatureFlags() {
-        guard BuildEnvironment.current != .debug else { return }
+    func updateRemoteFeatureFlags(forceReload: Bool = false) {
+        if BuildEnvironment.current == .debug, !forceReload { return }
 
         if FeatureFlag.errorLogoutHandling.enabled != Settings.errorLogoutHandling {
             ServerConfig.avoidLogoutOnError = FeatureFlag.errorLogoutHandling.enabled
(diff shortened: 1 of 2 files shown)


podcasts/AppDelegate.swift
```diff
@@ -312,8 +312,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
     }
 
-    func updateRemoteFeatureFlags() {
-        guard BuildEnvironment.current != .debug else { return }
+    func updateRemoteFeatureFlags(forceReload: Bool = false) {
+        if BuildEnvironment.current == .debug, !forceReload { return }
```
danielebogo: Any preferences about this?
I bet this can be also written as `guard BuildEnvironment.current != .debug || forceReload else { return }`

bjtitus: I always prefer the `guard`, personally.

SergioEstevao: I agree it's a small nit, but guard always feels better in this type of logic

SergioEstevao (approved):
Working correctly.

Just the small nit regarding guard/if but feel free to merge it.

SergioEstevao (approved):
:shipit:

PR #3301, opened 2025-06-30
Enable 'Warn before using data' by default

### LINKED ISSUE #3227: Enable 'Warn before using data' by default
### Description

The setting **Storage & data use > Warn before using data** is not enabled by default.

This has led to some users accidentally using a lot of mobile data.

In one case this was because they thought the setting **Auto Download > Only on unmetered WiFi** was a general setting, not only applicable to auto downloads.

We should enable 'Warn before using data' by default, and deep link to the setting in the warning, so users can choose to disable it if they want to.

Slack Reference: p1749029285430589-slack-C028JAG44VD

Android issue: https://github.com/Automattic/pocket-casts-android/issues/4080

_No response_

### Did you search for existing list?

- [x] I have searched for existing issues.

danielebogo:
Fixes #3227

This PR enables by default the Warn User data usage option. It also updates the screen we show when the data usage toggle is on and the device is not connected, with a link to our settings.

| Settings | Sheet |
| -------- | ------- |
| IMG_0209  | IMG_3871 |

## To test

- Use a real device with a sim to switch between wifi and another connection
- Build this branch on top of current Production or beta
- Run the app and navigate to our `Settings -> Storage & Data Use`
- Confirm you see `Warn Before Using Data` on
- Go to a podcast and open it
- Select an episode and start the download
- You should not see any Sheet presented
- Switch off the Wifi and stay on your sim connection
- Try to download the another episode from the podcast view
- Confirm you see the Sheet (as the screen above)
- Tap on Settings in the text
- Confirm you are redirected to the `Storage & Data Use` settings page

## Checklist

### DIFF
--- a/podcasts/MultiSelectHelper.swift
+++ b/podcasts/MultiSelectHelper.swift
@@ -243,6 +243,7 @@ class MultiSelectHelper {
                 actionDelegate.multiSelectActionBegan(status: status)
                 queueEpisodes(downloadableEpisodes, actionDelegate: actionDelegate)
             }
+            queueAction.outline = true
 
             if !Settings.mobileDataAllowed() {
                 warningMessage = L10n.downloadDataWarning + "\n" + warningMessage
--- a/podcasts/MultiSelectHelper.swift
+++ b/podcasts/MultiSelectHelper.swift
@@ -246,12 +246,10 @@ class MultiSelectHelper {
             queueAction.outline = true
 
             if !Settings.mobileDataAllowed() {
-                warningMessage = L10n.downloadDataWarningWithSettingsLink + "\n" + warningMessage
+                warningMessage = L10n.downloadDataWarningWithSettingsLink("pktc://settings/storage-and-data") + "\n" + warningMessage
             }
 
-            let attributes = [L10n.settings: URL(string: "pktc://settings/storage-and-data")!]
-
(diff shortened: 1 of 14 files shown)


podcasts/MultiSelectHelper.swift
```diff
@@ -243,11 +243,15 @@ class MultiSelectHelper {
...
             }
+            queueAction.outline = true
 
             if !Settings.mobileDataAllowed() {
-                warningMessage = L10n.downloadDataWarning + "\n" + warningMessage
+                warningMessage = L10n.downloadDataWarningWithSettingsLink + "\n" + warningMessage
             }
-            confirmPicker.addDescriptiveActions(title: L10n.notOnWifi, message: warningMessage, icon: "option-alert", actions: [downloadAction, queueAction])
+
+            let attributes = [L10n.settings: URL(string: "pktc://settings/storage-and-data")!]
```
SergioEstevao: Maybe this could be done using a Markdown link instead of this?

SergioEstevao (changes requested):
All is working correctly, but I left a comment regarding the way we are setting the link in the attributed string in order to try to have more failsafe way of doing the link connection.

danielebogo:
@SergioEstevao thanks for the heads up! Ready for another look!

SergioEstevao (approved):
:shipit: 

Thanks for addressing the issue with the url linking.

PR #3316, opened 2025-07-04
GRDB: enable only for TestFlight users

leandroalonso:
| 📘 Part of: #2773 | 
|:---:|

This PR adds the remote feature flag value for enabling GRDB for TestFlight users **only**. This is now done through the code since we downgraded Firebase due to crashes.

We will do a rollout, meaning not all users will have it enabled. We want to test that in a controlled way so we ensure everything goes smoothly.

## To test

1. [Comment this line](https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/AppDelegate.swift#L316) so the feature flag code runs in `DEBUG`
2. Change [isTestflight to `false`](https://github.com/Automattic/pocket-casts-ios/blob/trunk/Modules/Utils/Sources/PocketCastsUtils/Feature%20Flags/FeatureFlag.swift#L386)
3. Do a clean install of the app
4. ✅ Search for `Overriding grdb to true`. This log might be or might not be there. Try a few times, each time deleting and running a clean install of the app. This way you can also test that the rollout is working as intended.
6. ✅  You should should see a `Failed to set remote feature flag grdb: cannotBeOverridden` because this is not TestFlight
7. Change `isTestflight` to `true`
8. Delete the app, run it again until you see the `Overriding grdb to true`
9. ✅ This time there won't be a `Failed to set remote feature flag grdb: cannotBeOverridden` though
10. Re-run the app
11. ✅ You should see `[DataManager] Initialized using GRDB`

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Public/DataManager.swift
@@ -43,9 +43,11 @@ public class DataManager {
             config.busyMode = .timeout(10)
             dbQueue = GRDBQueue(dbPool: try! DatabasePool(path: DataManager.pathToDb(), configuration: config), logger: Self.logger)
             DataManager.setDatabaseFileProtectionToNone()
+            FileLog.shared.addMessage("[DataManager] Initialized using GRDB")
         } else {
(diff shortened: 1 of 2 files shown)


danielebogo (approved):
All good! I was able to see the override grdb set to false and true.
All logs were correct
:shipit:

PR #3360, opened 2025-07-28
Switch allowsCellularAccess to allowsExpensiveNetworkAccess

bjtitus:
Related to [PCIOS-40](https://linear.app/a8c/issue/PCIOS-40/bug-downloading-episodes-via-mobile-data-not-working-as-expected) but it probably doesn't fix it. `allowsCellularAccess` typically mimics `allowsExpensiveNetworkAccess` but there may be cases where it does not.

Since we're [checking](https://github.com/Automattic/pocket-casts-ios/blob/51d67cda29f7a5b741d057977a3a9de2a2007f9a/Modules/Utils/Sources/PocketCastsUtils/Networking/NetworkUtils.swift#L33) `NWPath.isExpensive` we should use the same when configuring our URLSessions or there may be situations where the prompt isn't shown but requests don't go out.

## To test

* Enable "Warn Before Using Data" setting
* Switch to cellular data
* Download an episode
* Approve the download from the warning
* Make sure the episode downloads

## Checklist

### DIFF
--- a/podcasts/DownloadManager.swift
+++ b/podcasts/DownloadManager.swift
@@ -90,7 +90,7 @@ class DownloadManager: NSObject, FilePathProtocol {
 
     lazy var wifiOnlyBackgroundSession: URLSession = {
         var config = URLSessionConfiguration.background(withIdentifier: "au.com.shiftyjelly.PCBackgroundSession")
-        config.allowsCellularAccess = false
+        config.allowsExpensiveNetworkAccess = false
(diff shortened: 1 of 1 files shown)


pocketcasts:
Version `7.95` has now entered code-freeze, so the milestone of this PR has been updated to `7.96`.

danielebogo (approved):
It's working correctly! :shipit:

PR #3383, opened 2025-08-01
Limit playback position changes when switching episodes

### LINKED ISSUE #1359: Investigate: After Scrubbing Playback to the Right, the Next Episode Starts From Middle/End or Is Skipped
### Description

User descriptions of the issue mention using the scrubber on the lock screen player to drag playback to the end. Instead of the next episode in their Up Next queue playing as normal, it either starts from the middle/end or is skipped entirely.

We haven't been able to replicate the issue so far. However, the user in #7475878-zen mentioned the following:

> I swiped the scrubber off the right edge of the screen from this view of the app (to be clear, I wasn't trying to bring the scrubber to the end of the episode, or even keep my finger on the displayport; I "grabbed" the scrubber and swiped across the entire device off the right):

_No response_

### Other Notes

We haven't been able to replicate so far, and the issue also seems to occur randomly for users.

We've started asking them for episodes they've experienced the issue with, in case the issue can be narrowed down to certain file types or hosts.

#7475878-zen:

App Version: 7.53.1
Device: iPhone16,1
OS: 17.2

They experienced the issue with the following episodes:

- Playing first: https://pca.st/episode/aed58424-32b5-4b47-8e4a-6ee7d025a46a
- Supposedly playing next: https://pca.st/episode/e2bbe789-0a4b-4716-9064-775494f0cc3d

They have since shared more episodes and logs:

Log UUID CDFF4EDD-805B-4EBE-87A4-4277754E8845, from around 2024-01-18 12:21:
- It looks like this was the episode that was playing first: 'Echo' Reactions and Armchair CEO: 'Daredevil' Edition | The Midnight Boys 
- And this is what was supposed to play next: Kurtis Blow

```
2024-01-18 12:21:44 Using DefaultPlayer
2024-01-18 12:21:44 activating audio session succeeded
2024-01-18 12:21:44 Sent position 6891 status 3 for episode 'Echo' Reactions and Armchair CEO: 'Daredevil' Edition | The Midnight Boys to server
2024-01-18 12:21:44 Remote control: changePlaybackPositionCommand
2024-01-18 12:21:44 seek to 6891.676734693878 startPlaybackAfterSeek false
2024-01-18 12:21:44 saving played up to 4760.633469387755 for episode Kurtis Blow
2024-01-18 12:21:45 Autoplay: reset the last playlist
2024-01-18 12:21:45 Finished playing Kurtis Blow
```

bjtitus:
Fixes #1359

When the changes position command from the lock screen happens while we are switching episodes, it seems that we sometimes run into race conditions where the episode is skipped. The remote command doesn't give us any info about the media item but we can track when the episode changes from our own internal notifications. For now, I've set a delay of 2 seconds since that seems like a reasonable amount of time where trying again would work.

The `limitPlaybackPositionChanges` flag controls these changes.

### Episode switching
* Load several episodes in your Up Next queue. Ideally these should be unplayed so they aren't pre-cached. Also, make sure you know the order so you know whether one is skipped.
* Play the first
* Scrub to the end of the episode
* ✅ Ensure the next episode from the queue is played
* Repeat scrubbing and skipping
* Try this with and without Trim Silence enabled

### Normal scrubbing
* Play an episode
* Scrub from the lock screen
* ✅ Ensure scrubbing works as expected and seeks to the proper position

One way you can tell this is being caught without checking the logs is that the scrub command may provide haptic feedback when it fails, this should mean the new code caught the case and, if you try again, scrubbing will work.

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -74,6 +74,9 @@ class PlaybackManager: ServerPlaybackDelegate {
 
     private(set) var transcriptsAvailable = false
 
+    /// The time the episode was last switched as tracked by handleCurrentlyPlayingEpisodeUpdated
+    private var episodeSwitchTime: Date?
+
     init() {
         queue = PlaybackQueue()
         queue.loadPersistedQueue()
@@ -93,6 +96,7 @@ class PlaybackManager: ServerPlaybackDelegate {
         NotificationCenter.default.addObserver(self, selector: #selector(refreshRemoteCommands), name: Constants.Notifications.remoteCommandSettingsChanged, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(updateNowPlayingInfo), name: Constants.Notifications.userEpisodeUpdated, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(updateAllNowPlayingData), name: .episodeEmbeddedArtworkLoaded, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(handleCurrentlyPlayingEpisodeUpdated), name: Constants.Notifications.currentlyPlayingEpisodeUpdated, object: nil)
 
         // run these on a background queue because some of them might call our singleton instance back, causing a crash because PlaybackManager.shared is called from the init method
         DispatchQueue.global().async {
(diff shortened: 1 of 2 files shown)


podcasts/PlaybackManager.swift
```diff
@@ -1731,7 +1735,17 @@ class PlaybackManager: ServerPlaybackDelegate {
                     if Settings.legacyBluetoothModeEnabled(), seekEvent.positionTime < 1 {
                         FileLog.shared.addMessage("Remote control: ignoring changePlaybackPositionCommand, it's to 0 and legacy bluetooth mode is on")
                     } else {
-                        FileLog.shared.addMessage("Remote control: changePlaybackPositionCommand")
+                        FileLog.shared.addMessage("Remote control: changePlaybackPositionCommand to \(seekEvent.positionTime)")
+
+                        if FeatureFlag.limitPlaybackPositionChanges.enabled {
+                            // Check if we're still in the limiting window after an episode change
+                            if let switchTime = episodeSwitchTime,
+                               Date().timeIntervalSince(switchTime) < 2.0 {
```
SergioEstevao: Just a small nit: Can we switch `Date()` to `Date.now` just find it a bit more explicit that is current date.

SergioEstevao (approved):
Working correctly just left a small nit.

PR #3404, opened 2025-08-06
GRDB: guard against ghost episodes

leandroalonso:
| 📘 Part of: #2773 | 
|:---:|

After we released GRDB we had a change in the way we convert database results into models. With `FMDB` if any row had a `NULL` value it would return something: `0`, an empty string `""`, etc.

`GRDB` is not lenient about this and it crashes — as it should.

One of the ~16 users affected contacted us and I could check that their database is corrupted. Specifically the episode index. This makes some queries to return "ghost episodes": episodes with all columns with `NULL` values.

This is a workaround solution: for episodes, we check if their `id` is valid, and if not we just don't return and don't parse them. This doesn't fix the underlying issue that is a user with a corrupted database, the other approach would be to delete their database — but that would cause data loss as some settings are saved there.

Copying database data to a non-corrupted database would be a proper solution but it's way more complex and I think it goes beyond of the scope of the GRDB migration — as GRDB is not the culprit here at all.

## To test

1. ✅  Run the app and check that episodes appear correctly

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/DB/GRDB/GRDBResultSet.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/DB/GRDB/GRDBResultSet.swift
@@ -57,7 +57,7 @@ class GRDBResultSet: PCDBResultSet {
     }
 
     func longLongInt(forColumn: String) -> Int64 {
-        row[forColumn]
+        row[forColumn] ?? 0
     }
 
     func bool(forColumn: String) -> Bool {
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/98">7.95 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

danielebogo (approved):
LGTM! All episodes are loaded correctly

PR #3405, opened 2025-08-06
Fix database export when using GRDB

bjtitus:
Fixes [PCIOS-85](https://linear.app/a8c/issue/PCIOS-85)

I think this is safe to include in the beta release because we still have some time and this _only_ affects the mostly debug export feature.

The `sqlite3-wal` file is also included:

> The WAL file exists for as long as any [database connection](https://www.sqlite.org/c3ref/sqlite3.html) has the database open. Usually, the WAL file is deleted automatically when the last connection to the database closes. However, if the last process to have the database open exits without cleanly shutting down the database connection, or if the [SQLITE_FCNTL_PERSIST_WAL](https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlpersistwal) [file control](https://www.sqlite.org/c3ref/file_control.html) is used, then the WAL file might be retained on disk after all connections to the database have been closed. The WAL file is part of the persistent state of the database and should be kept with the database if the database is copied or moved.

via [sqlite.org](https://www.sqlite.org/wal.html)

## To test

* Profile > Help & Feedback > ⋯ > Export Database
* Delete the app
* Re-import that file by opening from Files or drag and dropping it to the simulator
* Import database file
* Ensure that Podcasts and Settings are re-imported

## Checklist

### DIFF
--- a/podcasts/AppDelegate+UrlHandling.swift
+++ b/podcasts/AppDelegate+UrlHandling.swift
@@ -44,8 +44,8 @@ extension AppDelegate {
                 let alert = UIAlertController(title: "Import Podcasts and Settings", message: "Do you want to reset your podcasts and settings to this file?", preferredStyle: .alert)
                 alert.addAction(UIAlertAction(title: "Import", style: .default) { _ in
                     Task {
-                        let fileWrapper = try FileWrapper(url: url)
                         do {
+                            let fileWrapper = try FileWrapper(url: url)
                             try PCBundleDoc.performImport(from: fileWrapper)
                         } catch {
                             FileLog.shared.addMessage("File Import failed with error \(error)")
--- a/podcasts/AppDelegate+UrlHandling.swift
+++ b/podcasts/AppDelegate+UrlHandling.swift
@@ -45,7 +45,11 @@ extension AppDelegate {
                 alert.addAction(UIAlertAction(title: "Import", style: .default) { _ in
                     Task {
                         let fileWrapper = try FileWrapper(url: url)
-                        try PCBundleDoc.performImport(from: fileWrapper)
+                        do {
+                            try PCBundleDoc.performImport(from: fileWrapper)
+                        } catch {
+                            FileLog.shared.addMessage("File Import failed with error \(error)")
+                        }
                     }
                 })
                 alert.addAction(UIAlertAction(title: "Cancel", style: .cancel))
--- a/podcasts/PCBundleDoc.swift
+++ b/podcasts/PCBundleDoc.swift
@@ -10,7 +10,7 @@ struct PCBundleDoc: FileDocument {
     static var readableContentTypes = [UTType.pcasts]
 
     enum Constants {
-        static let databaseFilename = "database.sqlite"
+        static let databaseFilename = "database.sqlite3"
         static let preferencesFilename = "preferences.plist"
     }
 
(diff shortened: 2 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/98">7.95 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/PCBundleDoc.swift
```diff
@@ -10,7 +10,7 @@ struct PCBundleDoc: FileDocument {
     static var readableContentTypes = [UTType.pcasts]
 
     enum Constants {
-        static let databaseFilename = "database.sqlite"
+        static let databaseFilename = "database.sqlite3"
```
leandroalonso, line 13: @bjtitus I think there are two things we need to take into account with GRDB: the `shm` and the `wal` files. Eg.:

Image

bjtitus: Ok, yes. I was thinking these would be cleaned up with closing the database but it looks like that isn't always the case.

> The WAL file exists for as long as any [database connection](https://www.sqlite.org/c3ref/sqlite3.html) has the database open. Usually, the WAL file is deleted automatically when the last connection to the database closes. However, if the last process to have the database open exits without cleanly shutting down the database connection, or if the [SQLITE_FCNTL_PERSIST_WAL](https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlpersistwal) [file control](https://www.sqlite.org/c3ref/file_control.html) is used, then the WAL file might be retained on disk after all connections to the database have been closed. The WAL file is part of the persistent state of the database and should be kept with the database if the database is copied or moved.

via [sqlite.org](https://www.sqlite.org/wal.html)

bjtitus: @leandroalonso This is now updated with the wal file handling.

From what I can tell, the `shm` shared-memory file does not need to be included:
> The shared-memory file contains no persistent content. The only purpose of the shared-memory file is to provide a block of shared memory for use by multiple processes all accessing the same database in WAL mode. 

I also looked at the [backup method](https://github.com/groue/GRDB.swift#backup) provided by GRDB. I think that might be worth adding at some point but it seems to mostly work for now. What do you think?

podcasts/AppDelegate+UrlHandling.swift
```diff
@@ -45,7 +45,11 @@ extension AppDelegate {
                 alert.addAction(UIAlertAction(title: "Import", style: .default) { _ in
                     Task {
                         let fileWrapper = try FileWrapper(url: url)
-                        try PCBundleDoc.performImport(from: fileWrapper)
+                        do {
```
danielebogo, line 47: @bjtitus Would it make sense to include `let fileWrapper = try FileWrapper(url: url)` in `do { ... }` to catch also any possible error?

bjtitus: Good point. Changed!

danielebogo (approved):
Tested as described and it worked

PR #3428, opened 2025-08-19
Do not switch file if we already have the streamed file being played

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS-55 

## To test

1. Start
2. Enable the FF `doNotSwitchToDownloadedFile` in the Profile -> Settings -> Beta Features menu
3. Ensure that you do not have Trim Silence active
4. Start playing an episode of a podcast that you still didn't download
5. Check if you hear any "pop" sound when the episode downloads. You can start the download of episode after you start playing it to check the exact moment
6. Enable "Trim Silence" globally
7. Play another episode that is not downloaded
8. Check that this time your hear the "pop" when the download is finished. this is because the app needs to switch to the EffectsPlayer
9. Now disable the Trim Silence effect while the episode is playing
10. Switch to another episode that you didn't played before
11. Check that it plays correctly

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -2055,6 +2055,18 @@ class PlaybackManager: ServerPlaybackDelegate {
             // the current episode we were playing has downloaded, switch to playing the downloaded version
             let currentlyPlaying = playing()
             recordPlaybackPosition(sendToServerImmediately: false, fireNotifications: true)
+
+            let episodeIsChanging = refreshedEpisode.uuid != currentEpisode()?.uuid
+
+            if FeatureFlag.doNotSwitchToDownloadedFile.enabled,
+               FeatureFlag.streamAndCachePlayingEpisode.enabled,
+               !episodeIsChanging,
+               effects().trimSilence == .off,
+               !playerSwitchRequired(),
+               !refreshedEpisode.videoPodcast() {
+                return
+            }
+
             load(episode: refreshedEpisode, autoPlay: currentlyPlaying, overrideUpNext: false, saveCurrentEpisode: false)
             if refreshedEpisode.videoPodcast() {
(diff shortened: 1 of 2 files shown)


podcasts/PlaybackManager.swift
```diff
@@ -2055,6 +2055,18 @@ class PlaybackManager: ServerPlaybackDelegate {
             // the current episode we were playing has downloaded, switch to playing the downloaded version
             let currentlyPlaying = playing()
             recordPlaybackPosition(sendToServerImmediately: false, fireNotifications: true)
+
+            let episodeIsChanging = refreshedEpisode.uuid != currentEpisode()?.uuid
+
+            if FeatureFlag.doNotSwitchToDownloadedFile.enabled,
```
danielebogo: can we wrap this into a method in case we need to re-use it? Also would be great to add few logs including the param values. Wdyt?

danielebogo (approved):
Following your instruction, it seems to work correctly and no pop sound was emitted during the play with trim silence off 🎉

podcasts/PlaybackManager.swift
SergioEstevao: Do you want to log the episode uuid?

SergioEstevao: Something like `File downloaded for episode uuid is the same one has the current playing, skipping reload`

danielebogo: Maybe the episode title and the reasons of why is switching. In case we get some reports of pop sound and we find on log it didn't switch

danielebogo (approved):
All good! I got the log when switching player:
`Playback Manager: Needs to reload current episode [Cottingley Fairies - 346d38b3-4a85-43de-8140-90d838fd602d].
 Possible Reasons: Trim silence: low, Player switch required: false, Video podcast: false`

SergioEstevao:
Close in favour of https://github.com/Automattic/pocket-casts-ios/pull/3432

PR #3436, opened 2025-08-20
App Clips: add `pocketcasts.net` for testing

leandroalonso:
When working on App Clips it would be good to be able to use `pocketcasts.net` (Staging) for testing. Since this was missing I added.

## To test

Just code review.


bjtitus (approved):
LGTM! :shipit:

PR #3451, opened 2025-08-28
[Playlist] Smart Playlist detail and options

danielebogo:
| 📘 Part of: [Playlist](https://linear.app/a8c/project/ios-playlists-b287949437df/overview)
|:---:|

Fixes PCIOS-53
Fixes PCIOS-44
Fixes PCIOS-25

Adds the new Smart Playlist detail with all the options and edit.

| 1 | 2 | 3 | 4 |
| :-------: | :-------: | :-------: | :-------: |
| IMG_0324 | IMG_0325 | IMG_0326 | IMG_0327 |
| IMG_0328 | IMG_0329 | IMG_0330 | IMG_0331 |

## To test

- Make sure you have the playlist rebranding ff enabled
- Open Playlists and select the detail
- Test all the functionalities:
  - Make sure you can scroll
  - You can interact with the list episodes
  - The swipe gestures work
  - The options sheet works and also each option
  - The search work
  - The multiselection work
  - The edit from the detail opens the new rules screen and editing the rules update the detail screen
  - The Play all works and add all the episodes to the queue and start playing

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:
<details><summary>App Icon📲 You can test the changes from this Pull Request in <b>Pocket Casts Prototype Build</b> by scanning the QR code below to install the corresponding build.</summary>
<table>
<tr>
  <td><b>App Name</b></td><td>Pocket Casts Prototype Build</td>
</tr>
<tr><td><b>Build Number</b></td><td><code>11563</code></td></tr>
<tr><td><b>Version</b></td><td><code>PR #3451</code></td></tr>
<tr><td><b>Bundle ID</b></td><td><code>au.com.shiftyjelly.podcasts.prototype</code></td></tr>
<tr><td><b>Commit</b></td><td>3a0a36b93941b5bd14dd9383cf9b46f0985c7169</td></tr>
<tr><td><b>Installation URL</b></td><td><a href='https://appdistribution.firebase.google.com/testerapps/1:124902176124:ios:3999f43bfa1fcd1b1620f9/releases/11vbn9c07j3po?utm_source=fastlane'>11vbn9c07j3po</a></td></tr>
</table>
<em>Automatticians: You can use our internal self-serve MC tool to give yourself access to those builds if needed.</em>
</details>

Modules/DataModel/Sources/PocketCastsDataModel/Public/PlaylistHelper.swift
```diff
@@ -1,6 +1,204 @@
 import Foundation
 
 public class PlaylistHelper {
```
SergioEstevao: I don't know if the this class name was already present, but I think that `PlaylistQueryBuilder` could be a better name.

Modules/DataModel/Sources/PocketCastsDataModel/Public/PlaylistHelper.swift
```diff
@@ -1,6 +1,204 @@
...
+
+    private static func add(episodeUuidToAdd uuids: String?) -> QueryResult {
+        if let uuids {
+            return .value("AND ((episode.uuid = '\(uuids)') OR (", true)
+        }
+        return .value("AND (", false)
+    }
+
+    private static func add(sortFor sortType: Int32) -> String? {
+        if sortType == PlaylistSort.oldestToNewest.rawValue {
```
SergioEstevao: Suggestion: Why not write something like this:
```
guard let sort = PlaylistSort(rawValue: sortType) else {
    return nil
}

switch sort {
...
}
```

This way if new sort types are added you get warned to update this method.

Modules/DataModel/Sources/PocketCastsDataModel/Public/PlaylistHelper.swift
```diff
@@ -1,6 +1,204 @@
...
+            return "ORDER BY episode.duration DESC, episode.addedDate DESC"
+        }
+        return nil
+    }
+
+    private static func add(smartRulesFor playlist: EpisodeFilter) -> QueryResult {
+        var queryString = ""
+        var haveStartedWhere = false
+        // Playing Status
+        if !(playlist.filterUnplayed && playlist.filterPartiallyPlayed && playlist.filterFinished), playlist.filterUnplayed || playlist.filterPartiallyPlayed || playlist.filterFinished {
```
SergioEstevao: Could we refactor each `if` to a different method something like 

``` buildFilterPlayedStatus() -> String```

File: podcasts/New Detail/Playlist Header/PlaylistHeaderView.swift [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+                    PlaylistArtworkView(items: viewModel.images, imageSize: 192)
+                        .frame(width: 192.0, height: 192.0)
+                        .padding(.top, 5.0)
+                        .shadow(color: .black.opacity(0.2), radius: 30, x: 0, y: 2)
+                    Spacer()
+                }
+
+                VStack(spacing: 10.0) {
+                    Text(viewModel.playlist.playlistName)
+                        .font(size: 22.0, style: .body, weight: .bold)
```
SergioEstevao: Should this be base on the `.body` style or something larger? like `.title` ?

podcasts/New Detail/PlaylistDetailViewController+EditActions.swift
```diff
@@ -0,0 +1,172 @@
...
+
+        let optionsPicker = OptionsPicker(title: nil)
+
+        let chromecastAction = OptionAction(label: "Chromecast", icon: "nav_cast_off") {
+            Analytics.track(.filterOptionsModalOptionTapped, properties: ["option": "chromecast"])
+            self.castButtonTapped()
+        }
+        optionsPicker.addAction(action: chromecastAction)
+
+        let MultiSelectAction = OptionAction(label: L10n.selectEpisodes, icon: "option-multiselect") { [weak self] in
```
SergioEstevao: nitpick: the variable name should start with lowercase:  `multiSelectAction`

Suggestion: could each option be created in a separated function or dynamic var? in order to make this method smaller?

SergioEstevao:
Detected a theming bug when switching themes and having a playlist detail opened. Check the video bellow:


https://github.com/user-attachments/assets/2a93b141-b28b-4828-8ed7-992d457e23c2

podcasts/New Detail/PlaylistDetailViewController.swift
```diff
@@ -0,0 +1,383 @@
...
+            searchController.searchDelegate = self
+            searchController.view.translatesAutoresizingMaskIntoConstraints = false
+            addChild(searchController)
+        }
+    }
+    lazy private(set) var searchHeaderView: UIView = {
+        let header = UIView(frame: .zero)
+        header.backgroundColor = AppTheme.colorForStyle(.primaryUi02)
+        return header
+    }()
```
SergioEstevao, line 25: Can we have a space separating the properties/methods in this class?

podcasts/PlaylistsViewController.swift
```diff
@@ -82,8 +82,12 @@ class PlaylistsViewController: PCViewController, FilterCreatedDelegate {
 
             if let lastFilterUuid = UserDefaults.standard.string(forKey: Constants.UserDefaults.lastFilterShown), let filter = DataManager.sharedManager.findFilter(uuid: lastFilterUuid) {
                 DispatchQueue.main.async {
-                    let playlistViewController = PlaylistViewController(filter: filter)
-                    self.navigationController?.pushViewController(playlistViewController, animated: false)
+                    if FeatureFlag.playlistsRebranding.enabled {
+                        self.showFilter(filter)
+                    } else {
```
SergioEstevao: Do we need to have the `if FeatureFlag` here ? I see that inside the showlFilter method you also have an if based on the FF.

SergioEstevao (changes requested):
I tested the functionality and from what I can see it's working correctly. I only saw one small bug with theming.

Code wise I left some comments that i think can improve the code, let me know what do you think about them.

CookieyedCodes:
One thing on Android for a few years auto download for filters has had the option for 3 due to a community submission,
This I don't think was ever added to iOS, something that likely should be added :)

podcasts/New Detail/PlaylistDetailViewController+EditActions.swift
danielebogo: @SergioEstevao  I changed the name to be lowercase but this is the legacy code we have in the legacy filter. I didn't change anything here, just duplicated this part. Many parts are duplicated as long we will use the FF. I believe in general the way we handle the options should be refactored as the same code is duplicated in each VC

podcasts/New Detail/Playlist Header/PlaylistHeaderView.swift
danielebogo: We don't have a 1:1 match and I use our custom modifier to represent the same values set in Figma. We should all use this modifier as it supports and apply the dynamic text as well

Modules/DataModel/Sources/PocketCastsDataModel/Public/PlaylistHelper.swift
danielebogo: Yes this is an old class. I initially thought this could have been part of the model itself as class functions.

danielebogo:
@SergioEstevao would you mind give it another look? Especially at [this part](https://github.com/Automattic/pocket-casts-ios/pull/3451/commits/3a0a36b93941b5bd14dd9383cf9b46f0985c7169)

podcasts/New Detail/Playlist Header/PlaylistHeaderView.swift
SergioEstevao: I use it too, but I do think the style you select influence the way the sizing is done when changing font sizes in the system.
I normally try to align the style to the size the system uses  for the default Text Size.
You can check the values here: https://developer.apple.com/design/human-interface-guidelines/typography#Specifications

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
```diff
@@ -146,6 +146,10 @@ class EpisodeDataManager {
         loadMultiple(query: "SELECT * from \(DataManager.episodeTableName) WHERE \(customWhere)", values: arguments, dbQueue: dbQueue)
     }
 
+    func findSmartPlaylistEpisodesWhere(query: String, arguments: [Any]?, dbQueue: PCDBQueue) -> [Episode] {
```
bjtitus, line 149: Would it make sense to just expose `loadMultiple`? Why do we need this wrapper that just passes through the result of `loadMultiple`?

podcasts/EpisodeTableHelper.swift
```diff
@@ -15,6 +15,18 @@ struct EpisodeTableHelper {
         return newData
     }
 
+    static func loadSmartPlaylistEpisodes(tintColor: UIColor = AppTheme.appTintColor(), query: String, arguments: [Any]? = nil) -> [ListEpisode] {
+        let loadedEpisodes = DataManager.sharedManager.findSmartPlaylistEpisodesWhere(query: query, arguments: arguments)
+
+        var newData = [ListEpisode]()
```
bjtitus: Should we break out this shared code into something like a `mapEpisodes` function?

podcasts/FilterEditOptionsViewController.swift
```diff
@@ -19,8 +19,18 @@ class FilterEditOptionsViewController: PCViewController, UITableViewDelegate, UI
     private let buttonCellId = "ButtonCell"
     private let settingsCellId = "SettingsCell"
     private enum TableRow: Int { case filterName, color, icon, autodownload, autoDownloadLimit, siriShortcut }
-    private static let tableDataAutoDownloadDisabled: [[TableRow]] = [[.filterName], [.color, .icon], [.autodownload]]
-    private static let tableDataAutoDownloadEnabled: [[TableRow]] = [[.filterName], [.color, .icon], [.autodownload, .autoDownloadLimit]]
+    private static let tableDataAutoDownloadDisabled: [[TableRow]] = {
```
bjtitus, line 23: Shouldn't these be `static var` so they are run each time instead of only on load? I think any FeatureFlag updates won't get propagated if these are `static let`.

podcasts/New Detail/Playlist Header/PlaylistHeaderView.swift
```diff
@@ -0,0 +1,100 @@
+import SwiftUI
+
+struct PlaylistHeaderView: View {
+    @EnvironmentObject var theme: Theme
+    @StateObject var viewModel: PlaylistDetailViewModel
```
bjtitus: I believe this should be `ObservedObject` because it is being injected from the outside.

podcasts/New Detail/Playlist Header/PlaylistBlurHeaderView.swift
```diff
@@ -0,0 +1,18 @@
+import SwiftUI
+
+struct PlaylistBlurHeaderView: View {
+    @EnvironmentObject var theme: Theme
+    @StateObject var viewModel: PlaylistDetailViewModel
```
bjtitus: I believe this should be `ObservedObject` because it is being injected from the outside.

bjtitus (changes requested):
Overall, functionality looks good to me. I left several code cleanup changes that I think are worth making.

This trailing nav bar button appears like it's disabled to me so I'm not sure if it should use our normal tint color but maybe it's just me:
Image

Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
danielebogo: For few reasons: we never expose any of the manager but we access each trough the`DataManager`. The query used and injected is completely different.

podcasts/New Detail/Playlist Header/PlaylistHeaderView.swift
danielebogo: @SergioEstevao  Do you assign the style only or event the font size?

podcasts/FilterEditOptionsViewController.swift
danielebogo: @bjtitus I think we should probably use the other way and make the FF value being static as we don't reload the view when the FF changes. I would force the reload only when the view loads, which means each time the vc is pushed

podcasts/New Detail/Playlist Header/PlaylistHeaderView.swift
SergioEstevao: I do both, style and size

Modules/DataModel/Sources/PocketCastsDataModel/Public/PlaylistHelper.swift
SergioEstevao: Thanks for updating to the switch

podcasts/New Detail/Playlist Header/PlaylistHeaderView.swift
danielebogo: @SergioEstevao for now I only set the correspondent style. Let me know wdyt

danielebogo:
@bjtitus @SergioEstevao ready for another round of 👀 🙏

bjtitus (approved):

SergioEstevao (approved):
Thanks for addressing all the feedback!
Working great! :shipit:

PR #3548, opened 2025-10-06
Up Next clearing 

bjtitus:
Fixes [PCIOS-37](https://linear.app/a8c/issue/PCIOS-37/investigate-upnext-playlist-issue-on-iphone-and-ipad)

* Logs the current up next count on `PlaybackQueue: Saving replace`
* Removes `PlaybackQueue.insertIntoUpNext` which is now unused. This was confusing my log analysis so removing it makes this clear.
* Reloads the cached `PlaybackQueue.episode` if it is `nil` just to be _sure_ since a `nil` `currentEpisode` will result in the Up Next queue being overwritten.

This is a shot in the dark based on this user's logs. There was at least one instance where the code path chaged in be6b83ba replaced the queue with a new episode. Unsure whether this was the case where the Up Next queue was cleared for them but it seems possible the cache could get out of date in some situations.

## To test

* Test that adding episode to the Up Next queue works
* Empty the playback queue
* Play an episode
* Make sure the episode plays and is shown in the mini player and Up Next queue screens

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -547,6 +547,12 @@ class PlaybackManager: ServerPlaybackDelegate {
             AnalyticsEpisodeHelper.shared.episodeAddedToUpNext(episode: episode, toTop: toTop)
         }
 
+        // If we don't have a current episode, reload the persisted queue to updated our cache just in case
+        // We're getting reports from users about Up Next being cleared where this line is indicated by the logs
+        if currentEpisode() == nil {
+            queue.loadPersistedQueue()
+        }
+
         guard let playingEpisode = currentEpisode() else {
             // if there's nothing playing, just play this
             load(episode: episode, autoPlay: false, overrideUpNext: true)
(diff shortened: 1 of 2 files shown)


bjtitus:
I considered adding a Feature Flag for this but it's really just refreshing the cache so I'm not sure it's a huge deal.

danielebogo (approved):
- Episodes added correctly
- Episodes played correctly and added to the miniplayer

podcasts/PlaybackManager.swift
```diff
@@ -547,6 +547,12 @@ class PlaybackManager: ServerPlaybackDelegate {
             AnalyticsEpisodeHelper.shared.episodeAddedToUpNext(episode: episode, toTop: toTop)
         }
 
+        // If we don't have a current episode, reload the persisted queue to updated our cache just in case
+        // We're getting reports from users about Up Next being cleared where this line is indicated by the logs
+        if currentEpisode() == nil {
```
danielebogo, line 552: @bjtitus  I agree about not adding the FF. Maybe add a log for when this happens? Wdyt?

bjtitus: Good idea! I added this.

PR #3612, opened 2025-10-27
Conditionally show toolbar top trailing item in bookmarks screen

mokagio:
That is, only add the button if there are bookmarks on which it can act.

Before
image

After
image

Fixes PCIOS-234

## To test

Run the app using a free account or an paid account that has no bookmarks. Notice how there is no empty button in the bookmarks screen.

When adding bookmarks, notice the button appears and still works:

image

(_Also notice that the three dots are tinted blue/primary while the back button is white. Not sure how to fix that just yet._)

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary. — N.A.


bjtitus (approved):
This looks good. Thank you, Gio!

PR #3703, opened 2025-11-17
[Playlists] Improve Playlist list loading + fix tip source view

danielebogo:
| 📘 Part of: [Playlists](https://linear.app/a8c/project/ios-playlists-b287949437df/issues?layout=board&ordering=priority&grouping=workflowState&subGrouping=none&showCompletedIssues=all&showSubIssues=true&showTriageIssues=false)
|:---:|

Fixes PCIOS-299
Fixes PCIOS-297

| New List | List Indigo + Tip |
| :-------: | :-------: |
| IMG_0483 | IMG_0482 |

This PR adds:
- The Playlist query builder uses a new enum to return the first distinct podcasts based on the playlist. This will avoid filtering the whole episodes list.
- A new `PlaylistMetadataLoader`, a new `actor`, is now in charge of loading the the playlist episode count and the first distinct podcasts. This will work as cache, avoiding task re-entrancy, race conditions and empty states each time the view is reloaded. This also reduced the cell view model at its minimum requirements. 
- To adopt this, I created 3 new classes: `NewPlaylistCell`, `NewPlaylistCellView`, `NewPlaylistCellViewModel`. The plan is to replace the previous classes in other PRs and remove the `New` prefix after all.
- This PR also fixes the top left icon color
- The vc also uses now the cell image source to show the tip

## To test

- Start from a fresh install and login with a user that has multiple playlists
- Go to Playlists and the list should load
- Smoke test the Playlists list by entering inti a detail, change order of episodes to change the artwork 
- Navigate back and confirm it changes properly
- Create a new manual playlist (empty is fine)
- Go back and the tooltip for drag and drop should appear like the screenshot above

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/109">8.0 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/PlaylistsViewController+Table.swift
```diff
@@ -24,14 +24,25 @@ extension PlaylistsViewController: UITableViewDelegate, UITableViewDataSource {
...
-            let cell = cell(tableView, for: PlaylistCell.reuseIdentifier) as! PlaylistCell
+            let cell = cell(tableView, for: NewPlaylistCell.reuseIdentifier) as! NewPlaylistCell
             if let playlist = playlists[safe: indexPath.row] {
-                cell.configure(playlist: playlist, isLastRow: indexPath.row == playlists.count - 1)
+                if cell.tag != indexPath.row { cell.resetAllMetadata() }
+                cell.tag = indexPath.row
+                cell.set(playlistName: playlist.playlistName, isManualPlaylist: playlist.manual)
+                Task { [weak self] in
+                    await self?.playlistMetadataLoader.loadMetadata(for: playlist) { count in
+                        guard let cell = tableView.cellForRow(at: indexPath) as? NewPlaylistCell else { return }
```
danielebogo: @bjtitus @SergioEstevao any better idea to avoid detecting the cell for that specific indexpath when the block is called and set the value? This avoids setting values in wrong cells

danielebogo:
@bjtitus @SergioEstevao I pushed a change here [56a010f](https://github.com/Automattic/pocket-casts-ios/pull/3703/commits/56a010f0eb8a3d134b8cd54d82aad6d27bac7464) tome the loading logic into the cell and cancel the task when the prepareForReuse is called

podcasts/NewPlaylistCell.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    func loadMetadata(for playlist: EpisodeFilter) {
+        if playlistID == playlist.uuid {
+            return
+        }
+
+        playlistID = playlist.uuid
+        playlistCountLoadTask = Task { [weak self] in
+            guard let self else { return }
+            let count = await self.playlistMetadataLoader.loadCount(for: playlist)
```
SergioEstevao, line 118: What about adding here: 
```
let loadingPlaylist = playlistID
let count = await self.playlistMetadataLoader.loadCount(for: playlist)
if self.playlistID != loadingPLaylist {
   return // it's not worth updating because meanwhile we got a call to update to other index position/playlist
}
```

SergioEstevao (changes requested):
I tested the PR and I found the following:

- The loading of counter and image is much improved, and for majority of scenarios with relative size playlists up to 20K element the loading is much improved and faster.
- When having a very large playlist 40K and above I saw some refresh, but only if that row is visible on playlist screen
- The images are also much quicker to show, and again if no playlist with 40 is visible they show quickly

Taking in account that 40K is a bit of an edge case, maybe this can be handle in another PR?

The Tip display is now pointing to the image on the playlist

danielebogo:
Thanks @SergioEstevao , we can continue improving the list, maybe dropping UIKit entirely. Though I wouldn't block this TBH if there's some flash when the images changes as the queries are faster and this also contains the tip showed correctly.

bjtitus:
@danielebogo 
I think there's an empty AND clause bug but I don't see any issue in the UI itself:

- Create a new account
- Subscribe to a few podcasts (DON'T UNSUBSCRIBE FROM ANY)
- Create a new smart playlist with all default settings
- Notice this in the Console:

```
EpisodeDataManager.loadMultiple Episode error: SQLite error 1: near ")": syntax error - while executing `WITH numbered_episodes AS (
    SELECT episode.*,
           ROW_NUMBER() OVER (
             PARTITION BY episode.podcast_id
            ORDER BY episode.publishedDate DESC, episode.addedDate DESC
           ) AS rn
    FROM SJEpisode episode
    LEFT JOIN SJPodcast podcast
      ON episode.podcast_id = podcast.id
    WHERE episode.archived = 0 AND ( )
)
SELECT *
FROM numbered_episodes
WHERE rn = 1
ORDER BY numbered_episodes.publishedDate DESC, numbered_episodes.addedDate DESC
LIMIT 4
```

I added a unit test to reproduce and a potential fix in [this branch](https://github.com/Automattic/pocket-casts-ios/tree/bjtitus/playlist-query-builder-tests) but I'm not totally sure why the image grid still appears.

SergioEstevao (approved):
Did another round of tests and refreshing is now looking much better.
Going in and out of the details without changes work seamless and changing between main tabs also shows playlists without the need of refresh.
Great work!

bjtitus (approved):
Re-tested and I don't see the error any more.

Going in and out of the Playlist list looks much better 👍

PR #3717, opened 2025-11-19
[Tooling] Improve `new_hotfix_release` lanes to work when there's no release tag

iangmaia:
Related to AINFRA-1518

## Description

This PR aims to add more robustness and require less manual intervention to the hotfix creation lane `new_hotfix_release`.

We have observed that sometimes it is possible that the release has been finalized but a release tag isn't present yet for some reason (likely due to the fact that the release hasn't been published yet), so release managers will get an error and require manual intervention.

This PR changes the code so that, if no release tag is present, we start the hotfix branch based on the corresponding release branch.

## Testing instructions

These are some test scenarios that can be simulated:
1. There is a release tag that can be used to create the hotfix branch
2. There's no tag, but there's still a release branch that can be used to create the hotfix branch
3. There's no tag or branch (error)
4. The provided version doesn't have a patch component like `30.6` instead of `30.6.1` (error)

To test it, you can create test branches / tags (⚠️ delete them later or, even better, use [this technique](https://github.com/bloom/DayOne-Apple/pull/27670#issuecomment-3553138802) of pointing your working copy’s `origin` remote to a local bare repo instead ⚠️ ) and run `bundle exec fastlane new_hotfix_release skip_confirm:false version:"$VERSION"` and then answer `no` after the prompt so that the lane execution stops and check the printed values.

For example:
1. Create and push release branch `release/30.6` (which doesn't exist and has no corresponding tag for that version)
2. Run `bundle exec fastlane new_hotfix_release skip_confirm:false version:30.6.1`
3. It should use the `release/30.6` branch instead of trying to use a tag. It should print:
```
ℹ️  Tag '30.6' not found on the remote. Using release branch 'release/30.6' as the base for hotfix instead.

New hotfix branch from release/30.6: release/30.6.1

Current release version: 8.0
New hotfix version: 30.6.1

Current build code: 8.0.0.3
New build code: 30.6.1.0


Do you want to continue? (y/n)
```

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -502,19 +502,21 @@ platform :ios do
 
     # Parse the provided version into an AppVersion object
     parsed_version = VERSION_FORMATTER.parse(version)
+    # Validate that this is a hotfix version (must have a patch component > 0)
+    UI.user_error!("Invalid hotfix version '#{version}'. Must include a patch number.") unless parsed_version.patch.to_i.positive?
     previous_version = VERSION_FORMATTER.release_version(VERSION_CALCULATOR.previous_patch_version(version: parsed_version))
 
     release_branch_name = "release/#{version}"
     previous_release_branch = "release/#{previous_version}"
 
     # Determine the base for the hotfix branch: either a tag or a release branch
-    base_ref_for_hotfix = if git_tag_exists(tag: previous_version)
+    base_ref_for_hotfix = if git_tag_exists(tag: previous_version, remote: true)
                             previous_version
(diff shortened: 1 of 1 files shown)


mokagio (approved):
Tested a few scenarios locally. I didn't take screenshots because of the many logs lines.

iangmaia:
@AliSoftware final ping on this just FYI -- given @mokagio 's last pass, I think it's ok that he merged the PR (worst case I can open a new one). The last change we discussed about validating that the hotfix branch doesn't exist was implemented on hhttps://github.com/Automattic/pocket-casts-ios/pull/3717/commits/aadae186c3ca1f30b43fd6566e35093658f3e9a8.

PR #3777, opened 2025-12-05
[Playlists] Use Playlists cache before query

bjtitus:
Part of [POC-431](https://linear.app/a8c/issue/POC-431/playback-pauses-while-scrolling-playlists)

This doesn't directly solve the query performance issues but uses a cache + update scheme similar to Android.

The upside is we can show something quicker and downgrade the QoS of the query so it doesn't impact Effects Player.

The downside is that we will be showing out of date data some of the time which will eventually update and lead to pop in.

## Checklist

### DIFF
--- a/podcasts/PlaylistMetadataLoader/PlaylistMetadataLoader.swift
+++ b/podcasts/PlaylistMetadataLoader/PlaylistMetadataLoader.swift
@@ -1,4 +1,5 @@
 import PocketCastsDataModel
+import PocketCastsUtils
 
 actor PlaylistMetadataLoader {
     private var counts: [String: Int] = [:]
@@ -118,7 +119,7 @@ actor PlaylistMetadataLoader {
         let playlist = playlist
         let dataManager = self.dataManager
 
-        return await Task(priority: .userInitiated) {
+        return await Task(priority: FeatureFlag.playlistDataCacheBeforeQuery.enabled ? .medium : .userInitiated) {
             dataManager.allPlaylistEpisodeCount(
                 for: playlist,
                 episodeUuidToAdd: playlist.episodeUuidToAddToQueries(),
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/PlaylistMetadataLoader/PlaylistMetadataLoader.swift
```diff
@@ -36,12 +37,12 @@ actor PlaylistMetadataLoader {
         self.episodesDataManager = episodesDataManager
     }
 
-    func cachedCount(for playlistID: String) -> Int {
-        return counts[playlistID] ?? 0
+    func cachedCount(for playlistID: String) -> Int? {
```
bjtitus, line 40: These weren't being used before

podcasts/PlaylistMetadataLoader/PlaylistMetadataLoader.swift
```diff
@@ -115,7 +119,7 @@ actor PlaylistMetadataLoader {
         let playlist = playlist
         let dataManager = self.dataManager
 
-        return await Task(priority: .userInitiated) {
+        return await Task(priority: FeatureFlag.playlistDataCacheBeforeQuery.enabled ? .medium : .userInitiated) {
```
bjtitus, line 122: Downgrading QoS seems to prevent the audio skip, for me at least, with Effects Player.

podcasts/PlaylistMetadataLoader/PlaylistMetadataLoader.swift
danielebogo: No this is used only in the detail. It was causing issues in the list. The idea was to always return a value while fetching for a new one

PR #3829, opened 2025-12-23
Resource loader use memory buffer to fulfil requests

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

This PR improves the `MediaExporterResourceDelegate` to allow fulfilling of media requests using the media buffer.
This allows the following benefits:
- Larger in memory streaming cache, because slow networks will not delay first answer until first write
- Less write operations to disk, because we have a larger buffer before doing a write

## To test

1. Start the app
2. Start playing an episode that you don't have downloaded
3. Check if audio playback works correctly
4. Test this with slower connections, you can use the Network Conditioner or a device on a slower cellular connection

## Checklist

### DIFF
--- a/podcasts/MediaExporterResourceLoaderDelegate.swift
+++ b/podcasts/MediaExporterResourceLoaderDelegate.swift
@@ -8,7 +8,7 @@ import PocketCastsUtils
 /// MediaExporterItemConfiguration global configuration.
 private enum MediaExporterItemConfiguration {
     /// How much data is downloaded in memory before stored on a file.
-    public static var downloadBufferLimit: Int = 256.KB
+    public static var downloadBufferLimit: Int = FeatureFlag.streamAndDownloadReadFromMemoryBuffer.enabled ? 256.KB : 16.KB
 
     /// How much data is allowed to be read in memory at a time.
     public static var readDataLimit: Int = 10.MB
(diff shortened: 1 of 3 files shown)


podcasts/MediaExporterResourceLoaderDelegate.swift
```diff
@@ -8,7 +8,7 @@ import PocketCastsUtils
 /// MediaExporterItemConfiguration global configuration.
 private enum MediaExporterItemConfiguration {
     /// How much data is downloaded in memory before stored on a file.
-    public static var downloadBufferLimit: Int = 16.KB
+    public static var downloadBufferLimit: Int = 256.KB
```
danielebogo: Do you think it's better to change the buffer limit only with the new FF being on?

danielebogo (approved):
It looks good! I tested also with a very bad connection profile and it worked. Also switching back to the previous episode I was playing it did load it very quick.
I left one small comment though, let me know what you do think

podcasts/MediaExporterResourceLoaderDelegate.swift
SergioEstevao: Good catch! Update here: [7052b19](https://github.com/Automattic/pocket-casts-ios/pull/3829/commits/7052b19b8618d2fa8ceb551052ced67c82465d9d)

PR #3833, opened 2025-12-24
Switch to allowsCellularAccess from allowsExpensiveNetworkAccess

bjtitus:
Fixes [PCIOS-406](https://linear.app/a8c/issue/PCIOS-406/downloading-on-cellular-data-even-when-set-to-only-on-unmetered-wifi)

This is feature flagged behind `useCellularNetworkAPIs`

## To test

1. Make sure “Only on Unmetered Wifi” is enabled for Auto Download 
2. Make sure “Up Next” is enabled in Auto Download 
3. Make sure “Warn Before Using Data” is enabled in Storage & Data Use 
4. Ensure you’re only connected to cellular data 
5. Add an episode to Up Next 
6. ✅ Verify episode download is queued

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -268,7 +268,7 @@ public enum FeatureFlag: String, CaseIterable {
     case activateAudioSessionInBackground
 
     /// Use cellular-specific network APIs instead of expensive network APIs
-    case useCellularNetworkAPIs
+    case useCellularNetworkApis
 
     public var enabled: Bool {
         if let overriddenValue = FeatureFlagOverrideStore().overriddenValue(for: self) {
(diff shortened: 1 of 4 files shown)


danielebogo (approved):
It's working fine during my tests

Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
```diff
@@ -267,6 +267,9 @@ public enum FeatureFlag: String, CaseIterable {
     /// activates the audio session in the background to avoid locks in the main thread
     case activateAudioSessionInBackground
 
+    /// Use cellular-specific network APIs instead of expensive network APIs
+    case useCellularNetworkAPIs
```
SergioEstevao: Just a warning that this name of feature flag will be translated to `use_celullar_network_a_p_is` so maybe change to `Apis`

danielebogo: Good catch @SergioEstevao ! We can also force the remote key value by adding the case in `public var remoteKey: String?`

SergioEstevao (approved):
Working correctly, just left that little nitpick regarding the FF name.

PR #3836, opened 2025-12-28
Up Next reorder delay

bjtitus:
| | |
|--|--|
| **Android** | https://github.com/Automattic/pocket-casts-android/pull/4862 |

Fixes [PCIOS-419](https://linear.app/a8c/issue/PCIOS-419/dragging-up-next-and-syncing-at-the-same-time)

When we reorder episodes, a timer is started to sync after 5 seconds. However, if other episodes are being reordered while sync is occurring, the data sets can get out of sync. The fix here is to immediately cancel the timer when dragging begins.

https://github.com/user-attachments/assets/1a9205a0-a1d6-4ad6-89ab-0fd072385942

## To test

* Add many items to the Up Next queue
* Use Network Link Conditioner to set network profile to 3G
* Reorder episodes in the Up Next queue using drag and drop in relatively quick succession
* ✅ Verify that episode maintain ordering

## Checklist

### DIFF
--- a/PocketCastsTests/PlaybackQueueTests.swift
+++ b/PocketCastsTests/PlaybackQueueTests.swift
@@ -34,6 +34,36 @@ final class PlaybackQueueTests: XCTestCase {
                        "Should not include stale episode UUID in replacement list")
     }
 
+    func testRecentUserInteractionReturnsFalseWhenNoPreviousInteraction() {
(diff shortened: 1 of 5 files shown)


SergioEstevao (approved):
Tested this with the conditioner and order was always kept.
:shipit:

PR #3846, opened 2026-01-06
Widget logging update and ensure is run on background

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

This PR updates the Intent code for ensure that runs in background mode and we have extra logging

## To test

1. Start the app in a real device
2. Add some episode to the UpNext queue
3. Add an Up Next widget to the device that shows the current episode and the next ones
4. Kill the app
5. On the device widget tap play on one of the episodes
6. Check that plays correctly and the app is not sent to the foreground. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Add Deselect action from long press [#3786](https://github.com/Automattic/pocket-casts-ios/pull/3786)
 - Fix WatchManager to only report unknown message if it was truly unknown / unprocessed [#3748](https://github.com/Automattic/pocket-casts-ios/issues/3748)
 - Improves Add Episode performance [#3842](https://github.com/Automattic/pocket-casts-ios/pull/3842)
+- Ensure widget playback intent runs in the background [#3846](https://github.com/Automattic/pocket-casts-ios/pull/3846)
 
 8.2
 -----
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/112">8.3 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

bjtitus (approved):
This worked correctly but two follow ups:

1. Should we add this to the Changelog? Is there a user facing bug here?
2. I don't think the FileLog actually shows up anywhere visible to us since we don't send this log. Is that alright for now?

SergioEstevao:
@bjtitus 

> Should we add this to the Changelog? Is there a user facing bug here?

I will add it the purpose here is to try to fix the issue when play from the Widget does not work at first, and add extra logging for us to find out what is the root cause

> I don't think the FileLog actually shows up anywhere visible to us since we don't send this log. Is that alright for now?

I'm confused isn't this the one that the users can send to us? My idea is to ask user in Beta to test this version and if it fails get the logs from them to see what steps are being executed

bjtitus:
> I'm confused isn't this the one that the users can send to us?

I believe the widget logs to a separate file that isn't copied back over to the app. I have some stashed changes trying to share the log file using the App Group container.

PR #3899, opened 2026-01-27
Remove sorting from `cachePodcasts`

bjtitus:
While testing SQL Query changes with unit tests, I caught this slightly confusing behavior:

Since `PodcastDataManager.cachedPodcasts` uses a Dictionary, the results are not sorted, despite `PodcastDataManager.cachePodcasts` including `ORDER BY sortOrder ASC`.

Podcast sort order: Changed `PodcastDataManager.cachePodcasts` to remove the `ORDER BY` since we are not maintaining sort order anyway.

# Tests

It seems confusing that the results of the `cachePodcasts` query are sorted but then that sort is not really used. This removes `ORDER BY` to make this clearer. It should also improve query performance on this as well.

* Test podcast sort order in the Podcasts tab
* Test podcast sort order from the Folder. selection screen
* Test podcast sort order from the Add Episodes screen in a Manual Playlist

## Checklist

### DIFF
--- a/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
+++ b/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
@@ -667,7 +667,7 @@ class PodcastDataManager {
 
         dbQueue.read { db in
             do {
-                let resultSet = try db.executeQuery("SELECT * from \(DataManager.podcastTableName) ORDER BY sortOrder ASC", values: nil)
+                let resultSet = try db.executeQuery("SELECT * from \(DataManager.podcastTableName)", values: nil)
                 defer { resultSet.close() }
 
                 var newPodcasts = [String: Podcast]()


bjtitus:
@claude

leandroalonso:
@bjtitus I don't think this change is necessary. I was under the impression we do sort on the code and double-checked with Claude:

1. allPodcastsOrderedByTitle() and allPodcastsOrderedByAddedDate() (lines 99, 116) — both explicitly .sorted(by:) the array before returning.
2. allPodcastsInFolder() (line 237) — always applies its own sort (lines 256–264), including when using podcast1.sortOrder < podcast2.sortOrder for the custom sort case.
3. allPodcastsOrderedByNewestEpisodes() and allPodcastsOrderedByLastPlayedEpisodes() — bypass the cache entirely and query the DB with their own ORDER BY.
4. All other consumers (search, counts, settings bulk updates, image loading, etc.) don't careabout order at all.
5. `allPodcasts()` (line 85) — every caller either applies its own sort or only filters/counts the
  results. No one relies on iteration order.

About 5, some more context:

  - HomeGridDataHelper — calls allPodcasts() then passes results through
  gridItems(orderedBy:sortedPodcasts:) which applies its own sort based on LibrarySort type.
  - SearchResultsModel — iterates podcasts for substring matching, only cares about finding matches,
   not order.
  - ImageManager — iterates to collect image URLs for caching, order irrelevant.
  - PodcastManager — filters for podcasts with specific settings (push enabled, auto-download),
  doesn't use order.
  - DownloadSettingsViewController / NotificationsViewController / AutoAddToUpNextViewController —
  all filter podcasts by specific settings, order doesn't matter.
  - RefreshManager — refreshes all podcasts, order irrelevant.
  - ImportExportViewController — exports podcast UUIDs, order doesn't matter.
  - FoldersCoordinator — only checks podcast count for feature eligibility.

If the change was very simple I'd say go for it, but it adds one more dependency to the project. Not a critical thing but adding one more dependency for an issue that doesn't exist seems a bit too much? Up to you though.

If you really want to move ahead with this one let me know.

bjtitus:
> I was under the impression we do sort on the code

I think this is fair but if this is the intent, should we remove `ORDER BY` from [the `cachePodcasts` query](https://github.com/Automattic/pocket-casts-ios/blob/1ef192dc6ba6eac6f09ddad24899184b4b1d737c/Modules/DataModel/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift#L670)? I was hesitant to do this but I also had Claude review and it seemed to verify the callsites. I couldn't find any places where this seemed to cause an issue either way but it seems wasteful to do the sort in the query if we aren't going to ensure the order is maintained.

My thinking was that the `OrderedDictionary` at least keeps the intent from the query.

@leandroalonso What do you think?

leandroalonso:
@bjtitus hmm, I'd vote to remove the order.

I think this PR per se is already a good case of us getting confused about it, adding `OrderedDictionary` doubles down on this confusion. 😬

leandroalonso (approved):

PR #3908, opened 2026-02-01
GRDB Queries: Macros

bjtitus:
| 📘 Part of: [PCIOS-491](https://linear.app/a8c/issue/PCIOS-491/refactor-sql-queries-to-use-grdb-apis) |
|:---:|

- Add `GRDBMacros` Swift package with macros to reduce boilerplate:
  - `@GRDBRecord`
  - `@GRDBColumn`
  - `@GRDBIgnore`
 
- Add QueryInterface convenience methods to GRDBQueue for type-safe database operations
- Add `grdbQueryInterface` feature flag (disabled by default)

### GRDBRecord

`@GRDBRecord(table:SJEpisode)`

Depending on the type different implementations are generated:

  For NSObject subclasses (e.g., Episode, Podcast):
  - databaseTableName static property (if table: parameter provided)
  - CodingKeys enum matching @objc properties
  - init(from decoder: Decoder) implementation
  - Columns enum for type-safe query building
  - Extension conformances: FetchableRecord, TableRecord, Decodable

  For Codable types (e.g., PlaylistEpisode):
  - databaseTableName static property (if table: parameter provided)
  - Columns enum with string-based column names
  - Extension conformance: TableRecord

### GRDBColumn

`@GRDBColumn("columnName")` - Maps property to a different database column name

### GRDBIgnore
`@GRDBIgnore` - Excludes transient properties from code generation

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

bjtitus:
@SergioEstevao @leandroalonso I'd love your thoughts on using this Macro approach. When I first implemented this earlier in the week, I found that creating new Records led to a TON of additional code bloat but, more importantly, duplicated a bunch of the properties so that future changes would be very confusing. I thought adding Macros could help here, at least until we switch more to Codable (which I don't believe is possible right now given our Objective-C runtime usage).

I also prepared [User Episodes](https://github.com/Automattic/pocket-casts-ios/pull/3909) and [Bookmarks](https://github.com/Automattic/pocket-casts-ios/pull/3910) to give an idea of this.

Each of those also include Query tests which output the raw SQL for both versions. I created a normalization function just to get the tests passing, but included a table to compare before and after for those. I was thinking the tests are more of a temporary thing to help in reviewing these changes rather than something we would keep in the project longer term.

Modules/DataModel/Package.swift
```diff
@@ -17,14 +17,16 @@ let package = Package(
...
+        .package(path: "../GRDBMacros/")
     ],
     targets: [
         .target(
             name: "PocketCastsDataModel",
             dependencies: [
                 .product(name: "GRDB", package: "GRDB.swift"),
-                .product(name: "PocketCastsUtils", package: "Utils")
+                .product(name: "PocketCastsUtils", package: "Utils"),
+                .product(name: "GRDBMacros", package: "GRDBMacros")
```
leandroalonso, line 29: Are we adding this as a package because of the tests? I think it would make sense to have it on the DataModel directly.

leandroalonso (commented):
@bjtitus changes look good and the idea is interesting. Added a small comment about using it as a Package.

Modules/DataModel/Package.swift
bjtitus: I think the main reasons are:

1. I'm unsure of the Macro approach and whether we might want to drop it in the future so it seemed best to keep it isolated.
2. Keeping the macro code isolated from the DataModel code. Again, I think a sub-target could work fine but it seems more obvious for the package to be separate.

I don't think there's any limitation on adding a test target for Macros to DataModel if that makes more sense.

My initial reaction is that I'd rather keep it as isolated as possible but I don't feel strongly either way.

leandroalonso: Got it! Your reasoning makes sense.

SergioEstevao (approved):
The code is looking good.

Just to clarify is the final idea to refactor one by one our database managers to use this macro approach? Is there a place where I can read about the advantages of this?

bjtitus:
> Just to clarify is the final idea to refactor one by one our database managers to use this macro approach?

Yep, the ideas is to use the macros on our Data Model classes mostly to reduce the duplicated code we would need to make our NSObjects conform to `FetchableRecord`, `TableRecord`, and `Decodable`. Then I will create a PR for each of the models and the corresponding database manager. So those PRs should basically be the macros + query changes + unit tests.

> Is there a place where I can read about the advantages of this?

If you open an example like `UserEpisode.swift` in [this PR](https://github.com/Automattic/pocket-casts-ios/pull/3909) and right click the `GRDBRecord` macro at the top, then "Refactor" > "Inline Macro", it will generate the code in the file. That's the code for conforming to `FetchableRecord`, `TableRecord`, and `Decodable` that we would otherwise need to implement and keep updated. This seemed like a middle ground to keep our existing models but extend with GRDB support.

PR #3917, opened 2026-02-05
Remove notify nodes for Prototype Build steps

mokagio:
This way, PRs that have not unblocked the Prototype Build won't be stuck in the yellow state on GitHub because those jobs have not reported (and never will unless the Prototype Build is unblocked).

Before: 

image

With this PR:

TBD


bjtitus (approved):
Yay! This is awesome!👏

PR #3952, opened 2026-02-14
[Watch Up Next] Handle loginDetailsRequest in WatchConnectivity session

bjtitus:
| 📘 Part of: [Watch + Up Next Sync](https://linear.app/a8c/project/ios-watch-up-next-sync-3f36270786fd) |
|:---:|

When WatchConnectivity delivers loginDetailsRequest to the wrong delegate method (without reply handler), push data to watch instead of logging unknown message. This prevents excessive 'Unknown message type: loginDetailsRequest' log spam.

## To test

While debugging with Watch connected run the following with LLDB:

```
po WatchManager.shared.session(WCSession.default, didReceiveMessage: ["messageType": "loginDetailsRequest"])
```

Verify log appears:

```
WatchManager: loginDetailsRequest received without reply handler, pushing data to watch
```

## Checklist

### DIFF
--- a/podcasts/WatchManager.swift
+++ b/podcasts/WatchManager.swift
@@ -204,6 +204,13 @@ class WatchManager: NSObject, WCSessionDelegate {
             if DateUtil.hasEnoughTimePassed(since: ServerSettings.lastRefreshEndTime(), time: 30.minutes) {
                 RefreshManager.shared.refreshPodcasts()
             }
+        } else if WatchConstants.Messages.LoginDetailsRequest.type == messageType {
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
Looking good on my tests.

Does the co-pilot warning about the login session relevant?

bjtitus:
Yes, I think it's probably safer over all to avoid resetting that value. It should be set any time `loginDetailsRequest` occurs anyway so this is probably redundant in the majority of cases and could possibly cause a problem depending on the timing issue if reset elsewhere.

PR #3983, opened 2026-02-20
Skip migrations when User Defaults is unavailable

bjtitus:
Fixes [PCIOS-527](https://linear.app/a8c/issue/PCIOS-527/user-logged-out-unexpectedly)

This avoids logging users out when we shouldn't. 

See more context in the ticket but the basic rundown from the logs:

1. Keychain access failed with error `-25308` (`errSecInteractionNotAllowed`)
2. `AppDelegate.checkDefaults v5Run clearTokensFromKeyChain` is logged indicating the v5run migration occurred _after_ keychain errors even though there are many logs prior, indicating this isn't a new install.
3. User has to sign in again after all of this with `SyncSigninViewController.startSignIn`

This may be caused by [application prewarming](https://framna.com/articles/solving-prewarming).

## Questions

Should we do anything else here? Is there something else we should do other than just skip the migration?

## To test

* Test installing the app from fresh to ensure `v5run` occurs
*  Test starting the app again to ensure code isn't triggered

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,8 @@
 8.7
 -----
-
-
-8.7
------
 - Playlists: Add episode to a playlist from episode details [##3943](https://github.com/Automattic/pocket-casts-ios/pull/3943/) 
 - Refresh Now Playing info when Bluetooth device connects [#3964](https://github.com/Automattic/pocket-casts-ios/pull/3964) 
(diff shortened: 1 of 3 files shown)


SergioEstevao (approved):
Tested the scenarios in the PR description and all worked correctly.
Code looks good!
:shipit:

PR #3992, opened 2026-02-20
PCIOS-453: iOS: Podcast description can't be collapsed

pocketcasts:
Resolves https://linear.app/a8c/issue/PCIOS-453/ios-podcast-description-cant-be-collapsed

## Summary
- The tap gesture recognizer on the podcast description `RichExpandableLabel` (a `WKWebView` subclass) was being removed from the view whenever the description was expanded, making it impossible to collapse it by tapping again.

## Changes
- In `RichExpandableDescriptionView.swift`: moved `addGestureRecognizer(linkTapGesture)` from the `update()` method (where it was only called when `collapsed == true`) into `commonInit()`, so it is registered once and remains attached in both the collapsed and expanded states.
- Removed the corresponding `removeGestureRecognizer(linkTapGesture)` call from the expanded branch of `update()`.

## Testing
- Open the podcast detail page for any podcast with a description longer than 3 lines.
- The description starts collapsed (3 lines visible).
- Tap the description — it expands with an animation. ✅
- Tap the expanded description again — it collapses with the same easeInOut animation. ✅ (was broken before this fix)
- Links within the description still open correctly via the `WKNavigationDelegate`. ✅

---
🤖 This PR was created autonomously by linear-solver.

### DIFF
--- a/podcasts/RichExpandableDescriptionView.swift
+++ b/podcasts/RichExpandableDescriptionView.swift
@@ -58,6 +58,7 @@ class RichExpandableLabel: WKWebView {
         isUserInteractionEnabled = true
         scrollView.isScrollEnabled = false
         navigationDelegate = self
+        addGestureRecognizer(linkTapGesture)
         updateStyle()
     }
 
@@ -178,13 +179,11 @@ class RichExpandableLabel: WKWebView {
 
     private func update() {
         if collapsed {
-            addGestureRecognizer(linkTapGesture)
             let font = UIFont.preferredFont(forTextStyle: .body)
             let newHeight = Self.estimateHeightFor(maxLines: maxLines, lineHeightMultiple: desiredLinedHeightMultiple, font: font)
             heightConstraint.constant = newHeight
             heightChanged?(newHeight)
         } else {
-            removeGestureRecognizer(linkTapGesture)
             heightConstraint.constant = contentHeight
             heightChanged?(contentHeight.rounded(.up))
         }


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

leandroalonso:
@SergioEstevao while this fixes the issue, the animation is all wonky. I've spent some time trying to improve but I'm having issues with the text moving in the web view.

Question: is this something you experienced and why the collapse on tap is not available on iOS?

SergioEstevao:
> @SergioEstevao while this fixes the issue, the animation is all wonky. I've spent some time trying to improve but I'm having issues with the text moving in the web view.
> 
> Question: is this something you experienced and why the collapse on tap is not available on iOS?

From what I remember the UITableView animation for the SwiftUI hosted header expanding/collapsing was very sensitive, you need to start/end animation on it to work properly. Another thing that is tricky is that the WKWebView internal sizing once it expands always wants to report the higher height for the content.

leandroalonso:
@SergioEstevao got it. While this PR works, the collapse animation is quite bad IMHO. Not sure if we'd like to go with it. What do you think?

SergioEstevao:
> @SergioEstevao got it. While this PR works, the collapse animation is quite bad IMHO. Not sure if we'd like to go with it. What do you think?

The animation is not great, we could add it to trunk and see if our designers are ok with it, or if we should try to get a better solution.

SergioEstevao:
Looking at it a bit more, I think better to keep this out, until we get a good animation, it just feels weird that it jumps around and text moves in a strange way.

leandroalonso:
Yeah, agree, closing it.

PR #4017, opened 2026-02-25
Episode artwork network handling + prefer Show Notes image

bjtitus:
Fixes [PCIOS-482](https://linear.app/a8c/issue/PCIOS-482/podcast-artwork-is-displayed-instead-of-episode-artwork-in-episode)

The mp3 file episode artwork was preferred over the feed artwork. This PR changes this to prefer the feed artwork.

I noticed two issues testing this:
* The `NetworkUtils.shared.isConnected()` call was returning `false` for some reason. I think, instead, we should rely on the result of the connection to see if the request succeeds rather than pre-emptively checking connection.
* The Android and Web clients prefer the artwork from the Show Notes endpoint, rather than the one from the audio file. [Here's](https://github.com/Automattic/pocket-casts-android/blob/main/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/images/PocketCastsImageRequestFactory.kt#L98-L101) where Android does this:

```
    private fun RequestType.PodcastEpisode.data(context: Context) = if (useEpisodeArtwork) {
        episode.imageUrl ?: EpisodeFileMetadata.artworkCacheFile(context, episode.uuid).takeIf(File::exists) ?: episode.podcastArtworkUrl(context)
    } else {
        episode.podcastArtworkUrl(context)
    }
```

Also removes the `episodesInfoCacheReloadPolicy` Feature Flag which has been in production since December.

## To test

* Enable the Episode Artwork setting
* Play each of these episodes and ensure episode artwork appears in the player:
  * https://pca.st/episode/383f0386-f506-482d-b245-a6c73675be22
  * https://pca.st/episode/a45b9424-b448-40f6-b7ca-9a778e17e61b

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Playlists: Add episode to a playlist from episode details [##3943](https://github.com/Automattic/pocket-casts-ios/pull/3943/) 
 - Refresh Now Playing info when Bluetooth device connects [#3964](https://github.com/Automattic/pocket-casts-ios/pull/3964) 
 - Fix logout issue when refreshing in background while device is locked [#3983](https://github.com/Automattic/pocket-casts-ios/pull/3983) 
+- Fix Episode Artwork loading when audio asset image differs from feed [#4017](https://github.com/Automattic/pocket-casts-ios/pull/4017)
 
 8.6
 -----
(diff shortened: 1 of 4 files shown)


SergioEstevao (approved):
Tested both podcasts and episode and artwork displayed correctly. 
Turned device offline and the cache was used correctly.
:shipit:

PR #4028, opened 2026-02-27
[Watch + Up Next] Remove Watch Up Next sync limit

bjtitus:
| 📘 Part of: [Watch + Up Next Sync](https://linear.app/a8c/project/ios-watch-up-next-sync-3f36270786fd) |
|:---:|

Currently, Up Next sync to Apple Watch is limited to 50 episodes via `Constants.Limits.maxListItemsToSendToWatch`. This creates sync issues:

1. When phone has >50 episodes, UpNextListComparator receives truncated data (phoneUpNextCount > phoneEpisodes.count) and returns .notEnoughInformation
2. This prevents accurate sync decisions and can lead to data loss
3. Episodes beyond position 50 are never visible to watch

This adds a feature flag `unlimitedWatchUpNextSync` to allow syncing the full Up Next queue to watch. When enabled, all episodes in Up Next will be synced to watch instead of just the first 50.

In case this causes issues I've:
* Added Sentry logging so we can catch any `payloadTooLarge` errors we might receive.
* Added fallback logic to drop back to the 50 episode limit in cases where we receive `payloadTooLarge`.

## To test

I tested this with 1000+ episodes between iPhone 16 Pro and Apple Watch Series 11 and it was able to sync Up Next within a second or two. Adding episodes to top and bottom of playlist seemed to work without issue but the Up Next queue on the watch only shows the first 900 or so episodes (I think we're maxing out the list?).

* Install app to phone and watch (this code is only phone, but it doesn't hurt to make sure you're using most up to date code)
* Use the Phone source
* Test adding and removing from the Up Next queue
* Ensure episodes are synced as expected between watch and phone

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -292,6 +292,9 @@ public enum FeatureFlag: String, CaseIterable {
     /// Use transferUserInfo API for watch-to-phone actions and sendMessage for phone-to-watch state updates
     case watchTransferUserInfoApi
 
+    /// Remove the 50-episode limit when syncing Up Next to Apple Watch
+    case unlimitedWatchUpNextSync
+
     public var enabled: Bool {
         if let overriddenValue = FeatureFlagOverrideStore().overriddenValue(for: self) {
             return overriddenValue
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
Testing this with a list of around 200 episodes and moving episodes in and out of the phone and watch, and all worked correctly.

Code looks good!

:shipit:

PR #4043, opened 2026-03-02
Revert Watch App Release signing to fix App Store builds

iangmaia:
Fixes AINFRA-2128

## Summary
- Reverts the project file changes from 463f7dd4e that hardcoded Development signing settings (`CODE_SIGN_IDENTITY`, `DEVELOPMENT_TEAM`, `PROVISIONING_PROFILE_SPECIFIER`) in the Watch App Release configuration, causing App Store builds to fail looking for a dev profile.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR has the <code>Releases</code> label: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

SergioEstevao (approved):
:shipit:

podcasts.xcodeproj/project.pbxproj
```diff
@@ -12650,10 +12650,8 @@
 				CLANG_WARN_DIRECT_OBJC_ISA_USAGE = YES_ERROR;
 				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
 				CLANG_WARN_OBJC_ROOT_CLASS = YES_ERROR;
-				"CODE_SIGN_IDENTITY[sdk=watchos*]" = "iPhone Developer";
```
AliSoftware, line 12653: Nice catch!

I wonder if that is a one-time thing (i.e. it only happened because the Watch App target got recently modified for good reason, and this was introduced accidentally as part of this refactor of the targets, but we don't expect this to happen again anytime soon)?
Or if this is an issue that could happen too easily again in the future (e.g. Xcode randomly deciding to add that setting every time a developer opens to the "Signing" tab of the project and Xcode sees that setting empty and decides to fill it for us while it shouldn't?)

If there's a risk that Xcode might re-introduce that incorrect setting from time to time under us, maybe adding some lint rule or `Dangerfile` check to avoid this to happen again could help?

Another possibility I can think of for why this might have accidentally happened is if the Watch App doesn't have any Developer Code Signing setup currently (not even via `.xcconfig` files) which means that if a developer who has a physical Apple Watch wants to test the app on device, they couldn't unless they tried to change the Code Signing settings themselves to make it work (without realizing that this would break the Release Code Signing if not done via separate `.xcconfig` properly). If that's what happened, then we need to fix this to allow devs to test the Watch App on real device if they need to.

AliSoftware (approved):

podcasts.xcodeproj/project.pbxproj
iangmaia: > this might have accidentally happened is if the Watch App doesn't have any Developer Code Signing setup currently

Good point. Worth asking @bjtitus what happened to cause that change.
Tomorrow I'll check the `.xcconfig` files and the dev configurations to see if anything can be improved there.

twstokes: Related: https://github.com/Automattic/pocket-casts-ios/pull/3541

PR #4052, opened 2026-03-05
Ensure that Temporary download files are properly deleted from the system

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS-554 

This PR updates our download code so that files are removed and moved properly, without leaving temporary files lingering around in the system.

These files while stored on cache folder, were not being removed by the OS quickly and were being reported as space used on the system  Settings -> Storage -> Pocket Casts.

## To test

1. Start the app from scratch on a real device

### Test user explicit downloads or auto-downloads
3. Download some episodes from podcasts
4. Go to Profile -> Downloads 
5. Tap on the ... button on the top right
6. Tap on Clean Up
7. Check the space being used by all files with all the episode options active
8.  Now switch to the system Settings -> Storage -> Pocket Casts and check the storage usage for Documents & Data
9. Go back to the app, and tap on Clean Up
10. Go back to Settings and check if the storage for Pocket Cast Documents & Data went down for that amount

### Test streamed downloads 

12. Now go back to the app and play some episodes 
13. Now switch to the system Settings -> Storage -> Pocket Casts and check the storage usage for Documents & Data
14. Go back to the app, and tap on Clean Up
15. Go back to Settings and check if the storage for Pocket Cast Documents & Data went down for that amount

### Test if previous orphanated files are removed from the system

- Run the a previous version of the app that does not have this fix
- Download explicitly a large chunk of files
- Check the used space on Settings
- Now install this branch
- For a sync on the app
- Check if you get extra free space on Settings
 
## Checklist


podcasts/DownloadManager+URLSessionDelegate.swift
```diff
@@ -183,6 +183,7 @@ extension DownloadManager: URLSessionDelegate, URLSessionDownloadDelegate {
         let fileSize = FileManager.default.fileSize(of: location) ?? 0
         guard isEpisodeFileValid(contentType: contentType, fileSize: fileSize) else {
             markEpisode(episode, asFailedWithMessage: L10n.downloadErrorContactAuthorVersion2, reason: .suspiciousContent(fileSize))
+            StorageManager.removeItem(at: location)
```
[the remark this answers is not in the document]
SergioEstevao: Solved here: https://github.com/Automattic/pocket-casts-ios/pull/4052/commits/3fdfec1f3b0513d2592a1492e71f7be4de415d8e

podcasts/DownloadManager.swift
```diff
@@ -445,7 +445,9 @@ class DownloadManager: NSObject, FilePathProtocol {
             }
             if exportStatus.error == nil {
                 FileLog.shared.addMessage("DownloadManager stream and download: end downloading \(episode.uuid) successfully")
-                processEpisode(episode, downloadedFile: outputURL, reportedContentType: exportStatus.reportedType)
+                processEpisode(episode, downloadedFile: outputURL, reportedContentType: exportStatus.reportedType, copyFile: true)
+                //Now that the file is downloaded and copied we can mark it for deletion on release
+                customLoaderDelegate.deleteFileOnRelease = true
```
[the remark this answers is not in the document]
SergioEstevao: Solved here: https://github.com/Automattic/pocket-casts-ios/pull/4052/commits/3fdfec1f3b0513d2592a1492e71f7be4de415d8e

podcasts/DownloadManager.swift
```diff
@@ -445,7 +445,9 @@ class DownloadManager: NSObject, FilePathProtocol {
...
             } else {
                 FileLog.shared.addMessage("DownloadManager stream and download: failed downloading \(episode.uuid) -> \(exportStatus.error?.localizedDescription ?? "")")
```
leandroalonso, line 454: `deleteFileOnRelease` is never set in this case. Can the temp file always be there?

podcasts/DownloadManager+URLSessionDelegate.swift
```diff
@@ -191,7 +194,11 @@ extension DownloadManager: URLSessionDelegate, URLSessionDownloadDelegate {
         let destinationUrl = URL(fileURLWithPath: destinationPath)
 
         do {
-            try StorageManager.copyItem(at: location, to: destinationUrl, options: [.overwriteExisting])
+            if copyFile {
+                try StorageManager.copyItem(at: location, to: destinationUrl, options: [.overwriteExisting])
```
leandroalonso, line 198: If this throws the temp file will also not be deleted, right?

podcasts/DownloadManager.swift
```diff
@@ -445,7 +445,9 @@ class DownloadManager: NSObject, FilePathProtocol {
             }
             if exportStatus.error == nil {
                 FileLog.shared.addMessage("DownloadManager stream and download: end downloading \(episode.uuid) successfully")
-                processEpisode(episode, downloadedFile: outputURL, reportedContentType: exportStatus.reportedType)
+                processEpisode(episode, downloadedFile: outputURL, reportedContentType: exportStatus.reportedType, copyFile: true)
+                //Now that the file is downloaded and copied we can mark it for deletion on release
```
leandroalonso:
```suggestion
                // Now that the file is downloaded and copied we can mark it for deletion on release
```

leandroalonso (commented):
Left a few comments.

@SergioEstevao will that work only for new downloads? Does that mean that downloads executed prior of this change will have the file lingering there indefinitely?

SergioEstevao:
> @SergioEstevao will that work only for new downloads? Does that mean that downloads executed prior of this change will have the file lingering there indefinitely?

@leandroalonso  So the temporary files are stored in a cache directory so the iOS system will eventually delete them if it needs the space.

They will also be deleted if the user archives the episodes or unfollows the podcast, when the cleaup method runs

I also considered adding a new option to Clean Up Screen to delete all temporary file but think that will only have use for a short amount of time and only for specific users that may have a lot of those files lingering. 

Do you think it's worth to implement something to purge those files? Like a startup task, that goes and check of any lingering files that does not match any valid episode?

podcasts/DownloadManager.swift
SergioEstevao: The resource delegate is already deleting any file that does  not fully completes the download.

leandroalonso:
@SergioEstevao based on the original issue the user says "Pocket Casts is taking up 28GB of storage. Their downloaded files shows 20MB."

Shouldn't the system had deleted those?

I think the changes here are mostly fine, but my concern is that maybe nothing will change for this user specifically.

podcasts/DownloadManager+URLSessionDelegate.swift
SergioEstevao: So when we are copying we don't want to remove it because the file it's being used by the streaming player, when we have move ( it's user requested download) I added code [here](https://github.com/Automattic/pocket-casts-ios/pull/4052/commits/43f38313839c1455fcfb243c2a2ee9a5e7409d11) to delete the temp folder.

SergioEstevao:
> @SergioEstevao based on the original issue the user says "Pocket Casts is taking up 28GB of storage. Their downloaded files shows 20MB."
> 
> Shouldn't the system had deleted those?
> 
> I think the changes here are mostly fine, but my concern is that maybe nothing will change for this user specifically.

So we already had code on the `EpisodeManager->cleanupAllUnusedEpisodeBuffers` to remove tmp files. This was being called on the sync process but there could be cases where files where orphan in the tmp folder because of this existing bug of not removing tmp files.
 I now added an extra method that will remove any files on the tmp folder that are not matched to an episode.

SergioEstevao:
@leandroalonso do you mind to test and review this again? I added an extra code and test steps to cater for users that already have a lot of "orphan" files in their tmp folder.

podcasts/EpisodeManager.swift
```diff
@@ -378,6 +378,24 @@ class EpisodeManager: NSObject {
                 deleteDownloadedFiles(episode: episode)
             }
         }
+
+        // Remove any files in the temporary folder that were not part of the above, that should be orphan files
+        let filePaths = Set<String>(episodes.map { DownloadManager.shared.tempPathForEpisode($0) })
```
leandroalonso: `tempPathForEpisode` returns full path, right? And `enumerator(atPath:)` return relative filename. So the comparison in the line `394` will never succeed. Or am I missing something?

podcasts/DownloadManager+URLSessionDelegate.swift
```diff
@@ -191,14 +194,22 @@ extension DownloadManager: URLSessionDelegate, URLSessionDownloadDelegate {
...
 
             let newDownloadStatus: DownloadStatus = autoDownloadStatus == .playerDownloadedForStreaming ? .downloadedForStreaming : .downloaded
             dataManager.saveEpisode(downloadStatus: newDownloadStatus, sizeInBytes: fileSize, downloadTaskId: nil, episode: episode)
             dataManager.saveEpisode(downloadStatus: newDownloadStatus, lastDownloadAttemptDate: Date.now, autoDownloadStatus: autoDownloadStatus, episode: episode)
             EpisodeFileSizeUpdater.updateEpisodeDuration(episode: episode)
             NotificationCenter.postOnMainThread(notification: Constants.Notifications.episodeDownloaded, object: episode.uuid)
         } catch {
+            if !copyFile {
+                // Lets try remove the file so we don't have a pending file on the tmp folder
+                try? FileManager.default.removeItem(at: location)
```
leandroalonso: Any reason why not to use `StoreManager.removeItem`?

leandroalonso (commented):
@SergioEstevao

1. I think we should add logs into all this logic. This way we can check how things are going if any user contact us about that. Without logs we're in the dark;
2. Since this might affect streaming/download I believe it would be good to put it behind a feature flag. Or are we completely sure this won't introduce any other issues?

podcasts/DownloadManager+URLSessionDelegate.swift
SergioEstevao: Changed it here: [00e1ceb](https://github.com/Automattic/pocket-casts-ios/pull/4052/commits/00e1cebaafa48335e198a8af8bddc69f459969dd)

podcasts/EpisodeManager.swift
SergioEstevao: Good catch! I fixed it here: [Compare full paths and add logging](https://github.com/Automattic/pocket-casts-ios/pull/4052/commits/d663e1f97981329add68b7b2633e1854f9e908d2)

SergioEstevao:
@claude can you review this?

podcasts/DownloadManager+URLSessionDelegate.swift
```diff
@@ -191,14 +194,22 @@ extension DownloadManager: URLSessionDelegate, URLSessionDownloadDelegate {
...
+                try StorageManager.moveItem(at: location, to: destinationUrl, options: [.overwriteExisting])
+            }
 
             let newDownloadStatus: DownloadStatus = autoDownloadStatus == .playerDownloadedForStreaming ? .downloadedForStreaming : .downloaded
             dataManager.saveEpisode(downloadStatus: newDownloadStatus, sizeInBytes: fileSize, downloadTaskId: nil, episode: episode)
             dataManager.saveEpisode(downloadStatus: newDownloadStatus, lastDownloadAttemptDate: Date.now, autoDownloadStatus: autoDownloadStatus, episode: episode)
             EpisodeFileSizeUpdater.updateEpisodeDuration(episode: episode)
             NotificationCenter.postOnMainThread(notification: Constants.Notifications.episodeDownloaded, object: episode.uuid)
         } catch {
+            if !copyFile {
```
SergioEstevao, line 209: I don't think we need to FF flag this removal because it's for cases when we failed moving the file when a explicit download fails.

SergioEstevao:
@bjtitus can you give this one a looK?

bjtitus (approved):
Everything looked good in testing from what I could tell.

✅ Verified in-app storage matched system storage for downloads

✅ Verified that streamed episode files appear in storage usage after playback and are cleared with Clean Up

✅ Verified that previous orphaned files seemed to be removed on upgrade

My only other feedback is whether we should add a few unit tests to verify the temp file deletion.

SergioEstevao:
> My only other feedback is whether we should add a few unit tests to verify the temp file deletion.

@bjtitus how could we add those? add some random files and run the method?

bjtitus:
> how could we add those? add some random files and run the method?

Yeah, I was thinking the test can just make the file and run the method. Is that too much work just to verify?

This was the basic idea:
```swift
func testProcessEpisode_RemovesTempFile_WhenMoveSucceeds() {
    // Given: A successfully downloaded episode
    let testEpisode = Episode()
    testEpisode.uuid = "test-move-\(UUID().uuidString)"
    testEpisode.podcastUuid = podcast.uuid
    testEpisode.podcast_id = podcast.id
    testEpisode.autoDownloadStatus = AutoDownloadStatus.notSpecified.rawValue
    dataManager.save(episode: testEpisode)

    let fileManager = FileManager.default
    let tempFilePath = downloadManager.tempPathForEpisode(testEpisode)
    let destinationPath = downloadManager.pathForEpisode(testEpisode)

    // Create a temp file with test data (must be at least 10KB to pass validation)
    let testData = Data(repeating: 0, count: 10 * 1024)
    fileManager.createFile(atPath: tempFilePath, contents: testData)

    // Verify temp file exists
    XCTAssertTrue(fileManager.fileExists(atPath: tempFilePath), "Temp file should exist before processing")

    // When: Processing the episode with copyFile: false (move operation)
    let tempFileURL = URL(fileURLWithPath: tempFilePath)
    downloadManager.processEpisode(testEpisode, downloadedFile: tempFileURL, reportedContentType: "audio/mpeg", copyFile: false)

    // Then: Temp file should be removed (moved to destination)
    XCTAssertFalse(fileManager.fileExists(atPath: tempFilePath), "Temp file should be removed after move")
    XCTAssertTrue(fileManager.fileExists(atPath: destinationPath), "Destination file should exist")

    // Cleanup
    try? fileManager.removeItem(atPath: destinationPath)
    dataManager.delete(episodeUuid: testEpisode.uuid)
}
```

podcasts/EpisodeManager.swift
```diff
@@ -378,6 +378,42 @@ class EpisodeManager: NSObject {
...
+            let filePaths = Set(episodes.map { DownloadManager.shared.tempPathForEpisode($0) })
+            // Also protect actively downloading episodes
+            let downloadingEpisodes = DataManager.sharedManager.findEpisodesWhere(
+                customWhere: "episodeStatus == \(DownloadStatus.downloading.rawValue) OR episodeStatus == \(DownloadStatus.queued.rawValue)",
+                arguments: nil
+            )
+            let downloadingEpisodePaths = Set(downloadingEpisodes.map { DownloadManager.shared.tempPathForEpisode($0) })
+            // Also protect episodes that are being streamed and cached in parallel
+            let streamingEpisodes = DownloadManager.shared.downloadAndStreamEpisodes.values
+            let streamingEpisodePaths = Set(streamingEpisodes.map { DownloadManager.shared.tempPathForEpisode($0) })
```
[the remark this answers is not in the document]
SergioEstevao: Solved here: [be1fae4](https://github.com/Automattic/pocket-casts-ios/pull/4052/commits/be1fae4eab38022ddbabb652507209426737f454)

PR #1355, opened 2024-01-16
Player transition: fix UI freeze when Open Player Automatically is enabled

leandroalonso:
The latest #1350 attempt to fix this issue was not fruitful, so I'm back! But this time I reproduced the issue and know where it is. 😁

This PR fixes:

1. The player being displayed on the wrong `window`, causing the whole UI to freeze
2. The tap on miniplayer to open the player to fail due to player state being set to `open` while is not

The issue was happening because we were using `mapFirst` which sometimes returned a `UIInputWindowController` instead of the app main window.

You can see the freeze happening in this video:

https://github.com/Automattic/pocket-casts-ios/assets/7040243/f6782ab1-d32c-46f0-b8cb-16c19bd6d7fe

## To test

1. Run the app
2. Go to Settings > General > enable Open Player Automatically
3. Add two items on your Up Next
4. Swipe up to display the full player and then swipe down to dismiss it
5. Tap the Up Next icon on the miniplayer
6. Tap the episode that is not playing, then tap play
7. ✅ Nothing should happen and you should still be able to interact with the UI
8. Do this a few more times (swipe up to show full player, swipe down to dismiss, up next icon > play the episode that is not playing) — if you do that on `release/7.55` eventually the app will freeze)
9. Tap the Up Next icon
10. Play the other episode
11. This time, tap the miniplayer (instead of swiping)
12. ✅ The full player should display

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1350, opened 2024-01-15
Player transition: another attempt on fixing freezing issues

leandroalonso:
Back in December we had an issue that was blocking the UI when dismissing the player: #1288

However, we're still receiving a few reports of the issue happening:

1. https://www.reddit.com/r/pocketcasts/comments/196l8l8/ui_freezing_after_a_couple_of_minutes/
2. I'm also in contact with a user that reported this through TestFlight feedback

Last time, completing the transition without returning `true` was causing the `UITransitionView` to never goes away. I believe this might be happening when the `toViewController.view` is `nil`, so I changed it.

Unfortunately, I haven't been able to reproduce the issue.

## To test

1. To simulate the issue, go to `MiniPlayerToFullPlayerAnimator.swift` and before `guard let playerView = toViewController.view else {` add:

```swift
        if !isPresenting {
            transitionContext.completeTransition(false)
            return
        }
```

2. Then, open the full player
3. Dismiss it
4. 💥 Check that you can't interact with the UI anymore
5. Stop the app
6. Change the completeTransition to return true in the code you added: `transitionContext.completeTransition(true)`
7. Run the app
8. Open the full player
9. Dismiss it
10. ✅ You should be able to interact with the app

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1310, opened 2023-12-18
Player: add a spring animation when dismissing

leandroalonso:
This PR adds a spring animation when dismissing the player.

Besides the spring, there are additional changes:

1. All animations are now grouped into 3 animation blocks, instead of many different ones;
2. We do a setup phase (in which we position the views, take snapshot, etc) and only then animate (this was mixed before)
3. Given we were presenting the full player view itself, a lot of constraint warnings were shown. Now we snapshot it and use the snapshot instead
4. These changes makes the transition way smoother
5. The mass of the spring is reduced if the spring's initial velocity is too high. This is done to prevent the mini player from jumping too high.

### Normal speed
https://github.com/Automattic/pocket-casts-ios/assets/7040243/4fef0624-a302-4a2b-9c78-6546b98ff90e

### Slow speed

https://github.com/Automattic/pocket-casts-ios/assets/7040243/db561686-c23d-4753-a3e3-57806f1b1bbe

## To test

1. Open the player
2. Dismiss the player
3. ✅ The dismiss should be followed by a spring animation

Please test on multiple devices, light/dark mode.

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #1090, opened 2023-09-08
Player: custom transition

leandroalonso:
Address the remaining of https://github.com/Automattic/pocket-casts-ios/discussions/1037

This PR builds upon the last one and adds a custom transition from miniplayer to the full player. You can see the demo below in the normal speed and low speed.

https://github.com/Automattic/pocket-casts-ios/assets/7040243/dffbb573-9442-40f2-9ce8-cfcc0b3a99a2

I tried to tweak the animations to feel as good and smooth as possible. The player dismissal logic was also changed, in a way that if the user swipes down aggressively, the player will be dismissed fast.

## To test

I highly recommend testing this on real devices as it will help you get the real feeling of the new transition.

1. Run the app
2. Play any podcast
3. Tap the miniplayer
4. ✅ The full player should appear
5. Swipe down
6. ✅ The full player should dismiss
7. Reopen the player
8. Drag down to 15 or 20% of the screen and let it go
9. ✅ Check that the player dismiss and the artwork moves to the artwork in the miniplayer
10. Reopen the full player
11. Swipe down as quickly as you can
12. ✅ The player should dismiss as quickly as you swiped
13. Play a video podcast
14. Dismiss the player
15. ✅ There are no artwork animations and the full player is dismissed
16. Tap the miniplayer
17. ✅ The full player should open without the artwork animation
18. Open the full player and swipe tabs then dismiss it
19. ✅ The animation should perform

Please, test on small and big devices. :)

## Breakdown

It seems a simple animation but there's a lot going on. :) In terms of animation what is happening is:

1. The artwork moves from the small artwork to the big artwork, if the current episode **is not** a videocast
2. The miniplayer "grows" to the size of the full player
3. The full player appears with a fade in animation
4. When presenting the full player, we're using easy in-out. This means that the animation starts slowly, gets faster, and then slows down again.
5. However, when dismissing we use easy out, the animation starts fast and then slows down. I played a lot with that and this was the most natural result.

## Possible improvements

There are a few improvements/ideas we can also make so this is even better:

1. We can also hide the tab bar as the miniplayer is "growing"
2. We could do a bounce animation when dismissing
3. Sometimes it seems that the swipe left/right gestures get in the way of the swipe down gesture. This isn't a behavior added in this PR but it's something we can improve so the gesture is even more natural

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
[...]

PR #4054, opened 2026-03-08
Fix mini player to full player transition animation

marlonjames71:
## Summary

A while ago, the animation/transition of the mini player to the full player was changed with PR: #1090 and other changes were made to it with PRs: #1310, #1350, and #1355. There was another one, but I'd have to track that down again. 

Somewhere in between the time that new transition was initially introduced up until now the mini player to full player transition has acquired two main visual glitches and one minor visual occurrence that made the app feel not as polished as it should feel:

1. **Artwork flicker on dismiss** – dragging the player down slowly causes the artwork to flash before animating
2. **Swipe-up delay** – swiping up on the mini player doesn't begin the transition until the finger is lifted
3. **Semi-transparent background on present** – the player background briefly appears translucent during expand (minor, but noticeable and unnecessary in my opinion)

(see videos below)

This PR addresses all three by reworking the snapshot strategy and gesture handling in the custom view controller transition:

- Replace `snapshotView(afterScreenUpdates: true)` with `drawHierarchy(in:afterScreenUpdates:)` for the present snapshot to avoid flicker-inducing render-server commits
- Skip the player snapshot entirely on dismiss — the artwork overlay is the visual anchor, and removing this render eliminates setup delay
- Use `afterScreenUpdates: false` for mini player and tab bar snapshots on dismiss for instant capture
- Trigger `openFullScreenPlayer()` on gesture `.began` instead of `.ended` so the transition starts immediately on swipe-up making the experience feel much more responsive. This will also feel more familiar as this is the same way Apple Podcasts and Apple Music works.
- Use the opaque `mainView.backgroundColor` instead of the clear `view.backgroundColor` for the background transition
- Tune spring parameters (stiffness 500, damping 35) for a snappier, more natural dismiss feel

This PR keeps the spirit of the same newer transition that was introduced, but tweaks it to be more performant, snappier, and without the visual glitches.

## Before / After

https://github.com/user-attachments/assets/32e34b00-8849-41ff-9dec-c07dc3ee70d4

https://github.com/user-attachments/assets/86ed165e-9b73-4cca-b2ee-78524ec129c7

https://github.com/user-attachments/assets/445f1562-2091-4013-9942-335b093a5a0c

https://github.com/user-attachments/assets/034a1e58-b572-4890-b186-ef39ecee24ec

https://github.com/user-attachments/assets/75224912-9fb3-47ea-bb80-9a6097d740b5

## To test

1. Play any podcast so the mini player is visible
2. Tap the mini player to expand — verify no background flash or skeleton flicker
3. Swipe down slowly on the full player, then release — verify no artwork flicker
5. Flick the full player down quickly — verify smooth spring animation
6. Collapse the player, then swipe up on the mini player — verify the transition begins immediately without waiting for finger lift
7. Repeat steps 2–5 in both dark mode and light mode

## Checklist

- [X] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.

### DIFF
--- a/podcasts/MiniPlayerToFullPlayerAnimator.swift
+++ b/podcasts/MiniPlayerToFullPlayerAnimator.swift
@@ -290,40 +290,28 @@ class MiniPlayerToFullPlayerAnimator: NSObject, UIViewControllerAnimatedTransiti
             transitionContext.completeTransition(true)
         }
 
-        // For presenting, use a separate ease animation for properties
-        // that don't need spring physics
-        if isPresenting {
-            UIView.animate(withDuration: duration, delay: 0, options: .curveEaseInOut) {
-                backgroundTransitionView.backgroundColor = toColor
-                backgroundTransitionView.layer.cornerRadius = 0
-                toView?.layer.opacity = 1
-                tabBarSnapshot?.frame = hiddenTabBarFrame
-            } completion: { _ in
-                tabBar?.isHidden = false
-            }
-        }
     }
 
-    /// When presenting use curveEaseInOut. If dismissing, use spring animation
+    /// Use spring animation for both present and dismiss.
+    /// Dismiss carries gesture momentum via initialVelocity; present starts from rest.
     private func animate(withDuration duration: TimeInterval, animations: @escaping () -> Void, completion: ((Bool) -> Void)? = nil) {
-        if isPresenting {
-            UIView.animate(withDuration: duration, delay: 0, options: .curveEaseInOut, animations: animations, completion: completion)
-        } else {
-            // stiffness 500 → snappy response; damping 35 → ζ ≈ 0.78 (subtle bounce).
-            // initialVelocity carries the drag momentum directly into the spring.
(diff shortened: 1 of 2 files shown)


leandroalonso (changes requested):
@marlonjames71 thank you so much for these tweaks! We have a minor issue (which is not present on the previous version): when dismissing we have a duplicated artwork, one in the miniplayer and the one performing the animation:

https://github.com/user-attachments/assets/d9feaa64-fb88-43e7-bf53-cb0a713894a7

This is how it was:

https://github.com/user-attachments/assets/c7e6e97e-4856-4832-b574-04891821efc3

Once we fix this we should be good to go!

marlonjames71:
@leandroalonso great catch. I will fix that. Gonna try and get to that tonight. Thanks!

marlonjames71:
@leandroalonso Pushed up a change that fixes the double artwork on the mini player. Give it a try and see if that fixes it on your end.

leandroalonso (approved):
Thank you so much @marlonjames71! Looks superb!

marlonjames71:
Awesome! Thanks!

PR #3864, opened 2026-01-11
Airplay speaker fix

bjtitus:
Fixes [PCIOS-164](https://linear.app/a8c/issue/PCIOS-164/ios-player-no-longer-provides-the-ability-to-airplay-to-multiple-rooms)

This PR contains two changes:
* Fixes the multi-speaker selection by preparing the audio session when the route picker is presented under the `activateAudioSessionForRoutePicker` flag.
* Prevents automatic playback when audio routes changes if an episode was not already playing under the `dontAutoplayOnRouteChange` flag.

I feature flagged both of these because the API around AirPlay route changes is a bit opaque and we could run into issues.

### Test with nothing playing
* Open the Now Playing screen
* Tap the Airplay Route Selector button
* Tap on an Airplay speaker
* ✅ Verify you see the multi-select option visible
* ✅ Verify nothing begins playing

### Test with other media playing
* Open the Now Playing screen
* Tap the Airplay Route Selector button
* Tap on an Airplay speaker
* ✅ Verify you see the multi-select option visible
* ✅ Verify your playing media switches to speakers

### Test with playing podcast
* Open the Now Playing screen
* Tap the Airplay Route Selector button
* Tap on an Airplay speaker
* ✅ Verify you see the multi-select option visible
* ✅ Verify your playing episode switches to speakers

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #4062, opened 2026-03-11
Fix AirPlay route change bug

bjtitus:
Fixes [PCIOS-553](https://linear.app/a8c/issue/PCIOS-553/audio-stops-after-1-second-when-switching-playback-to-homepod-via-in)

I [added](https://github.com/Automattic/pocket-casts-ios/pull/3864) the `dontAutoplayOnRouteChange` Feature Flag with a change to the autoplay behavior when switching routes to avoid issues I noticed where triggering Airplay would autoplay without the episode currently playing. However, that change disabled all autoplay which a player switch was required (Trim Silence). This now checks whether the player should play, which was the intended behavior.

I've disabled the `dontAutoplayOnRouteChange` Feature Flag for now but it can be re-enabled with 8.8 after this change.

## To test

* Play an episode with Trim Silence enabled
* Airplay to a speaker
* Ensure playback continues without needing to be restarted in the app 

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -1968,7 +1968,18 @@ class PlaybackManager: ServerPlaybackDelegate {
 
         let reason = changeReason.uintValue
         if let currEpisode = currentEpisode(), playingOverAirplay() && playerSwitchRequired() {
-            let autoPlay = FeatureFlag.dontAutoplayOnRouteChange.enabled ? false : true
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
:shipit:

SergioEstevao:
@bjtitus will it not better to have this on 8.7?

bjtitus:
@SergioEstevao the only reason I didn't is that it's behind the feature flag so it's not an issue no that the flag has been disabled.

PR #4102, opened 2026-03-31
Fix playback resuming unexpectedly after phone calls

### LINKED ISSUE #1248: Playback: Audio resuming after phone calls
### Description

> I’ve encountered a new bug with iOS 17. The app resumes even manually paused audio playback when a phone call is ended. This includes rejecting incoming calls, ending a taken call, or Live voicemail. Unless the pocket cast app is manually swiped closed the previous audio will resume after the interaction with the phone app. Latest version 7.52.1 installed and confirm this doesn’t affect other audio apps, just pocket casts.

7380495-zd-a8c

Reporting it here based on the feedback here: https://github.com/Automattic/pocket-casts-ios/issues/45#issuecomment-1833797248

1. Pause playback of Pocket Casts
2. Answer a phone call
3. End the phone call, and notice the playback resumes

_No response_

App Version: 7.52.1
Device: iPhone14,5
OS: 17.1.1

leandroalonso:
Fixes #1248

When a phone call comes in on devices paired with Bluetooth accessories that have mic/speaker (e.g. Garmin Venu 2 Plus), iOS triggers audio route changes that send spurious `playCommand` events to the app. This caused paused playback to resume unexpectedly when the call ended.

Confirmed via user debug logs showing the sequence:
- `handleAudioInterrupt began` (phone call)
- Route change to Bluetooth device
- `Remote control: playCommand, treating as playPause` → starts playback

Two changes to the `playCommand` handler in `PlaybackManager`:

1. **Ignore during interruptions**: Early return when `interruptInProgress` is true, preventing playback during active phone calls.
2. **Play-only, not toggle**: Changed the default branch from `playPause()` to `play()`. A `playCommand` should only start playback — `togglePlayPauseCommand` and `pauseCommand` handle the other directions.

## To test

This has very specific conditions to reproduce. Our best bet is to release this through TestFlight and ping affected users so they can help us testing.

1. Pair a Bluetooth device with mic/speaker (ideally Garmin Venu 2 Plus)
2. Open Pocket Casts with an episode loaded but **paused**
3. Receive a phone call → silence it or let it ring to voicemail
4. Verify playback does **not** resume when the call ends
5. Verify normal play/pause from lock screen, headphones, and Control Center still works

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -1692,6 +1692,14 @@ class PlaybackManager: ServerPlaybackDelegate {
         commandCenter.playCommand.addTarget { [weak self] _ -> MPRemoteCommandHandlerStatus in
             guard let strongSelf = self, let _ = strongSelf.currentEpisode() else { return .noActionableNowPlayingItem }
 
+            // Don't start playback during an active audio interruption (e.g. phone call).
+            // Bluetooth devices like Garmin watches with mic/speaker trigger route changes
+            // during calls, which cause iOS to send spurious playCommand events.
+            if strongSelf.interruptInProgress {
+                FileLog.shared.addMessage("Remote control: playCommand, ignored because an audio interruption is in progress")
+                return .success
+            }
+
             strongSelf.analyticsPlaybackHelper.currentSource = strongSelf.commandCenterSource
 
             if Settings.legacyBluetoothModeEnabled() {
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
Tested this using my Sony 1000XM4 and all worked correctly:
- Phone call started and ended with audio paused, no play started at end of call
- Phone call started with audio playing, playback resume after the call
- Headphone commands worked correctly: double tap play and pause, skip forward and backward.

Do you want to put this behind a FF? and test on Testflight beta next week?

leandroalonso:
@SergioEstevao we have a user on the Beta Slack that can consistently reproduce this issue. If it fixes it for them I think we should fix, otherwise we remove it and iterate. Wdyt?

SergioEstevao:
> @SergioEstevao we have a user on the Beta Slack that can consistently reproduce this issue. If it fixes it for them I think we should fix, otherwise we remove it and iterate. Wdyt?

Ok then, we can merge this and then next week ask the user to test in Beta.

PR #4112, opened 2026-04-06
Disable watch crash explicit logging

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

This PR disables explicit crash logging on watch sync to avoid flood of warning logs on Sentry.
This is done with a FeatureFlag so we can be sure we are not hiding other real logs crashes coming from the watch or sync errors.

## To test

1. Start the watch app on the simulator or device
2. Smoke test the behaviour of app 

## Checklist

### DIFF
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -307,7 +307,7 @@ public enum FeatureFlag: String, CaseIterable {
     /// Track network data usage per episode/connection type in the NetworkDataUsage table
     case trackNetworkDataUsage
 
-    /// If enabled we send critical watch crash logs to Sentry
+    /// If enabled, send watch-related error and crash logs to Sentry
     case watchSentryCrashLog
 
     public var enabled: Bool {
--- a/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Utils/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -307,6 +307,9 @@ public enum FeatureFlag: String, CaseIterable {
     /// Track network data usage per episode/connection type in the NetworkDataUsage table
     case trackNetworkDataUsage
 
+    /// If enabled we send critical watch crash logs to Sentry
+    case watchSentryCrashLog
+
     public var enabled: Bool {
         if let overriddenValue = FeatureFlagOverrideStore().overriddenValue(for: self) {
             return overriddenValue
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/124">8.9 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

twstokes (commented):
👋 The code change makes sense to me @SergioEstevao – I'll leave it to @leandroalonso for a final review.

It seems now you'll either:
- Send a ton of logs for this error to Sentry if the feature flag is on (which will now hopefully get rate limited on the Sentry side)
- Not send anything for this error if it's off

So it makes me wonder: 
- What's the value of getting this error in the first place? 
- Should an error only be sent if the fallback / limited Up Next sync fails?

I also wanted to share a couple other options:
- This error could be caught and dropped by Sentry's [`beforeSend` event in their SDK](https://docs.sentry.io/platforms/apple/configuration/filtering/#filtering-error-events)
- These errors could be [individually filtered on the Sentry.io side](https://docs.sentry.io/concepts/data-management/filtering/)

Client-side is always preferred, though. Another idea is you could catch the error in the `beforeSend` event and apply your own rate limit just to that event – e.g. if it's the `payloadTooLarge` one:

```
return Int.random(in: 1...10) == 1 ? event : nil
```

to rate limit on the client side to 10%.

All that said, what you have seems effective. 👍

SergioEstevao:
So it makes me wonder:

> What's the value of getting this error in the first place?
Brandon implemented adding Sentry to the Apple Watch app, I think the rationale behind this logging was to try to get an idea how much this kind of error was happening and create a first fix for it. Now that we have the perception of the volume we can try to address it further.

> Should an error only be sent if the fallback / limited Up Next sync fails?

This could be an option, to see if the fallback has limited the number of errors

leandroalonso:
@SergioEstevao I agree with @twstokes comment about:

> What's the value of getting this error in the first place?

I feel like we're leaving this code there just for the sake of leaving it. If we don't have a very good reason to follow-up on those warnings I'd rather remove these events entirely.

If there's any reasoning behind keeping that it would be good to state in the PR description.

SergioEstevao:
@twstokes and @leandroalonso I will update the PR to remove the explicit logging completely and don't use the FF.
In the future if we need to address the underlying warnings we will add the logging away to monitor the changes made.

leandroalonso (approved):

PR #4119, opened 2026-04-09
Ensure stall observer is removed

SergioEstevao:
| 📘 Part of: # |  
|:---:|

This resulted from the investigation on the resume playback when stalls happened. Claude was reporting that the observer was being removed because we were not storing it. But in reality it’s the other way around it’s being kept around because the notification observer was retaining it and we were never removing it.

This PR ensures that stall removal is correctly removed on cleanup and adds some logging to stall removal actions

The Stall observer was not being removed on clean so it stayed dormant on the notification. The guard for weak self was protecting of crashes and retain cycle but better do the clean up operation correctly by storing the observer and removing on clean up
 
## To test

1. Start the app
2. Play an episode
3. Pause the episode
4. Play another episode ( if you want you can add a breakpoint on DefaultPlayer->cleanUpPlayer-> line 887 to see if removal happens
5. If you want to check the logging, activate the Network Link Conditioner and use the `Very Bad Network` setting to cause stall on playback of an episode
6. Check if logs show up

## Checklist

### DIFF
--- a/podcasts/DefaultPlayer.swift
+++ b/podcasts/DefaultPlayer.swift
@@ -860,7 +860,7 @@ class DefaultPlayer: PlaybackProtocol, Hashable {
             PlaybackManager.shared.playbackDidFail(error: .playbackError(logMessage: errorMessage, isLocalFile: isPlayingLocalFile))
         }
 
-        _ = nc.addObserver(forName: NSNotification.Name.AVPlayerItemPlaybackStalled, object: nil, queue: nil) { [weak self] _ in
+        playStalledObserver = nc.addObserver(forName: NSNotification.Name.AVPlayerItemPlaybackStalled, object: nil, queue: nil) { [weak self] _ in
             guard let self = self else { return }
 
             if self.shouldKeepPlaying {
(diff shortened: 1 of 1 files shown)


podcasts/DefaultPlayer.swift
```diff
@@ -860,10 +860,11 @@ class DefaultPlayer: PlaybackProtocol, Hashable {
             PlaybackManager.shared.playbackDidFail(error: .playbackError(logMessage: errorMessage, isLocalFile: isPlayingLocalFile))
         }
 
-        _ = nc.addObserver(forName: NSNotification.Name.AVPlayerItemPlaybackStalled, object: nil, queue: nil) { [weak self] _ in
+        playStalledObserver = nc.addObserver(forName: NSNotification.Name.AVPlayerItemPlaybackStalled, object: nil, queue: nil) { [weak self] _ in
             guard let self = self else { return }
-
+            FileLog.shared.addMessage("Received notification of playback stall")
```
leandroalonso, line 865: This might fire a lot depending on network conditions. Which might pollute the whole log.

But maybe that's what we want anyways?

leandroalonso (approved):

podcasts/DefaultPlayer.swift
SergioEstevao: I think this will be helpful to try to find if the stall is happening a lot to certain user and may be the cause of playback issues? Should we FF the log or check the results during beta and adjust?

PR #4121, opened 2026-04-09
Update proto buffer script output path

SergioEstevao:
| 📘 Part of: # |  
|:---:|

The latest changes made on project structure because of modularization changed the project modules folders.
This PR updates path to the Server Module code so that the script can generate at the correct place

## To test

1. Try to run the protobuf scripts running: `make update_proto API_PATH={API_PATH}`
2. Replace the  {API_PATH} with the full path to a local copy of the `pocket-casts-sync-api` project and then  `/api/modules/protobuf/src/main/proto` folder. EX: `make update_proto API_PATH=../pocket-casts-sync-api/api/modules/protobuf/src/main/proto`
3. Check if all runs correctly

## Checklist

### DIFF
--- a/scripts/update_proto.sh
+++ b/scripts/update_proto.sh
@@ -20,5 +20,5 @@ else
     echo "Brew is not installed. Make sure protoc + protoc-gen-swift is installed."
 fi
 
-protoc --swift_out=./Modules/Server/Sources/PocketCastsServer/Private/Protobuffer --proto_path=$API_BASE_FOLDER/ $API_BASE_FOLDER/api.proto
-protoc --swift_out=./Modules/Server/Sources/PocketCastsServer/Private/Protobuffer --proto_path=$API_BASE_FOLDER/ $API_BASE_FOLDER/files.proto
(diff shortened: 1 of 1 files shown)


mokagio (approved):
Tested this on my end. Reproduced the failure on `trunk`, verified the fix here.

@SergioEstevao does the Pocket Casts Server interface change often? Would we benefit from running this in CI to ensure it's up to date? Or is this something low frequency where we can deal with the occasional roundtrip?

SergioEstevao:
> Tested this on my end. Reproduced the failure on `trunk`, verified the fix here.
> 
> @SergioEstevao does the Pocket Casts Server interface change often? Would we benefit from running this in CI to ensure it's up to date? Or is this something low frequency where we can deal with the occasional roundtrip?

The change on protobuffer API is not very common, and normally it's associated with some new feature and change. I don't think we should run the script in CI every build.

mokagio:
> The change on protobuffer API is not very common, and normally it's associated with some new feature and change. I don't think we should run the script in CI every build.

Sounds good.

PR #4124, opened 2026-04-13
Install protobuf packages if missing in update_proto

mokagio:
When I tried to test #4121 , I got an error about `swift-protobuf` not being installed. The README (and AGENTS) document that the tool needs to be installed via `brew`, but I didn't read any because this is a repo I've had on my machine for a while, so I missed that information. 

This PR updates the protobuf generator script to install the required tools via `brew` if they are not available.

I think it's a neat improvement, but not a mandatory one. After all, installing the tool is a one-off on a new dev machine, I could live with it staying the way it is if you think the added complexity in the script is not worth the trade.


SergioEstevao (approved):
Good addition to automate the setup.

PR #4125, opened 2026-04-13
Ensure that we only send data back when the response object is available

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS-558 

This PR updates the streaming media loader to ensure that the `contentInformationRequest` when requested is properly filled with existing response data. 
The previous code could run into data race scenario that a `response` was not ready when `contentInformationRequest` was being filled,  but then filling media data that was already written to memory or disk and incorrectly fulfil the request as mark it as `finishLoading` without filling the correct data and making the AVPlayer to fail with bad media data.

## To test

1. Start the app using a real device
2. Select some episodes and start downloading them
3. Tap on play on one of those episodes that is downloading.
4. Check if playback happens correctly

## Checklist


SergioEstevao:
@claude can you review this?

leandroalonso (approved):

PR #4174, opened 2026-04-22
Fix animations in Playlist Details

kean:
I'm working on fixing the animations in the swipe actions. In order to do that and make it sensible to the user, you first need to fix animations in the list. This is what this PR is about. The foundation with `StagedChangeset` is already there but is used a bit incorrectly.

**What changed**
  - `numberOfSections(in:)` now reads from `viewModel.dataSource.count` instead of a hardcoded `numberOfSection` (removed).
  - `PlaylistDetailViewModel` seeds `dataSource` at init with the real initial shape (header + archive-if-manual + empty episodes). This prevents incorrect diff being created on the first reload.
  - Extracted `makeSections(episodes:)` so the init seed and `buildChangeSet` build sections through one code path.
  - First reload after init is forced non-animated; subsequent reloads animate as before.

**Why**

The table view's section count (2 or 3) disagreed with `dataSource ([])`, so the first `StagedChangeset` tried to insert sections UIKit already reported as present. That threw `NSInternalInconsistencyException`, which `SJCommonUtils.catchException` silently swallowed and fell back to `reloadData()` - masking the bug and losing the fade animation. Seeding the data source removes the mismatch; unifying section construction keeps the initial state and all future diffs in sync; skipping animation on the first reload avoids the header/archive rows visibly popping in on open.

**Future Changes**

The animations on manual reload currently still appear a bit off. I plan to slightly refactor `reloadPlaylistAndEpisodes` and `reloadEpisodeList` to add a bit of throttling. The similar mechanism will also be used in the future to temporary disable reloads when the swipe menu is shown and re-enable later.

## To test

- Open a Smart Playlsit
- Make a change on a different device
- Refresh on the first device
- Verify that animations are performed

## Recording

Before (no animations)

https://github.com/user-attachments/assets/3f542731-ace9-40f0-bb90-f924b52e5726

After (automatic animations; will be improved with debounce and delay in the future PRs)

https://github.com/user-attachments/assets/689fc10f-c18b-434f-abe8-16bc18e64dde

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #4187, opened 2026-04-24
Fix Playlist reload animations during refresh and actions (and more)

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS-606.

**Changes:**

- Fix an issue with `refreshControl` triggering invoking throttled refresh instead of a forced one
- Improve animations on reload that happens after pull-to-refresh by delaying the refresh until animations settle (previous recording was here https://github.com/Automattic/pocket-casts-ios/pull/4174)
- Fix an issue with reload happening while swipe actions are open (original issue)
- Improve performance by throttling reloads triggered by the DB notifications (see `ReloadScheduler`). The same mechanism is used for pausing the reloads during user actions.

From the user experience, it matches how Apple does it in apps like Apple Mail.

## To test

- **Verify** that pull-to-refresh aways triggers force reload (like in other screens)
- **Verify** that pull-to-refresh is dismissed after refresh finishes
- **Verify** animations work as described in changes – might require a bit of setup.

## Recording

**Refresh Animation**

The animations are now performed after the control is hidden and everything is settled, so it's less jarring:

https://github.com/user-attachments/assets/d4272e49-cad0-4b84-987c-745767bc50d8

**Reload During Swipe Action**

Reload is now delayed after you finish what you are doing

https://github.com/user-attachments/assets/10309c16-025b-482f-a2f4-639dd68ee794

## Checklist


Modules/Sources/PocketCastsUtils/General/ReloadScheduler.swift
```diff
@@ -0,0 +1,108 @@
...
+        resumeTask = Task { [weak self] in
+            try? await Task.sleep(for: duration)
+            guard !Task.isCancelled else { return }
+            self?.resume()
+        }
+    }
+
+    private func scheduleFlush() {
+        guard !isPaused, flushTask == nil else { return }
+        let delay: Duration = lastFlushAt.map {
```
kean, line 75: This is a little bit complicated by it achieves so-called instant leading edge, throttled trailing edge – best for UI updates.

**What's below is generated:**

Here's an example timeline you can paste into the PR description. Assumes interval = 100ms.

```
  t=0     request(A)       → delay=0 (no prior flush)
  t=~0    FLUSH            → onReload(A)         [leading edge]
                             lastFlushAt=0

  t=20    request(B)       → delay=80 (100−20)
  t=40    request(C)       → coalesced into pending (B∪C)
  t=100   FLUSH            → onReload(B∪C)       [trailing edge]
                             lastFlushAt=100

  t=250   request(D)       → delay=0 (quiet > interval)
  t=~250  FLUSH            → onReload(D)
```

  Key properties the timeline illustrates:

  - Leading edge: the first request after a quiet window ≥ interval fires immediately (t=0, t=250).
  - Coalescing: requests that arrive while a flush is already scheduled union into pending and don't reschedule (t=40 folds into the t=100 flush).
  - Trailing edge: a burst produces exactly one follow-up flush at lastFlushAt + interval (t=100), carrying the union of everything received during the window.
  - Cooldown: the minimum spacing between two onReload calls is interval (100ms between t=0 and t=100).

podcasts/New Detail/PlaylistDetailViewController.swift
```diff
@@ -404,12 +414,12 @@ class PlaylistDetailViewController: FakeNavViewController {
...
 
     private func reload(data: StagedChangeset<PlaylistDetailViewModel.DataSourceValue>, animated: Bool, contentChanged: Bool) {
-        removeLoadingIndicators()
+        loadingIndicator.stopAnimating()
 
         if animated, contentChanged {
             do {
                 try SJCommonUtils.catchException {
-                    tableView.reload(using: data, with: .automatic) { newData in
+                    tableView.reload(using: data, with: .fade) { newData in
```
kean, line 421: The original was `.fade`. I briefly tried `.automatic` (change was made [here](https://github.com/Automattic/pocket-casts-ios/pull/4174)), but it performed worse.

podcasts/New Detail/PlaylistDetailViewController.swift
```diff
@@ -404,12 +414,12 @@ class PlaylistDetailViewController: FakeNavViewController {
     }
 
     private func reload(data: StagedChangeset<PlaylistDetailViewModel.DataSourceValue>, animated: Bool, contentChanged: Bool) {
-        removeLoadingIndicators()
+        loadingIndicator.stopAnimating()
```
kean, line 416: This was previously removing both the refresh control and the loading indicator. We need to remove the refresh control when the refresh actually finished – not the table randomly reloads.

podcasts/New Detail/PlaylistDetailViewController.swift
```diff
@@ -391,9 +396,14 @@ class PlaylistDetailViewController: FakeNavViewController {
 
         refreshControl = CustomRefreshControl()
         refreshControl?.customTintColor = AppTheme.colorForStyle(.secondaryText02)
-        refreshControl?.perform = { refreshControl in
+        refreshControl?.perform = { [weak self] refreshControl in
             refreshControl.set(text: L10n.refreshControlFetchingEpisodes.uppercased())
-            RefreshManager.shared.refreshPodcasts()
+            self?.reloader.pause()
+            RefreshManager.shared.refreshPodcasts { [weak self] _ in
+                DispatchQueue.main.async {
```
kean, line 402: The completion is called on the background thread. Didn't want to mess with the original manager.

pocketcasts:
Version `8.11` has now entered code-freeze, so the milestone of this PR has been updated to `8.12`.

SergioEstevao (approved):
Tested this with multiple scenarios on playlists and the animations worked correctly.

PR #4191, opened 2026-04-24
POC-539: Streaming fingerprint support for synced transcripts

leandroalonso:
| 📘 Part of: POC-539 |
|:---:|

## Summary

Synced transcripts previously only fingerprinted **downloaded** episodes — the one-shot reader opened the local file, read to EOF, and stopped. For an episode the listener is streaming, the on-disk buffer is still growing as AVPlayer writes it, so fingerprinting never produced anchors and highlighting never turned on.

This PR adds a streaming fingerprint path that treats the buffer as a growing file:

- **Grow-loop** (`streamFingerprintGrowing`): reopens `AVAudioFile` each pass to pick up new bytes, seeks to where it left off, feeds PCM through the existing `StreamingWindowedFingerprinter` pipeline, and bails after `bufferGrowMaxStallSeconds` of no new bytes (a later playback-progress tick will restart it).
- **Routing** (`resolveAudioSource`): `.downloaded` for episodes with a complete local file (`pathForEpisode` or the completed stream-download at `streamingBufferPathForEpisode`) → unchanged one-shot path. `.streaming` for anything currently growing (`tempPathForEpisode` from stream-and-cache, or the legacy streaming buffer) → new grow-loop.
- **Start-position alignment** fix: the grow-loop waits until the on-disk file has reached `startSeconds * sampleRate` before initialising the streamer, then anchors at exactly that frame — so window timestamps line up with reference checkpoints. (Clamping `lastProcessedFrame` down to `audioFile.length` when the file was behind playback would have misattributed windows and produced all-red matches.)
- **`processProgress` guard**: if `playbackToReference` is empty, skip the drift-restart. The grow-loop gets progress ticks every second while still building its first window; without this, each tick would cancel the in-flight work before any anchor landed.
- **Tap-to-seek toast**: when the fingerprint mapping has no anchors yet and `playbackTime(forReferenceTime:)` returns nil (typical in the first few seconds of any new playback, especially streaming), show `"Tap to seek not available while streaming"` instead of falling back to identity `referenceTime`.

All gated by the existing `FeatureFlag.syncedTranscripts`. The tap handler only installs its gesture recogniser when the flag is on, and `prepareForEpisode` bails early when it's off, so the streaming grow-loop never runs with the flag disabled.

## To test

You can use Network Link Conditioner to simulate slow connectivity.

1. Enable `FeatureFlag.syncedTranscripts` (Beta menu or debug overrides).
2. Run the app in `StagingDebug` (I recommend running fresh to avoid picking up buffered files)
3. Login with a Plus account
5. Pick a podcast episode with a synced transcript that is **not** downloaded.
6. Start playback.
7. Within a few seconds of playback, the transcript highlight should start tracking. `FileLog` will show `"streaming grow-loop starting at …s"` and later `"streaming buffer opened …"`.
8. Tap a transcript cue *before* the mapping has produced any anchor (immediately after starting playback) → toast `"Tap to seek not available while streaming"` appears; no seek happens.
9. After a few more seconds, tap a cue → seek fires to the fingerprinted playback time.
10. Regression check: fully download the same episode, restart playback, confirm highlighting and tap-to-seek behave exactly as on `trunk`.
11. Flag off: confirm the transcript loads normally and the tap gesture does nothing

## Checklist

### DIFF
--- a/podcasts/Fingerprint/FingerprintTimingManager.swift
+++ b/podcasts/Fingerprint/FingerprintTimingManager.swift
@@ -241,28 +241,6 @@ final class FingerprintTimingManager: NSObject {
         }
     }
 
-    /// Strict-range variant of `playbackTime(forReferenceTime:)`. Returns `nil`
-    /// when `referenceTime` falls outside the already-mapped range instead of
-    /// extrapolating with an identity offset. Use this for tap-to-seek where a
-    /// wrong guess would land the listener in the wrong place.
-    func playbackTimeIfCovered(forReferenceTime referenceTime: Double) -> Double? {
-        dispatchPrecondition(condition: .notOnQueue(queue))
-        return queue.sync {
-            guard let first = referenceToPlayback.first,
-                  let last = referenceToPlayback.last,
-                  referenceTime >= first.referenceTime,
-                  referenceTime <= last.referenceTime else {
-                return nil
-            }
-            return Self.interpolate(
-                time: referenceTime,
-                in: referenceToPlayback,
-                keyPath: \.referenceTime,
-                valuePath: \.playbackTime
-            )
-        }
-    }
-
     #if DEBUG
     var totalDuration: Double? {
         dispatchPrecondition(condition: .notOnQueue(queue))
--- a/podcasts/Fingerprint/FingerprintTimingManager.swift
+++ b/podcasts/Fingerprint/FingerprintTimingManager.swift
@@ -29,6 +29,10 @@ final class FingerprintTimingManager: NSObject {
     private struct GenerationContext {
         let episodeUuid: String
         let audioFileURL: URL
+        /// True when `audioFileURL` points at a streaming buffer that may still be
+        /// growing. The fingerprint loop polls for new bytes instead of quitting
+        /// at EOF.
+        let isStreaming: Bool
         let duration: Double
         let matcher: CheckpointMatcher
         let isCancelled: () -> Bool
@@ -166,6 +170,15 @@ final class FingerprintTimingManager: NSObject {
         lastProgressPosition = playbackTime
 
         if isWithinMappedRange(playbackTime) { return }
+
+        // If we haven't produced any anchors yet, the current stream is still
+        // building the first window around the listener's initial position —
+        // restarting every progress tick (which happens each second while the
+        // map is empty) cancels that in-flight work before it can finish. The
+        // delta-based branch above still catches real seeks. Once we have
+        // coverage the drift-restart resumes as before.
+        if playbackToReference.isEmpty { return }
+
         #if DEBUG
         FileLog.shared.addMessage(
             "FingerprintTimingManager: playback at \(String(format: "%.1f", playbackTime))s outside mapped range — restarting"
@@ -194,6 +207,7 @@ final class FingerprintTimingManager: NSObject {
         let newContext = GenerationContext(
             episodeUuid: ctx.episodeUuid,
             audioFileURL: ctx.audioFileURL,
+            isStreaming: ctx.isStreaming,
             duration: ctx.duration,
             matcher: ctx.matcher,
             isCancelled: { flag.isCancelled }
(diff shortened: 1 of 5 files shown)


podcasts/en.lproj/Localizable.strings
```diff
@@ -4670,6 +4670,9 @@
 /* Transcript error message when transcript is empty*/
 "transcript_error_empty" = "Sorry, but it looks this transcript is empty";
 
+/* Toast shown when the user taps inside the transcript but the fingerprint mapping has no anchors yet, so we can't resolve an accurate seek target. */
+"transcript_tap_to_seek_streaming_unavailable" = "Tap to seek not available while streaming";
```
leandroalonso: Any ideas for a better message?

podcasts/TranscriptViewController.swift
```diff
@@ -930,11 +930,9 @@ class TranscriptViewController: PlayerItemViewController, AnalyticsSourceProvide
         }
         let referenceTime = cue.startTime + fraction * (cue.endTime - cue.startTime)
 
-        let seekTime: TimeInterval
-        if let playbackTime = FingerprintTimingManager.shared.playbackTime(forReferenceTime: referenceTime) {
-            seekTime = playbackTime
-        } else {
-            seekTime = referenceTime
+        guard let seekTime = FingerprintTimingManager.shared.playbackTime(forReferenceTime: referenceTime) else {
+            Toast.show(L10n.transcriptTapToSeekStreamingUnavailable)
```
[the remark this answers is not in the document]
leandroalonso: Not changing, as much cases are due to streaming.

podcasts/Fingerprint/FingerprintTimingManager.swift
```diff
@@ -500,6 +525,177 @@ final class FingerprintTimingManager: NSObject {
...
+    }
+
+    /// Sleep on the generation queue in small slices so cancellation is
+    /// observed within at most one slice.
+    private func sleepWithCancellation(seconds: TimeInterval, context ctx: GenerationContext) throws {
+        let sliceSeconds: TimeInterval = 0.2
+        let slices = max(1, Int(ceil(seconds / sliceSeconds)))
+        for _ in 0..<slices {
+            if ctx.isCancelled() { throw StreamError.cancelled }
+            Thread.sleep(forTimeInterval: sliceSeconds)
```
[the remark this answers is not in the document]
leandroalonso: Done in 8b657233b2d6d44404f34c955377548ecd792733

podcasts/Fingerprint/FingerprintTimingManager.swift
```diff
@@ -807,22 +1003,47 @@ final class FingerprintTimingManager: NSObject {
...
+        if FileManager.default.fileExists(atPath: tempPath) {
+            return .streaming(URL(fileURLWithPath: tempPath))
+        }
         if FileManager.default.fileExists(atPath: streamingPath) {
-            return URL(fileURLWithPath: streamingPath)
+            return .streaming(URL(fileURLWithPath: streamingPath))
         }
-        return nil
+        let preferred = FeatureFlag.streamAndCachePlayingEpisode.enabled ? tempPath : streamingPath
+        return .streaming(URL(fileURLWithPath: preferred))
```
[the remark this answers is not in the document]
leandroalonso: Valid point. Leaving this for a follow-up — the grow-loop's stall/exit conditions need test doubles for `FileManager`, `AVAudioFile`, and the sleep primitive, which is a larger refactor than I want to land in this PR.

leandroalonso:
## Review feedback addressed

- **FingerprintTimingManager.swift:690** (@Copilot): `sleepWithCancellation` could oversleep by up to one slice — switched to remaining-time tracking with `min(sliceSeconds, remaining)` per iteration. Done in 8b657233b2d6d44404f34c955377548ecd792733
- **FingerprintTimingManager.swift routing/grow-loop tests** (@Copilot): Deferred to a follow-up. The grow-loop needs test doubles for `FileManager`, `AVAudioFile`, and the sleep primitive before it's unit-testable — a larger refactor than belongs in this PR.
- **TranscriptViewController.swift:934** (@Copilot): Already declined — this toast is the agreed behaviour for this PR; most of the nil-mapping cases are the streaming grow-loop's first-window warmup, so the streaming-specific copy reads correctly.

podcasts/en.lproj/Localizable.strings
stefanosala: I think your message is good, but we should add something to inform the user about what to do in these cases. Wait? Manually download the episode?

stefanosala (approved):
Tested locally, it's working well. 👍

podcasts/en.lproj/Localizable.strings
leandroalonso: @stefanosala "Download the episode to tap to seek"?

stefanosala: I think that works 👍

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/127">8.11</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

PR #4204, opened 2026-04-27
Liquid Glass: Mini Player – first iteration

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS-479.

This PR updates the "Mini Player" with the new design – this was the first thing that was broken the most with `UIDesignRequiresCompatibility` off.

The new design spec also included a couple of minor changes to the layout. The controls are now strictly on the right, so it's now much easier to tap the bar to open the full player – it requires much less precision. The "next up" was removed as there is a dedicated tab bar item for that, and it's available in a full player. The progress view is also replaced by "time left" label.

**Note**: it's the first iteration and it's under the FF. There a couple of open discussion in the channel around it.

Screenshot 2026-04-27 at 4 42 46 PM

## Test Instructions

**Setup**: Enable “Liquid Glass” FF and set `UIDesignRequiresCompatibility` to `NO` in podcasts-Info.plist

- **Verify** that legacy “Mini Player” works with no changes
- **Verify** that buttons have a nice large tappable area
- **Verify** that hide/show animations work well (had some hiccups there)
- **Verify** that is does NOT resize for the dynamic size (just like the tab bar)

## Checklist

### DIFF
--- a/podcasts/MiniPlayerViewController.swift
+++ b/podcasts/MiniPlayerViewController.swift
@@ -37,6 +37,11 @@ class MiniPlayerViewController: SimpleNotificationsViewController {
 
     private let analyticsPlaybackHelper = AnalyticsPlaybackHelper.shared
 
+    private var glassContainer: UIVisualEffectView?
+    private var episodeTitleLabel: UILabel?
+    private var episodeTimeLeftLabel: UILabel?
+
+
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -44,7 +49,11 @@ class MiniPlayerViewController: SimpleNotificationsViewController {
 
         view.isHidden = false
 
-        setupCorners()
+        if FeatureFlag.liquidGlass.enabled, #available(iOS 26.0, *) {
+            setupLiquidGlassLayout()
+        } else {
+            setupCorners()
+        }
         addUINotificationObservers()
         playbackStateDidChange()
         themeChanged()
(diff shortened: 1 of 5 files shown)


podcasts/MiniPlayerViewController.swift
```diff
@@ -55,6 +64,107 @@ class MiniPlayerViewController: SimpleNotificationsViewController {
         mainView.layer.masksToBounds = true
     }
 
+    @available(iOS 26.0, *)
+    private func setupLiquidGlassLayout() {
```
kean, line 69: I tried a couple of different options how to implement it in terms of the architecture, but this one turned out to be the easier to understand and maintain.

SergioEstevao (approved):
Looking well with the FF on and off. 

I don't think it's part of the scope of this PR work but I noticed the following issues when testing:

- The error banner overlaping the tab bar. You may need to adjust the color being used in it.
Image

- When using a dark theme there is some strange color changes in the tab bar:

https://github.com/user-attachments/assets/4fec539b-92bc-4c2c-abe2-dba9410d435d

kean:
Thank you! I opened two new Linear issues to track these. The latter is probably because of the extra gradient.

rviljoen:
@kean Looks great! Really hard to go back to the old version once you've tried this one 😃 

Couple thoughts:
 - Not related to this PR, but I wonder why the play button in mini-player is animated, but skip forward/back is not? These are all animated in full-screen player. May be good to explore as part of this change.
 - Have you explored [UITabAccesory](https://developer.apple.com/documentation/uikit/uitabaccessory) for the mini-player (similar to what Music uses for their mini-player: https://developer.apple.com/videos/play/wwdc2025/284/?time=165)? It gives some free functionality, like minimising into the tab bar when scrolling..

kean:
Hey, @rviljoen.

> back to the old version once you've tried this one 

Yeah, I've also been testing it on the go yesterday, and it felt off the first five minutes but now I don't want to go back.

> but skip forward/back is not?

Great suggestion! I added it to Linear, and will do it in the scope of this project. 

> Have you explored [UITabAccesory](https://developer.apple.com/documentation/uikit/uitabaccessory) for the mini-player (Have you explored [UITabAccesory](https://developer.apple.com/documentation/uikit/uitabaccessory) for the mini-player (

I used it in a prototype app, but I'm not exactly sure we want to collapse the tab bar like other apps do. The collapsed versions usually have downsides – can't squeeze just as many controls into to them, so you'd have to compromise.

We're also discussing adding a progress view back – thoughts?

image

rviljoen:
> We're also discussing adding a progress view back – thoughts?

Hey @kean  - Progress bar looks much better there than the old mini-player design (bottom of player). Looking at the screenshot above, one nit would be to see if you can better align the time left and the title (right-align time left), but with short titles it will be tricky.  If you want to keep just one, then time left label is more useful to me than the progress bar.

Another thing to consider for the mini-player is to show the chapter title, and fall back to the episode title if no chapter info is available.

CookieyedCodes:
Small reminder, long pressing the mini player artwork brings up a menu to clear upnext and mark the current episode as played don't forget to add that functionality:)

kean:
> long pressing the mini player

The menu does seem to be working. We're planning to extend these context menus though, and use Apple's design there as well.

> Another thing to consider for the mini-player is to show the chapter title, and fall back to the episode title if no chapter info is available.

That is interesting, thank you.

rviljoen:
> We're planning to extend these context menus though, and use Apple's design there as well.

Yes, the native controls blend in much better than the existing `OptionsPicker`. I played around with `UIAlertController` when I added Liquid Glass to my local build a few months back:
image

kean:
@rviljoen, here a prototype of `UITabBarAccessory` in case you want to try it out: https://github.com/Automattic/pocket-casts-ios/pull/4208.

PR #4273, opened 2026-05-06
Optimize PodcastViewController

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

Cuts down the first render time by about 10%.

Implement the following optimizations:

```
1. Stop scanning every pixel of the artwork on main (~39–46 ms) — UIImage+Contrast.swift:5

  contrastColorForPodcastImage calls UIImage.isDark, which walks every byte of the cached image on the main thread inside setupRefreshControl().
  Two cheap fixes, either is fine:
  - Move the computation off-main with Task.detached, install the refresh control synchronously with a default tint, then update the tint when the
   result lands.
  - Cache the boolean per-podcast-uuid (in ImageManager or on Podcast itself) — isDark is purely a function of artwork bytes, recompute only when
  artwork changes.


  4. Cache the markdown-parsed displayCategoryAndAuthor (~11 ms × N redraws) — PodcastHeaderModel.swift:68

  AttributedString(markdown:) is parsed on every body evaluation. It depends only on podcast.podcastCategory and podcast.author. Compute once in
  the view model init, store as a let, return that.


  5. Drop the synchronous layoutIfNeeded() calls in PodcastHeaderCell.configureCellFromSwiftUIView — PodcastHeaderCell.swift:69, 82

  You force a UIHostingController + cell layout before the cell is in the view hierarchy. The table will lay it out anyway. Each is one extra full
   SwiftUI render pass on top of the inherent ~127 ms hosting cost — safe to delete both.
```

The latter matches exactly how `PlaylistHeaderViewCell` works.

## To test

- **Verify** that header view size not changed (other stuff is straightforward)

## Checklist

### DIFF
--- a/podcasts/Podcasts/Podcast Page/Cells/PodcastHeaderCell.swift
+++ b/podcasts/Podcasts/Podcast Page/Cells/PodcastHeaderCell.swift
@@ -1,7 +1,6 @@
 import Foundation
 import SwiftUI
 import UIKit
-
 import PocketCastsDataModel
 import PocketCastsServer
 import PocketCastsUtils
@@ -15,7 +14,7 @@ class PodcastHeaderCell: UITableViewCell {
     let podcast: Podcast
     weak var viewController: PodcastViewController?
     let viewModel: PodcastHeaderViewModel
-    var firstTime = true
+
     init(podcast: Podcast, vc: PodcastViewController) {
         self.podcast = podcast
         self.viewController = vc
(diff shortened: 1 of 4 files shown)


podcasts/Podcasts/Podcast Page/Cells/PodcastHeaderCell.swift
```diff
@@ -24,55 +23,16 @@ class PodcastHeaderCell: UITableViewCell {
...
-        swiftUICellViewController.view.translatesAutoresizingMaskIntoConstraints = false
-        NSLayoutConstraint.activate([
-            cell.contentView.topAnchor.constraint(equalTo: swiftUICellViewController.view.topAnchor),
-            cell.contentView.bottomAnchor.constraint(equalTo: swiftUICellViewController.view.bottomAnchor),
-            cell.contentView.leftAnchor.constraint(equalTo: swiftUICellViewController.view.leftAnchor),
-            cell.contentView.rightAnchor.constraint(equalTo: swiftUICellViewController.view.rightAnchor),
-        ])
-        swiftUICellViewController.didMove(toParent: viewController)
-        swiftUICellViewController.view.layoutIfNeeded()
+        contentConfiguration = UIHostingConfiguration {
```
SergioEstevao: Unfortunately these changes break the animation on the header when expanding/contracting the image.

My first implementation of this code was exactly like this, but then I faced the animation issues and needed to revert to the old method. If I recall well the issue is the way UIHostingConfiguration expands the cells when animating between states. It increases the size from the middle instead of bottom expansion.

Check the animations Before and After:

Before:

https://github.com/user-attachments/assets/65e8ff16-48ac-43f3-a57f-fd099942edfd

After:

https://github.com/user-attachments/assets/6a2eba25-2ab4-498d-9810-d8627eb738f4

SergioEstevao (changes requested):
The changes for markdown and color cache work correctly, unfortunately the changes to optimize layout passes break the animation so those need to be reverted or fixed.

kean:
I reverted the commit that change the header layout. I'll look into other ways to improve upon it later.

SergioEstevao (approved):
Animations back to normal! Thanks for restoring it. :shipit:

PR #4275, opened 2026-05-06
Liquid Glass: Update MultiSelectFooterView

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS-640.

Just adds Liquid Glass effect and brings it closer to the bottom to match the spacing for other bars.

**Note**: it should probably, ideally, replace the tab bar (out of the scope of the PR).

Screenshot 2026-05-06 at 5 15 17 PM

## Checklist

### DIFF
--- a/podcasts/Common Components/MultiSelect/MultiSelectFooterView.swift
+++ b/podcasts/Common Components/MultiSelect/MultiSelectFooterView.swift
@@ -77,6 +77,11 @@ class MultiSelectFooterView: UIView, MultiSelectActionOrderDelegate {
 
     @IBOutlet var statusLabel: ThemeableLabel!
     @IBOutlet var activityIndicator: ThemeLoadingIndicator!
+
+    @IBOutlet var containerLeadingConstraint: NSLayoutConstraint!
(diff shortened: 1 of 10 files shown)


SergioEstevao (approved):
Looking good, but maybe the size should be adjusted on LiquidGlassUI to match the mini-player and tab bar dimensions?

What do you think @david-gonzalez-a8c ?

david-gonzalez-a8c:
> Looking good, but maybe the size should be adjusted on LiquidGlassUI to match the mini-player and tab bar dimensions?

Agree. I would be good to have the same width as the mini player and tab bar (and the same height as the tab bar if it's not already that)

kean:
It looks nicer. _Update_: I'll move it a notch down.

Screenshot 2026-05-11 at 1 15 56 PM

When you scroll, the tab bar collapses, leaving more space.

Screenshot 2026-05-11 at 1 17 40 PM

Ideally, it might be worth hiding the tab bar entirely, but it's also an improvement.

rviljoen:
How about collapsing the tab bar as soon as multi-select is activated?

PR #4279, opened 2026-05-07
SwiftLint: Disable prefer_zero_over_explicit_init

kean:
| 📘 Part of: # |  
|:---:|

Would it be worth reverting this one? Writing `CGPoint(x: 0, y: 0)` is pretty common. I just ran into it in some clean generated code:

Screenshot 2026-05-07 at 9 57 30 AM

More broadly, I wonder if we could be a bit selective about which rules we opt into – focusing on ones that catch genuinely problematic patterns (like `.sorted({ ... }).max()`) and skipping the ones that are more stylistic. My main concern is that warnings showing up on otherwise harmless code, especially in generated files, could become noisy over time.

## Checklist

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -72,9 +72,6 @@ only_rules:
   # the declaration.
   - opening_brace
 
-  # Prefer `0` over explicit `Int(0)` / `CGFloat(0)` etc.
-  - prefer_zero_over_explicit_init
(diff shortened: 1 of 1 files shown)


mokagio (approved):
I'm happy with whatever you prefer @Automattic/pocket-casts-ios

PR #4282, opened 2026-05-07
Liquid Glass: Fix swipe to dismiss in Podcast, Playlist

kean:
| 📘 Part of: # |  
|:---:|

If you hide the back button, the view controller doesn't seem to set up the edge swipe gesture.

The solution is a bit of a mess, but it simplifies things moving forward.

- On iOS 26, we now simply show the standard back button (minimum hacks)
- On iOS 18 and earlier, I re-enabled this workaround in `PCNavigationControler`:

```swift
    override func setNavigationBarHidden(_ hidden: Bool, animated: Bool) {
        super.setNavigationBarHidden(hidden, animated: animated)
        enableInteractivePopGestureWorkaround()
    }

    func enableInteractivePopGestureWorkaround() {
        interactivePopGestureRecognizer?.delegate = self
    }
```

It would previously get enabled automatically when the controller would call `setNavigationBarHidden`. Since I removed this  call in previous PRs it broke this workaround.

## Checklist

### DIFF
--- a/podcasts/New Detail/PlaylistDetailViewController.swift
+++ b/podcasts/New Detail/PlaylistDetailViewController.swift
@@ -247,14 +247,22 @@ class PlaylistDetailViewController: PCViewController, UIScrollViewDelegate {
         )
         customRightBtn = defaultRightBarButton
 
-        defaultBackBarButton = FakeNavBarButton.makeBarButtonItem(
-            image: UIImage(systemName: "chevron.backward"),
-            accessibilityLabel: L10n.back,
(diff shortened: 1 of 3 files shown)


SergioEstevao (approved):
The swipes back are working correctly on both versions.

But is it just me or there is some tint on the Cast button?

Image

kean:
> But is it just me or there is some tint on the Cast button?

Not sure where it's coming from. I'm going to remove it in a follow-up PRs.

PR #4328, opened 2026-05-14
TV: Refresh podcasts on update notifications and server refresh on entry

SergioEstevao:
## Summary

On the tvOS app's Podcasts screen:

- **React to podcast update notifications.** `PodcastsViewModel` now subscribes to `Constants.Notifications.podcastUpdated`, `podcastAdded`, `podcastDeleted` and `ServerNotifications.podcastsRefreshed`, re-fetching the grid via `HomeGridDataHelper` whenever any of them fire so the UI stays in sync with database changes (e.g. after a sync, subscribe/unsubscribe, or background refresh).
- **Trigger a server data refresh when entering the screen.** `load()` (called from the view's `.task`) now calls `RefreshManager.shared.refreshPodcasts()` after the initial local fetch. `RefreshManager` already throttles to a 15s minimum between refreshes, so re-entering the tab quickly is safe.

While in here, two small correctness fixes in `fetchPodcasts()`:
- The original used a fire-and-forget inner `Task`, which meant the `async` function returned before the data was loaded. Switched to `Task.detached { ... }.value` so callers (including the notification handlers) can serialize properly.
- State now becomes `.empty` when there are no items, matching the existing `.empty` UI branch in `PodcastsView`.

The mock view model is unchanged.

## To test

1. Run the **Pocket Casts TV App** target on an Apple TV simulator (or device) while signed in.
2. Open the **Podcasts** tab — confirm the grid populates.
3. Trigger a server refresh from another device (or wait for the refresh to complete) and confirm the grid updates without leaving the tab.
4. Subscribe / unsubscribe to a podcast from another device — confirm `podcastAdded` / `podcastDeleted` notifications cause the grid to update.
5. Leave the Podcasts tab and come back — confirm a server refresh is triggered each time (subject to the 15s throttle in `RefreshManager`).
6. Sign out and back in — confirm no crashes and the grid loads.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Podcasts/PodcastsViewModel.swift
+++ b/Pocket Casts TV App/UI/Podcasts/PodcastsViewModel.swift
@@ -31,19 +31,41 @@ class PodcastsViewModel: PodcastsViewModelProtocol {
 
     init(dataManager: DataManager = DataManager.sharedManager) {
         self.dataManager = dataManager
+        observePodcastUpdates()
     }
 
     func load() async {
         await fetchPodcasts()
+        RefreshManager.shared.refreshPodcasts()
     }
 
     private func fetchPodcasts() async {
-        Task {
-            let gridItems = HomeGridDataHelper.gridItems(orderedBy: .titleAtoZ)
-            await MainActor.run {
-                self.items = gridItems
-                self.state = .ready
-            }
+        let gridItems = await Task.detached {
+            HomeGridDataHelper.gridItems(orderedBy: .titleAtoZ)
+        }.value
+
+        await MainActor.run {
+            self.items = gridItems
+            self.state = gridItems.isEmpty ? .empty : .ready
+        }
+    }
+
+    private func observePodcastUpdates() {
+        let notificationsToObserve: [Notification.Name] = [
+            Constants.Notifications.podcastUpdated,
+            Constants.Notifications.podcastAdded,
+            Constants.Notifications.podcastDeleted,
+            ServerNotifications.podcastsRefreshed
+        ]
+
+        for name in notificationsToObserve {
+            NotificationCenter.default.publisher(for: name)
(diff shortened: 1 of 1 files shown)


Pocket Casts TV App/UI/Podcasts/PodcastsViewModel.swift
```diff
@@ -31,20 +31,44 @@ class PodcastsViewModel: PodcastsViewModelProtocol {
...
         self.dataManager = dataManager
+        observePodcastUpdates()
     }
 
     func load() async {
         await fetchPodcasts()
+        RefreshManager.shared.refreshPodcasts()
     }
 
     private func fetchPodcasts() async {
```
kean, line 42: (nit) the new idiomatic way to do is looks more like this:

```swift
@MainActor @Observable
PodcastsViewModel {

    func load() async {
        items = await fetchPodcasts()
        state = items.isEmpty ? .empty : .ready
    }

    @concurrent private func fetchPodcasts() async -> [GridItem] {
        HomeGridDataHelper.gridItems(orderedBy: .titleAtoZ)
    }
}
```

It's generally best to avoid spawning unstructured tasks (won't participate in cooperative cancellation).

Having said that, it depends on what you are going for. It looks like `load` is invoke from `task`:

```swift
        .task {
            await model.load()
        }
```

The task will get cancelled when the view disappears and will restart when it appears. If you need `load` to be called on every appear, that's fine. If not, it might be worth fetching elsewhere.

kean (approved):
Approved, left minor nit.

Pocket Casts TV App/UI/Podcasts/PodcastsViewModel.swift
SergioEstevao: This is a good suggestion, let me give a look into this.

PR #4333, opened 2026-05-14
[PC TV] Fix folders locations for caches on tvOS

SergioEstevao:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

tvOS has limitations regarding disk caches and folders locations.
All the storage used needs to be in the caches folder, other location will fail when trying to create files on it.
This PR moves the database folder and the image cache folder to the application caches folder so they can be created.

This was only noticed now, because the simulator for tvOS doesn't enforce those limitation this is only detected on a real device.

## To test

1. Start the app on a real device
2. Check if database creation is done. ( app should launch)
3. Login with an user
4. Check if image cache for podcast is working

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Public/DataManager.swift
+++ b/Modules/Sources/PocketCastsDataModel/Public/DataManager.swift
@@ -1160,6 +1160,7 @@ public class DataManager {
 
     private static func pathToDbFolder() -> String {
         #if os(tvOS)
+        //tvOS does not allow the use of the application support or documents directory on a real device so we need to use the caches directory.
         let typeOfDirectory: FileManager.SearchPathDirectory = .cachesDirectory
         #else
         let typeOfDirectory: FileManager.SearchPathDirectory = .applicationSupportDirectory
--- a/Modules/Sources/PocketCastsDataModel/Public/DataManager.swift
+++ b/Modules/Sources/PocketCastsDataModel/Public/DataManager.swift
@@ -1174,7 +1174,7 @@ public class DataManager {
         do {
             try FileManager.default.createDirectory(atPath: pathToDbFolder(), withIntermediateDirectories: true, attributes: nil)
         } catch {
-            print("Unable to create database folder: \(error)")
+            FileLog.shared.addMessage("Unable to create database folder: \(error)")
         }
     }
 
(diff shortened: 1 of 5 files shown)


podcasts/ImageManager.swift
```diff
@@ -16,10 +16,18 @@ class ImageManager {
...
     lazy var subscribedPodcastsCache: ImageCache = {
+        #if os(tvOS)
+        let path = (NSSearchPathForDirectoriesInDomains(.cachesDirectory, .userDomainMask, true).last! as NSString).appendingPathComponent("artworkv3")
+        #else
         let path = (NSHomeDirectory() as NSString).appendingPathComponent("Documents/artworkv3")
+        #endif
         let url = URL(fileURLWithPath: path)
         subscribedPodcastsCache = try! ImageCache(name: "subscribedPodcastsCache", cacheDirectoryURL: url)
+#if os(tvOS)
+        subscribedPodcastsCache.diskStorage.config.sizeLimit = UInt(50.megabytes)
```
kean: It's probably too little, no? tvOS has fairly large disks.

SergioEstevao: Just wanted to limit this to leave more space for the DB, but what do you think it's reasonable? 200 MB?

Modules/Sources/PocketCastsDataModel/Public/DataManager.swift
```diff
@@ -1159,7 +1159,12 @@ public class DataManager {
...
-        let documentsPath = NSSearchPathForDirectoriesInDomains(.applicationSupportDirectory, .userDomainMask, true).last as NSString?
+        #if os(tvOS)
+        let typeOfDirectory: FileManager.SearchPathDirectory = .cachesDirectory
+        #else
+        let typeOfDirectory: FileManager.SearchPathDirectory = .applicationSupportDirectory
+        #endif
+        let documentsPath = NSSearchPathForDirectoriesInDomains(typeOfDirectory, .userDomainMask, true).last as NSString?
         let mainFolder = documentsPath?.appendingPathComponent("Pocket Casts")
 
         return mainFolder!
```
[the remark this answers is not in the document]
SergioEstevao: tvOS forbids Application Support so we need to use caches. Adding check for relaunches to do resync will be done on another PR

kean (approved):

PR #4346, opened 2026-05-15
Liquid Glass: Native Alerts – Part 2

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

https://github.com/user-attachments/assets/89ce3de7-4d4f-454b-a910-1e6b88595177

## Checklist

### DIFF
--- a/podcasts/UploadedViewController.swift
+++ b/podcasts/UploadedViewController.swift
@@ -3,6 +3,7 @@ import SwiftUI
 import PocketCastsDataModel
 import PocketCastsServer
 import UIKit
+
 class UploadedViewController: PCViewController, UserEpisodeDetailProtocol {
     private let episodesDataManager = EpisodesDataManager()
     private var cancellables = Set<AnyCancellable>()
@@ -305,7 +306,7 @@ class UploadedViewController: PCViewController, UserEpisodeDetailProtocol {
 
     func showDeleteConfirmation(userEpisode: UserEpisode) {
         Analytics.track(.userFileDeleteShown)
-        UserEpisodeManager.presentDeleteOptions(episode: userEpisode, preferredStatusBarStyle: preferredStatusBarStyle, themeOverride: nil, dismissCallback: {
+        UserEpisodeManager.presentDeleteOptions(episode: userEpisode, from: self, dismissCallback: {
             Analytics.track(.userFileDeleteDismissed)
         }) { deletedLocal, deletedRemote in
             Analytics.track(.userFileDeleted, properties: ["local": deletedLocal, "remote": deletedRemote])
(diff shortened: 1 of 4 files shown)


podcasts/UploadedViewController.swift
```diff
@@ -317,8 +318,6 @@ class UploadedViewController: PCViewController, UserEpisodeDetailProtocol {
                 self.reloadLocalFiles()
             }
         }
-
-        dismiss(animated: true, completion: nil)
```
kean, line 321: Why?

-----

UploadedViewController — the dismiss was a no-op. UploadedViewController lives in a tab/nav stack (not modally presented) and UserEpisodeDetailViewController uses its own UIWindow
  (via animateIn()), so it isn't in self.presentedViewController either. Calling dismiss(animated:) on UploadedViewController had nothing to dismiss. Safe to drop.

  UpNextViewController — the dismiss was doing real work. UpNext is presented modally as a formSheet (MiniPlayerViewController.swift:547). The original flow closed the UpNext modal so
  only the OptionsPicker (in its own window) remained visible.

  I removed it because in the new flow it would dismiss the alert itself — present(alert, animated:) makes the alert self.presentedViewController, and the very next line
  dismiss(animated:) would close it.

  If you want to preserve the original intent (close UpNext after triggering delete), the options are:

  1. Leave as-is — the alert simply appears on top of the UpNext modal; UpNext stays open. Arguably cleaner UX (less disorientation when canceling).
  2. Dismiss UpNext, then present alert from the underlying VC — would need the presenter to be presentingViewController instead of self, and the dismiss to happen first.

podcasts/UserEpisodeManager.swift
```diff
@@ -249,53 +249,51 @@ struct UserEpisodeManager {
...
-            let deleteCloudAction = OptionAction(label: L10n.deleteFromCloud, icon: nil, action: { [] in
-                UserEpisodeManager.deleteFromCloud(episode: episode)
-                actionCallback?(false, true)
-            })
+        let deleteCloudAction = UIAlertAction(title: L10n.deleteFromCloud, style: .destructive) { _ in
+            UserEpisodeManager.deleteFromCloud(episode: episode)
+            actionCallback?(false, true)
+        }
 
-            let deleteDeviceLabel = episode.uploaded() && episode.downloaded(pathFinder: DownloadManager.shared) ? L10n.deleteFromDeviceOnly : L10n.deleteFromDevice
```
kean, line 264: It's easier to see the diff locally. I instructed to keep the structure exactly the same – and it is.

SergioEstevao (approved):
Tested the behaviour on LiquidUI and non-LiquidUI and all is working correctly.
:shipit:

PR #4359, opened 2026-05-18
Liquid Glass: Add Up Next animations

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

- Add "Up Next" badge showing how many items are in the queue directly in the tab bar (backed in the image)
- Add animation (hacks on top of hacks, but it works)

https://github.com/user-attachments/assets/f4939c1c-dda0-44ba-8614-3e29868a6cf3

## To test

- Verify that when tab bar is minimized, the animation does not run

## Checklist

### DIFF
--- a/podcasts/Main/MainTabBarController+Alert.swift
+++ b/podcasts/Main/MainTabBarController+Alert.swift
@@ -1,3 +1,5 @@
+import UIKit
+
 extension MainTabBarController {
     func presentLoader() {
         DispatchQueue.main.async { [weak self] in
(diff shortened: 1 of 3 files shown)


SergioEstevao (changes requested):
While this is working correctly, I'm bit concerned of the hacks we are doing on the tabar view in order to support this.
If in the future Apple changes the way the tabbar is organized this can stop working or have some side effects.
Do we need to implement this?

I also find the space a bit tight when having 100 or more elements:

Image

kean:
> Apple changes the way the tabbar is organized this can stop working or have some side effects.

I'm adding a bit more defensive code right now to ensure the worse case scenario it starts working and we have to fix it, which I think is a small price to pay. 

I feel that we need to make the "Up Next" just as fun and compelling as the current implementation in the mini player, which is now gone.

>  I also find the space a bit tight when having 100 or more elements

Good catch. Let's change the clamp from 999 to 99. That's as much as would fit.

Screenshot 2026-05-20 at 12 16 37 PM

SergioEstevao (approved):
Thanks for adding more defensive code.

PR #4401, opened 2026-05-22
Liquid Glass: Add context menus in Up Next

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

Adds context menu to "Up Next" cells. It gives you more options and prevents you from accidentally playing an episode by long-pressing and expecting either a context menu (standard on iOS) or multi-select (standard in PC), none of which happen.

https://github.com/user-attachments/assets/920bf2c7-42e4-4c64-9297-b44efcdfd5ed

## To test

- Verify that all actions work
- Verify that long-press during multi-select still invokes the multi-select helper menu

## Checklist


kean:
I'm also adding it for the current episode in the next commit.

CookieyedCodes:
Some people might prefer the existing behaviour it might be good that it's a setting toggle

kean:
> Some people might prefer the existing behaviour it might be good that it's a setting toggle

It might be worth it, let us revisit it.

The number of folks using long-press is exceedingly small, and we can't really tell how many of these are accidental inputs where you expect either a context menu (standard on iOS) or multi-select (standard in PC). The rationale is that  If you do use long-press, you'd be able to tap "Play Now" in the menu. It's a rare operation, so an extra step is not too much to ask.

There is also an existing "Play on Tap" setting. It's a good existing option if you want to be able to quickly play the episodes from "Up Next" – even faster than long-press.

Screenshot 2026-05-22 at 10 04 11 AM

It flips the behavior in production. When it's on, tap = play, long-press = details. When it's off (default), tap = details, long-press = play.

CookieyedCodes:
I actually agree but the one thing that's nice about it is it's faster then looking at a context menu or swiping left to right-reordering upnext-, tbh if this dose get enabled I will probably flip it to be context and long press to bring up the episode card, 
I probably would be fine with it if add to playlist was an option tho.

Using the flipped behavior is too fast and leads to more misstaps imo and are you sure that is no longer going to be an option due to this new menu?

kean:
Descoping it form 8.13 for now.

Context menu on tap is an interesting option.

> leads to more misstaps

That could be the case. No plans to remove the existing option.

CookieyedCodes:
Fair fair, tbh I could see a slightly condensed context menu if 4 icons were placed on top of the context (starred, mark as played(delete for files), add to playlist, remove from upnext)

PR #4418, opened 2026-05-27
Track transcript text highlighting

yaelirub:
| 📘 Part of: # |  |
|:---:|

Adds an analytics event (`transcriptTextHighlighted`) that fires the first time a user creates a non-empty text selection in the transcript view. This is a measurement-only change to validate whether users highlight transcript text often enough to justify follow-up features (e.g. suggesting a clip/bookmark on highlight, or "popular highlights" overlays). No UI change, no data leaves the device beyond the existing event payload.

The event only fires on the empty→non-empty edge of `textViewDidChangeSelection`, so dragging the selection handles or tapping to clear does not spam events. One event per highlight gesture.

Phase 2 (suggest clip/bookmark on highlight) and Phase 3 ("popular highlights" Kindle-style underlines) are tracked separately and are gated on whatever this event shows over the next few weeks.

## To test

1. Play any episode that has a transcript.
2. Open the transcript view.
3. Long-press on the transcript text until iOS selects a word.
4. Confirm a single `transcript_text_highlighted` event is recorded (via Tracks logging / debug overlay).
5. Drag the selection handles to extend the selection — confirm no additional events fire.
6. Tap elsewhere to clear the selection, then long-press again — confirm a new event fires.

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to \`CHANGELOG.md\` if necessary.

### DIFF
--- a/podcasts/Analytics/AnalyticsEvent.swift
+++ b/podcasts/Analytics/AnalyticsEvent.swift
@@ -922,6 +922,7 @@ enum AnalyticsEvent: String {
     case episodeDetailTranscriptCardTapped
     case episodeTranscriptShown
     case transcriptShared
+    case transcriptTextHighlighted
     case syncedTranscriptSeekUsed
     case syncedTranscriptPreparationStarted
     case syncedTranscriptPreparationCompleted
--- a/podcasts/TranscriptViewController.swift
+++ b/podcasts/TranscriptViewController.swift
@@ -1170,9 +1170,9 @@ extension TranscriptViewController: UIScrollViewDelegate {
 
 extension TranscriptViewController: UITextViewDelegate {
     func textViewDidChangeSelection(_ textView: UITextView) {
-        let wasEmpty = !hadNonEmptySelection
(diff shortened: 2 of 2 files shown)


kean (approved):
Tested on iOS 18.6 

It works and sends the event the first time you make a selection, which I presume is the desired outcome.

Note: `textViewDidChangeSelection` gets called instantly as you open the screen with empty selection, but the implementation handles it as `hadNonEmptySelection` stays `false`.

yaelirub:
Much appreciate the quick review, @kean! 🙏 
This shouldn't be approved as the checklist items are not checked. Especially: 
"I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics."

I'm also updating the PR based on Copilot's suggestions, which were valid, especially the one about the reset of the flag. Fixing with Claude now.

PR #4427, opened 2026-05-28
Add explicit content badge to discover and search

dcamozzi:
## Summary
- Add explicit content ("E") badge to podcasts across discover lists, shelf cards, horizontal collections, search results, and search history
- Persist the `isExplicit` flag in the local database (migration v74) and parse it from discover, cache, and search API responses
- Centralize badge rendering in `ExplicitBadgeHelper` with theme-aware colors (`primaryIcon03`)
- All badge UI is gated behind the `showExplicitBadges` feature flag (default off)

## Changes
- **Data model**: `isExplicit` property on `Podcast`, DB migration v74, JSON parsing from discover/cache/search APIs
- **Discover**: Badge on `LargeListCell` (UIImageView in stack), `SmallListCell` and `DiscoverPodcastTableCell` (attributed string), `HorizontalCollectionList` (SwiftUI inline text)
- **Search**: Badge on `SearchResultCell` and `SearchHistoryCell` (SwiftUI inline text); added `CodingKeys` to `PodcastFolderSearchResult` and `CombinedSearchResult` mapping `isExplicit` to the `"explicit"` JSON key
- **Shared**: `ExplicitBadge` SwiftUI view, `ExplicitBadgeHelper` enum with UIImage generation, inline text helpers, attributed string helpers, and `resolvedIsExplicit` DB sync helper
- **Feature flag**: `showExplicitBadges` (default `false`) gates all badge UI rendering

## Test plan
- [ ] Enable the `showExplicitBadges` feature flag in Settings > Developer > Feature Flags
- [ ] Browse Discover — verify badge appears on explicit podcasts in small list, large shelf, table, and horizontal collection views
- [ ] Search for an explicit podcast — verify badge appears in results and autocomplete
- [ ] Check search history — verify badge appears for previously searched explicit podcasts
- [ ] Switch themes — verify badge color updates correctly across all 9 themes
- [ ] Verify badge does not scale with Dynamic Type (fixed 11pt size)
- [ ] Verify badge has correct accessibility label ("Explicit content")
- [ ] Disable the `showExplicitBadges` feature flag — verify no badges appear anywhere

## Demo
image


podcasts/Common Components/ExplicitBadge.swift
```diff
@@ -0,0 +1,81 @@
+import SwiftUI
+import PocketCastsDataModel
+import PocketCastsServer
+import PocketCastsUtils
+
+struct ExplicitBadge: View {
```
kean: Unused code.

podcasts/Common Components/ExplicitBadge.swift
```diff
@@ -0,0 +1,81 @@
...
+            .accessibilityLabel(L10n.podcastExplicitContent)
+    }
+}
+
+enum ExplicitBadgeHelper {
+    static let badgeSize: CGFloat = 11
+
+    static func badgeImage(for theme: Theme.ThemeType? = nil) -> UIImage {
+        let bgColor = ThemeColor.primaryIcon03(for: theme)
+        let renderer = UIGraphicsImageRenderer(size: CGSize(width: badgeSize, height: badgeSize))
```
kean, line 21: The app should ideally cache these (per theme) like we now do for other reused images.

Modules/Sources/PocketCastsServer/Public/Search/CombinedSearchTask.swift
```diff
@@ -18,6 +18,12 @@ public struct CombinedSearchResult: Decodable, Hashable {
     public let podcastUuid: String?
     public let podcastTitle: String?
     public let author: String?
+    public let isExplicit: Bool?
+
+    enum CodingKeys: String, CodingKey {
```
kean: (nit) Let's name the fields `explicit` to avoid introducing custom coding keys, same applies to `PodcastSearchTask`.

podcasts/Common Components/ExplicitBadge.swift
```diff
@@ -0,0 +1,81 @@
...
+        }
+        return Text(title)
+    }
+
+    static func resolvedIsExplicit(for podcast: DiscoverPodcast) -> Bool {
+        let discoverExplicit = podcast.isExplicit ?? false
+        if let uuid = podcast.uuid, let localPodcast = DataManager.sharedManager.findPodcast(uuid: uuid, includeUnsubscribed: true) {
+            if discoverExplicit, !localPodcast.isExplicit {
+                localPodcast.isExplicit = true
+                DataManager.sharedManager.save(podcast: localPodcast)
```
kean: The reads (`DataManager.sharedManager.findPodcast`) seem generally OK as there is an in-memory cache and that's what other code in Discover is doing, but the DB saves seem to happen synchronously on the main thread as you scroll.

Should Discover update the local podcasts? If yes, why only the `isExplicit` field, or why here? It would be acceptable performance-wise to do it close to where the app decodes the data: decode, then update local podcasts in the background in a batch.

kean (changes requested):
Hey, nice change. The labels are showing up. The code might need a bit of work before it's mergable. I could help with it, but not sure how to prioritize the work right now.

@SergioEstevao , can you also take a quick look, please, especially around the DB migration part? 

@dcamozzi , just an observation – would it make sense to show these labels in "Podcast Details" as well and should they be shown on episodes?

Modules/Sources/PocketCastsDataModel/Public/Model/Podcast+fromDictionary.swift
```diff
@@ -57,6 +57,11 @@ extension Podcast {
         if let isPrivate = podcastJson["is_private"] as? Bool {
             podcast.isPrivate = isPrivate
         }
+        if let explicit_ = podcastJson["explicit"] as? Bool {
```
SergioEstevao: Why the `_` on the variable name?

Modules/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
```diff
@@ -78,6 +78,11 @@ extension ServerPodcastManager {
         if let isPrivate = podcastJson["is_private"] as? Bool {
             podcast.isPrivate = isPrivate
         }
+        if let explicit_ = podcastJson["explicit"] as? Bool, explicit_ {
```
SergioEstevao: Same question here why the `_` ?

podcasts/Common Components/ExplicitBadge.swift
SergioEstevao: I agree,  This is not a pattern that we see in the app. DiscoverPodcasts are like a stub that get resolved when we want to see the detail of them with the full Podcast objects.

dcamozzi:
> Hey, nice change. The labels are showing up. The code might need a bit of work before it's mergable. I could help with it, but not sure how to prioritize the work right now.
> 
> @SergioEstevao , can you also take a quick look, please, especially around the DB migration part?
> 
> @dcamozzi , just an observation – would it make sense to show these labels in "Podcast Details" as well and should they be shown on episodes?

Thanks, Alex. I’m having trouble getting it to show consistently on the podcast detail page. Right now, the icon appears only a few seconds after the page loads, rather than on initial load. Maybe this is something you could help me with? 

For episode-level support, we do want to show it there as well. However, @geekygecko  and I agreed to start with the show level first since that data is already available.

I’m not sure where this falls in terms of prioritization. @geekygecko, do you have more context?

podcasts/Common Components/ExplicitBadge.swift
dcamozzi: Done in 4c3141955a1ad4c2bbaf744198e51eac68a85f99

podcasts/Common Components/ExplicitBadge.swift
dcamozzi: Done in cf9001f7940902cd0439fdd38d50cc628c8bf6b4

Modules/Sources/PocketCastsServer/Public/Search/CombinedSearchTask.swift
dcamozzi: Done in 12a5065d16803d499b2a0025975d5725064cf7a9

podcasts/Common Components/ExplicitBadge.swift
dcamozzi: Done in eb638fcea10580a963da3464cf9343fa76afc586 — removed resolvedIsExplicit entirely. Callers now use the DiscoverPodcast.isExplicit value directly.

Modules/Sources/PocketCastsDataModel/Public/Model/Podcast+fromDictionary.swift
dcamozzi: Done in 7842402812d79cd2e244e4e5c0eb3ff1f873ca6d

Modules/Sources/PocketCastsServer/Public/ServerPodcastManager+Update.swift
dcamozzi: Done in 7842402812d79cd2e244e4e5c0eb3ff1f873ca6d

dcamozzi:
## Review feedback addressed

- **ExplicitBadge.swift:6** (@kean): Remove unused `ExplicitBadge` SwiftUI view — Done in 4c3141955a1ad4c2bbaf744198e51eac68a85f99
- **ExplicitBadge.swift:25** (@kean): Cache `badgeImage` per theme — Done in cf9001f7940902cd0439fdd38d50cc628c8bf6b4
- **CombinedSearchTask.swift:23** (@kean): Rename `isExplicit` → `explicit` on search types to drop custom coding keys (also applied to `PodcastSearchTask`) — Done in 12a5065d16803d499b2a0025975d5725064cf7a9
- **ExplicitBadge.swift:59** (@kean, @SergioEstevao): Remove `resolvedIsExplicit` and its DB saves from Discover cells — Done in eb638fcea10580a963da3464cf9343fa76afc586
- **Podcast+fromDictionary.swift:60** (@SergioEstevao): Rename `explicit_` → `isExplicit` — Done in 7842402812d79cd2e244e4e5c0eb3ff1f873ca6d
- **ServerPodcastManager+Update.swift:81** (@SergioEstevao): Rename `explicit_` → `isExplicit` — Done in 7842402812d79cd2e244e4e5c0eb3ff1f873ca6d

geekygecko:
> I’m not sure where this falls in terms of prioritization. @geekygecko, do you have more context?

I asked Ellie about the episode explicit badge and it won't be prioritise right now, so it will only be available for the podcasts in the current version.

dcamozzi:
> > I’m not sure where this falls in terms of prioritization. @geekygecko, do you have more context?
> 
> I asked Ellie about the episode's explicit badge, and it won't be prioritized right now, so it will only be available for the podcasts in the current version.

@geekygecko does this mean we'll ship what we have right now, even if it's not showing on the podcast detail page?

geekygecko:
> does this mean we'll ship what we have right now, even if it's not showing on the podcast detail page?

I might continue this discussion in Slack as it's a bit easier. p1780553727875719-slack-C0AUKRHJT5J

podcasts/New Search/Views/Results/SearchResultCell.swift
```diff
@@ -180,7 +180,7 @@ private extension SearchResultCell {
                     .foregroundColor(AppTheme.color(for: .primaryText02, theme: theme))
                     .lineLimit(1)
                 } else if let result = model.podcastFolder {
-                    Text(result.titleToDisplay)
+                    ExplicitBadgeHelper.inlineTitle(result.titleToDisplay, isExplicit: result.kind == .podcast && result.resolvedIsExplicit, theme: theme.activeTheme)
```
kean, line 183: (nit) There is a bit of duplication that could be eliminated in follow up PRs.

podcasts/EpisodeAIChatEntryPointMockup.swift
```diff
@@ -0,0 +1,211 @@
+import SwiftUI
+
+// MARK: - Player AI Chat Entry Point Mockup (Spotify-style)
+// Self-contained mockup — delete after review
```
kean: Looks like a temporary file that was committed by accident. I removed it and will push this change.

kean (approved):
I reviewed the updates, and everything's been addressed, including the DB writes on main. 

There was a presumably accidentally committed `PlayerAIChatMockup` file that was unused and that I already removed and pushed the change in this branch. The rest of the code looks nice and maintainable. I smoke tested the changes as well.

I also updated the CHANGELOG.md file to include the change.

podcasts/New Search/Views/Results/SearchResultCell.swift
dcamozzi: Noted — will clean up the duplication in a follow-up PR.

podcasts/EpisodeAIChatEntryPointMockup.swift
dcamozzi: Thanks for catching and removing this!

PR #4433, opened 2026-05-28
Liquid Glass: Options Picker Sheet Presentation

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

There are three changes.

**1. Fix layout when only one item is shown**

In a special case when there is only one item, we don't want to add top padding and want it centered. Here's before and after.

Screenshot 2026-05-28 at 5 02 44 PM

**2. Fix layout in IncomingShareListViewController

Here's before and after. I believe it's a production issue, which is why I'm adding it to `trunk` only. It's rarely used.

You can apply this patch to test it [incoming-share-list-test.patch](https://github.com/user-attachments/files/28368130/incoming-share-list-test.patch).

Screenshot 2026-05-28 at 5 03 48 PM

**3. Update a few more places to use sheet presentation

Screenshot 2026-05-28 at 4 09 41 PM
Screenshot 2026-05-28 at 4 38 05 PM
Screenshot 2026-05-28 at 4 50 01 PM

## Checklist

### DIFF
--- a/podcasts/Episode Cells/EpisodeCell.swift
+++ b/podcasts/Episode Cells/EpisodeCell.swift
@@ -546,7 +546,6 @@ class EpisodeCell: ThemeableSwipeCell, MainEpisodeActionViewDelegate {
     func errorTapped() {
         guard let episode else { return }
 
-        let statusBarStyle = playlistUuid == nil ? UIStatusBarStyle.lightContent : AppTheme.defaultStatusBarStyle()
         if episode.playbackError() {
             let optionsPicker = OptionsPicker(title: nil)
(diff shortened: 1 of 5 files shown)


SergioEstevao (approved):
All working correctly, except for Refresh Artwork that is not showing the reload icon:

Image

But in production is also not showing the icon, so maybe a test from your part when saving the screenshot?

kean:
> so maybe a test from your part when saving the screenshot?

I temporarily added while working on it. I guess it does look better with an icon. I'll add it back – seems like a clear cut change.

PR #4440, opened 2026-05-29
Analytics: Move work to background

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

I took a simpler approach and removed a global `AnalyticsActor`. Instead, every `AnalyticsAdapter` now determines its own isolation. It can be: an actor, `@MainActor`, no isolation (but still concurrent).

- Hot call sites no longer block on adapter work (Tracks DB saves, date formatting, buffer reduction).
- Launch lighter – `TracksContextManager` init runs in a background Task
- `LiveAnalyticsStreamer` saves one Task hop per event (was `nonisolated` → `@AnalyticsActor Task`; now direct actor call).

## To test

Smoke test analytics.

## Checklist


podcasts/UserSatisfactionSurveyManager.swift
```diff
@@ -4,6 +4,7 @@ import SwiftUI
 import PocketCastsUtils
 import StoreKit
 
+@MainActor
 public class UserSatisfactionSurveyManager: NSObject {
```
kean, line 8: This and `NotificationsCoordinator` are now guaranteed to process events on the main thread (for synchronization). There are some bits that could be done in the background. I'll do this in a follow up PR.

SergioEstevao (approved):
Tested the multiple analytics adapter and performance is looking and code correct.
:shipit:

PR #4467, opened 2026-06-04
Fix transcript search bar background on iOS 26

leandroalonso:
Fixes PCIOS-715

| iOS 26 | iOS 18 |
| ------ | ------- |
| Simulator Screenshot - iPhone 17 - 2026-06-04 at 10 45 47 | Simulator Screenshot - iPad (A16) - 2026-06-04 at 10 45 26 |

The transcript search bar (`TranscriptSearchAccessoryView`) is a `UIInputView` created with `inputViewStyle: .keyboard`, which used to paint its own keyboard-style background. On iOS 26 that style no longer fills the background, so the bar rendered transparent and showed the transcript through it.

This restyles the bar **on iOS 26 only**, matching the native Liquid Glass find bar:

- The search field and the up/down chevrons are each wrapped in a `UIGlassEffect` capsule that floats over the keyboard.
- The field capsule matches the height of the chevron capsule (stack `alignment = .fill`).
- The **Done** button becomes a blue checkmark glass icon (`UIButton.Configuration.prominentGlass()`), with "Done" kept as its accessibility label.

Everything is gated behind `if #available(iOS 26.0, *)`, so older iOS keeps the previous flat layout backed by the keyboard style.

## To test

1. Run on an iOS 26 device/simulator.
2. Open an episode transcript (player or episode detail).
3. Tap **Search**.
4. Verify the search field and chevrons render as glass capsules over the keyboard, the field and chevron capsules are the same height, and the Done control is a blue checkmark.
5. Run on an iOS 18 (or earlier) simulator and confirm the search bar looks/behaves as before.

## Checklist

### DIFF
--- a/podcasts/TranscriptSearchAccessoryView.swift
+++ b/podcasts/TranscriptSearchAccessoryView.swift
@@ -45,13 +45,18 @@ class TranscriptSearchAccessoryView: UIInputView {
         return stackView
     }()
 
-    private lazy var doneButton: UIButton = createButton(
-        title: L10n.done,
-        action: #selector(done),
-        titleColor: .label,
-        highlightedTitleColor: .systemGray,
-        contentInsets: .init(top: 8, leading: 16, bottom: 8, trailing: 16)
-    )
+    private lazy var doneButton: UIButton = {
+        if #available(iOS 26.0, *) {
+            return createDoneGlassButton()
+        }
+        return createButton(
+            title: L10n.done,
+            action: #selector(done),
+            titleColor: .label,
+            highlightedTitleColor: .systemGray,
+            contentInsets: .init(top: 8, leading: 16, bottom: 8, trailing: 16)
+        )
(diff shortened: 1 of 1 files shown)


kean (approved):
Worked well.

One small nit – the standard "Done" button looks a bit out of place there and visually larger than the rest of the fields. Ideally, I would replace the "Search" (and share) buttons in the navigation bar with "Done" during search. You shouldn't be able to tap "Search" while you are already searching, so it seems like a natural place for it.

Image

leandroalonso:
@kean the problem I see with "Done" at the top is that is harder to reach when searching — which all of the elements are at the bottom of the screen.

Also, the "Done" is the same size as the input, but it might seem bigger.

I'll go ahead and merge this one but I believe this screen also needs some Liquid Glass so we can revisit those decisions. Even tho because it's currently broken visually.

PR #4471, opened 2026-06-04
[tvOS] Add error handling for QR code

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS-684.

- Show a `ContentUnavailableView` with the error message and a "Try Again" button when fetching the QR sign-in code
  fails, instead of hanging on a spinner.
- Scope the QR polling to the login tab via `.task(id:)` so it cancels (and resets state) when switching to Password,
  preventing the QR error from leaking across modes.

**Note**: the original ticket also included retry attempts. Do we really need them? Is the server known to initially fail but recovered quickly and we are working something around? If not, I'd suggest to keep it simple.

Screenshot 2026-06-04 at 1 58 38 PM

## To test

1. On Apple TV, open Sign In with QR selected, go offline (or otherwise force the code request to fail), and confirm the error view appears with a focused "Try Again" button that retries successfully once back online.
2. Trigger the QR error, switch to the Password tab, and confirm no error is shown there; switch back to QR and confirm a fresh code/flow starts.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Auth/SignInView.swift
+++ b/Pocket Casts TV App/UI/Auth/SignInView.swift
@@ -84,7 +84,9 @@ struct SignInView: View {
             .padding(.top, 80)
             .offset(y: -64)
         }
-        .task {
+        .task(id: loginType) {
+            // Clear any leftover error so it doesn't leak across login modes.
+            model.state = .start
             switch loginType {
             case .qr:
                 await model.thirdPartyApprovalSignin()
--- a/Pocket Casts TV App/UI/Auth/SignInView.swift
+++ b/Pocket Casts TV App/UI/Auth/SignInView.swift
@@ -64,15 +64,19 @@ struct SignInView: View {
                         usernamePasswordLogin
                             .padding(.top, 64)
                     case .qr:
-                        Text(L10n.tvSignInSubtitle)
-                            .font(.headline)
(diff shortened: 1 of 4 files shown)


Pocket Casts TV App/UI/Auth/SignInView.swift
```diff
@@ -64,23 +64,29 @@ struct SignInView: View {
...
-                            .font(.headline)
-                            .foregroundStyle(Color.pcTextSecondary)
-                        qrCodeDigits
+                        if case .error(_, let message) = model.state {
+                            qrCodeError(message: message)
+                        } else {
+                            Text(L10n.tvSignInSubtitle)
+                                .font(.headline)
+                                .foregroundStyle(Color.pcTextSecondary)
+                            QRCodeView(url: model.pairURLString)
```
kean, line 73: It shows a default QR code with a https://pocketcasts.com/pair link when the actual URL is loading. Shouldn't it show a spinner instead?

Image

SergioEstevao (approved):
Working correctly, showing error when connection fails

Pocket Casts TV App/UI/Auth/SignInView.swift
SergioEstevao: Humm good point maybe all the section should be a progress view until we get the code?

PR #4497, opened 2026-06-10
[PC TV] Require account for Up Next / mark played / archive actions

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-692.

Extends the tvOS `requireAccount` gate — previously only on **Follow podcast** — to the remaining episode actions that mutate account-bound state synced to the server. Logged out, these would silently fail to sync (and in some cases never sync even after a later sign-in), so they now prompt account creation instead.

### Actions now gated behind account creation
- **Play Next in Up Next** (`.default` + `.upNext` menus)
- **Play Last in Up Next** (`.default` + `.upNext` menus)
- **Mark as Played**
- **Archive / Unarchive**

All live in the shared `EpisodeRowWithActions` menu, so the gate now applies consistently across Podcast detail, Up Next, Playlists, Starred, and History.

## To test
1. Sign out of the tvOS app.
2. Open a podcast, focus an episode, and press the **…** (more) button.
3. Select **Mark as Played**, **Archive**, or **Play Next/Last in Up Next** → the Create Account flow should appear instead of the action running.
4. Sign in, repeat → the action runs normally.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Podcasts/EpisodeRow.swift
+++ b/Pocket Casts TV App/UI/Podcasts/EpisodeRow.swift
@@ -110,6 +110,7 @@ struct EpisodeRowWithActions: View {
     let model: EpisodeRowViewModel
     var context: Context = .default
 
+    @Environment(\.requireAccount) private var requireAccount
     @FocusState private var focusedElement: FocusElement?
     @State private var isPlaying = false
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
:shipit:

PR #4513, opened 2026-06-12
[PC TV] Report tvOS platform in LiveAnalyticsStreamer

kean:
Fixes PCIOS-727.

The Live analytics adapter is already registered on the tvOS target, but `LiveAnalyticsStreamer` hardcodes `platform: "iOS"` — so every event sent from tvOS is mislabeled as iOS.

This resolves the platform string at compile time via `#if os(tvOS)`, mirroring the existing `pcios`/`pctvos` split in `TracksAdapter`. iOS behavior is unchanged.

### DIFF
--- a/podcasts/Analytics/Adapters/LiveAnalyticsStreamer.swift
+++ b/podcasts/Analytics/Adapters/LiveAnalyticsStreamer.swift
@@ -49,6 +49,14 @@ actor LiveAnalyticsStreamer: AnalyticsAdapter {
 
     private let encoder = JSONEncoder()
 
+    private static let platform: String = {
+#if os(tvOS)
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
:shipit:

SergioEstevao (changes requested):
While code looks good I'm unable to see the logs on server admin area. Could it be that tvOS needs to be added there?
Where you able to check on the server the events?

pocketcasts:
Version `8.15` has now entered code-freeze, so the milestone of this PR has been updated to `8.16`.

SergioEstevao (approved):
After the changes on server side this is now working correctly!

:shipit:

PR #4526, opened 2026-06-16
Configure Dependabot bundler updates and Ruby review routing

mokagio:
Sets up Dependabot to check Ruby dependencies daily so we can more easily address security updates. See [AINFRA-2437](https://linear.app/a8c/issue/AINFRA-2437).

Also routes review of the Apps Infra surface (dependency manifests, CI config, toolchain pins) to @Automattic/apps-infra-tooling via CODEOWNERS — GitHub retired the dependabot.yml `reviewers` key, and CODEOWNERS is its designated replacement. The routing is path-based, so it applies to any PR touching those files, not just Dependabot's; deliberate, since no source-scoped alternative exists.

---

Posted by Claude Code, Opus 4.8 (1M context) on behalf of @mokagio.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: bundler
+    directory: "/"
+    schedule:
+      interval: daily
+    groups:
+      ruby-minor-and-patch:
+        update-types: [minor, patch]
+    open-pull-requests-limit: 5
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: bundler
     directory: "/"
     schedule:
-      interval: daily
+      interval: weekly
     groups:
       ruby-minor-and-patch:
         update-types: [minor, patch]
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,22 @@
 * @Automattic/pocket-casts-ios
+
+# Ruby dependency surface
+Gemfile*              @Automattic/apps-infra-tooling
+.bundle/              @Automattic/apps-infra-tooling
+.rubocop*.yml         @Automattic/apps-infra-tooling
+fastlane/             @Automattic/apps-infra-tooling
+# Store-listing localization is content, not infra.
+fastlane/metadata/
+
+# CI and automation
+.buildkite/           @Automattic/apps-infra-tooling
+.github/workflows/    @Automattic/apps-infra-tooling
+.github/dependabot.yml @Automattic/apps-infra-tooling
+Dangerfile*           @Automattic/apps-infra-tooling
+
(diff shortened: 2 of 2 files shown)


kean:
> dependencies daily so we can more easily address security updates

Hey, will surface security updates or every patch? It's be nice to avoid being bombarded with PRs the way other repos are. Would it perhaps be possible to set dependabot to run once a day and batch the updates?

mokagio:
Hey @kean 

> Hey, will surface security updates or every patch? 

This should surface every Ruby update. 

> It's be nice to avoid being bombarded with PRs the way other repos are. 
Fair. FWIW, this setup has been open for a week+ in other repos and I haven't seen a bombardment. Plus, with the `CODEWONERS` setting it's Apps Infra that should get the ping. However... maybe this repo has a different setting elsewhere:

image

> Would it perhaps be possible to set dependabot to run once a day and batch the updates?

The setting is to once a day, [`interval: daily`](https://github.com/Automattic/pocket-casts-ios/pull/4526/changes#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R6), at the moment. Patch and minor updates will be batched ([`groups`](https://github.com/Automattic/pocket-casts-ios/pull/4526/changes#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R7-R9)). Major updates will be standalone. We could change to group major, minor, and patch. Alternatively, I'm investigating the best way to skip major updates because those should be intentional on our end.

Also FWIW, we plan to follow up this wave of config update with [AINFRA-2503: Define reusable auto-merge workflow in dangermattic](https://linear.app/a8c/issue/AINFRA-2503/define-reusable-auto-merge-workflow-in-dangermattic)

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: bundler
+    directory: "/"
+    schedule:
+      interval: daily
```
mokagio:
```suggestion
      interval: weekly
```

@kean if the concern is on a constant drip of PRs, then maybe bumping from daily to weekly would help.

kean: Thank Gio, I appreciate you taking this into consideration 🙇 

Let's also make sure @SergioEstevao is aware and agrees with it.

I think weekly is good.

SergioEstevao (approved):
I agree with Alex, weekly should be enough.

mokagio:
@kean @SergioEstevao switched to weekly as discussed in https://github.com/Automattic/pocket-casts-ios/pull/4526/commits/90a2b3843530f84b849b52ccb3309e47483b9ad3 . Thank you for your prompt feedback.

Goes without saying, if the configuration as it is now will result in too much noise in the PRs queue, I and the team will be happy to iterate.

PR #4527, opened 2026-06-16
Fix Playlist artwork header popping back during search

dericleeyy:
Fixes PCIOS-609

When searching within a Playlist/Filter, the artwork header would pop back into view if the search returned a short results list: the table was too short to keep the ~335pt header scrolled off-screen, so it rubber-banded back to the top.

This aligns the Playlist search behaviour with the Podcast detail page's search (which already handles this):

- On search begin, add a clear `tableFooterView` so the table always has room to scroll the artwork header fully off-screen and keep it there even when few results match (removed again on search end). Mirrors `PodcastViewController.didActivateSearch()`.
- Dismiss the keyboard when the user starts scrolling the results (`scrollViewWillBeginDragging` → `view.endEditing(true)`), matching the Podcast page.

The search bar remains non-sticky (scrolls with the content) and the artwork header stays present during search — consistent with the Podcast view. Changes are in `PlaylistDetailViewController+Search.swift` and `PlaylistDetailViewController.swift`.

## To test

1. Tap the **Filters** tab and open a filter (smart filter) that contains many episodes.
2. Scroll down so the artwork header is off-screen.
3. Tap the search bar and type a query that returns only **1–2 matching episodes**.
- [x] Verify the artwork header does **not** pop back into view; the search bar stays at the top, just below the navigation bar (not under the status bar / Dynamic Island).
4. Clear the query and try one that returns **0 results**, then one that returns **many results**.
- [x] Verify no jarring header pop-back in any case.
5. While viewing results, start scrolling the list.
- [x] Verify the on-screen keyboard dismisses on scroll.
6. Tap **Cancel** to exit search.
- [x] Verify the artwork header returns cleanly and the scroll position is sane.
7. Repeat with a **manual playlist**.
- [x] Confirm the behaviour matches the **Podcast** detail page's episode search.

## Checklist

> ⚠️ **TODO (manual):** No unit tests added — this is UI scroll/layout behaviour and `PlaylistDetailViewModel`'s section builder is private with `DataManager` dependencies; verified manually instead. Remove this note or add coverage if you disagree.

- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics. — n/a, no analytics changes


dericleeyy:
I made the behaviour consistent with the Podcast view.

podcasts/New Detail/PlaylistDetailViewController+Search.swift
```diff
@@ -7,11 +7,21 @@ extension PlaylistDetailViewController: PCSearchBarDelegate {
         }
         viewModel.startSearch()
 
+        // Pad the bottom so the table has enough room to scroll the artwork header fully off-screen
+        // and keep it there even when the search returns few results — otherwise a short list
+        // rubber-bands back to the top and the header pops back into view (PCIOS-609). Mirrors
+        // PodcastViewController.didActivateSearch(); 335 is the PlaylistHeaderViewCell height.
+        let tableBounds = tableView.bounds
+        let footer = UIView(frame: CGRect(x: 0, y: 0, width: tableBounds.width, height: tableBounds.height - 335))
```
SergioEstevao, line 15: Nit: Maybe we can make the 335 a constant on PlaylistHeaderViewcell to that can be use there and here in tandem and avoid any future change of the value causing issues.

SergioEstevao (approved):
Working correctly thanks for fixing this!
:shipit:

PR #4535, opened 2026-06-16
Sync playback progress locally between Apple Watch and phone on pause

dericleeyy:
Fixes PCIOS-196

Playback progress made on the Apple Watch (Watch play source) didn't appear on the iPhone — and changes on the phone didn't appear on the watch — without manually tapping "Refresh now" on both devices, because the two apps only reconciled through a server sync/refresh.

Currently the Watch source only syncs to the cloud server, with a 2 minute throttle limit update. When the phone receives the update, it does not reflect it immediately and require an app restart. The idea now is for the watch to also send the update to the phone locally, without depending on cloud. The cloud sync is untouched.

On pause, each device now pushes its playback position directly to the other over WatchConnectivity and applies it to the local player/DB, so Now Playing and episode progress update without a manual refresh. Conflicts use last-write-wins on `playedUpToModified`, the actively-playing device is never overridden, and the existing server sync still propagates to other devices (web, iPad). Gated behind a new `watchPlaybackProgressLocalSync` feature flag (default on).

**Changes**

- Add `watchPlaybackProgressLocalSync` feature flag (`FeatureFlag.swift`).
- Add a `PlaybackProgressUpdate` WatchConnectivity message and a `nowPlayingPlayedUpToModified` now-playing key (`WatchConstants.swift`).
- Watch → phone: on pause the watch sends its position (`WatchSyncManager.playbackPaused` → `SessionManager.sendPlaybackProgress`); the phone applies it and moves the live mini-player via `seekToFromSync` (`WatchManager.handlePlaybackProgressUpdate`).
- Phone → watch: the phone includes `playedUpToModified` in its now-playing state, and the watch applies the phone's paused position to its local player in `SessionManager.handleStateUpdate` (new `WatchDataManager` accessor).

## To test

Use a paired iPhone + Apple Watch (or simulators) signed into the same Pocket Casts Plus account, with the `watchPlaybackProgressLocalSync` flag enabled.

Watch → phone:

1. On the watch, set the play source to **Watch** and play an episode.
2. Let it play ~15s, then pause on the watch.
3. Open that episode / Now Playing on the phone **without** tapping Refresh.
- [x] Verify the phone shows the watch's position within a second or two.

Phone → watch:

1. With the same episode loaded on the watch (Watch source), play it on the **phone**, then pause.
- [x] Verify the watch's Now Playing position updates to the phone's position without relaunching the watch app.

Flag off:

- [x] Verify that with the flag disabled, progress only syncs after a manual "Refresh now" (previous behavior).

Phone to Watch local sync

https://github.com/user-attachments/assets/bb6e708b-aade-4de7-982d-37eb3244e0e4

Watch to Phone local sync

https://github.com/user-attachments/assets/b779383d-e9da-4b76-99f1-57cbbfc9620c

## Checklist


pocketcasts:
Version `8.15` has now entered code-freeze, so the milestone of this PR has been updated to `8.16`.

SergioEstevao (changes requested):
I tested this today by installing the branch on my Watch and Phone.

[x] Update from Watch to Phone: where working correctly, when moving the position of the now play episode on the watch source and pausing after a couple of seconds the phone updated.

[x] Update from Phone to Watch: this didn't work for me. I move around the phone play position, paused. The Watch only updated when I force a refresh on the home screen on the watch.

dericleeyy:
> I tested this today by installing the branch on my Watch and Phone.
> 
> [x] Update from Watch to Phone: where working correctly, when moving the position of the now play episode on the watch source and pausing after a couple of seconds the phone updated.
> 
> [x] Update from Phone to Watch: this didn't work for me. I move around the phone play position, paused. The Watch only updated when I force a refresh on the home screen on the watch.

Was this on a physical device? Updating from phone to watch on the SIM seemed to work for me. Unfortunately, I dont have a physical Apple Watch to test on.

dericleeyy:
Quick context on what's changed since your review, @SergioEstevao: the commit after your review ("Fix last-write-wins behavior") addressed a Copilot comment on the **watch→phone** path (preserving the payload's `modifiedAt` timestamp) — that's the direction you said *worked*, so it won't have changed what you saw.

I traced the phone→watch path and the phone side looks correct: on pause it bumps `playedUpToModified` and sends a fresh now-playing payload (with position + timestamp) via `sendStateToWatch`, triggered on both `playbackPaused` and `playbackPositionSaved`. So the update is being sent. The failure is on the watch's apply side, in `applyPhoneNowPlayingProgress()` — but every precondition there bailed with a silent `return`, so there was no way to tell *why* it didn't apply for you.

I've just pushed a commit that splits those guards and logs the specific reason (with values) on each reject path. Could you re-run the phone→watch test with this build and grab the watch logs? We're looking for a `SessionManager: …phone progress…` line — it'll either show which guard rejected it (e.g. "not newer", "episode mismatch", "no local current episode") or be absent entirely, which would tell us the message isn't reaching `handleStateUpdate` in the first place.

One thing I want to confirm on that note: the whole apply path is gated behind the `watchTransferUserInfoApi` flag — was that enabled in your build alongside `watchPlaybackProgressLocalSync`? If it was off, that alone would explain it.

pocketcasts:
Version `8.16` has now entered code-freeze, so the milestone of this PR has been updated to `8.17`.

SergioEstevao (approved):
I tested this again today, and now it looks is working correctly both ways:

Phone to Watch
Watch to Phone

Changing episode status on any of them update the other.
This was tested on real device and I will send you the logs by email for you to check.

PR #4539, opened 2026-06-16
Re-state unchecked Sendable conformance on Operation subclasses

kean:
Re-instates `@unchecked Sendable` conformance on our `Operation` subclasses (the API tasks, refresh, and playlist-fetch operations) plus a couple of `UIActivityItemProvider` subclasses. These were dropped while cleaning up Xcode warnings, but `Operation` still requires the conformance to build without concurrency warnings, so we need to put it back for now.

Longer term we should redesign these tasks and remove the `Operation` usage entirely rather than keep papering over it with `@unchecked Sendable`.

## To test

1. Build the app — there should be no Sendable/concurrency warnings on the affected types.

## Checklist


SergioEstevao (approved):
:shipit:

PR #4554, opened 2026-06-18
[PC TV] Auto-play video previews on Discover video cells

SergioEstevao:
Fixes [PCIOS-745](https://linear.app/a8c/issue/PCIOS-745/fix-button-spacing-and-enable-video-autoplay-in-made-for-tv-row)
|:---:|

Adds an auto-playing, muted video preview to the Discover video episode cells on Pocket Casts TV. When a video episode card gains focus it begins playing a short preview that cross-fades in from the thumbnail; when focus is lost (or the preview reaches its time limit) the audio fades out and the card cross-fades back to the thumbnail.

### What changed
- **`DiscoverVideoEpisodeCell`** — the card background now swaps between the thumbnail and a `VideoPlayer` based on focus/playback state, wrapped in a `Group` with an opacity `.transition`/`.animation` so the swap cross-fades smoothly. Focus changes drive `model.play()` / `model.pause()`.
- **`DiscoverVideoEpisodeModel`** — owns the preview `AVPlayer`: sets it up on load, enforces a `maxPreviewTime` cap via a periodic time observer, and fades the audio out on pause via a step timer. Preview audio is muted when something else is already playing so it doesn't talk over the user's current playback.

### Why
The Discover video row previously showed only static thumbnails; previewing the video on focus makes the row feel alive and helps users decide what to play, matching the richer Discover experience.

### Reviewer notes
- The cross-fade is keyed on `model.isPlaying` (not raw focus) so the video only appears once playback has actually started, avoiding a flash of an empty player.
- The fade timer is invalidated on re-focus and in `deinit` so a quick unfocus→refocus can't let a stale fade pause a freshly-started preview.
- `VideoPlayer` (AVKit) tears its controller down immediately on removal, so the video→thumbnail direction may still snap on some hardware even with the transition in place — worth a look on-device.

## To test
1. Open the TV app and go to the Discover tab.
2. Scroll to a video episodes row.
3. Move focus onto a video card — the preview should fade in and start playing (muted if you already have audio playing).
4. Move focus away — audio should fade out and the card should cross-fade back to the thumbnail.
5. Leave a card focused past the preview time limit — it should fade out/pause automatically.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Discover/DiscoverVideoEpisodeCell.swift
+++ b/Pocket Casts TV App/UI/Discover/DiscoverVideoEpisodeCell.swift
@@ -34,10 +34,11 @@ struct DiscoverVideoEpisodeCell: View {
         static let cardHeight = CGFloat(402)
         static let cardWidth = CGFloat(716)
         static let fadeDuration: TimeInterval = 0.5
+        static let playDelay: TimeInterval = 2
     }
 
     init(episode: DiscoverEpisode, listId: String? = nil, source: String = "") {
-        _model = State(wrappedValue: DiscoverVideoEpisodeModel(episode: episode, fadeDuration: Layout.fadeDuration))
+        _model = State(wrappedValue: DiscoverVideoEpisodeModel(episode: episode, fadeDuration: Layout.fadeDuration, playDelay: Layout.playDelay))
         self.listId = listId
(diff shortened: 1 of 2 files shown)


kean (approved):
This feels much better!

PR #4562, opened 2026-06-18
[PC TV] Add slower playback speeds down to 0.5x

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-471

The tvOS player's playback speed menu started at 1.0x, so slower speeds were unavailable. This extends the menu down to 0.5x to match iOS. The bounds now come from the shared `SharedConstants.PlaybackEffects` min/max constants instead of being hardcoded, keeping tvOS in sync with iOS. The backing `PlaybackManager` already supported these speeds — only the tvOS menu was capping the lower bound.

Screenshot 2026-06-18 at 2 13 57 PM

**Note**: asking David if we need this much granularity.

## To test

1. On Apple TV, start playing an episode and open the Now Playing screen.
2. Open the playback speed menu.
3. Verify speeds now range from 0.5x to 3.0x in 0.1 increments.
4. Select a speed below 1.0x (e.g. 0.5x) and confirm playback slows accordingly and the toast/checkmark reflects the selection.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Player/NowPlayingView.swift
+++ b/Pocket Casts TV App/UI/Player/NowPlayingView.swift
@@ -133,7 +133,7 @@ private struct NowPlayingPlayerRepresentable: UIViewControllerRepresentable {
     }
 
     private func makePlaybackSpeedMenu() -> UIMenu {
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
:shipit:

PR #4564, opened 2026-06-18
[PC TV] Fix snackbar confirmation copy

kean:
Fixes PCIOS-741

The tvOS episode-action toasts were reusing the iOS button-label strings (imperatives like "Mark as Played", "Remove From Up Next"), so the confirmations read as commands rather than results. This adds dedicated tvOS confirmation strings (matching the existing `tv_player_*` toast style) and wires up a missing confirmation for unarchive.

| Action | Before | After |
|---|---|---|
| Mark as played | "Mark as Played" | "Marked as played" |
| Archive | "Archived" | "Archived" |
| Unarchive | _(no toast)_ | "Unarchived" |
| Remove from Up Next | "Remove From Up Next" | "Removed from Up Next" |
| Play next | "Play next in Up Next" | "Will play next" |
| Play last | "Play last in Up Next" | "Will play last" |

The shared iOS strings are left untouched (they're correct as button labels). Follow/unfollow and play keep no toast — the Follow button flips to "Following" and Play opens the full-screen player, so both already have clear in-place feedback.

## To test

1. On the tvOS app, open a podcast and focus an episode, then open the "…" actions menu.
2. Trigger each action and confirm the toast in the top-right reads as a result:
   - Mark as played → "Marked as played"
   - Archive → "Archived"; then unarchive the same episode → "Unarchived"
   - Play Next → "Will play next"; Play Last → "Will play last"
3. In the Up Next tab, remove an episode → "Removed from Up Next".

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Podcasts/EpisodeRowViewModel.swift
+++ b/Pocket Casts TV App/UI/Podcasts/EpisodeRowViewModel.swift
@@ -92,7 +92,7 @@ class EpisodeRowViewModel: Identifiable {
 
     func markAsPlayed() {
         EpisodeManager.markAsPlayed(episode: episode, fireNotification: true)
-        ToastManager.shared.show(L10n.markPlayed)
+        ToastManager.shared.show(L10n.tvEpisodeMarkedAsPlayed)
     }
 
     var canArchive: Bool {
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
:shipit:

PR #4567, opened 2026-06-18
Add soft progressive-blur edge to podcast grid

kean:
Under Liquid Glass, the system bottom scroll edge effect samples the scrolling content, so over a grid of colorful podcast artwork it washes out and barely registers — leaving the floating tab bar / mini player hard to read against the artwork. The system removes the blur when the app needs it the most.

https://github.com/user-attachments/assets/04802f16-0bfb-4692-8f7a-4eb464074b54

This adds a dedicated material-blur overlay (`ProgressiveBlurView`) pinned to the bottom of the grid that fades in via a gradient mask, giving a consistently visible soft edge regardless of the content behind it. It's shown for the grid layouts only; list view keeps the system default. The overlay is non-interactive, so it never intercepts touches on the cells underneath.

https://github.com/user-attachments/assets/5ff9acf0-9867-4b7f-ab21-4c3b8161fa4f

This change also adds a thin layer of color under the mini player itself because, as it's further from the bottom, it benefits less from the progressive blur.

## To test

1. Open the **Podcasts** tab with several subscriptions, in **Large grid** or **Small grid** layout.
2. Scroll the grid and confirm the artwork fades out softly behind the tab bar / mini player (instead of clashing with it), in both light and dark themes.
3. Switch to **List** layout and confirm the custom blur is not shown.

## Screenshots

**Before / After**

Screenshot 2026-06-18 at 7 11 54 PM

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix the Filters search header popping back when few episodes match [#4527](https://github.com/Automattic/pocket-casts-ios/pull/4527)
 - Drop iOS 16 and watchOS 9 [#4521](https://github.com/Automattic/pocket-casts-ios/issues/4521)
 - Fix incorrect section heading when opening an episode on the Apple Watch [#4528](https://github.com/Automattic/pocket-casts-ios/pull/4528)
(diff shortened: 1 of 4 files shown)


SergioEstevao (approved):
Tested this on all themes and legibility is looking much better on the mini-player.
:shipit:

PR #4595, opened 2026-06-23
Fix build on Xcode 27 beta 2

kean:
| 📘 Part of: # |
|:---:|

Reorders property initialization in `HorizontalCarousel.init`. `visibleIndex` is a `@State` property, so assigning it via `self.visibleIndex = …` goes through the wrapped-value setter, which references `self` and therefore requires all stored properties to be initialized first. The assignment previously ran before `items` and `content` were set; it's now moved to the end of `init`.

## To test

1. Open any screen that uses `HorizontalCarousel` (e.g. the onboarding/upsell carousels).
2. Confirm it builds and the carousel renders, swipes between pages, and reports the correct current index.

## Checklist


SergioEstevao (approved):
Working correctly on Xcode 26 and 27. :shipit:

PR #4597, opened 2026-06-23
Improve color generation

kean:
Updates the theme color generator (`scripts/themes/generate_themes.rb`) to emit Xcode `#colorLiteral` swatches with `// #HEX (r,g,b)` comments instead of plain `UIColor(hex:)`, so colors are previewable inline in `ThemeColor.swift`. Behavior-identical – `#colorLiteral` resolves to the same `UIColor(red:green:blue:alpha:)`.

The new implementation is a little bit faster as it doesn't need to parse string literals, and it should slightly reduce the binary size.

## To test

1. Build the app (a build phase runs the generator).
2. Open `podcasts/ThemeColor.swift` and confirm each color shows an inline swatch + `// #HEX (r,g,b)` comment.
3. Run and switch themes to confirm colors render exactly as before.

### Before
Screenshot 2026-06-23 at 3 33 16 PM

### After
Screenshot 2026-06-23 at 3 33 02 PM

## Checklist

### DIFF
--- a/scripts/themes/generate_themes.rb
+++ b/scripts/themes/generate_themes.rb
@@ -13,85 +13,109 @@ class String
   end
 end
 
-def write_theme_value(hex_val, opacity, token_name, file_path, theme_name)
-  if token_name.start_with?('filterU') || token_name.start_with?('filterI') || token_name.start_with?('filterT')
-    str = ''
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
Fantastic update!
:shipit:

PR #4602, opened 2026-06-24
[PC TV] Fix focus clipping in episode row with actions

kean:
| 📘 Part of: # |
|:---:|

On tvOS, `EpisodeRowWithActions` applied an extra `clipShape(RoundedRectangle(cornerRadius: 12))` to the whole row (play button **and** the ellipsis "More" button) whenever the row was focused. That outer clip cut off the focus highlight/scale on the row's contents. The inner play button already clips itself, so the focus-time clip was both redundant and visually wrong. This removes it.

It also made context menus appear more reliable on long-press – the `if` was messing with it.

## To test

1. Build and run the tvOS app (`Pocket Casts TV App`).
2. Open a podcast and scroll down to the episode list.
3. Move focus across the episode rows.
4. Verify the focused row (and its ellipsis "More" button) renders its focus highlight fully, without being clipped at the rounded corners.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Podcasts/EpisodeRow.swift
+++ b/Pocket Casts TV App/UI/Podcasts/EpisodeRow.swift
@@ -172,9 +172,6 @@ struct EpisodeRowWithActions: View {
         .contextMenu {
             EpisodeActionButtons(model: model, context: context, isShowingShowNotes: $isShowingShowNotes)
         }
-        .if(isFocused) { content in
-            content.clipShape(RoundedRectangle(cornerRadius: 12))
-        }
         .animation(.easeInOut(duration: 0.2), value: shouldShowMoreButton)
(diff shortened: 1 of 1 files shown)


Pocket Casts TV App/UI/Podcasts/EpisodeRow.swift
```diff
@@ -172,9 +172,6 @@ struct EpisodeRowWithActions: View {
         .contextMenu {
             EpisodeActionButtons(model: model, context: context, isShowingShowNotes: $isShowingShowNotes)
         }
-        .if(isFocused) { content in
```
SergioEstevao, line 175: .if strikes again...

SergioEstevao (approved):
Working better! :shipit:

PR #4603, opened 2026-06-24
[PC TV] Track signInShown from entry buttons instead of onAppear

kean:
On Apple TV the `.signInShown` analytics event was firing unreliably and often multiple times for a single visit. It was tracked from `SignInView`'s `.onAppear`, but the view is presented inside `navigationDestination` / `fullScreenCover` content builders. Those builders get re-invoked as their parent re-evaluates (the QR pairing flow constantly mutates state), so `SignInView` was repeatedly torn down and rebuilt — its `@State` reset each time, defeating an "only track once" guard, and `.onAppear` re-fired on each rebuild.

Rather than fight the view-identity churn, this moves the event to the two buttons that actually lead to the sign-in screen, so it fires exactly once per tap:

- Welcome screen → "Log In" link
- Profile menu → "Log In" item (signed-out)

The `.onAppear` tracking (and its now-unused guard) is removed.

## To test

1. Launch the TV app signed out.
2. On the Welcome screen, tap **Log In** → verify `signInShown` fires exactly once.
3. Go back, then start browsing without an account; open the profile menu and tap **Log In** → verify `signInShown` fires exactly once.
4. On the sign-in screen, switch between the QR and Manual tabs and let the QR code load → verify no additional `signInShown` events fire.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Auth/SignInView.swift
+++ b/Pocket Casts TV App/UI/Auth/SignInView.swift
@@ -118,9 +118,6 @@ struct SignInView: View {
                 break
             }
         }
-        .onAppear {
-            Analytics.track(.signInShown)
-        }
         .background(Color.pcBackgroundBase)
     }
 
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

SergioEstevao (approved):
:shipit:

PR #4610, opened 2026-06-25
[PC TV] Add empty state for folder detail view

SergioEstevao:
| 📘 Part of: # |
|:---:|

Simulator Screenshot - Apple TV 4K (3rd generation) - 2026-06-25 at 15 11 43

## Summary

Adds an empty state to the TV app's folder detail view. When a folder contains no podcasts, the view now shows a `ContentUnavailableView` directing the user to edit the folder in the mobile app, with an OK button that dismisses the view. The folder detail view model gains a `loading / ready / empty` state that drives the view, and the toolbar/tab-bar presentation is fixed while a detail is shown.

This PR also includes a small follow-up cleanup: the new `tv_folder_empty_message` string now has the required translator comment, and a typo ("Edit you folder" → "Edit your folder") was corrected.

## To test

1. On the TV app, open the Podcasts tab.
2. Open a folder that has no podcasts in it.
3. Verify the empty state appears with the message and an OK button, and that tapping OK dismisses the folder view.
4. Open a folder that has podcasts and verify the grid still displays as before.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Podcasts/FolderDetailView.swift
+++ b/Pocket Casts TV App/UI/Podcasts/FolderDetailView.swift
@@ -5,6 +5,10 @@ struct FolderDetailView: View {
 
     @State var model: FolderDetailViewModel
 
+    @Environment(\.dismiss) var dismiss
+
+    @Environment(MainTabViewModel.self) var tabRouter: MainTabViewModel
+
     init(folder: Folder) {
         self.model = FolderDetailViewModel(folder: folder)
     }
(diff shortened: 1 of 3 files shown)


Pocket Casts TV App/UI/Podcasts/FolderDetailView.swift
```diff
@@ -19,6 +23,30 @@ struct FolderDetailView: View {
...
+                case .ready:
+                    podcastList
+            }
+        }
+        .task {
+            model.load()
+        }
+        .toolbar(.hidden, for: .tabBar)
+        .onAppear {
+            tabRouter.isShowingDetail = true
```
kean, line 41: What does it do? Is it necessary?

kean (approved):
Looks good, left one question.

Pocket Casts TV App/UI/Podcasts/FolderDetailView.swift
SergioEstevao: It's to hide the profile and pc logo. We use it on other non-root level screens. I think we may need to implement a modifier that does in a more streamlined way.

PR #4616, opened 2026-06-26
Show episode artwork on episode details page

kean:
| 📘 Part of: #482 |
|:---:|

Fixes PCIOS-482

When the "episode artwork" setting is enabled, the episode details page showed the podcast artwork instead of the episode's own artwork, even though the player shows episode artwork. This displays the episode artwork there too (matching the web app).

## To test

1. Enable Settings → Appearance → "Use embedded artwork".
2. Open an episode that has its own artwork (e.g. podcast "La Bonne Auberge", episode "Cyberpunk - Choomba Night").
3. The details page shows a gray placeholder briefly, then the episode artwork (not the podcast artwork).
4. Open an episode without dedicated artwork → it falls back to the podcast artwork.
5. Disable the setting and reopen an episode → the podcast artwork is shown, as before.

https://github.com/user-attachments/assets/3a05aa28-1eba-47b1-940a-767cb9f46b10

## Checklist


SergioEstevao (approved):
Working great!
Do you wonder if we could also use this on the TV app?

kean:
> Do you wonder if we could also use this on the TV app?

That'd be ideal.

PR #4619, opened 2026-06-26
Fix crash from concurrent episode artwork loading

kean:
## Summary

`EpisodeArtwork` mutated its in-progress-loads dictionary from both the calling thread and the load task's background executor — a data race that could crash. It's now a `@MainActor final class`, so that state is only touched on the main actor. Also switches embedded-artwork extraction to the async `load(.commonMetadata)` / `load(.dataValue)` APIs (the sync ones were deprecated in iOS 16).

Fixes a crash surfaced in 8.15 TestFlight crash reports - it's a pre-existing production issue, not beta-specific, but probably worth fixing earlier than later.

## To test

1. Settings → Appearance → enable **Use Embedded Artwork**.
2. Stream/play an episode with embedded (or show-notes) artwork; confirm it appears and nothing crashes.
3. Queue several downloads and rapidly switch episodes to exercise concurrent artwork loads.

## Checklist


podcasts/DownloadManager.swift
```diff
@@ -285,7 +284,11 @@ class DownloadManager: NSObject, FilePathProtocol {
 
         // try and cache the episode embedded artwork
         #if !os(watchOS)
-        episodeArtwork.loadEmbeddedImage(asset: nil, podcastUuid: episode.parentIdentifier(), episodeUuid: episode.uuid)
+        let artworkPodcastUuid = episode.parentIdentifier()
+        let artworkEpisodeUuid = episode.uuid
+        Task { @MainActor in
+            episodeArtwork.loadEmbeddedImage(asset: nil, podcastUuid: artworkPodcastUuid, episodeUuid: artworkEpisodeUuid)
```
kean, line 290: The manages heavily used from UI need to be synchronized on MainActor as well, but it is not in the scope of the PR.

SergioEstevao (approved):
:shipit:

PR #4622, opened 2026-06-26
Fix watch crash creating background sync tasks on an invalidated URLSession

kean:
Fixes a watchOS crash when starting a background refresh. `URLSession.downloadTask(with:)` can throw an Objective-C `NSGenericException` ("Task created in a session that has been invalidated") when watchOS invalidates a freshly created background session mid-setup — typically when the app is being suspended as a WatchConnectivity update triggers the refresh. The first download task is created fine, then the session flips to invalid before the next one. Because the exception comes from Foundation, Swift's `do/catch` can't catch it.

- Wrap the two watch entry points (`ExtensionDelegate` and `WatchSyncManager`) in `SJCommonUtils.catchException`, logging and skipping the refresh instead of crashing.
- Advance the 5-minute refresh throttle only after the tasks are successfully resumed, so a failed attempt retries on the next refresh instead of being suppressed for 5 minutes.
- Invalidate the session on the early-return path so it isn't leaked.

## To test

1. Build and run the Watch App target (staging) on a paired Apple Watch signed in with a Plus account.
2. Confirm normal sync still works: change Up Next / playback state on the phone and verify the watch reflects it after a refresh.
3. While the phone is pushing updates, background the watch app (return to the watch face) and repeat a few times — the app should no longer crash with "Task created in a session that has been invalidated".
4. In the watch logs, when the system invalidates a session you should see "Skipped background refresh, URLSession was invalidated" rather than a crash, and the next scheduled refresh should still run.

## Checklist


SergioEstevao (approved):
Changes look solid, I tested this on phone and watch and all looks is working correctly.
:shipit:

kean:
Thanks for testing! I'm not super happy that it now basically ignores the errors, but it's better than silently crashing in the background.

PR #4623, opened 2026-06-26
Fix crash scrolling the podcast page after it's popped

kean:
Fixes a crash on the podcast page reported via Sentry.

**Note:** I'm not sure why `assertionFailure` is seemingly firing in production, but it's a separate issue.

A scroll view can keep firing `scrollViewDidScroll` during momentum deceleration *after* its view controller is popped off the navigation stack. By that point `navigationController` is `nil`, so `setTransparentNavBarScrolled` has no bar to style. The `guard` already handled this gracefully with an early `return`, but the `assertionFailure("navigationBar is missing")` on that path crashed Debug/TestFlight builds for what is actually a normal lifecycle race.

This removes the bogus assertion (keeping the graceful `return`) and documents why `nil` is expected here.

## To test

This is hard to reproduce reliably (it's a deceleration-vs-pop race), but to exercise the path:

1. Open a podcast page.
2. Scroll down so the nav bar blurs (past the header).
3. Flick to start a fast scroll, then immediately tap the back button mid-deceleration to pop the screen.
4. Verify there's no crash (on `main`/Debug this could hit the assertion).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 8.15
 -----
+- Fix a rare crash when leaving a podcast page while it's still scrolling [#4623](https://github.com/Automattic/pocket-casts-ios/pull/4623)
 - Fix "Download All" on a season not warning before using cellular data [#4523](https://github.com/Automattic/pocket-casts-ios/pull/4523)
 - Fix Share swipe action not dismissing on episode cells [#4522](https://github.com/Automattic/pocket-casts-ios/pull/4522)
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/131">8.15 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

SergioEstevao (approved):
:shipit:

PR #4624, opened 2026-06-26
Fix autoplay skipping unfollowed-podcast episodes in manual playlists

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-645

Autoplay built its "play next" list with the legacy playlist query, which ignores `EpisodeFilter.manual` and filters out episodes from podcasts you don't follow. In a manual playlist this dropped unfollowed-podcast episodes (so playback just stopped) or returned unrelated episodes from default smart-filter rules. Autoplay now uses the same query the playlist screen uses, so it plays exactly the list you see.

## To test

1. Enable Autoplay in Profile > Settings > General.
2. Create a manual playlist with episodes from podcasts you follow **and** podcasts you don't follow.
3. Make sure your Up Next queue is empty.
4. Play episodes from the manual playlist.
5. When an episode from an unfollowed podcast would be next, confirm Autoplay continues to that episode (in playlist order) instead of stopping.

## Checklist


SergioEstevao (approved):
Tested with manual playlists filled with followed and non-followed podcasts and they all played in order using auto-play.

PR #4639, opened 2026-06-29
Fix auto-add-to-Up-Next setting overwriting every podcast

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-800.

`PodcastDataManager.updateAutoAddToUpNext(to:for:)` is meant to set the "Add to Up Next" behaviour for a specific set of podcasts. A SQL typo (`AND` instead of `WHERE`) left the `UPDATE` with no `WHERE` clause, so SQLite assigned the boolean expression `? AND uuid IN (...)` to **every** podcast row:

- Podcasts not in the selection were reset to **Off** (silent data loss).
- Targeted podcasts set to **Add to top** (`addFirst`) were clamped to **Add to bottom** (`addLast`).

This fixes the statement to use `WHERE uuid IN (...)` so the update is scoped and the value is stored verbatim. It also rewrites the malformed, flag-gated `json_patch` settings statement (a no-op `SELECT` with invalid `WHERE`/`FROM` ordering) to the `json_set` pattern used by the rest of the file.

## To test

Covered by unit tests in `PodcastDataManagerTests` (run both SQL and GRDB paths):
1. `testUpdateAutoAddToUpNextOnlyUpdatesTargetedPodcasts` — a scoped update must not touch other podcasts.
2. `testUpdateAutoAddToUpNextStoresExactValue` — `addFirst` round-trips exactly.

Manual:
1. Subscribe to at least two podcasts; set Podcast A's Up Next behaviour to *Add to bottom*.
2. Change Podcast B's Up Next behaviour (Podcast settings → Up Next) to *Add to top*.
3. Confirm Podcast A is **unchanged** and Podcast B is **Add to top** (not downgraded to bottom).

## Checklist


SergioEstevao (approved):
Great fix! :shipit:

PR #4642, opened 2026-06-29
Fix force-unwrap crash when episode deleted mid-download-notification

kean:
Fixes a force-unwrap crash in `PlayerCell`. The cell's `episode` was an implicitly-unwrapped optional, but the download/status notification handlers reassign it from `DataManager.findBaseEpisode(uuid:)` (which returns `BaseEpisode?`). If the episode was deleted between the notification being posted and the handler running, `updateDownloadStatus()` force-unwrapped `nil` and crashed.

This converts the property to a true optional and adds the missing `nil` guard in `updateDownloadStatus()`, matching the sibling `UpNextNowPlayingCell`.

## To test

1. This is a hard-to-reproduce race; verify there's no regression in normal use.
2. Open a filter or list using `PlayerCell` (e.g. a Filter screen).
3. Start downloading an episode and confirm the downloading/downloaded indicators and info text still update correctly.
4. Star/unstar an episode in the list and confirm the star indicator updates.

## Checklist


SergioEstevao (approved):
Good catch!

PR #4644, opened 2026-06-29
Remove PocketCastsDependencyInjection module and @Dependency wrapper

kean:
Removes the `PocketCastsDependencyInjection` module and its `@Dependency` property wrapper. The container kept dependencies in mutable `static` state that triggered a concurrency warning, and the wrapper was only used in one place (playlist metadata loading). Replaced with plain `.shared` singletons.

Going forward, let's use SwiftUI's environment for dependency injection instead.

## To test

1. Open the Filters/Playlists tab — playlist cells still show episode counts and artwork.
2. Open a playlist's detail screen — episodes load correctly.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>

.</td>
    </tr>
  </tbody>
</table>

</p>

SergioEstevao (approved):
Good removal, this library never got traction in terms of use, and now with SwiftUI we can do this in a different way.

PR #4645, opened 2026-06-30
Fix force-unwrapping a failable audio buffer during playback

kean:
| 📘 Part of: # |
|:---:|

The audio read thread created an `AVAudioPCMBuffer` with its failable initializer and then force-unwrapped it (and optional-chained it) throughout `readFromFile()`. If allocation returned `nil` (memory pressure / invalid format), the force-unwrap trapped — crashing the read thread and bypassing the existing `NSException` + `do/catch` recovery.

This guards the allocation and routes a `nil` result to the existing `readErrorOccurred` error path (matching the nearby `channelCount == 0` recovery), so a failed allocation surfaces a read error instead of crashing. With the buffer now non-optional, all downstream `!`/`?` uses are removed.

## To test

This fixes a rare allocation failure that's hard to reproduce directly, so verify there's no regression in normal playback:

1. Play an episode and confirm audio plays normally.
2. Enable Trim Silence and confirm playback still works.
3. Enable a voice effect (VoiceBoost) and confirm playback still works.
4. Seek within an episode and confirm playback resumes correctly.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 8.16
 -----
+- Fix a rare crash during playback when an audio buffer fails to allocate [#4645](https://github.com/Automattic/pocket-casts-ios/pull/4645)
 - Add a "Remove from Up Next" multi-select action outside the Up Next screen [#4606](https://github.com/Automattic/pocket-casts-ios/pull/4606)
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
Good catch!
Testing in some different episodes and all worked correctly, when switching around different effects combinations.

PR #4648, opened 2026-06-30
Fix episode list jumping when a download finishes

kean:
Fixes PCIOS-804

The Filter detail (and other `EpisodeCell` lists) jumped when a background download finished, because the view controllers reloaded the whole table in response to `episodeDownloaded` / `episodeDownloadStatusChanged`. `EpisodeCell` already observes those notifications and updates itself, so the table reload was redundant. Removed those observers from the Filter detail, Starred, and Listening History screens.

**Note:** `PodcastViewController` (groups episodes by download status) and `DownloadsViewController` (download status *is* its membership) are intentionally left as-is. This will require more work, not sure how to tackle it.

## To test

1. Open a Filter (or the Starred / Listening History screen) with several episodes.
2. Start a download on an episode that isn't at the top, and scroll so it's visible.
3. While it downloads (and when it finishes), confirm the list no longer jumps/scrolls and the cell updates its download state in place.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 8.16
 -----
+- Fix episode lists jumping when a background download finishes [#4648](https://github.com/Automattic/pocket-casts-ios/pull/4648)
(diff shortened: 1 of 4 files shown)


SergioEstevao (approved):
Working correctly

PR #4650, opened 2026-06-30
Add episode sorting to the tvOS podcast screen

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-773

Adds episode sorting to the podcast screen on tvOS, matching iOS. A circular "More" (⋯) menu now sits next to the existing archived filter in the episode list header, offering the standard sort orders (Newest/Oldest, Title A–Z/Z–A, Shortest/Longest, Serial). The choice is persisted on the synced podcast settings, so it stays consistent with iOS and the rest of the app.

**Note**: I decided to not add it to "Up Next" in the scope of this PR because there was no clear way to add it (it is in a tab-based view)

## To test

1. On an Apple TV (or tvOS simulator), open a podcast with several episodes.
2. Focus the "All episodes" header and open the circular **More** (⋯) menu next to the archived filter.
3. Pick a sort order under **Sort By** — confirm the episode list reorders immediately and a checkmark marks the active order.
4. Leave and reopen the podcast — confirm the sort order is remembered.
5. Open the same podcast on iOS — confirm the sort order matches.

Screenshot 2026-06-30 at 4 34 37 PM

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Show episode artwork on the episode details page [#4616](https://github.com/Automattic/pocket-casts-ios/pull/4616)
 - Add support for generated chapters on episodes [#4630](https://github.com/Automattic/pocket-casts-ios/pull/4630)
 - Fix Autoplay stopping instead of continuing through manual playlists that contain episodes from podcasts you don't follow [#4624](https://github.com/Automattic/pocket-casts-ios/pull/4624)
+- Add episode sorting to the podcast screen on tvOS [#4650](https://github.com/Automattic/pocket-casts-ios/pull/4650)
 
 8.15
 -----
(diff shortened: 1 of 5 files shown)


SergioEstevao (approved):
Working correctly and code is clean.
@david-gonzalez-a8c  do you approve design wise?

david-gonzalez-a8c:
Nice job! I just have two notes:

image

It would be great if can make the margins between grouped buttons consistent. We can follow this group of buttons in this same view as an example:

image

I don't think the ellipsis icon makes a lot of sense for this case, I'd expect a group of different options under this menu. I suggest we either use a label ("Sort by") or a similar icon to the one in the mobile app for this option:

IMG_0518

kean:
Great suggestions; addressed both – much better.

Screenshot 2026-07-01 at 4 40 35 PM

PR #4651, opened 2026-06-30
Disable activateAudioSessionInBackground on tvOS

kean:
Fixes PCIOS-778

Defaults the `activateAudioSessionInBackground` feature flag to `false` on tvOS. On tvOS the audio session is now activated synchronously on the main thread (the original behavior) instead of being dispatched to a background queue. iOS, iPadOS, and watchOS are unaffected.

## To test

1. Build and run the tvOS target.
2. Play an episode and confirm playback starts normally.
3. Pause/resume and skip between episodes — audio session activation should behave as before with no regressions.
4. Confirm iOS/iPadOS behavior is unchanged (flag still defaults to `true` there).

## Checklist


SergioEstevao (approved):
Working correctly! Was this causing an crash on tvOS because of MainActor isolation on tvOS?

kean:
> Was this causing an crash on tvOS because of MainActor isolation on tvOS?

It was due to `AVPlayer` observations used by the native player.

PR #4653, opened 2026-06-30
Fix separator alignment and list style on the Discover category list

kean:
Fixes a few visual issues on the Discover category list (e.g. the "Popular" / charts list):

- Align the row separator with the artwork in ranked rows, so it no longer runs under the ranking number.
- Hide the separator on the featured cell so it renders edge-to-edge.
- Fix the list background color by making the table a `ThemeableTable` with the `primaryUi02` style.

## To test

1. Open the **Discover** tab.
2. Open a category list that shows ranking numbers (e.g. a "Popular" / charts list).
3. Confirm the row separators start at the podcast artwork, not under the ranking number.
4. Confirm the featured cell at the top has no separator and the list background uses the correct theme color (check in both light and dark themes).

Before / After

Screenshot 2026-06-30 at 7 31 30 PM Screenshot 2026-06-30 at 7 44 09 PM

Before / After

Screenshot 2026-06-30 at 7 31 39 PM Screenshot 2026-06-30 at 7 44 15 PM

## Checklist


SergioEstevao (approved):
Looking great! :shipit:

PR #4655, opened 2026-07-01
PCIOS-805: Fix transcript controls layout shift causing mis-taps

pocketcasts:
Resolves https://linear.app/a8c/issue/PCIOS-805/ios-transcript-controls-shift-position-as-the-transcript-loads-causing

## Summary
Fixes a layout shift in the transcript controls bar where buttons would reposition as the transcript loaded, causing taps intended for Share to land on Play.

## Changes
- Always add shareButton and playButton to the stack view upfront (instead of conditionally via if guards), using .isHidden for static feature flag / context gating
- Replace searchButton.isHidden toggling in loading/loaded states with alpha + isUserInteractionEnabled toggling — this keeps the button in the UIStackView layout (preserving its space) so the flexible spacer does not expand/contract and shift the other buttons
- Apply the same alpha-based visibility to all control buttons (share, play, search) during the loading transition so they all appear together

## Verification
- **Lint**: PASS
- **Tests**: N/A — pre-existing CarPlay build errors unrelated to this change
- **Diff review**: PASS — single file changed, minimal diff, no unintended changes
- **Secret scan**: PASS

## Confidence
**HIGH** — The root cause is clearly identified and the fix uses standard UIKit patterns.

---
Triage complexity: simple | [Linear issue](https://linear.app/a8c/issue/PCIOS-805/ios-transcript-controls-shift-position-as-the-transcript-loads-causing)

### DIFF
--- a/podcasts/TranscriptViewController.swift
+++ b/podcasts/TranscriptViewController.swift
@@ -277,13 +277,11 @@ class TranscriptViewController: PlayerItemViewController, AnalyticsSourceProvide
         stackView.addArrangedSubview(closeButton)
         stackView.addArrangedSubview(UIView())
 
-        if FeatureFlag.shareTranscripts.enabled {
-            stackView.addArrangedSubview(shareButton)
-        }
+        shareButton.isHidden = !FeatureFlag.shareTranscripts.enabled
+        stackView.addArrangedSubview(shareButton)
 
-        if showFromEpisode {
-            stackView.addArrangedSubview(playButton)
-        }
+        playButton.isHidden = !showFromEpisode
+        stackView.addArrangedSubview(playButton)
 
         stackView.addArrangedSubview(searchButton)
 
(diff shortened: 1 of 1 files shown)


SergioEstevao (changes requested):
While it solves the issue of moving elements, we have now a bit of a layout issue when the transcript load fails.

As you can see in the screenshot the play button now hangs in the toolbar. instead of being on the right.

Do you think this is better  than the moving buttons? If yes I can approve it.

Image

joashrajin:
Good catch @SergioEstevao. I attempted a fix in b7affc0: on a load error, the transcript-only controls (share/search) are now fully collapsed, so the play button sits flush at the trailing edge. Retrying restores the reserved-space layout so the original no-shift behaviour during loading is unchanged.

SergioEstevao (approved):
Thanks for addressing the issue!
All looking good now :shipit:

pocketcasts:
Version `8.17` has now entered code-freeze, so the milestone of this PR has been updated to `8.18`.

PR #4660, opened 2026-07-01
Exclude .claude harness directory from SwiftLint

kean:
SwiftLint lints the entire repo root, so it picks up file copies inside Claude Code git worktrees (`.claude/worktrees/`). The existing `fastlane` exclusion uses relative-path matching and doesn't cover the nested `fastlane` copy inside those worktrees, producing spurious warnings.

This adds `.claude` to the `excluded` list so transient worktrees and other harness files never pollute lint output. No source code is affected — the actual codebase already lints clean.

## To test

1. Ensure a Claude Code worktree exists under `.claude/worktrees/` (or any `.swift` file that would violate a rule).
2. Run `make lint`.
3. Confirm there are no warnings and no files under `.claude/` are scanned.

## Checklist

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -3,6 +3,8 @@ swiftlint_version: 0.63.2
 # Project configuration
 #
 excluded:
+  # Claude Code harness directory, including transient git worktrees
+  - .claude
   - BuildTools/.build
   - DerivedData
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
:shipit:

PR #4661, opened 2026-07-01
tvOS: Query new releases from subscribed podcasts on Home

SergioEstevao:
Fixes PCIOS-

The tvOS Home screen's "New Releases" row previously surfaced only the single latest episode from each of the first 8 subscribed podcasts. That approach missed newer episodes from podcasts beyond the first 8, and could show already-listened or stale episodes.

This replaces that logic with a dedicated `DataManager.findNewReleaseEpisodes(limit:)` query that returns, across **all** subscribed podcasts:

- episodes **not listened to** (`playingStatus = notPlayed` — excludes in-progress and completed)
- published within the **last two weeks** (matching the app's built-in New Releases smart filter window of `24 * 14` hours)
- excluding deleted and archived episodes
- sorted by **release date**, newest first
- capped at **12** episodes

The `HomeViewModel` now calls this query directly instead of looping over podcasts and fetching each one's latest episode.

## To test

1. On tvOS, subscribe to several podcasts (more than 8, with recent episodes).
2. Open the Home screen.
3. Confirm the "New Releases" row shows up to 12 unplayed episodes from across your subscriptions, most recent first.
4. Start playing one of the episodes (making it in-progress), then reload Home — confirm it no longer appears in New Releases.
5. Confirm episodes older than two weeks do not appear.

## Checklist


kean (approved):
Nice one.

PR #4662, opened 2026-07-01
Fix scrolled transparent nav bar ignoring the in-app theme on iOS 18

kean:
Fixes PCIOS-799.

On iOS 18, scrolling in smart playlists (and the podcast page) turned the nav bar the system light/dark color instead of the in-app theme — e.g. a light band under a dark theme when the system was set to Light. iOS 26 was unaffected.

`PCViewController.setTransparentNavBarScrolled(_:)` built its scrolled bar with `configureWithOpaqueBackground()` but never set a color, so iOS derived one from the trait collection — which follows the system appearance pre-Liquid Glass. Fixed by pinning it to the theme color, mirroring `setupNavBar`. Centralized, so it covers both the playlist and podcast pages.

Screenshot 2026-07-01 at 1 41 37 PM

## To test

1. On iOS 18.x, set **Settings > Display & Brightness** to **Light**.
2. In Pocket Casts, pick a dark theme under **Profile > Settings > Appearance**.
3. Open a smart playlist with enough episodes to scroll and scroll down.
4. Verify the header stays dark (matching the theme) instead of turning light. Repeat on a podcast page.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 8.16
 -----
+- Fix the smart playlist and podcast page header showing the system light/dark appearance instead of the in-app theme when scrolling on iOS 18 [#4662](https://github.com/Automattic/pocket-casts-ios/pull/4662)
 - Fix episode lists jumping when a background download finishes [#4648](https://github.com/Automattic/pocket-casts-ios/pull/4648)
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
:shipit:

PR #4663, opened 2026-07-01
Fix UserEpisode equality casting to the wrong type

kean:
`UserEpisode.isEqual(_:)` cast the object to `Episode`, but `UserEpisode` and `Episode` are siblings (neither subclasses the other), so the cast always failed and two equal `UserEpisode`s never compared equal. Fixed by casting to `UserEpisode`, mirroring `Episode.isEqual`.

**No user-facing impact:** latent since the initial commit — every episode comparison in the app goes through `uuid` strings, so no current code path hit the broken `isEqual`. This removes a landmine for future code and adds regression tests.

## To test

Covered by unit tests; no manual QA needed.

1. Run `make test_staging ONLY_TESTING=PocketCastsDataModelTests/EpisodeEqualityTests`.
2. Confirm all cases pass.

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary. (No user-facing impact — no CHANGELOG entry.)

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Public/Model/UserEpisode.swift
+++ b/Modules/Sources/PocketCastsDataModel/Public/Model/UserEpisode.swift
@@ -160,7 +160,7 @@ public class UserEpisode: NSObject, BaseEpisode {
     }
 
     override public func isEqual(_ object: Any?) -> Bool {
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
Good catch

PR #4664, opened 2026-07-01
Fix hasCustomImage decoding in UserEpisode.populateFromMap

kean:
`UserEpisode.populateFromMap(_:)` decoded `hasCustomImage` from the literal string `"hasCustomImage"` instead of `episodeMap["hasCustomImage"]`. `Bool("hasCustomImage")` is always `nil`, so the flag always decoded to `false`.

This map is only used to sync user episodes (uploaded Files) from the phone to the Apple Watch. **No user-facing symptom:** the Watch's artwork path (`UserEpisode.urlForImage`) keys off `imageColor`/`imageUrl` — both of which round-trip correctly — not `hasCustomImage`, and the decoded flag isn't read anywhere else on the Watch. Fixing anyway since it's a clear correctness bug and could bite a future consumer of the decoded object.

Added a unit test covering the encode/decode round-trip.

## To test

1. `make test_staging ONLY_TESTING=PocketCastsDataModelTests/UserEpisodeEncodingTests`
2. Test passes. Reverting the one-line fix in `UserEpisode+Encoding.swift` makes it fail.

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary. (Not needed — no user-facing behavior change.)
- [x] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. (No analytics changes.)

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Public/Model/UserEpisode+Encoding.swift
+++ b/Modules/Sources/PocketCastsDataModel/Public/Model/UserEpisode+Encoding.swift
@@ -49,7 +49,7 @@ public extension UserEpisode {
         uploadStatus = decodeInt32FromString(value: episodeMap["uploadStatus"])
         imageUrl = episodeMap["imageUrl"]
         imageColor = decodeInt32FromString(value: episodeMap["imageColor"])
-        hasCustomImage = decodeBoolFromString(value: "hasCustomImage")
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
:shipit:

PR #4666, opened 2026-07-01
Align generated chapters to the played audio timeline via fingerprint timing

SergioEstevao:
Fixes [POC-654](https://linear.app/a8c/issue/POC-654/use-fingerprint-to-adjust-timings)
|:---:|

## Summary

Generated chapters are timed against the **reference** (clean) audio the transcript was generated from. When the played file contains dynamic ads, the real audio positions shift, so the generated chapter markers drift out of sync with what's actually playing.

This PR re-maps each generated chapter's start time from the reference timeline onto the **played file's timeline** using `FingerprintTimingManager` (the same mapping already used for synced-transcript seeking), and recomputes chapter durations from the adjusted start times.

Because the fingerprint mapping is built asynchronously and grows around the current playback position during playback, chapters are re-aligned both at parse time and again whenever the mapping advances — via a new `fingerprintTimingMappingUpdated` notification. Chapters ahead of the covered region keep their reference times until coverage reaches them, then self-correct.

### Notes for reviewers

- `ChapterManager` compiles into targets that do **not** include the `Fingerprint` sources (Watch, App Clip, TV). To avoid breaking those builds, the dependency is decoupled behind the `ChapterReferenceTimeMapping` protocol + a `ChapterReferenceTimeMappingProvider` registry. The provider stays unset in those targets, so the adjustment is a no-op there and generated chapters simply keep their reference times.
- `FingerprintTimingManager` conforms to the protocol (`hasChapterReferenceMapping` is true once `.active`, and the existing `playbackTime(forReferenceTime:)` satisfies the mapping requirement) and registers itself as the provider in the main app.
- The adjustment is gated on `chaptersOrigin == .generated` **and** an active mapping. The active-mapping requirement implicitly covers "has generated transcripts", since the fingerprint reference only exists for generated-transcript episodes.
- No new analytics events were added.

## To test

1. Play an episode that has generated transcripts and generated chapters, where the played file contains dynamic ads.
2. Open the transcript so synced-transcript preparation kicks in (this builds the fingerprint mapping).
3. As playback progresses and the mapping reaches `.active`, confirm the chapter markers/boundaries line up with the real audio content rather than being offset by the inserted ads.
4. Regression: on an episode with embedded/native or Podcast Index chapters, confirm nothing changes.

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Playback/Chapters/ChapterManagerTests.swift
+++ b/PocketCastsTests/Tests/Playback/Chapters/ChapterManagerTests.swift
@@ -85,6 +85,50 @@ class ChapterManagerTests: XCTestCase {
         XCTAssertEqual(nextVisiblePlayableChapter, chapterInfo(startTime: 201, duration: 300, shouldPlay: true))
     }
 
+    /// Generated chapters are timed against the reference (clean) audio; when a
+    /// fingerprint mapping is available their start times should be re-mapped onto
+    /// the played file's timeline (here a flat +30s dynamic-ad shift).
+    func testGeneratedChaptersAreRemappedToPlaybackTimeline() async {
+        let parserMock = PodcastChapterParserMock() // no file chapters -> falls through to generated
+        let showInfoCoordinatorMock = GeneratedChaptersShowInfoCoordinatorMock(generatedChapters: [
+            GeneratedChapter(title: "One", timestamp: "00:00", startTime: 0),
+            GeneratedChapter(title: "Two", timestamp: "01:40", startTime: 100),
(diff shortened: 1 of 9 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:
Version `8.16` has now entered code-freeze, so the milestone of this PR has been updated to `8.17`.

SergioEstevao:
Closed in favour of this PR: https://github.com/Automattic/pocket-casts-ios/pull/4728

PR #4690, opened 2026-07-07
Use Xcode 26.6 and Ruby 3.4.9

mokagio:
See https://linear.app/a8c/issue/AINFRA-2605/use-ruby-349-in-pocket-casts-ios

An overdue tooling update: bump CI to Xcode 26.6 and Ruby 3.4.9, re-bundling with Bundler 4.0.15.
Also forces the Ruby gem platform so native gems compile from source under the new toolchain.


iangmaia:
Considering Xcode bumps might break other things in the project, asking for a ✅ from @Automattic/pocket-casts-ios.

SergioEstevao (approved):
Tested locally and it worked all correctly

PR #4696, opened 2026-07-07
Fix player opening on Bookmarks tab in RTL languages

### LINKED ISSUE #1952: Player don't appear as the first screen when using RtL language
When using RtL the Bookmarks screen is shown, not the player:

https://github.com/user-attachments/assets/0d91161f-8032-459c-a54a-7e95d3d02089

## Steps to reproduce

1. Set Xcode to use RtL Pseudo Language
2. Run the app
3. Tap mini player
4. Bookmarks is shown, not the player

mizadi:
Closes #1952

### Problem
In a right-to-left language, tapping the mini player opens the **Bookmarks** tab instead of **Now Playing**, and the tab-bar highlight desyncs from the visible page.

### Before / After
| Before — opens on Bookmarks | After — opens on Now Playing |
|:---:|:---:|
| pc_rtl_before | pc_rtl_after |

### Cause
The full-screen player is a horizontal paged carousel. Its tabs are laid out with `leading`/`trailing` constraints, but the paging is index-based (`offset == index * width`, reset via `setContentOffset(.zero)`). Under RTL, UIKit mirrors the scroll view, so `.zero` lands on the **last** page (Bookmarks) rather than page 0 (Now Playing).

### Fix
The tabs are a spatial carousel, not text, so force both the content scroll view (`mainScrollView`) and the tab bar (`PlayerTabsView`) to lay out left-to-right in every language. This keeps the existing index math correct and the tab order in sync, without mirroring the carousel. LTR behavior is unchanged (`forceLeftToRight` is a no-op there).

### Testing
1. Set the scheme's App Language to **Right-to-Left Pseudolanguage**.
2. Play any episode, tap the mini player to expand.
3. **Before:** opens on Bookmarks. **After:** opens on Now Playing.
4. Confirmed LTR is unaffected.

### DIFF
--- a/podcasts/PlayerContainerViewController.swift
+++ b/podcasts/PlayerContainerViewController.swift
@@ -23,6 +23,12 @@ class PlayerContainerViewController: SimpleNotificationsViewController, PlayerTa
 
     @IBOutlet var mainScrollView: RegionCancellingScrollView! {
         didSet {
+            // The player tabs are a spatial carousel, not text, so keep the paging
+            // left-to-right in every language. Otherwise UIKit mirrors the scroll
+            // view under RTL and the index-based paging math (offset == index * width)
+            // opens the last tab (Bookmarks) instead of Now Playing. See #1952.
+            mainScrollView.semanticContentAttribute = .forceLeftToRight
+
             // We don't need to handle the scroll view because it is not dismissable in App Clip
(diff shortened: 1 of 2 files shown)


kean (approved):
Nice fix, thank you!

PR #4702, opened 2026-07-07
Fix Download button not working in Search/Discover results

kean:
Fixes PCIOS-

In the new Search UI (Search tab and Discover search, behind `searchImprovements`), tapping the **Download** button on an episode result did nothing and the cell never updated. `SearchResultCellModel`'s `downloadTapped`/`stopDownloadTapped` were empty no-ops and it never observed download notifications.

This wires the cell model up to `PlaybackActionHelper` (fetching the episode into the DB first when needed) and observes download progress/status notifications, mirroring `EpisodeCell`.

## To test

1. Ensure the **Search Improvements** feature flag is enabled.
2. Open **Search** (or **Discover → search**) and search for a term that returns episode results.
3. Set your primary row action to **Download** (Settings → General → Row Action) so rows show the download button.
4. Tap **Download** on an episode result.
5. Verify the button shows progress and ends in the downloaded state.
6. Tap again mid-download and verify it cancels and returns to the Download state.

https://github.com/user-attachments/assets/29bbe588-2bfc-40b6-85ae-43b45a16e4c6

## Checklist

-  [x] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@
 - Fix the Playlists tab showing a blank screen instead of the empty state when there are no playlists [#4794](https://github.com/Automattic/pocket-casts-ios/pull/4794)
 - [tvOS] Improve thumbnail algorithm for videos [#4861](https://github.com/Automattic/pocket-casts-ios/pull/4861)
 - [tvOS] Show correct number of playlist items, and update playlists episode status on changes [#4869](https://github.com/Automattic/pocket-casts-ios/pull/4869)
+- Fix the Download button doing nothing in Search and Discover episode results [#4702](https://github.com/Automattic/pocket-casts-ios/pull/4702)
 
 8.16.1
 ------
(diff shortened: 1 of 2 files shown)


SergioEstevao (changes requested):
Good catch @kean .

But I think the download/retry button should have never showed on the search results.

The designs only have play/pause I think because we reused the episode cell. this ended up showing when changing the row action.

I think the real fix if possible is to ignore that configuration when displaying the cell on search.

kean:
I'm running this configuration for testing purposes that replaces "Play" with "Download".

Screenshot 2026-07-08 at 6 34 15 PM

kean:
> I think the real fix if possible is to ignore that configuration when displaying the cell on search.

@SergioEstevao what is the background for it? I'd probably expect cells to be the same. Shouldn't we just use `EpisodeCell` here? 🤔

pocketcasts:
Version `8.17` has now entered code-freeze, so the milestone of this PR has been updated to `8.18`.

pocketcasts:

pocketcasts:

PR #4709, opened 2026-07-08
Fix stale multi-selected episodes in Up Next when the queue changes

kean:
Fixes stale selections in Up Next multi-select. `upNextChanged()` tried to drop selected episodes that had left the queue, but the guard `index > selectedPlayListEpisodes.count` is unsatisfiable for an `enumerated()` index, so the removal never ran. Externally-removed episodes stayed selected, producing a wrong count badge and letting bulk actions target episodes no longer in the queue.

Replaced the broken loop with a single `removeAll(where:)` that keeps only selections still present in Up Next. With the prune working, the `changedViaSwipeToRemove` bypass is no longer needed and has been removed — swipe-to-remove now flows through `upNextChanged()` and updates the selection correctly.

## To test

1. Add at least 3–4 episodes to Up Next.
2. Open Up Next and enter multi-select (long-press an episode).
3. Select 2–3 episodes and confirm the action bar shows the correct count.
4. Swipe to remove one of the *selected* episodes from Up Next.
5. Verify the removed episode drops out of the selection immediately, the count badge is correct, and bulk actions only affect episodes still in the queue.
6. Repeat, but this time remove a selected episode externally (let it finish playing, or remove it via sync on another device) and confirm the selection prunes the same way.

Here was the behavior **before** the fix:

https://github.com/user-attachments/assets/b1fac9b6-2403-48f2-a902-860e16fa2809

## Checklist


SergioEstevao (approved):
Tested both scenarios and all working correctly. 
:shipit:

PR #4712, opened 2026-07-08
Fix Add to Playlist dropping selections made under an active search

kean:
When adding an episode to a manual playlist, `doneTapped` applied changes by iterating the *displayed* playlist list, which during an active search is only the filtered subset. Any playlist toggled under a prior search term that fell outside the current filter was silently never added or removed.

Fixed by keeping an authoritative `allManualPlaylists` list that is never filtered and applying changes against it, while the table keeps showing the filtered `manualPlaylists`.

## To test

1. Have at least two manual playlists (e.g. "Workout" and "Chill") that don't contain the episode.
2. Open an episode → **Add to Playlist**.
3. Search "Work", toggle **Workout** on.
4. Change the search to "Chill", dismiss the keyboard (Return key), tap **Done**.
5. Verify the episode was added to **Workout**.
6. Repeat with removal (start with the episode already in a playlist, toggle it off under a search that filters it out, then confirm it's removed).

## Checklist


SergioEstevao (approved):
Tested this all now episodes are correctly added to playlist.

PR #4714, opened 2026-07-08
Load episode details show notes and artwork earlier

kean:
Moves show notes and episode artwork loading from `viewDidAppear` to `viewWillAppear` so they start loading before the view is on screen and appear a little sooner.

Also fixes the `downloadingShowNotes` re-entrancy guard, which was checked but never actually set to `true`. Because the guard was dead, repeated/overlapping `loadShowNotes()` calls each re-ran the transcript branch and added a new `ThemedHostingController` without removing the previous one, stacking duplicate transcript excerpt views and leaking child view controllers. Setting the guard at load start makes moving the load into `viewWillAppear` (which can fire more than once) safe.

## To test

1. Open an episode details page for an episode that has show notes and a transcript.
2. Confirm show notes and artwork appear as the screen transitions in.
3. Navigate away and back, and tap "Try Again" a few times — confirm only a single transcript excerpt is shown (no stacked duplicates).

## Checklist

### DIFF
--- a/podcasts/Episode/EpisodeDetailViewController+ShowNotes.swift
+++ b/podcasts/Episode/EpisodeDetailViewController+ShowNotes.swift
@@ -33,6 +33,7 @@ extension EpisodeDetailViewController: WKNavigationDelegate, SFSafariViewControl
 
     func loadShowNotes() {
         if downloadingShowNotes { return }
+        downloadingShowNotes = true
 
         loadingIndicator.startAnimating()
         hideErrorMessage(hide: true)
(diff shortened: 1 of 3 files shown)


podcasts/Episode/EpisodeDetailViewController+ShowNotes.swift
```diff
@@ -33,6 +33,7 @@ extension EpisodeDetailViewController: WKNavigationDelegate, SFSafariViewControl
 
     func loadShowNotes() {
         if downloadingShowNotes { return }
+        downloadingShowNotes = true
```
kean, line 36: I don't think it's possible to break it right now, but this was setter was missing.

SergioEstevao (approved):
Working correctly.

PR #4721, opened 2026-07-09
Fix slow Bookmarks tab caused by blocking episode lookups on the main thread

kean:
Fixes PCIOS-817.

Opening the Bookmarks tab was slow when bookmarks referenced episodes missing locally: each orphaned bookmark fired a synchronous network lookup (10s timeout) on the main thread during row rendering.

- Added async `addMissingPodcastAndEpisode` to `ServerPodcastManager`; the old completion-based variant (synchronous networking) is soft-deprecated
- `BookmarkRow` now owns `BookmarkRowViewModel` as `@StateObject` and loads the missing episode in `.task(id:)`, off the main actor, once per episode
- Bookmark title/date/play-time render immediately from the bookmark; episode heading/artwork fill in when the load completes
- `PodcastViewController` creates its bookmarks list view model lazily on first switch to the Bookmarks tab instead of in `viewDidLoad`

## To test

To reproduce the missing-episode scenario locally, apply the attached debug patch — it deletes the local episodes referenced by the podcast's bookmarks every time the Bookmarks tab is opened, forcing rows through the `addMissingPodcastAndEpisode` path:

[debug-delete-bookmarked-episodes.patch](https://github.com/user-attachments/files/29901229/debug-delete-bookmarked-episodes.patch)

1. Download the patch and apply it from the repo root: `git apply ~/Downloads/debug-delete-bookmarked-episodes.patch`
2. Add a few bookmarks to episodes of a subscribed podcast (or sign in to an account that already has some)
3. Open the podcast page and switch to the Bookmarks tab — the tab should open instantly; each row shows the bookmark title, date, and play time right away, and the episode name + artwork fill in once fetched from the server
4. Scroll away and back — rows shouldn't refetch episodes they already loaded
5. Tap play on a bookmark and confirm playback starts at the bookmarked time
6. Discard the patch when done: `git checkout -- "podcasts/Podcasts/Podcast Page/PodcastViewController.swift"`

### DIFF
--- a/Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
+++ b/Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
@@ -147,6 +147,7 @@ public class ServerPodcastManager: NSObject {
 
     /// Soft-deprecated: performs synchronous networking, blocking the calling thread, and never calls
     /// the completion on failure. Use the async ``addMissingPodcastAndEpisode(episodeUuid:podcastUuid:shouldUpdateEpisode:)`` instead.
+    @available(*, deprecated, message: "Performs synchronous networking and blocks the calling thread. Use the async addMissingPodcastAndEpisode(episodeUuid:podcastUuid:shouldUpdateEpisode:) instead.")
     public func addMissingPodcastAndEpisode(episodeUuid: String, podcastUuid: String, shouldUpdateEpisode: Bool = false, completion: ((Episode?) -> ())? = nil) {
         let url = ServerConstants.Urls.cache() + "mobile/podcast/findbyepisode/\(podcastUuid)/\(episodeUuid)"
 
--- a/Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
+++ b/Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
@@ -1,5 +1,6 @@
 import Foundation
 import PocketCastsDataModel
+import PocketCastsUtils
 
 public class ServerPodcastManager: NSObject {
     private static let maxAutoDownloadSeperationTime = 12.hours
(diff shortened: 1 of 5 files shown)


Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
```diff
@@ -313,6 +337,27 @@ public class ServerPodcastManager: NSObject {
         return try decoder.decode(PodcastCollection.self, from: data)
     }
 
+    /// Sends the request and returns the parsed JSON, or nil when the server responds with 304 Not Modified
+    private func loadFrom(url: String) async throws -> [String: Any]? {
```
kean, line 342: This is just a copy of a non-async method. I initially refactored all call sites, but it became unwieldy, so I focused only on bookmarks. I added a note to address other call sites when time allows.

Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
```diff
@@ -144,6 +145,8 @@ public class ServerPodcastManager: NSObject {
         return nil
     }
 
+    /// Soft-deprecated: performs synchronous networking, blocking the calling thread, and never calls
+    /// the completion on failure. Use the async ``addMissingPodcastAndEpisode(episodeUuid:podcastUuid:shouldUpdateEpisode:)`` instead.
     public func addMissingPodcastAndEpisode(episodeUuid: String, podcastUuid: String, shouldUpdateEpisode: Bool = false, completion: ((Episode?) -> ())? = nil) {
```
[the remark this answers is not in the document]
SergioEstevao: This looks like a good suggestion, does it add a lot of noise in the compile logs?

Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
```diff
@@ -158,6 +161,27 @@ public class ServerPodcastManager: NSObject {
         }
```
SergioEstevao, line 162: When testing this with an episode/podcast that was completely removed from the database found out that this code is missing an 
```
} else  {
  completion?(nil)
}
```
Maybe something to fix in a follow up PR?

SergioEstevao (approved):
Working correctly, rows getting updated from non existing bookmarks podcasts/episodes on device

Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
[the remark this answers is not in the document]
kean: Let's add it. This method is quite unsafe. It does perform synchronous networking despite having a completion closure instead of a return value.

Modules/Sources/PocketCastsServer/Public/ServerPodcastManager.swift
kean: I deprecated this method and I plan to remove it entirely for this release.

PR #4722, opened 2026-07-09
Fix options picker labels wrapping to multiple lines when there is enough space

kean:
Fixes PCIOS-818

Options picker rows with a value on the right (e.g. "Sort By") could wrap their labels to multiple lines even when the row had enough space for both, because a constraint forced the title label to be at least as wide as the value label. This PR removes that constraint, letting compression resistance priorities decide which label gives way first, and widens the spacing between the labels.

## To test

1. Open a podcast page and tap the sort button (or Podcasts tab → ⋯ → Sort By).
2. Verify rows like "Sort By" show both labels on a single line when they fit.
3. Increase the Dynamic Type size and verify the labels only wrap when there genuinely isn't enough room.

Screenshot 2026-07-09 at 6 47 52 PM

## Checklist


SergioEstevao (approved):
:shipit:

PR #4729, opened 2026-07-10
Fix the disabled Transcript action rendering as invisible

kean:
Fixes PCIOS-

When the playing episode has no transcript, the Transcript action in the player's "More Actions" sheet (and the shelf button variant) was tinted with `playerContrast06` — white at 10% alpha, a background-fill color — making it render as essentially invisible. This switches the disabled state to `playerContrast04` (~30% alpha), which reads as disabled but stays legible on all themes. Tap behavior is unchanged: the disabled action still shows the "transcript not available" toast.

| Before | After |
|:---:|:---:|
| Screenshot 2026-07-10 at 1 38 35 PM (1)| Screenshot 2026-07-10 at 5 06 56 PM
 |

## To test

1. Play an episode that has no transcript.
2. Open the full-screen player and tap the "More Actions" (…) button.
3. Verify the Transcript row is dimmed but clearly visible.
4. Tap it and verify the "transcript not available" toast appears.
5. In the player, tap Edit in the More Actions sheet and move Transcript onto the shelf; verify the shelf icon is also dimmed but visible.

## Checklist


SergioEstevao (approved):
Visible on all themes. :shipit:

PR #4731, opened 2026-07-11
Adopt a8c-secrets for encrypted secrets

mokagio:
## What it does

Replaces `mobile-secrets`/`.configure` with [`a8c-secrets`](https://github.com/Automattic/a8c-secrets). Context: paaHJt-96q-p2

## How to test

**Automattic contributors only**

1. Checkout this branch
2. Notice that `.configure` and `.configure-files/` now appear as new, untracked files in the repo. **Delete them.** `rm -rf .configure .configure-files`
3. Set up [`a8c-secrets`](https://github.com/Automattic/a8c-secrets) following the instructions on the repository. You can find the dev key our internal secrets store searching for `a8c-secrets - pocket-casts-ios - dev private key`
4. Run `bundle exec fastlane configure_secrets`
5. Verify the app builds alright.


fastlane/Fastfile
```diff
@@ -227,6 +221,16 @@ platform :ios do
     )
   end
 
+  # Decrypt the app secrets into the local `a8c-secrets` directory the build reads from.
+  #
+  lane :configure_secrets do
+    unless FastlaneCore::CommandExecutor.which('a8c-secrets')
+      UI.user_error!('`a8c-secrets` was not found on your PATH. See https://github.com/Automattic/a8c-secrets for installation and usage instructions.')
```
mokagio: Earlier iterations had more detailed instructions. Apart from being uglier to look at in code, they seemed like a liability. Better to use the project's `README.md` as the single source of truth.

.buildkite/commands/build-tvos-app-store.sh
```diff
@@ -8,9 +8,6 @@ fi
 
 "$(dirname "${BASH_SOURCE[0]}")/shared_setup.sh"
 
-echo "--- :closed_lock_with_key: Installing Secrets"
-bundle exec fastlane run configure_apply
-
```
mokagio, line 13: Notice the removal of the decrypt call because now it runs as part of the shared setup script.

config/PocketCasts.base.xcconfig
```diff
@@ -11,3 +11,8 @@ CURRENT_PROJECT_VERSION = $(VERSION_LONG)
 PRODUCT_BUNDLE_IDENTIFIER_ROOT = au.com.shiftyjelly.podcasts
 
 DEVELOPMENT_TEAM = PZYM8XX95Q // Automattic, Inc. (Company)
+
+A8C_SECRETS_DIR = $(HOME)/.a8c-secrets/pocket-casts-ios@github.com@automattic
+FIREBASE_SECRETS_PATH = $(A8C_SECRETS_DIR)/GoogleService-Info.plist
+SECRETS_PATH = $(A8C_SECRETS_DIR)/pocket_casts_credentials.json
+GOOGLE_SIGN_IN_SECRETS_PATH = $(A8C_SECRETS_DIR)/GoogleSignIn.txt
```
mokagio, line 22: These pointed to `~/.configure/...` before. DRYed them here. While we can't rely on `a8c-secrets which` at this level (and setting up an aggregate target + build phase to run it seems a bit too much) at least we can centralize the place where the path is hardcoded.

pocketcasts:
<details><summary>App Icon📲 You can test the changes from this Pull Request in <b>Pocket Casts Prototype Build</b> by scanning the QR code below to install the corresponding build.</summary>
<table>
<tr>
  <td><b>App Name</b></td><td>Pocket Casts Prototype Build</td>
</tr>
<tr><td><b>Build Number</b></td><td><code>16480</code></td></tr>
<tr><td><b>Version</b></td><td><code>PR #4731</code></td></tr>
<tr><td><b>Bundle ID</b></td><td><code>au.com.shiftyjelly.podcasts.prototype</code></td></tr>
<tr><td><b>Commit</b></td><td>50047481f2db56c7a6b177080a84dbcc6196718d</td></tr>
<tr><td><b>Installation URL</b></td><td><a href='https://appdistribution.firebase.google.com/testerapps/1:124902176124:ios:3999f43bfa1fcd1b1620f9/releases/63bj7nueldc20?utm_source=fastlane'>63bj7nueldc20</a></td></tr>
</table>
<em>Automatticians: You can use our internal self-serve MC tool to give yourself access to those builds if needed.</em>
</details>

.buildkite/commands/install-secrets.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+echo "--- :closed_lock_with_key: Installing Secrets"
+curl -fsSL https://raw.githubusercontent.com/Automattic/a8c-secrets/main/install.sh | bash
```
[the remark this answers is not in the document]
AliSoftware: We should definitively install `a8c-secrets` inside all our CI agents at some point and reprovision them, so that repos adopting of `a8c-secrets` don't have to do this `curl` at all.

I see 2 solutions in the short term:
 - Either bite the bullet already and install `a8c-secrets` inside all our agents before going further. Technically that shouldn't be _that_ complicated (adding a shared Ansible role in our buildkite-ci repo that downloads and install the right binary based on architecture, then invoke that role in all our agents)
 - Or accept that we'll temporarily have our scripts download and install it via `curl` during each CI job, but in that case I agree with Copilot that we should also verify the expected checksum.
    - Sure one could say "eh there's little change of supply-chain attack because this `curl`'d URL is hosted by a repository owned by a8c in the first place"
    - But since this is a tool related to handling secrets, I'd still prefer for us to be extra careful and add more checks than less when it comes to security
    - And the argument like "this is temporary, as we'll soon remove this to use the one pre-installed in the agent" is not acceptable IMHO, because we all know how this kind of "this is temporary" cases can go

Besides, writing all that made me realize that even if we re-provision all our Agents with Ansible to include `a8c-secrets` pre-installed, this won't cover the case of jobs running inside Mac VMs, unless we update `hostmgr` to make it mount `a8c-secrets` as a shared volume inside the VM and add that shared volume into the `$PATH` (which would IMHO be a better approach than having to rebuild our VM images just to include `a8c-secrets` binary in them, and would also allow us to update `a8c-secrets` on our Mac hosts without having to rebuild VMs every time we update it)

fastlane/Fastfile
```diff
@@ -227,6 +228,21 @@ platform :ios do
...
+        UI.important('`a8c-secrets` was not found on the PATH, but given CI execution was detected, we assume secrets has already been decrypted. See https://linear.app/a8c/issue/AINFRA-2681.')
+        next
+      else
+        UI.user_error!('`a8c-secrets` was not found on your PATH. See https://github.com/Automattic/a8c-secrets for installation and usage instructions.')
+      end
+    end
+
+    sh('a8c-secrets', 'decrypt')
+  end
+
```
AliSoftware: I wonder if it would be worth it on the long term to create a dedicated Fastlane Action in `releases-toolkit` that would take care of (1) installing `a8c-secrets` binary if not present (providing a version and shasum) + (2) doing this validation dance + finally call `decrypt`…

This way we could DRY the setup across repos.

The drawback is that now we'd be relying on Ruby and Fastlane to do all that (including the install) while (1) some jobs might not require Ruby to run but might still need secrets, so installing and decrypting secrets directly via bash helpers in `.buildkite/commands` like you did is more universal and doesn't require to install Ruby and Fastlane just for that (2) so we'd still have 2 sources of truth to maintain…
So maybe not such a good idea after all?

scripts/build-phases/generate-credentials.sh
```diff
@@ -24,8 +24,8 @@ fi
 
 ## Validate Secrets!
 ##
-if [ ! -f $SECRETS_PATH ]; then
-    echo "error: $SECRETS_PATH not found! Please run \`bundle exec fastlane run configure_apply\`."
+if [ ! -f "$SECRETS_PATH" ]; then
+    echo "error: $SECRETS_PATH not found! Please run \`bundle exec fastlane configure_secrets\`."
```
AliSoftware: While I like that we have a `configure_secrets` lane that wraps the call to `a8c-secrets decrypt` in order to providing better error messages if the binary is not found before calling it, and in order for it to be automatically called before each lane is executed… the whole idea behind migrating to a Rust-based `a8c-secrets` was to avoid needing Ruby and for users to be able to run `a8c-secrets` binary with verbs that were sensible and easy to understand.

I wonder if nudging people to call `fastlane configure_secrets` (which effectively only calls `a8c-secrets decrypt`) would thus hide that simplicity we were aiming for with `a8c-secrets` behind something more obscure, making them become less familiar with the `a8c-secrets` CLI itself (e.g. "how do I edit a secret? Oh I need to call `a8c-secrets` command? But I never use it otherwise, I always use `fastlane configure_secrets`, so I've never heard of `a8c-secrets` command"…)

Maybe I'm overthinking it, but I think having the error messages printed by Xcode's build phase mention `a8c-secrets decrypt` directly rathen than `fastlane configure_secrets` could at least be a good middle ground there?

fastlane/Fastfile
```diff
@@ -199,15 +198,17 @@ GLOTPRESS_TO_ASC_METADATA_LOCALE_CODES = {
...
-  # Decrypt the secrets. This is redundant on dev machines most of the time,
-  # but it has such a negligible overhead that it's worth running it here to
-  # keep the individual lanes cleaner.
-  configure_apply
+  # Decrypt the secrets.
+  #
+  # This is redundant on dev machines most of the time, but it
+  # has such a negligible overhead that it's worth running it
+  # here to keep the individual lanes cleaner.
+  configure_secrets unless lane == :configure_secrets
```
AliSoftware: Interesting approach to have `a8c-secrets decrypt` be guaranteed to be run before every single Fastlane lane to be sure we're always up-to-date…

But that makes me wonder if it's always necessary for all lanes?

For example, while for lanes that compile the app I can see this being needed/useful, for other lanes (like if we were to call `configure_code_signing` to update certificates etc), I don't think we'd need any of the secrets files managed by `a8c-secrets` in this repo, so that would make those be decrypted for nothing  (and sure this doesn't take long to decrypt them in general, but that can also lead to errors if you don't have the private key configured on your machine, and I could see an AppsInfra engineer who runs `match` to update certs not necessary having the compilation secrets configured for every repo…)

Also I haven't thought about what happens if we have external contributors? Admittedly, I'm not sure they would be expected to call any `fastlane` lane to build and test the project, as they would probably instead only do that via Xcode. But still, that could still be another case where having `configure_secrets` being called before each lane could be a blocker for them if applied to all lanes even the ones not doing any compilation?

AliSoftware (commented):
I'm at the end of my day so won't have the time to properly run the test steps until tomorrow at the earliest, but overall the code logic looks good so far 👍

kean (commented):
I was assigned a a reviewer, but I don't really have much to add to the conversion here. No reason not to make this change.

fastlane/Fastfile
mokagio: The thought of moving to `release-toolkit` crossed my mind also but haven't had the time to work on it yet.

That's probably for the best, because the points you raise on adding an unnecessary dependency on Ruby tooling are strong. Add to that the desire to have _less_ of the devs workflow depend on Ruby, and pushing to `release-toolkit` would be shooting ourselves in the foot and create more migration work for later.

I actually like the suggestion you left [below](https://github.com/Automattic/pocket-casts-ios/pull/4731#discussion_r3589652245) to push the checks to the build phase(s) that need the secrets.

mokagio:
Back to draft because of the ongoing work for the `a8c-secrets` set up in CI toolkit, see https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/218

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/commands/install-secrets.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# `install_a8c-secrets_binary` comes from the a8c-ci-toolkit plugin. It pins the
+# a8c-secrets version and checks the download against a checksum vendored there.
```
mokagio:
```suggestion
```

config/PocketCasts.base.xcconfig
```diff
@@ -11,3 +11,8 @@ CURRENT_PROJECT_VERSION = $(VERSION_LONG)
 PRODUCT_BUNDLE_IDENTIFIER_ROOT = au.com.shiftyjelly.podcasts
 
 DEVELOPMENT_TEAM = PZYM8XX95Q // Automattic, Inc. (Company)
+
+A8C_SECRETS_DIR = $(HOME)/.a8c-secrets/pocket-casts-ios@github.com@automattic
```
mokagio, line 19:
```suggestion
# Notice that this assumes a default a8c-secrets setup:
# 
# - $(HOME)/.a8c-secrets is the default tool workspace/
# - the rest of the path should be in sync with .a8c-secrets/repo-id from the repo root
A8C_SECRETS_DIR = $(HOME)/.a8c-secrets/pocket-casts-ios@github.com@automattic
```

fastlane/Fastfile
mokagio: Addressed in https://github.com/Automattic/pocket-casts-ios/pull/4731/commits/c6aa0cadd0a80d9f2548b342f314b0770353543c by removing the lane.

pocketcasts:

pocketcasts:

PR #4640, opened 2026-06-29
Remove code behind FeatureFlag.newSettingsStorage and FeatureFlag.settingsSync

kean:
Removes all code gated by `FeatureFlag.newSettingsStorage` and `FeatureFlag.settingsSync`, along with the two flags themselves.

## To test

1. Build and run the app.
2. Open Settings → General and toggle a few options (keep screen awake, auto-open player, smart playback, open links in browser); confirm they persist across relaunch.
3. Open a podcast → Settings → playback effects; change custom speed / trim silence / volume boost and confirm they persist.
4. Change app appearance/theme and confirm it persists.
5. Confirm settings still sync to the server (skip times, marketing opt-in, home grid sort).

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics.

PR #4743, opened 2026-07-13
Remove redundant Podcast settings forwarding properties

kean:
Follow-up to [this comment on #4640](https://github.com/Automattic/pocket-casts-ios/pull/4640/changes#diff-3dea37fbffe08e15761497d649dc0a99fd91dd263d27ce82d6cbdfb3d3a573dc): with the settings-sync flags gone, the computed properties in `Podcast+Settings.swift` were trivial forwarders to the stored DB columns. Call sites now use the stored properties directly (`podcastSortOrder` stays — it does a real conversion). Also deletes the unused `PodcastSettings.autoUpNextSetting`.

## To test

1. Build and run the app.
2. Open a podcast → Settings; toggle notifications, custom playback effects, and auto-archive overrides; confirm they persist across relaunch.
3. On the podcast page, toggle "Show archived" and confirm it persists.

## Checklist


Modules/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
```diff
@@ -304,7 +304,7 @@ class PodcastDataManager {
         var podcastsOverrideArchive = [Podcast]()
         cachedPodcastsQueue.sync {
             for podcast in cachedPodcasts.values {
-                if podcast.isSubscribed(), podcast.isAutoArchiveOverridden {
+                if podcast.isSubscribed(), podcast.overrideGlobalArchive {
```
kean, line 307: If we ever need to add branching again, I'd go for a dedicated struct for settings:

```swift
podcast.settings.isAutoArchiveOverridden
```

The structure could provide conditional logic and Swifty naming. However, since there is no need, I'd stick close to the DB naming and ignore the non-swiftness.

SergioEstevao (approved):
Tested multiple podcast settings and properties were kept between launches correctly

:shipit:

PR #4745, opened 2026-07-14
Remove no-op PCDBResultSet.close()

kean:
As discussed, removing the no-op `PCDBResultSet.close()` since it's no longer needed with GRDB.

## To test

1. Run the app and browse around (podcasts, episodes, filters, Up Next, bookmarks) — all data loading goes through these result sets and should behave exactly as before.
2. Run the DataModel module tests.

## Checklist


SergioEstevao (approved):
Smoke tested the app and it looks all DB operations are working correctly
:shipit:

PR #4746, opened 2026-07-14
Fix Bookmark row artwork stretching when non-square

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS-832

Bookmark rows render the episode artwork with a plain `.resizable()` image inside a fixed square frame, so non-square artwork (e.g. embedded episode artwork) gets stretched. This adds `.aspectRatio(contentMode: .fill)` so the artwork is cropped to the square thumbnail instead, matching how `PodcastImageView` renders artwork elsewhere in the app.

## To test

1. Enable Settings → Appearance → Use Embedded Artwork.
2. Play an episode whose artwork is non-square (e.g. an episode with wide embedded artwork).
3. Add a bookmark, then open the Bookmarks tab in the full-screen player.
4. Verify the bookmark row's thumbnail is a square crop of the artwork with no stretching or distortion.

## Checklist


SergioEstevao (approved):
:shipit:

PR #4750, opened 2026-07-14
Add opt-in SQL statement tracing and remove no-op busy_timeout pragma

kean:
This PR addressed the following comment from Copilot: https://github.com/Automattic/pocket-casts-ios/pull/4745#discussion_r3575154591.

Adds an easy way to see every SQL statement the app runs, and removes a dead statement found along the way:

- **Opt-in SQL tracing**: launching a debug build with the `-PCSQLTracing` argument logs every SQL statement (with arguments) to the unified logging system under `category:SQL`, following the same OSLog pattern as `AnalyticsOSLogAdapter`. The argument is included in the shared `pocketcasts` scheme, disabled by default, so enabling it is a single checkbox. Compiled only in `DEBUG`.
- **Remove no-op `PRAGMA busy_timeout`**: the manual pragma in `DatabaseHelper.setup` was redundant — GRDB's `Configuration.busyMode = .timeout(10)` already configures the same 10-second busy timeout on every connection.

I verified with the new SQL tracing that `busy_timeout` pragma it not executed. With GRBD, [this code](https://github.com/Automattic/pocket-casts-ios/blob/700f6e395250e81e27160abb67ec318749fe9fe2/Modules/Sources/PocketCastsDataModel/Public/DataManager.swift#L49) handles the busy errors.

## To test

1. Open Edit Scheme ▸ Run ▸ Arguments for the `pocketcasts` scheme and enable `-PCSQLTracing`.
2. Run the app and browse around (open a podcast, play an episode).
3. In the Xcode console (or Console.app filtering on `category:SQL`), verify every SQL statement is logged with its arguments.
4. Disable the argument and relaunch — verify no SQL logging occurs.
5. Verify normal app behavior (database reads/writes, playback, sync) is unchanged.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
+++ b/Modules/Sources/PocketCastsDataModel/Private/Managers/Util/DatabaseHelper.swift
@@ -10,8 +10,6 @@ class DatabaseHelper {
         var databaseWasCreated = false
         queue.write { db in
             do {
-                _ = try db.executeQuery("PRAGMA busy_timeout = 10000", values: nil)
-
                 let startingSchemaVersion = db.pragmaUserVersion() ?? 0
                 databaseWasCreated = startingSchemaVersion < 1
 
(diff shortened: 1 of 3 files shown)


SergioEstevao (approved):
Smoke tested the app with the new setting and looking at the trace.

All working correctly.

:shipit:

PR #4753, opened 2026-07-14
Fix Stats screen stuck loading when the remote stats request fails

kean:
Fixes PCIOS-835

`StatsManager.loadRemoteStats` returned from its guard on the failure path (nil `remoteStats`) without calling the completion. `StatsViewController` sets its state to `.loading` and relies on that completion to flip it, so a failed request left the spinner showing forever. The failure path now calls `completion?(false)` so the view controller shows the failed state.

## To test

1. Sign in to an account.
2. Put the device in Airplane Mode (or otherwise make `user/stats/summary` fail).
3. Go to Profile → Stats.
4. Verify the screen shows the error state instead of an endless loading spinner.
5. Disable Airplane Mode, reload the screen, and verify stats load normally.

Screenshot 2026-07-14 at 1 54 52 PM

## Checklist


SergioEstevao (approved):
:shipit:

PR #4755, opened 2026-07-14
Fix intro-skipped time never syncing when it is the only changed stat

kean:
Fixes PCIOS-

`changedStats()` decides whether there are any listening stats worth syncing, but the guard checked `totalSkippedTime == nil` twice and never checked `totalIntroSkippedTime`. If intro/auto-skip time was the only non-nil stat, the guard bailed out and that stat was never synced. This fixes the copy-paste so the guard checks `totalIntroSkippedTime`.

## To test

1. Sign in to a sync account and reset local stats (or use a fresh install) so all stats start at zero.
2. Play an episode with intro skip / auto-skip configured so only intro-skipped time accumulates (no listening, skipping, or speed savings).
3. Trigger a sync and verify the device stats record is sent with `timeIntroSkipping` populated (previously no stats record was sent at all).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 8.17
 -----
+- Fix intro-skipped time not syncing to your account when it is the only listening stat that changed [#4755](https://github.com/Automattic/pocket-casts-ios/pull/4755)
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
Great catch!

:shipit:

PR #4756, opened 2026-07-14
Add frequency property to plus_promotion_upgrade_button_tapped

yaelirub:
Fixes PCIOS-

The `plus_promotion_upgrade_button_tapped` analytics event does not include the billing frequency (monthly vs yearly), even though `purchase_successful` does. Without it, the upgrade funnel can't be segmented by frequency — we can see how many purchases succeeded per frequency but not how many users tapped upgrade for each.

This adds a `frequency` property to the event at both live call sites, using the same value shape as `purchase_successful` (`PlanFrequency` raw value → `monthly` / `yearly`):

- `PlusLandingViewModel.unlockTapped(_:)` — from the tapped product (`product.frequency`).
- `UpgradeAccountViewModel.purchaseTapped()` — from `selectedProduct.frequency`, mirroring exactly what `purchase_successful` reads (`productId.frequency`).

(The third reference to this event, `AnalyticsHelper.plusUpgradeConfirmed(source:)`, has no callers and no product/frequency data, so it was left untouched.)

Event Horizon schema companion PR: Automattic/EventHorizonSchemas#109

## To test

1. Enable `FeatureFlag.tracksLogging` so tracked events log to the console as `🔵 Tracked: plus_promotion_upgrade_button_tapped [... "frequency": ...]`.
2. Open an upsell / upgrade screen (e.g. Profile → Upgrade).
3. Select the **Yearly** plan and tap the upgrade button. Confirm the logged event includes `"frequency": "yearly"`.
4. Go back, select the **Monthly** plan and tap upgrade. Confirm the logged event includes `"frequency": "monthly"`.
5. Repeat across both entry points (the plus landing/upsell view and the account upgrade view) to cover both call sites.

## Checklist

### DIFF
--- a/podcasts/Onboarding/Plus/NewUpgrade/UpgradeAccountViewModel.swift
+++ b/podcasts/Onboarding/Plus/NewUpgrade/UpgradeAccountViewModel.swift
@@ -139,7 +139,7 @@ class UpgradeAccountViewModel: PlusPurchaseModel {
     }
 
     func purchaseTapped() {
-        track(.plusPromotionUpgradeButtonTapped)
+        track(.plusPromotionUpgradeButtonTapped, properties: ["frequency": selectedProduct.frequency.rawValue])
 
         guard SyncManager.isUserLoggedIn() else {
             presentLogin(with: ProductInfo(plan: upgradeTier.plan, frequency: selectedFrequency))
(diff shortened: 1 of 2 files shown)


kean (changes requested):
Hey, could you update the schema? https://github.com/Automattic/EventHorizonSchemas/blob/954ef8a2212931c7192cbb528ec21d322ba0072d/schema/pocket-casts.yml#L877

The iOS app doesn't use it yet, but the schema is the source of truth. It looks like `plus_promotion_upgrade_button_tapped` has multiple properties (on Android?) that aren't reflected on iOS. It is however missing `frequency`.

yaelirub:
@kean , sorry I should have done that before pushing. Working on it now.

yaelirub:
@kean , updated.

yaelirub:
@kean , added to Event Horizon and to Android (both merged already). Anything else needed here?

kean (approved):
Thanks!

PR #4757, opened 2026-07-14
Fix empty mini player after a non-animated full player dismissal

kean:
Fixes PCIOS-798 (**more details in the ticket!**)

On iOS 26, `PlayerZoomAnimator` hides the mini player's subviews while the full-screen player is presented and only restores them in its dismiss animation. Since UIKit skips the animator for `dismiss(animated: false)` (push notifications, deep links, stacked-sheet dismissals), the mini player stayed empty and frozen, and the player state was never closed out. (Zendesk 11349405)

The fix vends a custom `UIPresentationController` whose `dismissalTransitionDidEnd(_:)` fires for every completed dismissal; it performs the restore only when no transition animation ran, leaving animated dismissals untouched.

## To test

1. On an iOS 26 simulator, play an episode and open the full-screen player.
2. Trigger a non-animated dismissal: open `pktc://settings/storage-and-data` in Safari, or tap an episode push notification.
3. Verify the mini player still shows artwork/title/controls and keeps updating, and tapping it reopens the full player.
4. Sanity-check the animated flows: opening and swiping down the player should look unchanged.

## Checklist

### DIFF
--- a/podcasts/MiniPlayerViewController+TransitionDelegate.swift
+++ b/podcasts/MiniPlayerViewController+TransitionDelegate.swift
@@ -1,6 +1,17 @@
 import Foundation
 
 extension MiniPlayerViewController: UIViewControllerTransitioningDelegate {
+    func presentationController(forPresented presented: UIViewController, presenting: UIViewController?, source: UIViewController) -> UIPresentationController? {
+        guard let fullPlayer = presented as? PlayerContainerViewController else { return nil }
+
+        let presentationController = FullPlayerPresentationController(presentedViewController: fullPlayer, presenting: presenting)
+        presentationController.onNonAnimatedDismiss = { [weak self, weak fullPlayer] in
+            guard let fullPlayer else { return }
+            self?.fullScreenPlayerDidDismiss(fullPlayer)
+        }
+        return presentationController
+    }
+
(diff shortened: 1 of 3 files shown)


kean:
I tested normal open/close, push when the app is open, push when the app is closed.

podcasts/MiniPlayerViewController+TransitionDelegate.swift
```diff
@@ -43,3 +54,29 @@ extension MiniPlayerViewController: UIViewControllerTransitioningDelegate {
...
+
+    private var isDismissalAnimated = false
+
+    override func dismissalTransitionWillBegin() {
+        super.dismissalTransitionWillBegin()
+
+        isDismissalAnimated = presentedViewController.transitionCoordinator?.isAnimated ?? false
+    }
+
+    override func dismissalTransitionDidEnd(_ completed: Bool) {
```
kean, line 75: It is a bit complicated, but it does get the job done. I'm looking into refactoring the original implementation to eliminate the workaround. It's easier said than done though due to the complexity of the presentation.

SergioEstevao (approved):
Teste with push notification and open from Safari and animation worked correctly and mini-player was showing content correctly.

:shipit:

PR #4763, opened 2026-07-15
Add Generated chapters playback setting

leandroalonso:
Fixes PCIOS-836

Adds a **Generated chapters** toggle to Settings → General. When enabled (the default), AI-generated chapters are shown in the player; when disabled, they are hidden while publisher-provided chapters (embedded / Podcast Index / show-notes) remain available. Mirrors the [web player setting](https://github.com/Automattic/pocket-casts-webplayer/pull/4729) and [pocket-casts-android#5564](https://github.com/Automattic/pocket-casts-android/pull/5564).

The setting is a positive UI toggle stored as the inverse global named setting `disableAiChapters`, synced to the server via the new `disable_ai_chapters` named setting ([pocket-casts-sync-api#1424](https://github.com/Automattic/pocket-casts-sync-api/pull/1424)) — mirroring the existing `audioOnly` / `marketingOptIn` sync path. The row is gated behind `FeatureFlag.generatedChapters`, so it is hidden in production until that flag ships.

Toggling the setting re-parses the current episode's chapters via `PlaybackManager.forceUpdateChapterInfo()`, so the player Chapters tab, Now Playing info, lock screen, mini player, CarPlay, and Watch all react immediately without needing to reopen the episode.

## To test

1. Build the app using the `StagingDebug` scheme.
2. Login (or create an account).
3. Go to **Settings → General** and confirm a new **Generated chapters** toggle appears, **defaulting to on**, with the "Show automatically generated chapters when available." description.
4. Open **The Daily** and play an episode whose only chapters are AI-generated. Confirm the **Chapters** tab appears in the player and the generated chapters are listed.
5. While that episode is playing, return to **Settings → General** and turn **Generated chapters off**. Go back to the player and confirm the **Chapters** tab and chapter info disappear immediately (no reopen needed).
6. Turn **Generated chapters on** again and confirm the Chapters tab and generated chapter list return immediately.
7. Open a podcast episode with **publisher-provided** chapters (eg.: **The Changelog**) with the setting **off**, and confirm its Chapters tab and chapters are still shown and usable.

### Verify it syncs

8. Access `https://pocketcasts.net/podcasts`.
9. Login to the same account.
10. Go to `https://pocketcasts.net/settings/playback`.
11. Confirm **Generated chapters** matches the state set on iOS (toggle it on web and confirm iOS reflects it after a refresh/sync as well).

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
+++ b/Modules/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
@@ -1223,6 +1223,15 @@ nonisolated struct Api_ChangeableSettings: @unchecked Sendable {
   /// Clears the value of `audioOnly`. Subsequent reads from it will return its default value.
   mutating func clearAudioOnly() {_uniqueStorage()._audioOnly = nil}
 
+  var disableAiChapters: Api_BoolSetting {
+    get {_storage._disableAiChapters ?? Api_BoolSetting()}
+    set {_uniqueStorage()._disableAiChapters = newValue}
+  }
+  /// Returns true if `disableAiChapters` has been explicitly set.
+  var hasDisableAiChapters: Bool {_storage._disableAiChapters != nil}
+  /// Clears the value of `disableAiChapters`. Subsequent reads from it will return its default value.
+  mutating func clearDisableAiChapters() {_uniqueStorage()._disableAiChapters = nil}
+
   var unknownFields = SwiftProtobuf.UnknownStorage()
 
   init() {}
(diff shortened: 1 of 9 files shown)


leandroalonso:
@claude review

rviljoen:
At the risk of overcomplicating the setting, I think the following options would be nice. Use AI chapters:
- Never
- If creator chapters not available 
- Always (ie overrides creator chapters)

My personal setting would be to only show AI chapters if creator chapters are not available

leandroalonso:
@rviljoen Noted! I'll pass that along to the team.

However, the way it currently works we always default to the creator chapters. So if they're available, we don't show generated chapters.

rviljoen:
> @rviljoen Noted! I'll pass that along to the team.
> 
> 
> 
> However, the way it currently works we always default to the creator chapters. So if they're available, we don't show generated chapters. 

Ah, in that case I think my suggestion makes less sense. The current implementation is good 👍🏻

SergioEstevao (approved):
Tested following the steps and all worked correctly.

With the setting on:

- The Daily: Show generated chapters
- ATP: Show creator chapters

With the setting off:

- The Daily:  Don't show chapters
- ATP: Show creator chapters

:shipit:

PR #4769, opened 2026-07-16
Fix an issue with main thread blocked by networking when playing episodes from search

kean:
Playing an episode from search results used the deprecated `addMissingPodcastAndEpisode` (synchronous networking, silent failures). Now:

- `playEpisodeSearchResult` is async throwing and uses the async/await `addMissingPodcastAndEpisode`.
- The play button is disabled while the episode is fetched, showing a spinner after 0.5s (no flash on fast loads).
- Failures show a toast and fire an error haptic (haptic also added to the matching bookmark play flow).

## To test

Apply the testing patch attached in the comments — it deletes the local copy of the episode on every play tap so the server fetch always fires: `git apply search-result-play-force-server-fetch.patch`

1. Search for an episode and tap play — playback starts (spinner only appears if loading takes over 0.5s).
2. With a slow connection (e.g. Network Link Conditioner), tap play — the button turns into a spinner and is disabled until playback starts.
3. In Airplane Mode (or use breakpoints in Proxyman) tap play — an error haptic fires and a "The episode couldn't be loaded" toast appears at the bottom.
4. Revert the patch and confirm regular playback from search still works.

Screenshot 2026-07-16 at 1 48 03 PM Screenshot 2026-07-16 at 1 48 13 PM

## Checklist

### DIFF
--- a/podcasts/Bookmarks/List/BookmarkListViewModel.swift
+++ b/podcasts/Bookmarks/List/BookmarkListViewModel.swift
@@ -107,7 +107,7 @@ extension BookmarkListViewModel {
             do {
                 try await self?.router?.bookmarkPlay(bookmark)
             } catch {
-                HapticsHelper.triggerEpisodeLoadFailedHaptic()
+                HapticsHelper.triggerErrorHaptic()
                 Toast.show(L10n.discoverEpisodeFailToLoad)
             }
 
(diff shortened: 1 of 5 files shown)


SergioEstevao (approved):
:shipit:

PR #4766, opened 2026-07-15
Add on-device bookmark enrichment using Apple Foundation Models

kean:
Fixes PCIOS-839 and PCIOS-840

Adds `BookmarkFoundationModelEnricher`, an on-device alternative to the `/mobile/bookmark/enrich` endpoint for generating bookmark titles/summaries using Apple Foundation Models (iOS 26+). Uses `@Generable` guided output, `prewarm()` to hide model-load latency, `NLLanguageRecognizer` so the output matches the transcript's language, and a prompt tuned for specific, well-formatted titles (with optional podcast/episode context). Not wired into bookmark creation yet.

Stacked on #4765.

Benchmarked on an iPhone 16 Pro against staging: pre-warmed on-device generation averaged ~0.95s per bookmark vs ~1.1–1.3s (with occasional multi-second outliers) for the server, with comparable output quality.

## To test

1. From the repo root: `git apply smart-bookmarks-fm-benchmark.patch` (attached below)
2. Build the **Pocket Casts Staging** scheme on an iOS 26+ device with Apple Intelligence enabled, logged in with a **staging** account (a production account gets 401s and is logged out).
3. Launch the app; after ~5 seconds the console prints a markdown report between `SMART BOOKMARKS BENCHMARK BEGIN/END` markers (also written to the app's Documents as `smart-bookmarks-benchmark.md`).
4. Verify the `server` and `on-device` rows show sensible titles/summaries and timings — including the non-English snippets responding in their own language.
5. Revert: `git apply -R smart-bookmarks-fm-benchmark.patch`

[smart-bookmarks-fm-benchmark.patch](https://github.com/user-attachments/files/30067910/smart-bookmarks-fm-benchmark.patch)

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4772, opened 2026-07-16
Generate bookmark title suggestions from transcripts in the add-bookmark sheet

kean:
Fixes PCIOS-838

Generates a title suggestion for newly created bookmarks from the transcript around the bookmark position. Stacked on #4766, gated on `FeatureFlag.smartBookmarks` (debug-only).

- **Snippet extraction** (`BookmarkTranscriptSnippetExtractor`): 25s back / 5s forward of the bookmark (backward-weighted for reaction time), snapped to sentence boundaries, ≥10-word guard. For generated transcripts, playback time is first resolved to the reference timeline via a one-shot fingerprint match (`FingerprintTimingManager.resolveReferenceTime`), falling back to the raw time.
- **Generation**: on-device Foundation Models first, server enrich fallback (`BookmarkManager.generateTitle`). The on-device enricher now produces **only the title, as a plain string**: [`.permissiveContentTransformations`](https://developer.apple.com/documentation/foundationmodels/systemlanguagemodel/guardrails/permissivecontenttransformations) — needed so transcripts on heavy topics aren't rejected with "May contain sensitive content" — only applies to plain string output, not `@Generable` guided generation. The summary wasn't used anyway.
- **UI (temporary)**: spinner on the right side of the title while generating; the suggestion replaces the default title if untouched, otherwise shows as a tappable row under the field. The loading behavior hasn't been specified yet, so treat this UX as a placeholder.

## To test

1. Build the Staging scheme on an iOS 26+ device with Apple Intelligence enabled, logged in with a staging account (for the server fallback).
2. Play an episode with a transcript and create a bookmark from the player.
3. Verify a spinner appears at the right of the default title, then the suggested title replaces it.
4. Create another bookmark and start typing before generation finishes — the suggestion should appear under the field, and tapping it applies it.
5. Create a bookmark on an episode without a transcript — the default title stays and the sheet behaves as before.

## Checklist

### DIFF
--- a/podcasts/Bookmarks/BookmarkTranscriptSnippetExtractor.swift
+++ b/podcasts/Bookmarks/BookmarkTranscriptSnippetExtractor.swift
@@ -1,0 +1,107 @@
+import Foundation
+import NaturalLanguage
+import PocketCastsDataModel
+import PocketCastsUtils
+
+/// Extracts the transcript text surrounding a bookmark's position, used as the
+/// input for generating a bookmark title/summary.
+///
+/// Generated transcripts are timed against a reference copy of the episode
+/// audio, which dynamic ads shift relative to what the listener actually
+/// hears — so the bookmark's playback time is first resolved to the reference
+/// timeline by fingerprinting the local audio around it (see
+/// `docs/transcripts.md`), falling back to the raw playback time when a
+/// confident mapping isn't available.
+struct BookmarkTranscriptSnippetExtractor {
+
+    /// The capture window is asymmetric: by the time a user acts on a moment
+    /// they want to keep, that moment is already behind the playhead — so we
+    /// reach much further back than forward.
+    static let backwardWindowSeconds: TimeInterval = 25
+    static let forwardWindowSeconds: TimeInterval = 5
+
+    /// Snippets with fewer words than this carry too little signal to generate
+    /// a meaningful title from, so extraction aborts instead.
+    static let minimumWordCount = 10
+
+    /// Extracts the transcript snippet around `time` (a playback-timeline
+    /// position, e.g. `Bookmark.time`) for the given episode. Returns nil when
+    /// the episode has no usable timed transcript or the window holds fewer
+    /// than `minimumWordCount` words.
+    func snippet(forTime time: TimeInterval, episode: BaseEpisode) async -> String? {
+        let transcriptManager = TranscriptManager(episodeUUID: episode.uuid, podcastUUID: episode.parentIdentifier())
+        guard let model = try? await transcriptManager.loadTranscript() else {
+            return nil
+        }
+
+        // Only generated transcripts have a reference fingerprint published, so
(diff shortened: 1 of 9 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/Fingerprint/FingerprintTimingManager.swift
```diff
@@ -525,6 +525,132 @@ final class FingerprintTimingManager: NSObject {
...
+    /// this is one-shot and side-effect-free: it uses its own matcher,
+    /// cancellation flag, and scratch mapping, and never mutates the
+    /// continuous transcript mapping (`main`), `context`, or `state` — except
+    /// for the warm fast path, which reads the continuous mapping when it
+    /// already confidently covers `playbackTime`.
+    ///
+    /// Returns nil when no reference exists for the episode, the audio region
+    /// isn't local, no confident match is found, or the timeout expires —
+    /// callers should fall back to the raw playback time.
+    func resolveReferenceTime(forPlaybackTime playbackTime: Double, episode: BaseEpisode) async -> Double? {
```
kean, line 549: Hey, @leandroalonso. As a local fingerprinting expert, can you bless this implementation? I pointed it to https://github.com/Automattic/pocket-casts-ios/blob/trunk/docs/transcripts.md that you thoughtfully added, and this is what it came up with.

leandroalonso: @kean all good from what I can see. :)

podcasts/Bookmarks/BookmarkTranscriptSnippetExtractor.swift
```diff
@@ -0,0 +1,107 @@
...
+    func snippet(forTime time: TimeInterval, episode: BaseEpisode) async -> String? {
+        let transcriptManager = TranscriptManager(episodeUUID: episode.uuid, podcastUUID: episode.parentIdentifier())
+        guard let model = try? await transcriptManager.loadTranscript() else {
+            return nil
+        }
+
+        // Only generated transcripts have a reference fingerprint published, so
+        // only they can (and need to) be re-anchored to the reference timeline.
+        var center = time
+        if transcriptManager.isDisplayingGeneratedTranscript, FeatureFlag.syncedTranscripts.enabled,
```
SergioEstevao, line 31: When transcript is not generated we are going ahead and creating a suggestion using the current time and the transcript provided by the creator. Those can be out of sync because of dynamic ads so the bookmark suggestion can be off.

Something to check with product if we want to do this or not.

SergioEstevao (approved):
Code looks good, and it working correctly for generated transcripts.

I just left a question if we want to do this for non-generated transcripts where sync issues can happen because of Dynamic Ads

PR #4788, opened 2026-07-19
Add rewind after interruptions setting

mpgirro:
This PR adds an optional **"Rewind after interruptions"** setting: when playback resumes after an audio interruption (phone call, alarm, Siri, CarPlay navigation prompt or another app taking over audio), Pocket Casts seeks back a configurable number of seconds so listeners don't lose context. This is especially useful in the car, where navigation prompts regularly interrupt playback. A matching PR with identical behavior is open for Android: Automattic/pocket-casts-android#5611.

**Behavior**

- New picker in Settings > General, directly below Intelligent Playback Resumption: Off, 5, 10, 15, 30, 60 seconds (default: 5 seconds once the feature flag is enabled).
- Rewind applies on both automatic resume (system `.shouldResume`) and manual resume after an interruption-caused pause.
- When Intelligent Playback Resumption would also rewind the same resume, the **larger amount wins** - the two never stack.
- Interruptions while already paused, route disconnections (headphone unplug) and ducking don't trigger it; the rewind clamps at 0:00.
- CarPlay inherits the behavior automatically via the shared playback engine.
- Behind the new `interruptionRewind` feature flag (default: debug builds only). The setting is device-local; adding it to named-settings sync would need a server-side key and is left as a possible follow-up for maintainers.

**Implementation notes**

- The interruption handler now records the frozen playback position and cancels the progress timer before storing the pause state - without this, the 1-second progress timer kept moving `playedUpTo` during the interruption and `PlaybackCatchUpHelper`'s position guard rejected the adjustment on resume. This also benefits the existing Intelligent Playback Resumption after interruptions.
- `PlaybackCatchUpHelper` now computes the pause-length ladder and the interruption rewind separately and applies `max()` of the two.

## To test

1. Run a debug build (flag `interruptionRewind` is enabled by default there) and start playing an episode.
2. Go to Settings > General > Rewind after interruptions and pick e.g. 30 seconds.
3. Trigger an interruption while playing: easiest is setting a timer/alarm that fires, or receiving a phone call.
4. When the alarm is dismissed / call ends, playback resumes automatically and starts ~30 seconds before the interruption point.
5. For a long interruption where iOS doesn't auto-resume: press play manually - the rewind still applies.
6. Set the setting to Off and verify resume happens exactly where playback stopped (for pauses shorter than 5 minutes).
7. Verify no stacking: with interruption rewind 60s, interrupt for >5 minutes, resume - playback rewinds 60s (not 70s). A manual pause of >5 minutes still gets the unchanged Intelligent Playback Resumption rewind (10s).
8. Verify a manual pause or unplugging headphones followed by resume does not trigger the interruption rewind.

## Checklist

- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics. *Maintainer action needed: this PR adds the `settingsGeneralInterruptionRewindChanged` event (value property = chosen seconds); I can't edit the schema repo as an external contributor.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 8.17
 -----
+- Add an optional "Rewind after interruptions" setting that jumps back a little when playback resumes after calls, alarms, navigation and other audio [#4788](https://github.com/Automattic/pocket-casts-ios/pull/4788)
 - Fix intro-skipped time not syncing to your account when it is the only listening stat that changed [#4755](https://github.com/Automattic/pocket-casts-ios/pull/4755)
 - Fix Up Next multi-select keeping episodes selected after they leave the queue, causing a wrong count and bulk actions on episodes no longer queued [#4709](https://github.com/Automattic/pocket-casts-ios/pull/4709)
 - Fix the Help & Feedback screen following the system/in-app dark theme and using incorrect colors in the navigation bar [#4698](https://github.com/Automattic/pocket-casts-ios/pull/4698)
--- a/Modules/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
+++ b/Modules/Sources/PocketCastsUtils/Feature Flags/FeatureFlag.swift
@@ -320,6 +320,9 @@ public enum FeatureFlag: String, CaseIterable {
     /// Enable Smart Bookmarks
     case smartBookmarks
 
+    /// Enable the "Rewind after interruptions" setting: rewind playback after calls, alarms and other audio interruptions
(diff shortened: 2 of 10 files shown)


dcamozzi:
Thanks for the suggestion and for taking the time to put together a PR. We've discussed it with the team and, at least for now, we don't plan to add a separate interruption rewind setting. 

We appreciate the effort and will keep the feedback in mind as we continue to improve the listening experience.

PR #4793, opened 2026-07-20
Fix PodcastIndex chapter fetch retry after network failure

SergioEstevao:
Fixes the retrieval of PodcastIndex chapter data when a network request fails.

Previously, the entry in `dataRequestMap` was only cleared inside the fetch `Task`, on the success path (after the response was cached). If the network request threw, the failed `Task` stayed in the map, so every subsequent call for that URL returned the same failed task and chapters could never be re-fetched.

This moves the cleanup into a `defer` block in `loadChapters(_:)`, so the map entry is cleared once the call completes regardless of success or failure — a later call can retry. The now-redundant `setDataRequestMapToNil(for:)` helper is removed.

## To test

1. Play an episode that has PodcastIndex chapters while offline (or with the network otherwise failing).
2. Confirm chapters fail to load.
3. Restore connectivity and trigger a chapter fetch again for the same episode.
4. Verify the chapters now load rather than returning the earlier failed result.

## Checklist

### DIFF
--- a/podcasts/PodcastIndexChapterDataRetriever.swift
+++ b/podcasts/PodcastIndexChapterDataRetriever.swift
@@ -38,12 +38,15 @@ public actor PodcastIndexChapterDataRetriever {
             return try chapters(from: cachedResponse.data)
         }
 
+        defer {
+            dataRequestMap[urlString] = nil
+        }
+
         let task = Task<PodcastIndexEvelope, Error> { [weak self] in
             guard let self else { throw TaskError.nilSelf }
             let (data, response) = try await URLSession.shared.data(for: request)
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/133">8.17</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

kean (approved):

PR #4366, opened 2026-05-19
Fix swipe action animations when adding episodes to up next

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

Fix swipe actions animations for the "Up Next" action. Previously, the action would disappear instantly and the cell would be updated instantly as well. It also simplifies the code a bit because only `EpisodeCell` now need to observe "Up Next".

## To test

- Verify that swipe action animations work as expected

https://github.com/user-attachments/assets/e625e652-8c03-476f-8ff4-2e683a329c18

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #4795, opened 2026-07-20
Fix swipe-to-Up-Next hang in playlist episode lists

kean:
Fixes PCIOS-785

Fixes the swipe-to-Up-Next action hanging in playlist episode lists. The playlist reload path stalled the swipe animation in two ways the podcast page path doesn't:

- `PlaylistDetailFetchOperation` kept running its expensive DB queries and blocked on the main thread even after being cancelled, stalling the serial operation queue and delivering stale data. Added the missing `isCancelled` checks, matching `PodcastEpisodesRefreshOperation`.
- The fetch completion deferred its work with a nested `DispatchQueue.main.async` even though it already runs on main, letting the changeset be computed against stale data during SwipeCellKit's editing state and fall back to a full `reloadData`. It now runs synchronously.

**Note:** these were recommendations from Claude. They do make the code more sound, however, they are moot due to the following previous changes:

- https://github.com/Automattic/pocket-casts-ios/pull/4187 reduces the number of reloads and stopped them during swipe (with a delay)
- https://github.com/Automattic/pocket-casts-ios/pull/4366 removed reloads on “add to queue” actions entirely – only individual cells update

## To test

1. Open a playlist with a good number of episodes.
2. Swipe an episode row and tap the Up Next action (or do a full swipe).
3. Verify the swipe animation completes smoothly and the row updates without the list freezing.
4. Repeat quickly across several rows; verify there is no hang and the episodes land in Up Next.

## Checklist

### DIFF
--- a/podcasts/New Detail/Operation/PlaylistDetailFetchOperation.swift
+++ b/podcasts/New Detail/Operation/PlaylistDetailFetchOperation.swift
@@ -37,8 +37,12 @@ class PlaylistDetailFetchOperation: Operation, @unchecked Sendable {
                 episodeUuidToAdd: playlist.episodeUuidToAddToQueries()
             )
 
+            if self.isCancelled { return }
+
             DispatchQueue.main.sync { [weak self] in
                 guard let strongSelf = self else { return }
+                if strongSelf.isCancelled { return }
+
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
Working great

PR #4804, opened 2026-07-21
Migrate watch navigation to a value-based NavigationStack

kean:
The watch app still navigated through `NavigationLink(destination:tag:selection:)`, deprecated since watchOS 9. Every link was driven by a selection binding, and `NavigationManager` tracked the pushed screen as an `Int` index left over from the old WatchKit page-based API.

This moves the whole watch app to a value-based stack:

- `NavigationManager` owns the stack as `@Published var path: [WatchRoute]`. `navigateTo` replaces the stack (matching the old tag/selection behavior, where only one link could be active at a time); the new `push` appends, for screens opened from within Now Playing.
- New `WatchRoute` enum plus `WatchRouteView`, which maps a route to its screen and is registered once via `.navigationDestination(for:)` on the root stack, so any screen can push any route.
- All rows are now `NavigationLink(value:)`.
- `SourceInterfaceNavigationView` uses `NavigationStack(path:)` instead of the deprecated `NavigationView`.
- `NowPlayingContainerView` loses the `ZStack` of hidden zero-size `NavigationLink`s that existed only so buttons inside the paged `TabView` could push. Those buttons now call `navigationModel.push(_:)`, which also removes the `presentView` binding threaded through `NowPlayingOptions` and `NowPlayingControls`.
- `NowPlayingRow` contained its own `NavigationLink`, so in `InterfaceView` it was nested inside another link. The link moved out to the two call sites (`InterfaceView`, `UpNextView`).
- Removed `WatchInterfaceType.interfacePosition` / `.indexPosition` and the now-unused `NavigationManager.navigateToMainMenu()` / `topMostController()`, which drops the last `WKInterfaceController` usage from the navigation layer.

`.unknown`, `.interface`, and `.filter` never had a destination, so `WatchRoute.init?` returns `nil` for them and they stay no-ops rather than pushing a blank screen.

No intended behavior change. The app still launches into the current source's interface list, which the old code achieved by pre-seeding the selection binding and this preserves by seeding `path`.

## To test

Requires a paired Apple Watch or the watch simulator.

1. Launch the watch app. It should open directly on the current source's list (Phone or Watch), with the source picker one back-swipe away.
2. From the source picker, tap **Phone** and then **Watch** (Plus accounts). Each should push the matching list, and the source switch side effects should still run — the active-source indicator moves and a sync/refresh is triggered when the now playing episodes differ between the two.
3. From the interface list, open each row in turn: Now Playing, Up Next, Podcasts, Playlists, Downloads, Files. Each should push its screen, and back-swipe should return.
4. Open **Now Playing**, swipe left to the options page and tap **Episode Details**. The episode should push on top of Now Playing.
5. On the Now Playing controls page, tap the effects icon and then the Up Next icon. Both should push on top of Now Playing, not replace it.
6. In **Up Next**, tap the Now Playing row at the top — it should push Now Playing once, not twice.
7. Background the app from a screen other than the root, then relaunch. State restoration should land you back on that screen.
8. With the watch app closed, start playback on the phone so watchOS auto-launches the app. It should open on Now Playing for the phone source.

## Checklist

### DIFF
--- a/Pocket Casts Watch App/UI/Episode Lists/Up Next/UpNextView.swift
+++ b/Pocket Casts Watch App/UI/Episode Lists/Up Next/UpNextView.swift
@@ -11,7 +11,9 @@ struct UpNextView: View {
     var body: some View {
         ItemListContainer(isEmpty: viewModel.isEmpty, noItemsTitle: L10n.watchUpNextNoItemsTitle, noItemsSubtitle: L10n.watchUpNextNoItemsSubtitle) {
             List {
-                NowPlayingRow(isPlaying: $viewModel.isPlaying, podcastName: $viewModel.upNextTitle)
+                NavigationLink(value: WatchRoute.interface(.nowPlaying)) {
+                    NowPlayingRow(isPlaying: $viewModel.isPlaying, podcastName: $viewModel.upNextTitle)
+                }
                 EpisodeListView(title: L10n.upNext.prefixSourceUnicode, showArtwork: true, episodes: $viewModel.episodes, playlist: nil)
                     .padding(.vertical, 10)
             }
--- a/Pocket Casts Watch App/UI/Main Page/InterfaceView.swift
+++ b/Pocket Casts Watch App/UI/Main Page/InterfaceView.swift
@@ -6,8 +6,6 @@ import PocketCastsUtils
 import WatchKit
 
 struct InterfaceView: View {
-    @EnvironmentObject var navigationModel: NavigationManager
-
     @StateObject var upNextViewModel: UpNextViewModel
     @StateObject var downloadsViewModel: DownloadListViewModel
     private let source: Source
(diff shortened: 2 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:

pocketcasts:

kean:
I re-implemented in a separate PRs. The app now has zero warnings.

PR #4805, opened 2026-07-21
Split the smart bookmarks edit UI into its own view and view model

kean:
| 📘 Part of: # |
|:---:|

[#4772](https://github.com/Automattic/pocket-casts-ios/pull/4772) added title suggestions *into* `BookmarkEditTitleView`/`BookmarkEditViewModel`. The two implementations will keep diverging, so interleaving them behind flag checks in one file will only get messier.

Instead there are now two stacks, and `BookmarkEditTitleViewController` picks one based on `FeatureFlag.smartBookmarks`:

| Flag off | Flag on |
|---|---|
| `BookmarkEditTitleView` + `BookmarkEditTitleViewModel` | `BookmarkEditView` + `BookmarkEditViewModel` |

The flag-off files are an exact revert of what #4772 changed, with the view model renamed to match its view. The title generation infrastructure is untouched — the new view still uses it.

The new implementation also drops the `ContentSizeReader` height-sync hack (a hidden blank line pins the field height instead) and moves the title into the view model, removing the `autoApplySuggestion` Combine subject and the `pendingSuggestion`/`hasEdited` tracking in the view.

`BookmarkEditing.swift` is the glue the fence needs. When the flag ships, delete the two `...Title...` files and that glue — `BookmarkEditRouter` and `BookmarkEditTheme` live in them and will need moving over.

## To test

1. With `smartBookmarks` **off**, add and rename bookmarks from the player, an episode, and the Bookmarks tab — behaviour should be identical to trunk.
2. Turn it **on** and add a bookmark on an episode with a transcript: a spinner shows in the title field, and the title fills in automatically.
3. Repeat, but type before the suggestion arrives — your text must not be replaced; the suggestion appears below the field and applies when tapped.
4. Type a very long title in both modes — text scales down to fit, the field doesn't jump.

## Checklist

### DIFF
--- a/podcasts/Bookmarks/Editing/BookmarkEditTitleViewModel.swift
+++ b/podcasts/Bookmarks/Editing/BookmarkEditTitleViewModel.swift
@@ -1,0 +1,76 @@
+import Foundation
+import PocketCastsDataModel
+
+protocol BookmarkEditRouter: AnyObject {
+    func titleUpdated(title: String)
+    func dismiss()
+}
+
+class BookmarkEditTitleViewModel: ObservableObject {
+    weak var router: BookmarkEditRouter?
+
+    let editState: EditState
+    let maxTitleLength = Constants.Values.bookmarkMaxTitleLength
+    let originalTitle: String
+
+    /// Localized Strings
+    let headerTitle: String
+    let headerSubTitle: String
+    let saveButtonTitle: String
+    let placeholder: String = L10n.bookmarkDefaultTitle
+
+    @Published var didAppear = false
+
+    private let bookmarkManager: BookmarkManager
+    private let bookmark: Bookmark
+
+    var analyticsSource: BookmarkAnalyticsSource = .unknown
+
+    init(manager: BookmarkManager, bookmark: Bookmark, state: EditState) {
+        self.bookmarkManager = manager
+        self.bookmark = bookmark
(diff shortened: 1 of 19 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/Bookmarks/Editing/BookmarkEditTitleViewModel.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+import PocketCastsDataModel
+
+protocol BookmarkEditRouter: AnyObject {
+    func titleUpdated(title: String)
+    func dismiss()
+}
+
+class BookmarkEditTitleViewModel: ObservableObject {
```
kean, line 9: Just renaming the production version from `BookmarkEditViewModel`  to `BookmarkEditTitleViewModel` and reverting the changes made originally in https://github.com/Automattic/pocket-casts-ios/pull/4772 – no functional changes.

SergioEstevao (approved):
Working correctly with and without the flag

PR #4812, opened 2026-07-23
Fix set sleep timer via shortcut/siri when locked

### LINKED ISSUE #32: Siri: see if it's possible to invoke "Set sleep timer" command without unlocking phone
"Set sleep timer" seems to be the only Siri command we have that requires the phone to be unlocked. Is there a limitation here in what Apple makes available to us?

This has been reported in [#4965917-zen](https://woothemes.zendesk.com/agent/tickets/4965917), [#4525389-zen](https://woothemes.zendesk.com/agent/tickets/4525389) (2021), [#4517200-zen](https://woothemes.zendesk.com/agent/tickets/4517200) (2020) and possibly [#4495431-zen](https://woothemes.zendesk.com/agent/tickets/4495431) (2019).

lparry:
Fixes: https://github.com/Automattic/pocket-casts-ios/issues/32 

This PR fixes issue 32: that the set sleep timer shortcut/action cannot succeed when the device is locked.

This is a critical feature for those suffering from insomnia. When waking part way through the night, the last thing anyone wants to do is unlock their phone in order to make sure that the podcasts they need to get to sleep don't continue running unrestricted.

With this shortcut fixed and the correct bluetooth-device triggered automation configured, all one needs to do is reconnect their bluetooth sleep mask and they'll get another interval of podcasts with zero risk of them continuing to play for hours.

I also added the ability for the shortcut to natively configure the length of the sleep timer. I've never really be able to figure exactly where the old value came from, it was all a bit magic for me; so now it's possible for the shortcut to be the source-of-truth for what actually gets set when the shortcut runs.

## To test

1. Configure a shortcut that sets a sleep timer of x minutes when a simple trigger happens (eg. bluetooth device connects, or power is connected)
2. Lock the phone
3. Cause the trigger to fire by connecting the bluetooth device/connecting power
4. Unlock the phone and see that the sleep timer is enabled and configured to the right duration.

## Checklist


kean:
Hey, @lparry. Thanks for the PR! I will review it as soon as I finish with the current project.

lparry:
Thank you @kean, I've been running with a build from this git sha the past week, and the sleep timer shortcut has been happily running each time when locked now which has been great for my insomnia. 

Please let me know if anything needs to be changed or updated to get this over the line, I'm very keen to help get this into trunk and the official app 🙏

kean:
I did a quick test by setting up an Automation and it worked great. I can see how this makes it much more useful, so thanks for addressing it!

I have a couple of questions/observations about the potential changes to the behavior.

1. The "Set Sleep Timer" shortcut is not longer present in the app under "Profile / Settings / Siri Shortcuts". My understanding is that this is how you add it to Siri. Was the change intentional? Can we still support it?

Screenshot 2026-07-29 at 3 19 45 PM

2.  When you search for Pocket Casts in "Shortcuts" tab in the "Shortcuts" app, it shows only the new migrated actions first. If you select the app, it actually does surface the rest of the shortcuts. I guess it's acceptable considering this is the only shortcut that was migrated for now.

IMG_7825 IMG_7824

lparry:
Thanks @kean

re: 1, no, that wasn't intentional. I've found the line responsible and reverted that change, and have verified that it now shows back up within Pocketcasts and that the added Siri shortcut appears to run successfully with a locked device.

Screenshot 2026-07-31 at 23 05 54

re: 2, I think you're right and that once an app exposes one modern AppShortcut, the shortcut app begins hiding legacy INShortcut actions until you tap through and get shown the full list.

I could attempt to migrate a few more in a follow-up so that the single action surfaced at the top level doesn't look so lonely if you like?

kean:
> shows back up within Pocketcasts and that the added Siri shortcut appears to run successfully

Awesome, thank you!

> I could attempt to migrate a few more in a follow-up so that the single action surfaced at the top level doesn't look so lonely if you like?

If you're up for it, that would be ideal. I'd suggest doing it in a separate PR. It looks like we'd need to do it sooner than later anyway. I think it'll also be OK to merge as is once 8.18 release branch is cut and finish the migration during the next iteration.

I wonder if there is a way to connect `LocalizedStringResource` with https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/Strings%2BL10n.swift, which is used in the app for all localizations.

lparry:
> I wonder if there is a way to connect LocalizedStringResource with https://github.com/Automattic/pocket-casts-ios/blob/trunk/podcasts/Strings%2BL10n.swift, which is used in the app for all localizations.

@kean I've had a crack at getting an agent to do this, and have changed the language on my phone and can confirm that the shortcut will show up in Spanish, but I'm not really sure if this last commit is what you were meaning or not. Happy for it to be discarded if it's no good

kean:
Hey, I've opened a copy of PR with the same commit: https://github.com/Automattic/pocket-casts-ios/pull/4979, which is required by the CI in order to run automatics. It performed a review and left a few comments. Would you mind looking into those? Please, feel free to push in your local branch if you need to make some changes; I'll cherry-pick the changes.

lparry:
I know it's bad manners to force push something that someone has already committed on top of, but I realised my jj config from my work machine had leaked my work email into personal commits, so I've rewritten right back to the beginning. Since I was rewriting anyway, I've rebased it all onto a more recent trunk too. I'll dogfood the changes in response to claude's reviews for a few nights and report back

kean:
Nice work, thank you! I recreated the PR for CI https://github.com/Automattic/pocket-casts-ios/pull/4979 and smoke-tested the new changes. It worked well in the scenario I tested.

I noticed the following in the logs:

```
Sleep Timer Live Activity: unable to start activity: visibility
```

Not sure what it was referring to as the automation I used _was_ able to start the sleep timer successfully.

> 'll dogfood the changes in response to claude's reviews for a few nights and report back

If you find something, let's fix it forward. I'm going to merge the current changes.

lparry:
Thanks @kean, I was just popping on now to say I've used it for a few days now and think it's still solid with the latest round of changes and was pleasantly surprised to see you'd merged already 🙏 Thanks for helping me get this change in ❤️

PR #4819, opened 2026-07-23
Fix Bookmarks multi-selection presentation on iOS 26

kean:
Fixes PCIOS-860 – the Bookmarks multi-selection UI on iOS 26 (Liquid Glass) 

- Present the selection action bar via `safeAreaInset` under Liquid Glass, and hide the enclosing tab bar + mini player while multi-selecting (new `EnclosingTabBarHiddenModifier` / `HostControllerResolver`).
- Fix the navigation bar button colors in the profile Bookmarks list.
- Offset the row content when entering multi-select instead of shrinking it, so the title/subtitle keep the same available width and are not re-laid out. The play button collapses to just its time label (background, border, and play icon hidden) to make room.

Before / After

https://github.com/user-attachments/assets/42abc68a-9dc6-4f0f-946b-fd455ae537e1

https://github.com/user-attachments/assets/5e0b2b6f-113d-4f08-86d8-83b06a79d021

## To test

1. Open a podcast (or the profile Bookmarks screen, or the player's Bookmarks tab) that has several bookmarks.
2. Long-press a bookmark (or use the "..." > Select) to enter multi-selection.
3. Verify the row title/subtitle do **not** re-truncate or shift width — the content simply slides right to reveal the select circle, and each row's time button collapses to a plain time label (no pill background, border, or play icon).
4. Verify the selection action bar appears at the bottom and the tab bar + mini player are hidden while selecting; exit selection and confirm they return.
5. Repeat on iOS 18 and iOS 26 to confirm both the legacy overlay bar and the Liquid Glass presentation look correct.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 8.18
 -----
-- Fix the Bookmarks multi-select layout on iOS 26: rows keep their text width instead of shrinking, and the selection bar hides the tab bar and mini player correctly [#4819](https://github.com/Automattic/pocket-casts-ios/pull/4819)
+- Fix the Bookmarks multi-select layout: rows keep their text width instead of shrinking, the selection bar sits just above the mini player instead of floating too high, and on iOS 26 it hides the tab bar and mini player correctly [#4819](https://github.com/Automattic/pocket-casts-ios/pull/4819)
 
 
 8.17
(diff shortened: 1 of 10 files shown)


SergioEstevao (changes requested):
While it's looking now very good on iOS 26 it's not looking great on iOS 18:

| 1 | 2 | 3 |
| - | - | - |
| Image | Image | Image |

Can we fix this for iOS 18 too?

kean:
I fixed both of the issues on iOS 18:

Screenshot 2026-07-24 at 11 19 21 AM Screenshot 2026-07-24 at 11 34 22 AM


The design is what it is in production.

SergioEstevao (approved):
:shipit: 

Thanks for fixing it for iOS 18 too!

PR #4820, opened 2026-07-24
Store a bookmark's captured passage and its transcript location

kean:
Part of the Smart Bookmarks data layer (behind the `smartBookmarks` feature flag), built field by field. This adds the transcript passage a bookmark captures — and where that passage sits in the episode transcript — to the `Bookmark` model.

- **`Bookmark.passage`** — the captured transcript passage, moved off `BookmarkManager` and onto the model itself as a dynamic property. Still backed by `UserDefaults` for now; it moves into the database in a later step.
- **`Bookmark.passageLocation`** — the passage's start position in the transcript, kept only to disambiguate a passage that appears more than once. The passage text stays the source of truth.
- **`BookmarkTranscriptSnippetExtractor.passageRange(for:at:in:)`** — resolves the range to highlight for a stored passage: it tries the captured location first, then falls back to searching the whole transcript for the passage text.

No user-facing behavior changes on its own — this is plumbing for the feature, which stays behind the flag.

## To test

1. Enable the `smartBookmarks` feature flag.
2. Play an episode that has a generated transcript and add a bookmark.
3. Open the bookmark's edit sheet and confirm the captured passage is shown, and that re-picking a different passage updates it.
4. Close and reopen the bookmark; confirm the passage persists.
5. `passageLocation` and the `passageRange` resolver are covered by unit tests in `BookmarkTranscriptSnippetExtractorTests` (round-trip, duplicate disambiguation, text-search fallback, absent passage).

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
+++ b/Modules/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
@@ -43,6 +43,29 @@ extension Bookmark: Identifiable {
     public var id: String { uuid }
 }
 
+// MARK: - Passage
+
+extension Bookmark {
+    /// The transcript passage the bookmark captures.
+    ///
+    /// Temporarily backed by `UserDefaults`, until the passage is stored with the bookmark itself.
+    public var passage: String? {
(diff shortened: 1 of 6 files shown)


SergioEstevao (approved):
Working correctly. Confirmed that passage was correctly stored.

:shipit:

PR #4821, opened 2026-07-24
Start bookmark playback from the reference time

kean:
Fixes PCIOS-858.

**Note**: this is an initial PR. I'm also working on some performance optimizations and confirming if the UX could be improved in parallel.

A bookmark's stored time sits on this device's playback timeline, which dynamic ads shift per download — after a re-download or on another device it can point at the wrong content. This PR records the bookmark's position on the transcript's reference timeline when a passage is captured, and resolves it back to the local audio (via fingerprinting) when the bookmark is played, so playback starts on the bookmarked content.

**UI:** tapping play on a bookmark shows a spinner in place of the row's play icon while the episode loads into the player paused at the bookmark; once resolved, playback starts at the right spot and the spinner clears. Usually this is a few seconds at most; on a cold stream it can take longer (capped at 25s) since the resolver waits for the buffer to reach the bookmark — loading the episode is what starts that download, so non-local episodes work too. Seeking, pausing, or playing something else during the wait discards the pending start and leaves the player alone. Bookmarks without a reference time play immediately, as today.

Notes:

- `PlaybackManager.load`'s Up Next branch now passes the caller's `autoPlay` through to `switchTo` instead of hard-coding `true` — behavior-neutral for existing callers, needed to load the episode without autoplaying.
- New `bookmarkFingerprintCalculated` analytics event (same properties as the chapter one) so the timeouts can be tuned from field data.
- The reference time lives in `UserDefaults` alongside the captured passage for now, until it's persisted with the bookmark and synced.

## To test

Prerequisites: Smart Bookmarks feature flag; an episode with a generated transcript; bookmarks created on this device after capture (look for `[Bookmarks] Resolved bookmark …` in the logs).

1. While the episode is playing, tap play on one of its bookmarks — playback jumps to the bookmarked content near-instantly, no spinner.
2. With something else (or nothing) playing, play a bookmark of a **downloaded** episode — spinner replaces the play icon, episode appears paused in the player, then plays at the bookmarked content.
3. Repeat for a **non-downloaded** episode — same flow, longer spinner on slow networks, falling back to the stored time on timeout.
4. While the spinner is up, scrub, pause, or start another episode — playback stays where you put it and never jumps on its own afterwards.
5. With a non-empty Up Next queue, play a bookmark from a different episode — the current episode moves into Up Next; a normal episode tap with a queue still autoplays.
6. Generated-chapter taps and chapter skips still resolve and seek correctly.
7. Bookmarks without a reference time (older ones, or flag off) play immediately at the stored time.

https://github.com/user-attachments/assets/2c6e6248-ca98-4a55-84b7-f9e7a8c39e44

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
+++ b/Modules/Sources/PocketCastsDataModel/Public/Bookmarks/Bookmark.swift
@@ -77,12 +77,36 @@ extension Bookmark {
         }
     }
 
+    /// The bookmark's position on the transcript's canonical (reference) timeline, when the
+    /// episode has a generated transcript and a confident mapping was available when captured.
+    ///
+    /// Preferred over `time` wherever possible: dynamic ads shift the playback timeline per
+    /// device and download, so `time` can point at the wrong content elsewhere, whereas the
+    /// reference time is stable. Falls back to `time` when nil.
+    ///
+    /// Temporarily backed by `UserDefaults`, until it's stored with the bookmark itself.
+    public var referenceTime: TimeInterval? {
+        get { UserDefaults.standard.object(forKey: Self.referenceTimeKey(for: uuid)) as? TimeInterval }
+        nonmutating set {
+            let key = Self.referenceTimeKey(for: uuid)
+            if let newValue {
+                UserDefaults.standard.set(newValue, forKey: key)
+            } else {
+                UserDefaults.standard.removeObject(forKey: key)
+            }
+        }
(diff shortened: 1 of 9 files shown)


leandroalonso (approved):
@kean all those changes looks good to me. There's no difference in behavior.

Nice job!!

SergioEstevao (approved):
Tested with a set of bookmarks and worked correctly!

Code looks good!

Great work!

PR #4827, opened 2026-07-24
Remove MultiLineButton from SleepTimerViewController (10 warnings)

kean:
Removes the `MultiLineButton` subclass, which sized itself using the deprecated iOS 15 UIButton inset properties (`contentEdgeInsets`, `titleEdgeInsets`, `imageEdgeInsets`) — 10 deprecation warnings in total. The six sleep-timer preset buttons now use `ThemeableUIButton` directly. Outlets were already typed as `ThemeableUIButton`, so no other code changes are needed.

There is, evidently, no need for it. The current design has pretty straightforward and short rows with labels.

Screenshot 2026-07-24 at 5 38 15 PM

## To test

1. Open the player and start playback.
2. Open the Sleep Timer screen.
3. Verify the preset buttons (5/15/30/60 min, End of Episode, Custom) look and behave as before.
4. Increase the system text size (Settings → Accessibility → Display & Text Size → Larger Text) and re-open the Sleep Timer screen; confirm the preset button labels are still readable.

## Checklist


SergioEstevao (changes requested):
I think we still needed it see the images bellow, with before/after the changes using the larger fonts, and remember it can also happen on other languagues with larger strings
| Before | After |
| - | - |
| Image | Image |
| Image | Image |

kean:
> I think we still needed it see the images bellow, with before/after the changes using the larger fonts, and remember it can also happen on other languagues with larger strings

I plan to address it by limiting the maximum supposed font size. The current version is usually large for this type of a component.

kean:
I limited the maximum size to something Apple would use in similar components and also fixed the width on iPad – it was too narrow because of the values from xib. I tested it with double-length localization. It should now work for every "regular" language.

Screenshot 2026-07-27 at 1 07 18 PM

SergioEstevao (approved):
Working correctly now.

PR #4830, opened 2026-07-24
Migrate BadgeHelper to setBadgeCount (4 warnings)

kean:
`UIApplication.applicationIconBadgeNumber` is deprecated in iOS 17. Migrated all uses in `BadgeHelper` to `UNUserNotificationCenter.setBadgeCount(_:)`, which clears 4 deprecation warnings. Since `setBadgeCount` is thread-safe, the `DispatchQueue.main.async` wrappers (only needed for the main-thread-only `applicationIconBadgeNumber`) are gone.

Also removed the unused `clearNotifications()` method and its "bump the badge to 1 then back to 0" hack. That trick relied on legacy iOS coupling where zeroing the badge also cleared Notification Center entries — a behavior Apple decoupled years ago (the badge API only affects the icon badge now). The method was also dead code (nothing called it), so its branch was unreachable. `clearBadge()` now simply sets the count to 0. No behavior change.

## To test

1. In Settings → General, set the app icon badge to **Total Unplayed Episodes** (or a filter count).
2. Background the app and confirm the app icon badge shows the expected count.
3. Change the badge setting to **Off** and confirm the badge clears from the app icon.

## Checklist

### DIFF
--- a/podcasts/Utilities/BadgeHelper.swift
+++ b/podcasts/Utilities/BadgeHelper.swift
@@ -12,18 +12,20 @@ class BadgeHelper {
     private var cancelable: Cancellable?
 
     func setup() {
-        let notifications: [NSNotification.Name] = [Constants.Notifications.playlistChanged,
-                                                    Constants.Notifications.episodePlayStatusChanged,
-                                                    Constants.Notifications.episodeArchiveStatusChanged,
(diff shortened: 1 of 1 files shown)


SergioEstevao (approved):
Working correctly!

PR #4837, opened 2026-07-27
Generate a title and passage for bookmarks created offscreen

kean:
Fixes PCIOS-854.

Bookmarks made with a headphone button, in CarPlay, or while the app is in the background never open the edit sheet, so nothing generated a title or passage for them. They're now enriched and saved directly.

## To test

1. Enable the Smart Bookmarks feature flag and assign a headphone button to Add Bookmark (Settings → General → Headphone Controls).
2. Play an episode that has a transcript, then background the app.
3. Press the assigned headphone button.
4. Open the episode's bookmarks and confirm the new bookmark has a generated title and passage instead of "Bookmark".
5. Repeat with the app in the foreground but the full screen player closed — confirm the same, and that the toast's "Change Title" action opens the sheet showing the generated title.
6. With the full screen player open, add a bookmark and confirm the edit sheet still generates the title itself, with no duplicate generation.

## Checklist

### DIFF
--- a/podcasts/Bookmarks/BookmarkManager+TitleSuggestion.swift
+++ b/podcasts/Bookmarks/BookmarkManager+TitleSuggestion.swift
@@ -30,6 +30,30 @@ extension BookmarkManager {
         return await BookmarkTranscriptSnippetExtractor().snippet(forPassage: passage, at: bookmark.passageLocation, episode: episode)
     }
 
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
Tested this with my headphones and worked like magic!

:shipit:

PR #4846, opened 2026-07-28
Vectorize PCM interleaving in the fingerprint pipeline

kean:
Part of PCIOS-858. Stacked on top of #4821.

The PCM interleave step copied planar Float32 channels into a freshly allocated `[Float]` one sample at a time, per 5s chunk of decoded audio — 35% of a cold bookmark resolve. It now does a strided `vDSP_vsadd` copy per channel (`memcpy` for mono) into a scratch buffer reused across chunks. The output is byte-identical, so matching results are unchanged.

## To test

1. Play an episode with a fingerprint reference and dynamic ad insertion.
2. Create a bookmark, seek elsewhere, then tap the bookmark — playback should resume at the right spot.
3. Tap a transcript passage to seek and confirm it lands correctly.
4. Repeat on a streaming (not downloaded) episode.

## Checklist

### DIFF
--- a/podcasts/Fingerprint/FingerprintTimingManager.swift
+++ b/podcasts/Fingerprint/FingerprintTimingManager.swift
@@ -1,4 +1,5 @@
 import AVFoundation
+import Accelerate
 import Foundation
 import Fingerprint
 import PocketCastsDataModel
@@ -1216,6 +1217,7 @@ final class FingerprintTimingManager: NSObject {
             throw StreamError.bufferAllocationFailed
         }
 
+        var interleaved: [Float] = []
         while true {
             if ctx.isCancelled() { throw StreamError.cancelled }
             let nextChunkStartSeconds = Double(audioFile.framePosition) / format.sampleRate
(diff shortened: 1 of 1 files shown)


leandroalonso (approved):
@kean works correctly. However I do have a question, does this has any effect on release builds? Based on the review I did it seems for Release the compiler already does this work.

But it worths having it anyways.

kean:
I did confirm it using a release build. The compiler optimization do help massively in some of these steps, but there is still a big win thanks to the new `interleave`. I'm not sure the compiler is quite capable to do things like reuse the buffer, which the PR introduces.

```
// before
FingerprintTimingManager: bookmark resolve breakdown for 1002fc70-2338-46b2-956d-a712a814474c — 
total 1896ms: interleave 1172ms (62%), fingerprint 510ms (27%), audio decode 165ms (9%), 
match 33ms (2%), matcher build 4ms (0%), other 11ms (1%)

// after
FingerprintTimingManager: bookmark resolve breakdown for 1002fc70-2338-46b2-956d-a712a814474c — 
total 750ms: fingerprint 528ms (70%), audio decode 167ms (22%
), match 34ms (5%), interleave 6ms (1%), matcher build 4ms (1%), other 10ms (1%)
```

PR #4850, opened 2026-07-29
Add animated waveform to TV now playing screen

dcamozzi:
## Summary

Adds an animated waveform visualization to the tvOS now playing screen. Subtle vertical bars appear on either side of the podcast artwork, pulsing gently while audio is playing and smoothly fading in/out when playback starts/stops.

**Implementation details:**
- `NowPlayingWaveformView` — SwiftUI `Canvas` + `TimelineView` draws animated bars with sine-wave motion
- `isPlaying` state derived from `AVPlayer.timeControlStatus` via existing KVO observation, so it responds to transport bar play/pause
- Amplitude-based smooth transitions (2s ease in/out) for natural fade in/out
- Added as a `.background` modifier on the artwork image to avoid disrupting the existing blurred cover layout

## Demo

https://github.com/user-attachments/assets/c824d779-3586-47aa-bd34-5535ebd98164

TV2

TV0

Tv1

## To test

1. Open the app on Apple TV (or tvOS simulator)
2. Play a podcast episode
3. Observe animated waveform bars on either side of the artwork
4. Pause playback using the transport bar — bars should smoothly fade out over ~2 seconds
5. Resume playback — bars should smoothly fade back in
6. Verify the blurred cover background behind the artwork is positioned correctly (not shifted)

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/Player/MediaOverlayView.swift
+++ b/Pocket Casts TV App/UI/Player/MediaOverlayView.swift
@@ -32,6 +32,13 @@ struct MediaOverlayView: View {
                                 .blurredCoverBackground(size: proxy.size.height / scale) {
                                     Image(uiImage: uiImage)
                                 }
+                                .background {
+                                    NowPlayingWaveformView(
+                                        color: .white,
+                                        isAnimating: model.isPlaying
+                                    )
+                                    .frame(width: proxy.size.width * 0.75)
+                                }
                                 .animation(.smooth, value: isTransportBarVisible)
                             Spacer()
                         }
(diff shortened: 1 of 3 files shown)


dcamozzi:
All three review comments addressed in b9330f64d:

1. **Amplitude fade** — Replaced `withAnimation` + `@State amplitude` with a time-based lerp computed each Canvas frame. `beginFade()` captures the current amplitude and target, then each frame computes the interpolated value using an ease-in-out curve driven by `timeline.date`.

2. **Artwork size** — `NowPlayingWaveformView` now accepts `artworkSize: CGFloat` as a parameter. `MediaOverlayView` passes `proxy.size.height / scale` (the same value used for the actual artwork frame) so bars are anchored to the real artwork edge.

3. **Timeline pausing** — The `TimelineView` now uses `.animation(minimumInterval: nil, paused: !isActive)`. After a fade-out completes, a delayed task sets `isActive = false` to pause the timeline. If playback resumes before the timer fires, `beginFade(to: 1)` re-activates it and the condition check prevents the stale timer from pausing.

SergioEstevao:
Closed in favour of: https://github.com/Automattic/pocket-casts-ios/pull/4853

PR #4853, opened 2026-07-29
Add audio-reactive waveform to TV now playing screen

dcamozzi:
## Summary

Fixes [PCIOS-878](https://linear.app/a8c/issue/PCIOS-878/add-waveform-to-player)

Builds on #4850 to make the TV now playing waveform respond to actual audio levels. The bars keep their smooth sine wave motion but their peak height now scales with real audio from the episode.

**How it works:**
- Computes RMS audio levels from the existing `MTAudioProcessingTap` buffer data in `DefaultPlayer` (both volume-boost and passthrough paths)
- Exposes `currentAudioLevel` through `PlaybackManager` for UI consumption
- The waveform view polls the level each Canvas frame and uses it as a multiplier on the sine wave shape
- A `.task` polls for real audio data availability outside of Canvas (since `@State` can't be mutated inside Canvas draw closures). Once detected, locks into audio-reactive mode to avoid jarring transitions on pause
- Falls back to pure sine animation for HLS streams where no audio tap is available (`asset.tracks` is empty for HLS)

**Files changed:**
- `DefaultPlayer.swift` — Added `currentAudioLevel` property and `updateAudioLevel(from:frameCount:)` RMS computation in tap callbacks
- `PlaybackManager.swift` — Added `currentAudioLevel` computed property
- `NowPlayingWaveformView.swift` — Uses real audio level to scale sine wave bars, with HLS fallback

## Demo

https://github.com/user-attachments/assets/19113b88-39e1-4271-ae6f-da0b8d9f2786

## To test

1. Open the TV app and play a **downloaded** podcast episode (audio tap is only available for non-HLS content)
2. Observe bars responding to speech volume — louder parts produce taller bars
3. Pause playback — bars should smoothly fade out without abrupt jumps
4. Resume — bars fade back in and continue reacting to audio
5. Play a **streamed** episode (HLS) — bars should show the pure sine wave animation as a fallback

## Checklist


dcamozzi:
All three review comments addressed in 6638bc072:

1. **Data race** — Documented why the plain `Float` is acceptable: ARM64 aligned 32-bit loads/stores are atomic at the hardware level, and a lock is avoided because `updateAudioLevel` runs on the real-time audio thread where blocking is not safe.

2. **Stale level / buffer safety** — Reset `currentAudioLevel = 0` in `tapInit` so it doesn't carry over across episodes. Added a `mDataByteSize` bounds guard in `updateAudioLevel` to prevent over-reading, and documented the non-interleaved format assumption.

3. **Sticky `useAudioReactive`** — Now reset to `false` in `beginFade(to: 1)` (each play-start), and the polling `.task` is keyed on `isAnimating` so it restarts per episode. This handles the downloaded → HLS transition correctly.

kean (approved):
It looks great! The code is OK.

It's barely visible in light mode (General / Appearance / Dark), probably needs a different color. Approving since it could be done in a separate PR.

(unrelated) not sure if it's an existing behavior, but the logo seems to get a bit uncomfortably close to the episode title.

Screenshot 2026-07-29 at 2 58 30 PM

PR #4855, opened 2026-07-29
Anchor bookmarks to the transcript's reference timeline

kean:
Fixes PCIOS-852.

Dynamic ads shift the playback timeline per device, so a bookmark's stored time can point at the wrong content. Bookmarks now record a `referenceTime` on the generated transcript's canonical timeline (resolved via the fingerprint mapping at capture):

- Playing a bookmark resolves the reference time back to this device's audio before seeking, falling back to the stored time when there's no confident match.
- The transcript editor marks the bookmark's line with a glyph in the leading gutter.

**Note**: it probably needs more ways to access the transcript screen – I'll work on that

Screenshot 2026-07-29 at 11 46 23 AM

## To test

1. Enable the `smartBookmarks` and `syncedTranscripts` feature flags.
2. Play an episode with a generated transcript and create a bookmark.
3. In the edit sheet, tap **Edit** on the transcript — the bookmark's line is marked in the gutter.
4. Play the bookmark from the episode's Bookmarks tab — playback starts on the bookmarked content.

## Checklist

https://claude.ai/code/session_014spkvY4zB1ecL4zsaHVq4V

### DIFF
--- a/podcasts/Bookmarks/Editing/BookmarkEditView.swift
+++ b/podcasts/Bookmarks/Editing/BookmarkEditView.swift
@@ -69,7 +69,10 @@ struct BookmarkEditView: View {
     @ViewBuilder
     private var transcriptEditor: some View {
         if let transcript = viewModel.snippet?.transcript {
-            BookmarkTranscriptEditView(transcript: transcript, selection: $viewModel.transcriptRange, theme: theme)
+            BookmarkTranscriptEditView(transcript: transcript,
+                                       referenceTime: viewModel.referenceTime,
+                                       selection: $viewModel.transcriptRange,
+                                       theme: theme)
         }
     }
 
--- a/podcasts/Bookmarks/Editing/BookmarkEditViewModel.swift
+++ b/podcasts/Bookmarks/Editing/BookmarkEditViewModel.swift
@@ -54,6 +54,13 @@ class BookmarkEditViewModel: ObservableObject {
         snippet?.text ?? bookmark.passage
     }
 
+    /// The bookmark's position on the transcript's reference timeline. The stored value
+    /// covers bookmarks opened for editing; a snippet freshly captured for a new bookmark
+    /// carries the resolved time before it's saved.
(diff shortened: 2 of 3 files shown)


kean:
> Broad behavior change in load(...) (PlaybackManager.swift:197) 

I think we are good.

> Comment 1 (autoPlay at load:197) — concern is unfounded; the change is provably behavior-preserving
> 
> The switchTo branch only fires when the episode being loaded is different from the current one (currEpisode.uuid != episode.uuid at PlaybackManager.swift:196). I enumerated every caller of load(..., overrideUpNext: false):
> 
> - Every caller that passes (or can compute) autoPlay: false — loadCurrentEpisode (:254), play() (:295), reloadCurrentEpisodeSource (:938), internalPlayerForVideoPlayback (:954), handlePlaybackEffectsChanged (:1143), route change (:2335), remoteDeviceDisconnected (:2443), episode refresh (:2514) — loads currentEpisode() itself, so the same-uuid check means none of them can ever reach the switchTo line. remoteDeviceAutoConnected (:2461) has an explicit early-return at :2454 when the uuid differs.
> - Every caller that loads a different episode (all 14 external call sites in PlaybackActionHelper, Siri, CarPlay, watch, URL handling, Up Next VC, intents, plus internal ones at :1345, :2430, :2648) passes autoPlay: true.
> 
> So the hardcoded true happened to equal the caller's autoPlay in every previously reachable case, and the new bookmark path at :2840 is the first caller to exercise the branch with false. The reviewer's specific examples (254/295/954/2443) are precisely the ones that structurally cannot hit this line. No pre-existing "load paused with non-empty Up Next" scenario exists to regress. (Caveat worth stating in the reply: this is an unenforced invariant — but the new semantics are what the parameter always claimed to mean, so a future caller passing false gets the correct behavior.)

SergioEstevao (approved):
This is a great upgrade on the positioning of bookmarks!

Code looking good.

:shipit:

PR #4856, opened 2026-07-29
Load the podcast title for bookmarks from podcasts not in the local database

kean:
Fixes PCIOS-862.

Bookmark details resolved the podcast title only from the local database, so bookmarks synced from podcasts this device isn't subscribed to showed no podcast title at all. The header now shows a redacted placeholder while the title is fetched from the cache server, and the local lookup includes unsubscribed podcasts, which resolve without a network call.

## To test

1. Enable the Smart Bookmarks feature flag and sign in to an account that has a bookmark for a podcast this device isn't subscribed to (e.g. create one on another device or simulator, then sync).
2. Open Profile → Bookmarks and tap that bookmark.
3. Confirm the header shows a redacted placeholder where the podcast title goes, then the actual title once it loads.
4. Open a bookmark for a subscribed podcast and confirm its title still appears immediately.

Here I tested it with a podcast I'm not subscribed to:

Screenshot 2026-07-29 at 12 40 43 PM

## Checklist

### DIFF
--- a/podcasts/Bookmarks/Details/BookmarkDetailsView.swift
+++ b/podcasts/Bookmarks/Details/BookmarkDetailsView.swift
@@ -31,13 +31,7 @@ struct BookmarkDetailsView: View {
             artwork
 
             VStack(alignment: .leading, spacing: 2) {
-                viewModel.podcastTitle.map {
-                    Text($0)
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
Working correctly!

:shipit:

PR #4858, opened 2026-07-30
DRY secrets paths via xcconfig

mokagio:
See [AINFRA-2775](https://linear.app/a8c/issue/AINFRA-2775), under [AINFRA-1541](https://linear.app/a8c/issue/AINFRA-1541).

I was looking at #4731 to get it ready for a new round of review and grew overwhelmed by the various changes and the open question of how it should flow in CI vs local (when and how to call the decrypt?) and for internal vs external contributors.

As such, I've decided to extract the parts that are already good in dedicated PRs, and to rework the remainder in isolation.

This PR DRYs the build settings configuration to the paths of the secret files used at build time

### DIFF
--- a/config/PocketCasts.base.xcconfig
+++ b/config/PocketCasts.base.xcconfig
@@ -11,3 +11,8 @@ CURRENT_PROJECT_VERSION = $(VERSION_LONG)
 PRODUCT_BUNDLE_IDENTIFIER_ROOT = au.com.shiftyjelly.podcasts
 
 DEVELOPMENT_TEAM = PZYM8XX95Q // Automattic, Inc. (Company)
+
+SECRETS_DIR = $(HOME)/.configure/pocketcasts-ios/secrets
+FIREBASE_SECRETS_PATH = $(SECRETS_DIR)/GoogleService-Info.plist
+SECRETS_PATH = $(SECRETS_DIR)/pocket_casts_credentials.json
+GOOGLE_SIGN_IN_SECRETS_PATH = $(SECRETS_DIR)/GoogleSignIn.txt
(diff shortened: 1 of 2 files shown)


config/PocketCasts.base.xcconfig
```diff
@@ -11,3 +11,8 @@ CURRENT_PROJECT_VERSION = $(VERSION_LONG)
 PRODUCT_BUNDLE_IDENTIFIER_ROOT = au.com.shiftyjelly.podcasts
 
 DEVELOPMENT_TEAM = PZYM8XX95Q // Automattic, Inc. (Company)
+
+SECRETS_DIR = $(HOME)/.configure/pocketcasts-ios/secrets
+FIREBASE_SECRETS_PATH = $(SECRETS_DIR)/GoogleService-Info.plist
+SECRETS_PATH = $(SECRETS_DIR)/pocket_casts_credentials.json
```
AliSoftware, line 17: This `SECRETS_PATH` is quite a generic variable name for something specific to `pocket_casts_credentials.json`, and makes it not clear to distinguish from this `SECRET_PATH` not meaning "the path to the secrets folder" (which is what `SECRET_DIR` is for) or from this being any other secret file (like `GoogleService-Info.plist` or any other).

I'd thus recommend to rename the variable to something more explicit, like `PC_CREDENTIALS_JSON_PATH` or something.

(Maybe that's something you only had in mind for a separate, follow-up PR, though? 🤷 )

AliSoftware (approved):

config/PocketCasts.base.xcconfig
AliSoftware: Whoops I didn't realize auto-merge was on. I guess follow-up PR it will be, then 🙃

mokagio: We'll touch this file again when changing to `a8c-secrets`. I'll try to remember.

mokagio: > I'll try to remember.

Actually, that's a recipe for failure. Handed it over to an agent...

mokagio: Good call — done in #4868, as `CREDENTIALS_JSON_PATH`.

I skipped the `PC_` prefix: these build settings are Pocket Casts-only, so it'd be redundant here.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

PR #4859, opened 2026-07-30
Init `a8c-secrets` (identities only)

mokagio:
See [AINFRA-2776](https://linear.app/a8c/issue/AINFRA-2776), under [AINFRA-1541](https://linear.app/a8c/issue/AINFRA-1541).

I was looking at #4731 to get it ready for a new round of review and grew overwhelmed by the various changes and the open question of how it should flow in CI vs local (when and how to call the decrypt?) and for internal vs external contributors.

As such, I've decided to extract the parts that are already good in dedicated PRs, and to rework the remainder in isolation.

This PR sets up the `a8c-secrets` identities and `.gitattributes`. I may have gone a bit overboard with extracting this, too. But, I thought it'd be interesting to see how GitHub would render the upcoming `.age` files in the diff once this lands with the instruction to treat them as binaries.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,4 +3,4 @@ CHANGELOG.md merge=union
 
 podcasts/Strings+Generated.swift linguist-generated
 
-.a8c-secrets/**/*.age binary
+.a8c-secrets/*.age binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,5 @@ CHANGELOG.md merge=union
 *.pbxproj merge=union
 
 podcasts/Strings+Generated.swift linguist-generated
+
+.a8c-secrets/**/*.age binary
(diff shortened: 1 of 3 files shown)


.gitattributes
```diff
@@ -2,3 +2,5 @@ CHANGELOG.md merge=union
 *.pbxproj merge=union
 
 podcasts/Strings+Generated.swift linguist-generated
+
+.a8c-secrets/**/*.age binary
```
AliSoftware: Technically speaking `a8c-secrets` only accepts secret files as flat structure not directory—i.e. by design you can't `a8c-secrets encrypt foo/bar/Secrets.swift` as we refuse paths that are contain directories and are not just plain file names, and similarly `~/.a8c-secrets/<repoid>/` contains only a flat structure of decrypted files. This is intentional to reduce the risk of malicious attacks via `../` or synlink'ed dirs etc.

This means that `.a8c-secrets/` folder also will be a flat structure and all `.age` files in there will be directly under that directory, so no need for the extra `**`


```suggestion
.a8c-secrets/*.age binary
```

AliSoftware (approved):

.gitattributes
mokagio: Fair enough.

PR #4864, opened 2026-07-30
Delete the new bookmark when its edit sheet is swiped away

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-870

The bookmark edit sheet only ran its `onDismiss` callback from the close button, so swiping the sheet away kept the new bookmark with its default title instead of discarding it. The controller is now its own `UIAdaptivePresentationControllerDelegate` and routes `presentationControllerDidDismiss` through `viewModel.cancel()`, the same path the close button uses.

The second commit replaces the `((String, Bool) -> Void)?` callback with a `BookmarkEditOutcome` enum (`.saved(title:)` / `.cancelled`).

## To test

1. Play an episode, open the player, and create a bookmark.
2. Swipe the edit sheet down.
3. The bookmark should not appear in the player's Bookmarks tab. Before this change it was kept, titled "Bookmark".
4. Repeat, tapping X instead — same result. Repeat once more, saving a title — the bookmark is kept and the toast appears.
5. Edit an existing bookmark and swipe the sheet down — its title is left untouched.

## Checklist


SergioEstevao (approved):
:shipit:

PR #4873, opened 2026-07-31
Sync bookmark passage and reference time fields

kean:
Bookmark passages now sync. The server side landed in [pocket-casts-sync-api#1448](https://github.com/Automattic/pocket-casts-sync-api/pull/1448), which added `passage`, `passage_location`, and `reference_time` to `SyncUserBookmark` (fields 14–18); this wires the app up to it:

- **Storage** — `passage`, `passageLocation`, and `referenceTime` move from the temporary UserDefaults backing into real `Bookmark` columns, each group paired with a modified date for last-write-wins merging (schema v76).
- **Protobuf** — regenerated from the updated proto. `Api_SyncUserBookmark` switched to a heap-backed storage class because it crossed 16 fields, which is most of the diff there.
- **Upload** — a group is only emitted when its modified date is set, mirroring the server's emission rules (`passage` + `passage_location` share one timestamp and merge atomically).
- **Import** — a group is only applied when its modified timestamp is present, so records without the fields leave local values untouched.
- **Full sync** — `BookmarkResponse` deliberately doesn't carry these fields (same as ai_title/ai_summary in #1355), and the full-sync path deletes and re-adds every bookmark, so it now carries the local passage fields across the re-add instead of silently wiping them.

## To test

1. Enable the `smartBookmarks` and `syncedTranscripts` feature flags on two devices signed into the same staging account.
2. On device A, play an episode with a generated transcript and create a bookmark — it captures a passage.
3. Sync both devices (background/foreground the app).
4. On device B, open the bookmark's edit sheet — the passage and transcript anchor are there, and playing the bookmark seeks using the synced reference time.
5. Sign out and back in on device A (forces a full sync) — the passage survives.

## Checklist

https://claude.ai/code/session_015M8AZ4NDxm74z5EcR7g9b7

### DIFF
--- a/Modules/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
+++ b/Modules/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
@@ -6435,121 +6435,166 @@ nonisolated struct Api_SyncUserFolder: Sendable {
   fileprivate var _dateAdded: SwiftProtobuf.Google_Protobuf_Timestamp? = nil
 }
 
-nonisolated struct Api_SyncUserBookmark: Sendable {
+nonisolated struct Api_SyncUserBookmark: @unchecked Sendable {
   // SwiftProtobuf.Message conformance is added in an extension below. See the
   // `Message` and `Message+*Additions` files in the SwiftProtobuf library for
   // methods supported on all messages.
 
-  var bookmarkUuid: String = String()
+  var bookmarkUuid: String {
+    get {_storage._bookmarkUuid}
+    set {_uniqueStorage()._bookmarkUuid = newValue}
+  }
 
-  var podcastUuid: String = String()
+  var podcastUuid: String {
+    get {_storage._podcastUuid}
+    set {_uniqueStorage()._podcastUuid = newValue}
+  }
 
-  var episodeUuid: String = String()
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Modules/Sources/PocketCastsServer/Private/Protobuffer/api.pb.swift
```diff
@@ -7862,9 +7907,40 @@ nonisolated struct Api_SuggestedFolder: Sendable {
 
   var podcastUuids: [String] = []
 
+  var color: SwiftProtobuf.Google_Protobuf_Int32Value {
```
kean, line 7910: Are wonder where these new fields are coming form and why the app isn't using them. It's part of `Api_SuggestedFolder`. cc @SergioEstevao

SergioEstevao (approved):
Tested creating and editing a bookmark between two different devices.
Initial creation was correctly updated and follow up edits between different devices where sync correctly after refresh.

PR #4900, opened 2026-08-04
Add swipe actions for bookmarks

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-

Adds native SwiftUI `.swipeActions` to bookmark rows: leading edge "Share" (blue), trailing edge "Delete" (red, with the existing delete confirmation alert), using the app's custom icons.

`.swipeActions` only works inside a `List`, so this required converting the full-screen bookmark lists (Player tab, standalone Podcast bookmarks list, Episode bookmark list, Profile bookmarks list) from `ScrollView`/`LazyVStack` to `List`, styled to match the previous look.

**Known limitation:** the Podcast tab's embedded Bookmarks section (`PodcastViewController` → `BookmarksHostingCell`) stays on the non-scrolling `LazyVStack` path and does **not** get swipe actions in this PR — it's hosted non-scrolling inside a `UITableViewCell`, and a `List` doesn't reliably self-size in that configuration, so it was left out of scope rather than risk layout bugs there.

## To test

1. Open a podcast with bookmarks, go to the Player tab's Bookmarks section — swipe a row left for Delete (red), right for Share (blue).
2. Confirm swiping Delete shows a confirmation alert before the row disappears, and canceling leaves the row in place.
3. Repeat on the standalone Podcast bookmarks list, an episode's bookmark list, and the Profile bookmarks list.
4. Open the Podcast tab (not the Player tab) and view its embedded Bookmarks section — confirm rows behave as before (no swipe actions here, this is expected).

https://github.com/user-attachments/assets/9b4f62ca-8e82-4ca4-b34c-35d724653b30

## Checklist

### DIFF
--- a/podcasts/Bookmarks/List/BookmarksListView.swift
+++ b/podcasts/Bookmarks/List/BookmarksListView.swift
@@ -202,23 +202,54 @@ struct BookmarksListView<ListStyle: BookmarksStyle>: View {
 
     @ViewBuilder
     private var scrollView: some View {
-        ScrollView {
-            LazyVStack(spacing: 0) {
-                bookmarksRows
-            }
+        List {
+            bookmarksRows
         }
+        .animation(.default, value: viewModel.bookmarks)
+        .listStyle(.plain)
+        .scrollContentBackground(.hidden)
+        .environment(\.defaultMinListRowHeight, 0)
     }
 
     @ViewBuilder
     private var bookmarksRows: some View {
         ForEach(viewModel.bookmarks) { bookmark in
             BookmarkRow(bookmark: bookmark, style: style)
+                .swipeActions(edge: .leading, allowsFullSwipe: false) {
+                    if !viewModel.isMultiSelecting && viewModel.canShare(bookmark) {
+                        Button {
+                            viewModel.shareTapped(bookmark)
+                        } label: {
+                            Image("podcast-share")
+                        }
+                        .tint(style.shareSwipeTint)
+                    }
+                }
+                .swipeActions(edge: .trailing, allowsFullSwipe: false) {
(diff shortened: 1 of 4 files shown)


podcasts/Bookmarks/List/BookmarksListView.swift
```diff
@@ -202,23 +202,65 @@ struct BookmarksListView<ListStyle: BookmarksStyle>: View {
...
             Spacer(minLength: BookmarkListConstants.multiSelectionBottomPadding)
+                .listRowInsets(EdgeInsets())
+                .listRowSeparator(.hidden)
+                .listRowBackground(Color.clear)
+        }
+    }
+
+    @ViewBuilder
+    private func bookmarkRow(_ bookmark: Bookmark) -> some View {
+        if allowInternalScrolling {
```
SergioEstevao, line 238: Nit: can we use a guard here, and simplify the code layout

SergioEstevao (changes requested):
I'm only able to use the swipe actions on the Bookmark list inside profiles.

On the other places the swipe action, switches the tabs. 

Am I missing something?

kean:
> On the other places the swipe action, switches the tabs.

It is a known limitation. In "Podcast", it requires re-implementing the bookmark tab. It is currently rendered as one large `VStack` inside a containing scroll view. I have a draft PR to change it https://github.com/Automattic/pocket-casts-ios/pull/4904 and plan to finish it today.

In "Player", not sure if it needs to be added – will address it later.

SergioEstevao (approved):
After chatting, it's clear that the fix is only for the Bookmarks screen in Profile so let's :shipit:

PR #4902, opened 2026-08-04
Remove PCUnevenRoundedRectangle and use standard API for it

kean:
|  |  |
|:---:|

`PCUnevenRoundedRectangle` was a custom `Shape` fallback for `UnevenRoundedRectangle`, needed back when the app supported iOS 15. The app's deployment target is now iOS 17, so the fallback is dead weight — this removes it and switches `KidsProfileBannerView` to the standard `UnevenRoundedRectangle` API directly.

## To test

1. Build and run the app as a user eligible for the Kids Profile banner (or force it visible).
2. Confirm the banner's face image still renders with the bottom-trailing corner rounded and the other three corners square, same as before.

Screenshot 2026-08-04 at 3 45 04 PM

## Checklist


SergioEstevao (approved):
:shipit:

PR #4903, opened 2026-08-04
Fix crash when the onboarding flow is triggered mid-transition

kean:
| 📘 Part of: # |
|:---:|

Fixes an `EXC_BREAKPOINT` crash inside UIKit's `_pinInputViewsForKeyboardSceneDelegate`.

`showOnboardingFlow` could be invoked from inside a modal transition's completion window. When a full-screen onboarding modal is dismissed, `MainTabBarController.viewDidAppear` fires during `-[UIPresentationController transitionDidFinish:]`'s deferred-commit flush (`_runAfterCACommitDeferredBlocks`), and `showInitialOnboardingIfNeeded()` — which runs on every appearance, not just the first — navigates straight back into the onboarding flow. Dismissing `presentedViewController` from there made UIKit start the new transition synchronously inside that flush and trap while pinning the keyboard's input views.

Hopping a runloop turn with `DispatchQueue.main.async` starts the new transition from a clean runloop pass instead. The hop lives in `showOnboardingFlow` itself, so it covers all five entry points, not just the crashing one. The whole method body is deferred rather than only the dismiss: that keeps `OnboardingFlow.shared.begin(...)` adjacent to the presentation so the shared flow state isn't mutated a turn before the controller appears, and it re-reads `presentedViewController` at the moment it's used instead of capturing it while the modal stack is still settling. All callers are fire-and-forget, and every synchronous reader of `OnboardingFlow.shared.currentFlow`/`source` lives inside the presented flow's own controllers, so the one-turn delay is safe.

## To test

Repro is timing-dependent, so the most reliable check is that onboarding still behaves normally:

1. Delete the app and install a fresh build, then launch it. The onboarding/login flow should appear as usual.
2. Dismiss it, then background and foreground the app a few times — no crash, and onboarding should not re-appear unexpectedly.
3. Log out from Profile → Settings. The forced-logout onboarding flow should present correctly.
4. From the Profile header, tap the sign-in prompt. The login flow should present.
5. Tap a locked feature to reach the Plus upsell, then dismiss it. No crash on dismissal.

## Checklist

### DIFF
--- a/podcasts/Main/MainTabBarController.swift
+++ b/podcasts/Main/MainTabBarController.swift
@@ -739,14 +739,18 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
     }
 
     func showOnboardingFlow(flow: OnboardingFlow.Flow?) {
-        let controller = OnboardingFlow.shared.begin(flow: flow ?? .initialOnboarding, source: .onboarding)
-        guard let presentedViewController else {
-            present(controller, animated: true)
-            return
-        }
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
 
-        presentedViewController.dismiss(animated: true) {
-            self.present(controller, animated: true)
+            let controller = OnboardingFlow.shared.begin(flow: flow ?? .initialOnboarding, source: .onboarding)
+            guard let presentedViewController = self.presentedViewController else {
+                self.present(controller, animated: true)
+                return
+            }
+
+            presentedViewController.dismiss(animated: true) {
+                self.present(controller, animated: true)
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/134">8.18 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

podcasts/Main/MainTabBarController.swift
```diff
@@ -739,14 +739,18 @@ class MainTabBarController: UITabBarController, NavigationProtocol {
     }
 
     func showOnboardingFlow(flow: OnboardingFlow.Flow?) {
-        let controller = OnboardingFlow.shared.begin(flow: flow ?? .initialOnboarding, source: .onboarding)
-        guard let presentedViewController else {
-            present(controller, animated: true)
-            return
-        }
+        DispatchQueue.main.async { [weak self] in
```
kean, line 742: I suggest to refactor it to a use a dedicated `UIWindow` in the future, but for now, the best Fable (and me) came up with is a good-old `DispatchQueue.main.async`.

In general, I have a sense that presenting in `viewDidAppear` is bad based on my experience, so it seems like an OK workaround.

It worked in the scenarios I tested.

SergioEstevao (approved):
Tested on phone and iPad and it looks all is working correctly

PR #4904, opened 2026-08-04
Render the podcast Bookmarks tab as native table cells

kean:
| 📘 Part of: # |
|:---:|

Stacked on #4907 and #4908 — review those first. This is a prerequisite for adding swipe actions (#4900) to bookmarks in the "Podcast" screen.

`BookmarksHostingCell` hosted the whole list — search field, count header, rows, empty state — inside a single non-scrolling `UITableViewCell`, so there was no cell reuse, no table animations and no per-row swipe actions.

- The tab is now a header cell, a search cell and one cell per bookmark (or the empty/locked state), owned by the new `BookmarkListController`.
- It reuses the Episodes tab's search header instead of a SwiftUI lookalike, in its own section so typing doesn't lose keyboard focus.
- `BookmarksListView` drops the `allowInternalScrolling` / `showSearchField` / `useExternalActionBar` plumbing that only served the embedded case. The full-screen lists (Player tab, episode, profile) are unchanged.
- The multi select action bar is driven by the list view model directly, and rows are selectable again so the table draws its standard highlight on tap. Hiding the table's own footer for it no longer leaves the Episodes tab without one.

## To test

1. Open a podcast with bookmarks and switch to the Bookmarks tab: the search field matches the Episodes tab's and the count row reads "N bookmarks".
2. Type in the field — results filter and the keyboard stays up. Clear it with the (x), then search for something with no matches and use "Clear search" in the empty state.
3. Tap "…" for sort / multi select. Long press a row to enter multi select and confirm "…" is disabled while selecting.
4. Tap a row — it highlights and the bookmark details open.
5. Select a few bookmarks: the action bar appears above the mini player, with Edit/Share only for a single selection.
6. Cancel the selection, switch to Episodes and long press an episode: the multi select footer is there.
7. Switch between the Episodes / Bookmarks / You might like tabs a few times, and check the locked state with a free account.

Here's just on of the issue it's fixing:

https://github.com/user-attachments/assets/2fdc701a-a122-4526-b305-bce8d915ab03

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

podcasts/Bookmarks/List/BookmarksListView.swift
```diff
@@ -234,31 +171,6 @@ struct BookmarksListView<ListStyle: BookmarksStyle>: View {
         ))
     }
 
-    /// Legacy action bar handling for the previous (non-Liquid Glass) behavior.
-    /// Under Liquid Glass, `listView` adds `ActionBarView` directly via `.safeAreaInset`.
-    @ViewBuilder
-    private func actionBarView<Content: View>(_ content: @escaping () -> Content) -> some View {
```
kean, line 282: Fortunately, we no longer need all this extra complexity and conditional behavior!

podcasts/Podcasts/Podcast Page/Cells/BookmarksHostingCell.swift
```diff
@@ -1,48 +0,0 @@
...
-    override func setEditing(_ editing: Bool, animated: Bool) {
-        // Intentionally ignore editing state to avoid shifting/indenting this embedded SwiftUI view
-        super.setEditing(false, animated: animated)
-    }
-
-    func configure(with viewModel: BookmarkListViewModel, externalActionBarHandler: @escaping ((ExternalActionBarState) -> Void)) {
-        // Wrap in a top-aligned container to avoid temporary vertical centering while sizing updates.
-        contentConfiguration = UIHostingConfiguration {
-            VStack(spacing: 0) {
-                BookmarksListView(viewModel: viewModel,
```
kean, line 31: It was rendering the entire list in one cell.

podcasts/Podcasts/Podcast Page/PodcastSearchController.swift
```diff
@@ -0,0 +1,275 @@
+import UIKit
+
+/// The search header displayed above the lists of the podcast page.
+///
+/// It's only responsible for the presentation: the view controller hosting it configures what
+/// it displays and reacts to the events it reports.
+class PodcastSearchController: SimpleNotificationsViewController, UITextFieldDelegate {
```
kean: It's not a new class. I renamed `EpisodeListSearchController` to `PodcastSearchController`. Let me revert that to make the PR easier to review.

podcasts/Podcasts/Podcast Page/PodcastViewController+Options.swift
```diff
@@ -0,0 +1,171 @@
+import PocketCastsDataModel
+import PocketCastsServer
+import PocketCastsUtils
+import UIKit
+
+/// The options displayed by the overflow button of the episodes tab
+extension PodcastViewController {
+    func showEpisodeOptions() {
```
kean, line 8: I moved these from `EpisodeListSearchController`, which is now a component that just renders the UI.

SergioEstevao (approved):
Tested all different options inside Podcasts details and all working correctly.

:shipit:

PR #4906, opened 2026-08-05
AINFRA-2809: Filter tvOS TestFlight changelog

iangmaia:
| 📘 Part of: AINFRA-2809 |
|:---:|

Fixes AINFRA-2809

Separates platform-specific TestFlight changelogs from the current release notes:

- tvOS receives only entries marked with `[tvOS]`, with the marker stripped from the tester-facing notes.
- iOS excludes `[tvOS]` entries entirely.
- Empty tvOS markers are ignored, and either platform falls back to `Minor changes.` when no notes remain.

Adds five focused Minitest cases covering mixed-platform notes, exact prefix matching, marker stripping, empty markers, and fallback behavior. The tests also run in Buildkite's default queue.

## To test

1. Run `bundle exec ruby fastlane/test/helpers_test.rb`.
2. Verify all five tests pass.
3. Review the iOS and tvOS TestFlight changelog helpers against mixed and platform-only release notes.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

iangmaia:
Addressed the agentic review feedback:

- iOS now excludes tvOS-only notes; tvOS includes only marked notes and strips the marker.
- Empty tvOS markers correctly fall back to Minor changes.
- Added five helper tests; Buildkite runs them on the default queue only for Fastlane/Ruby changes, using the Ruby image derived from .ruby-version.

.buildkite/pipeline.yml
```diff
@@ -75,6 +75,17 @@ steps:
         agents:
           queue: mac
 
+      - label: ":ruby: Fastlane Helper Tests"
```
iangmaia: It crossed my mind that this is a standard that we could adopt for testing the Fastlane / automation part a bit more. I don't recall if we have similar tests in other repos.

.buildkite/shared-pipeline-vars
```diff
@@ -12,11 +12,22 @@ if [[ ! -s "$XCODE_VERSION_FILE" ]]; then
     exit 1
 fi
 XCODE_VERSION=$(sed -E 's/^~> ?//' "$XCODE_VERSION_FILE")
+RUBY_VERSION="$(cat .ruby-version)"
 
 CI_TOOLKIT_PLUGIN_VERSION="6.1.1"
 
 TEST_COLLECTOR_PLUGIN_VERSION="1.11.0"
 
+RUN_FASTLANE_TESTS=true
```
AliSoftware: So this means they will run on `trunk` (because set to `true` here then not a PR so not entering the `if` next line), right? Not that I'd expect those to take a lot of time, and they run on the `default` queue anyway so this should be fine, but not sure if strictly necessary 🙃

fastlane/lib/helpers.rb
```diff
@@ -31,3 +31,25 @@ def get_required_env!(key, env_file_path: USER_ENV_FILE_PATH)
     MSG
   end
 end
+
+# Builds the iOS TestFlight changelog without tvOS-only release notes.
+def ios_testflight_changelog(release_notes)
+  filtered_notes = release_notes.each_line.grep_v(/\A- \[tvOS\](?:\s|$)/).join.chomp
```
AliSoftware: Wouldn't it be more readable to use `start_with?`

```suggestion
  filtered_notes = release_notes.each_line.reject { |line| line.start_with?('- [tvOS]') }.join.chomp
```
Sure it's a bit longer, but it avoids having to try to decipher the RegExp for a human reading the code 🙃

fastlane/lib/helpers.rb
```diff
@@ -31,3 +31,25 @@ def get_required_env!(key, env_file_path: USER_ENV_FILE_PATH)
...
+# The marker is stripped because it is selection metadata and should not be shown to testers.
+def tvos_testflight_changelog(release_notes)
+  filtered_notes = release_notes.each_line.filter_map do |line|
+    next unless line.match?(/\A- \[tvOS\](?:\s|$)/)
+
+    note = line.sub(/\A- \[tvOS\][ \t]*/, '')
+    next if note.strip.empty?
+
+    "- #{note}"
+  end.join.chomp
```
AliSoftware: Similar to above, we could just use `start_with?`, and for the removal of the `[tvOS]` prefix, use `delete_prefix`.
Which would make this simplify to:

```suggestion
  filtered_notes = release_notes.each_line.filter_map do |line|
    next unless line.start_with?('- [tvOS]')

    note = line.delete_prefix('- [tvOS]').strip
    next if note.empty?

    "- #{note}"
  end.join("\n").chomp
```

AliSoftware (approved):
LGTM so approving to unblock; but the logic could be simplified before merging to be more readable imho

kean:
@iangmaia might be worth moving tvOS changes to a separate folder/file.

.buildkite/pipeline.yml
mokagio: We do https://github.com/Automattic/studio/blob/1e141614fc0c85480fa9bcd134e2a5e7a184a109/.buildkite/pipeline.yml#L50-L62

Definitely +1 on a standard/best-practice to encourage.

A nitpick on that note. There is a `:fastlane:` custom emoji in Buildkite. And fastlane, confusingly, should not be capitalized

image

iangmaia:
```suggestion
      - label: ":fastlane: fastlane Helper Tests"
```

fastlane/lib/helpers.rb
iangmaia: 👍 Updated on 7948ae1fc3823595aa5c8e495980afcae0463784

fastlane/lib/helpers.rb
iangmaia: I agree 👍 Updated on 7948ae1fc3823595aa5c8e495980afcae0463784

.buildkite/shared-pipeline-vars
iangmaia: Yeah, good point.  Running on trunk feels like too much, so I've set the default to be `false` on 7948ae1fc3823595aa5c8e495980afcae0463784

iangmaia:
@kean:
> @iangmaia might be worth moving tvOS changes to a separate folder/file.

Hey Alex, not sure if I get what you mean. You meant the helper functions in Ruby?

PR #4909, opened 2026-08-05
AINFRA-2806: Make localization downloads reviewable

iangmaia:
| 📘 Part of: [AINFRA-2806](https://linear.app/a8c/issue/AINFRA-2806/split-pocket-casts-ios-metadata-and-translation-downloads) |
|:---:|

Fixes AINFRA-2806
Releases V2 PR: https://github.a8c.com/Automattic/wpcom/pull/232628

Splits app translation and iOS/tvOS App Store metadata downloads out of beta preparation and release finalization.

A dedicated Fastlane lane and Buildkite pipeline now download the localized content into a phase-specific branch, lint it, and create one reviewable PR against the active release branch when changes exist.

Release finalization still checks translation progress, lints the reviewed files, and uploads iOS/tvOS metadata after the separate localization PR has been merged.

## Testing

- Triggered from ReleasesV2 the build https://buildkite.com/automattic/pocket-casts-ios/builds/17222 ([test release](https://dev-mc.a8c.com/releases-v2/release.php?product=PCiOS&version=0.0) with branch `release/0.0`)
  - This other run generated no PR (based on this branch, which has latest strings): https://buildkite.com/automattic/pocket-casts-ios/builds/17210
- Generated PR (I added some changes first to force a diff): https://github.com/Automattic/pocket-casts-ios/pull/4919
- The entire end to end flow will be tested in the next release cycle.

## Checklist


fastlane/Fastfile
```diff
@@ -424,14 +423,59 @@ platform :ios do
...
+    ensure_git_branch(branch: '^release/')
+
+    phase = phase.to_s
+    UI.user_error!('The localization update phase must contain only lowercase letters, numbers, and hyphens') unless phase.match?(/\A[a-z0-9-]+\z/)
+
+    version = release_version_current
+    base_branch = "release/#{version}"
+    pr_branch = "translations/release-#{version}-#{phase}"
+
+    Fastlane::Helper::GitHelper.create_branch(pr_branch, from: base_branch)
```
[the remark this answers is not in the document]
iangmaia: Done in 6e437b9419e0d2cbc5812f46bfd766d990f1dc6f. Reruns now delete the previous versioned localization branch (closing its PR) before recreating it from the release branch.

.buildkite/release-pipelines/download-translations-and-metadata.yml
```diff
@@ -0,0 +1,28 @@
...
+env:
+  IMAGE_ID: $IMAGE_ID
+
+steps:
+  - label: "🌐 Download Translations and Metadata"
+    command: |
+      echo '--- 🤖 Use bot for Git operations'
+      source use-bot-for-git
+
+      .buildkite/commands/checkout-release-branch.sh "${RELEASE_VERSION}"
```
mokagio: Now that CI toolkit ships the command, we should use it.

And, actually, https://github.com/Automattic/pocket-casts-ios/pull/4916 removes the command script, so we _have to_ use it.
```suggestion
      checkout_release_branch "${RELEASE_VERSION}"
```

fastlane/Fastfile
```diff
@@ -424,14 +423,61 @@ platform :ios do
...
+    else
+      push_to_git_remote(tags: false, set_upstream: true)
+      pr_url = find_or_create_pull_request(
+        repository: GITHUB_REPO,
+        base: base_branch,
+        head: pr_branch,
+        title: "Update translations and App Store metadata for `#{version}`",
+        body: "This PR integrates the latest app translations and iOS/tvOS App Store metadata from GlotPress into the `#{base_branch}` branch.",
+        labels: ['Releases'],
+        github_token: ENV.fetch('GITHUB_TOKEN')
```
mokagio:
```suggestion
        github_token: get_required_env!('GITHUB_TOKEN')
```

Sort of redundant because of the `require_env_vars!` call above, but seems good for consistency.

fastlane/Fastfile
```diff
@@ -424,14 +423,61 @@ platform :ios do
...
+    base_branch = "release/#{version}"
+    pr_branch = "translations/release-#{version}"
+
+    if Fastlane::Helper::GitHelper.branch_exists_on_remote?(branch_name: pr_branch)
+      UI.important("A previous `#{pr_branch}` branch exists on the remote. It will be deleted and GitHub will close any associated PR.")
+      Fastlane::Helper::GitHelper.delete_remote_branch_if_exists!(pr_branch)
+    end
+
+    Fastlane::Helper::GitHelper.delete_local_branch_if_exists!(pr_branch)
+    Fastlane::Helper::GitHelper.create_branch(pr_branch, from: base_branch)
```
mokagio: My research agent surfaced this, and I validated it with Opus 5 and GPT 5.6 Sol.

The two predicates use different sources of truth. In `wpmreleasetoolkit` 14.11.2:

- `branch_exists_on_remote?` queries the server with `git ls-remote`.
- `delete_remote_branch_if_exists!` checks only locally available `refs/remotes/*`.

The agent fetches only `release/<version>`. If another agent created `translations/release-<version>`, line 440 returns `true`, but line 442 finds no local tracking ref and silently returns `false`. The lane then recreates the branch locally, and the unforced push at line 458 fails non-fast-forward. This leaves the failure addressed by `6e437b9` unresolved.

Deleting the remote branch when a tracking ref is available is also the wrong lifecycle. It closes the existing PR; `find_or_create_pull_request` then finds no open PR and creates a new one. Each rerun therefore loses the existing PR number, discussion, approvals, and stable Buildkite annotation URL.

The toolkit documents the intended pattern for rolling automation: retain the same head branch and force-push refreshed contents so GitHub updates the existing PR. A force-push can still overwrite commits reviewers add directly to the automation-owned branch, so preserving those commits should not be claimed as a benefit. The blocker is destroying the review’s identity and history instead of refreshing it.

.buildkite/release-pipelines/download-translations-and-metadata.yml
iangmaia: Addressed in fd5c500ddf8f933785e6db93b1d523bbb66a4ba1. The pipeline now uses `checkout_release_branch`, consistently with the other release pipelines.

fastlane/Fastfile
iangmaia: Addressed starting in 90cb5769cbb5dbbdcd4e9c634efb9e9902173b93, with the final lifecycle refinements in aa87f3181be69182151d4b029d4b025405704248. Reruns now force-push the stable automation-owned branch, preserving the existing PR identity, and delete it only when no changes remain.

fastlane/Fastfile
iangmaia: Addressed in 40d4daf3f0d318e91c15d6cb892515838111a5a3. The PR helper now receives the token through `get_required_env!`.

mokagio (approved):
I think the [Claude review](https://github.com/Automattic/pocket-casts-ios/pull/4909#issuecomment-5268813174) comments are a bit over the board, like the one about a possible error in Releases V2. Or maybe not over the board, but something that we might want to address later in a hardening pass of the whole flow instead?

Some of the nits, however, might be worth applying. Example:

fastlane/Fastfile:509 — release_version is introduced but release_version_current is still called at :530, :531, :533. Half-migrated; either thread the local through or leave it out.

mokagio (approved):

PR #4911, opened 2026-08-06
Add CODEOWNERS entry for fastlane/metadata-tvos/

AliSoftware:
Turns out that https://github.com/Automattic/pocket-casts-ios/pull/4892 was not enough to stop our team getting ping'd on backmerge PRs we don't need to chime in on 😅  as I forgot that `fastlane/metadata-tvos` was a new thing after the recent introduction of tvOS in the repo.

## To test

Can't really test until the next backmerge PR happens


mokagio (approved):
Wonder if a glob would have worked, too. But maybe it's best to be specific and enumerate instead.

PR #4912, opened 2026-08-06
Add bookmark swipe actions to the podcast page

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-886

#4900 added Share / Delete swipe actions to the bookmark lists, but the podcast screen's Bookmarks tab is a `UITableView` (#4904) and had none.

The actions moved into `makeBookmarkSwipeActions`, next to the existing action bar factory, and both call sites render them: SwiftUI's `.swipeActions` in `BookmarksListView`, and `UIContextualAction`s in the podcast page's table.

Screenshot 2026-08-06 at 11 49 12 AM

## To test

1. Open a podcast with bookmarks and switch to the Bookmarks tab.
2. Swipe a row right — Share appears and opens the share sheet.
3. Swipe a row left — Delete appears and removes the bookmark, with the row animating out.
4. Confirm a full swipe on either edge doesn't fire the action on its own.
5. Long press a row to enter multi select and confirm rows no longer swipe.
6. Switch to the Episodes tab and confirm episode swipe actions still work.
7. Check the Player, Episode, and Profile bookmark lists still swipe as before.

## Checklist


SergioEstevao (approved):
Tested this on iOS 26 and iOS 18 and all working correctly!

PR #4914, opened 2026-08-06
Replace the MultiSelectListViewModel base class with a MultiSelectable protocol

kean:
`MultiSelectListViewModel` was a base class in the `ListViewModel` hierarchy, so every `SearchableListViewModel` inherited multi select whether or not it had a selection. It's now a `MultiSelectable` protocol with default implementations: a view model opts in by publishing `isMultiSelecting` and `selectedIDs`.

Selection is derived from the list at read time rather than mirrored into published counts, so `numberOfSelectedItems` and the `validateSelectedItems()` id pruning are gone. `tapped(item:)` moves to `BookmarkListViewModel`, where the bookmark specific behaviour already lived.

No user facing change. `BookmarkListViewModel` is the only conformer.

## To test

Bookmarks multi select is the only thing affected. Check it in all three lists: podcast page Bookmarks tab, Profile > Bookmarks, and the episode bookmarks list.

1. Long press a bookmark — the list enters multi select with that bookmark selected.
2. Long press another — the Select All Above / Select All Below picker appears; check both.
3. Tap rows to select and deselect; the action bar count follows.
4. With exactly one bookmark selected, Share and Edit appear in the action bar; with two or more, only Delete.
5. Use Select All / Deselect All in the header (podcast page: the nav bar button) and confirm the title flips between the two.
6. Delete the selected bookmarks — the list updates and multi select exits.
7. Outside multi select, tapping a bookmark still opens its details.
8. Enter multi select, select some bookmarks, then delete them all from another device and sync — multi select exits on its own.

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics.

PR #4913, opened 2026-08-06
Fix bookmark multi-select breaking when a selected bookmark changes

kean:
`Bookmark`'s `Hashable` contract was broken: `==` compared only `uuid`, but `hash(into:)` mixed in `title`, `time`, `passage` and the sync dates. Equal values hashed differently, so every `Set<Bookmark>` operation in `MultiSelectListViewModel` was unreliable — editing a selected bookmark's title (or a sync-driven refresh swapping the instance) dropped it from the set while the row still looked selected, leaving a wrong count and a row that couldn't be deselected.

`Bookmark` no longer conforms to `Hashable`/`Equatable`. The list view models are now constrained to `Identifiable` and track selection as a `Set<Model.ID>`; `selectedItems` is derived from `items`, so it always reflects the current instances.

## To test

1. Open a bookmarks list (Player tab, podcast Bookmarks tab, episode, or Profile).
2. Long press a bookmark to enter multi-select, and select a couple of bookmarks.
3. Tap Edit on a single selected bookmark, change its title, and save.
4. Verify the count stays correct and the row can still be deselected.
5. Verify Select All / Select All Above / Below, Share, and Delete still behave.

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics.

PR #4915, opened 2026-08-06
Match the table view long press options in the SwiftUI multi selection

kean:
Two fixes to the bookmarks long press.

**Select all above/below.** These were always offered, even for the first or last row, and never flipped to the Deselect variants once everything that way was already selected. The table view based lists have done both for a while, so this brings `MultiSelectable` in line with `UIViewController+MultiSelectActions`. Addresses https://github.com/Automattic/pocket-casts-ios/pull/4913#pullrequestreview-4876898738.

**Row animation.** The long pressed row's title snapped into place while the artwork and play button animated around it. `NonBlockingLongPressView` worked around `.onLongPressGesture` swallowing the drag inside a `ScrollView`, but the list became a `List` in #4900 and `BookmarkRow` was its only consumer left. Its `@State` press tracking re-rendered just the title mid update, dropping it out of the row's offset animation. Replaced with the stock `.onTapGesture` / `.onLongPressGesture(perform:onPressingChanged:)` pair, now on the whole row rather than just the title.

Stacked on #4914.

## To test

1. Open a podcast with several bookmarks and go to the Bookmarks tab.
2. Long press a bookmark in the middle of the list: the whole row, title included, animates into the multi selection together.
3. Long press it again: both *Select all above* and *Select all below* are offered.
4. Tap *Select all above*, then long press the same bookmark again: the option now reads *Deselect all above*, and tapping it clears that range.
5. Repeat 4 for *Select all below* / *Deselect all below*.
6. Long press the **first** bookmark in the list: only the below option is offered.
7. Long press the **last** bookmark in the list: only the above option is offered.
8. With a single bookmark in the list, long press it twice: no options sheet appears.
9. Drag the list to scroll, including starting the drag on a row, and swipe a row left and right for the Share and Delete actions.
10. Tap a row's artwork and title to open the bookmark, and its play button to play it.
11. Repeat a few of the above in the player's Bookmarks tab and in Profile → Bookmarks.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

SergioEstevao (approved):
Love that now the actions are consistent with episodes!

Thanks for addressing this!

:shipit:

PR #4920, opened 2026-08-07
Move fingerprint timing to Swift Concurrency (13 warnings)

kean:
Moves the fingerprint timing subsystem off GCD and onto Swift Concurrency, which clears the 13 concurrency warnings in `FingerprintTimingManager` and removes a main-thread stall.

`FingerprintTimingManager` becomes a `@MainActor` that owns only the mapping and the state machine. The background work moves onto two actors:

- **`FingerprintStreamProcessor`** — the continuous transcript pass. Owns the episode's `CheckpointMatcher`, stream-decodes the local file, and hands each batch of scored candidates back to the main actor to commit.
- **`FingerprintRegionProcessor`** — the bounded one-shot resolves (chapter taps and bookmarks). Matches into a local scratch accumulator, so nothing the highlighter depends on is ever touched.

Supporting value types were hoisted out of the manager into `FingerprintMapping.swift` and `FingerprintMatching.swift` so the actors can share them. The manager drops from 1984 to 1098 lines. Gone: `queue`, `generationQueue`, `onDemandQueue`, the `NSLock`-backed `CancellationFlag`, `dispatchPrecondition`, and the sliced `Thread.sleep` — cancellation is now `Task` cancellation, observed once per decoded chunk.

Two stalls fall out of it: the 60 Hz highlight tick no longer `queue.sync`s behind a whole match batch, and `PlaybackManager.currentTime()` is no longer read from the decode thread.

### Also fixes: every completed pass ended in failure

`AVAudioFile.read` throws at EOF rather than returning zero frames (reproduced on WAV and AAC), so `runComplete`'s `while true` loop always exited through the error path instead of its `break`. The tail flush was lost, the terminal state was `.failed` rather than `.unavailable`, and `persistMappingCache` never ran — so the mapping cache was never written for a downloaded episode and the "skip the stream, load from cache" path could never fire at all. Pre-existing, not a refactor regression. The loop is now bounded by the file's length; a genuine mid-file read failure still throws.

### Tests

26 new tests, 63 in the fingerprint suite. They synthesize an episode's audio and build its reference fingerprint from that same audio, then run the real pipeline — decode, matcher, score and dominance gates, drift filter — so the expected mapping is exact and drift fails an assertion rather than looking plausible. Covers the dynamic-ad case the subsystem exists for (an ad spliced into the listener's copy shifts the mapping by exactly its length, and highlighting stops while it plays), both one-shot resolves, the mapping-cache round trip, teardown, and the preparation guards. Plus unit tests for reference decoding and the PCM interleave.

### Behaviour changes

Otherwise a mechanical port, with a few deltas worth knowing about:

- A timed-out one-shot resolve now cancels the in-flight reference download. `FingerprintReferenceRetriever` dedupes by episode, so if transcript preparation is waiting on the same fetch it gets cancelled too and falls through to `.unavailable`. Narrow (needs no on-disk reference plus a concurrent chapter tap), but the old flag-based timeout never cancelled a `Task`. Happy to fix here if you'd rather not ship it.
- A chapter resolve that fails on "no usable checkpoints" now reports the real `is_streaming` instead of a hardcoded `false`.
- State transitions are synchronous instead of `DispatchQueue.main.async`, so `state` is correct immediately after `prepareForCurrentEpisode()` / `stop()`.
- DEBUG only: one-shot resolves no longer push their rejected candidates into the debug overlay.

Untouched: the drift filter, the score and dominance gates, `interpolate`, `isWithinMatchedContent`, `searchWindow`, the mapping cache format, the streaming grow-loop's stall logic, and every analytics event name.

## To test

Needs the `syncedTranscripts` feature flag on, plus `generatedChapters` for step 4.

1. Open a **downloaded** episode's transcript. It should reach the highlighted state and track playback as before. Scroll while it prepares — scrolling should stay smooth (this is the `queue.sync` that's gone).
2. Seek forward 5+ minutes mid-episode. Highlighting should drop out, then re-anchor at the new position within a few seconds.
3. Open the transcript for an episode that is **still streaming** (not downloaded). It should pick up the growing buffer and start highlighting once enough audio has arrived.
4. Open the chapters list on an episode with generated chapters and tap a chapter. It should seek to the right spot in the audio, not the raw reference time. Tap a second chapter while the first is still resolving — the second one wins and the first never seeks.
5. Tap a bookmark on an episode that isn't downloaded yet. Playback should start at the bookmarked content once the buffer reaches it.
6. Play a downloaded episode's transcript to the end, then close and reopen it — the second open should be instant (the mapping cache now actually gets written).
7. Close the transcript mid-preparation, then reopen it. No stuck spinner, no crash.

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Fingerprint/FingerprintFixtures.swift
+++ b/PocketCastsTests/Tests/Fingerprint/FingerprintFixtures.swift
@@ -1,0 +1,202 @@
+import AVFoundation
+import Fingerprint
+import Foundation
+
+@testable import podcasts
+
+/// The two on-disk inputs a fingerprint run needs: an audio file, and a reference
+/// fingerprint built from that same audio the way the server builds one from the
+/// publisher's copy of the episode.
+enum FingerprintFixtures {
+
+    static let sampleRate: Double = 44100
+
+    /// The reference grid `makeReference` fingerprints on, matching the shape of a
+    /// server-generated reference: a checkpoint every 2 s, each covering 8 s.
+    static let checkpointIntervalSeconds = 2
+    static let checkpointDurationSeconds = 8
+
+    enum FixtureError: Error {
+        case bufferAllocationFailed
+        case openForWritingFailed(String)
+        case openForReadingFailed(String)
+        case readFailed(String)
+    }
+
+    // MARK: - Audio
+
+    /// Writes `seconds` of deterministic audio to `url`.
+    ///
+    /// Every 250 ms the tones change, drawn from a seeded generator, so no two
+    /// moments in the file share a spectrum — the property a fingerprint relies on
+    /// to tie a window to one point on the timeline. A little broadband noise
+    /// underneath keeps the spectrum from being three bare spikes.
+    static func writeAudio(seconds: Double, to url: URL) throws {
+        // The file is only closed (and its header finalized) when the last
+        // reference to it goes away, so it stays scoped to this pool — the
+        // caller's next move is to open the same path for reading.
+        try autoreleasepool {
+            try write(seconds: seconds, to: url)
+        }
+    }
+
+    private static func write(seconds: Double, to url: URL) throws {
+        let settings: [String: Any] = [
+            AVFormatIDKey: kAudioFormatLinearPCM,
+            AVSampleRateKey: sampleRate,
+            AVNumberOfChannelsKey: 1,
+            AVLinearPCMBitDepthKey: 16,
+            AVLinearPCMIsFloatKey: false,
+            AVLinearPCMIsBigEndianKey: false,
+            AVLinearPCMIsNonInterleaved: false
+        ]
+        let file: AVAudioFile
+        do {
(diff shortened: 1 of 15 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:

kean:
I extracted the part with unit tests https://github.com/Automattic/pocket-casts-ios/pull/4957 and eliminated warnings without a massive rework https://github.com/Automattic/pocket-casts-ios/pull/4980. This PR is no longer necessary. It's too complex of a change.

PR #4928, opened 2026-08-07
Open the episode when a bookmark row's artwork is tapped

kean:
| 📘 Part of: #4761 |
|:---:|

Tapping a bookmark's artwork now opens the episode the bookmark was made in, so you can jump from a bookmark to its episode without leaving the list. The bookmark details screen gets the same tap on its header artwork.

This adds a `bookmarkEpisode(_:)` route to `BookmarkListRouter` with a default implementation for `UIViewController` conformers, so every bookmarks list gets it: the podcast screen's Bookmarks tab and the profile list. The episode opens in `EpisodeDetailViewController` as a form sheet, matching how Downloads, Starred and Up Next open episodes, and it's attributed to a new `bookmarks` analytics source on the existing `episode_detail_shown` event.

Two hosts already have the episode on screen, so opening it again would be a duplicate of what you're looking at. They opt out with `opensBookmarkEpisode`, a router property defaulting to `true`:

- **The player's Bookmarks tab**, which is showing the episode being played.
- **The episode's own Bookmarks tab**, which `EpisodeDetailViewController` embeds as a child. `present(_:)` on a child routes up to the presenting controller, so without this the tap would stack a second `EpisodeDetailViewController` for the episode you're already looking at.

Where the tap doesn't open an episode, the row falls back to its own behaviour — toggle selection while multi-selecting, otherwise open the bookmark details. That also covers **uploaded files**, because `UserEpisodeDetailViewController` needs a host that implements `UserEpisodeDetailProtocol`, which the bookmark routers don't, so it would show dead actions. On the details screen there's no fallback to make, so the artwork simply isn't a button in those cases.

The new route isn't gated by the `smartBookmarks` feature flag — it ships on its own, so the artwork tap works wherever a bookmarks list is shown today. The fallbacks route through `tapped(item:)`, which keeps its existing flag check, so with the flag off the artwork behaves exactly like the rest of the row.

## To test

1. Make sure the Smart Bookmarks feature flag is enabled.
2. Play an episode and add a couple of bookmarks to it.
3. Go to the podcast screen's Bookmarks tab. Tap the artwork on a bookmark row — the episode details sheet should open for that episode. Dismiss it.
4. Tap anywhere else on the row — the bookmark details should open as before. Tap the artwork in its header — the episode details sheet should open.
5. Repeat steps 3 and 4 from Profile → Bookmarks.
6. Open the player's Bookmarks tab and tap the artwork — the bookmark details should open, the same as tapping the rest of the row, and its header artwork shouldn't be tappable either.
7. Open an episode's details, go to its Bookmarks tab and tap the artwork — same as step 6, no second copy of the episode.
8. Enter multi-select (long press a row, or "…" → Select Bookmarks) and tap the artwork — it should toggle the row's selection instead of opening the episode.
9. Add a bookmark to an uploaded file, then tap its artwork in Profile → Bookmarks — it should open the bookmark details, not an episode.
10. Turn the Smart Bookmarks flag off and tap the artwork — the episode details should still open, since this change isn't gated by the flag.

## Checklist


CookieyedCodes:
I personally would prefer the files option to open files and scroll to the episode but fair enough if thats more work

pocketcasts:

PR #4929, opened 2026-08-07
Log to the console synchronously and buffer to file under a lock

kean:
| 📘 Part of: # |
|:---:|

Every `FileLog.addMessage` wrapped its work in an unstructured `Task`, so each of the ~844 call sites paid a `Task` allocation plus an actor hop to append a string to an array. Unstructured tasks aren't FIFO, so messages raced into the actor out of order — hence the timestamp sort on each chunk, which only fixed ordering *within* a chunk.

This splits the two concerns:

- **Console output is synchronous**, straight to the `Logger` on the calling thread, so Xcode's console shows messages as they happen.
- **File buffering uses an `OSAllocatedUnfairLock` instead of an actor**, so appends preserve call order and the sort is gone. Only the flush is dispatched, onto a serial queue.

That lock needs macOS 13, so this also raises the package's macOS platform from 11 to 13. Nothing ships for macOS, and the manifest was already inconsistent — it declared macOS 11 while depending on `WrappingHStack`, which requires 13, so SwiftPM refused to resolve the graph at all.

No user-facing change, so no `CHANGELOG.md` entry.

Follow-up stacked on this one: #4930, which drops Combine from `FileLog` entirely.

## To test

Note: the `Modules-Package` scheme doesn't build on `trunk` — `PocketCastsDataModelTests` fails to compile because `Bookmark` isn't `Equatable` — so running the tests locally needs that test target temporarily disabled.

1. Run `PocketCastsUtilsTests/LogBufferTests` and confirm all 7 pass (including the new concurrent-append test).
2. Run the app and log in, so a variety of subsystems write to the log.
3. Confirm messages appear in Xcode's console in the order the code executes.
4. Open the support screen (Profile → Help & Feedback) and tap the logs button in the navigation bar.
5. Confirm the log is ordered by timestamp with no entries out of sequence, and timestamps are formatted as before.

## Checklist

### DIFF
--- a/Modules/Package.swift
+++ b/Modules/Package.swift
@@ -6,7 +6,7 @@ import CompilerPluginSupport
 let package = Package(
     name: "Modules",
     platforms: [
-        .iOS(.v17), .watchOS(.v10), .macOS(.v11), .tvOS(.v17)
+        .iOS(.v17), .watchOS(.v10), .macOS(.v13), .tvOS(.v17)
     ],
     products: XcodeSupport.products + [
         .library(
--- a/Modules/Sources/PocketCastsUtils/Logging/FileLog.swift
+++ b/Modules/Sources/PocketCastsUtils/Logging/FileLog.swift
@@ -2,114 +2,27 @@ import Combine
 import Foundation
 import os
 
-/// The destinations a log message can be written to.
-public struct LogDestination: OptionSet, Sendable {
-    public let rawValue: Int
-
-    public init(rawValue: Int) {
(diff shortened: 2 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>

.</td>
    </tr>
  </tbody>
</table>

</p>

PR #4932, opened 2026-08-10
Guard the App Store description length budget

mokagio:
One of my agents noticed that `es-ES/description.txt` vanished in f543679c9.  The reason turned out to be that `gp_downloadmetadata` writes nothing for a locale whose translation overflows `max_size`, after deleting that locale's file.

This adds per-file source budgets that leave translations room to expand, enforced by `fastlane/test/helpers_test.rb` so copy that's too long fails on the PR that writes it for faster feedback.

It also regenerates `fastlane/AppStoreStrings.po`, which is generated from these same files and had not been regenerated since `2026-08-03`. Trimming the description changes what GlotPress should be handing translators, and the `app_store_desc` translator comment now carries the limit. Leaving the `.po` behind would reproduce the exact drift that stripped the Terms of Use and Privacy Policy links from every localized description in 8.18. #4931 was regenerating the `.po` on its own; it is closed in favour of doing it here, in the same commit range as the source change.

<details>
<summary>AI-generated details</summary>

`gp_downloadmetadata` writes nothing for a locale whose translation exceeds `max_size`, having already run `FileUtils.rm_f` over that locale's files. The locale ends up with **no** file — that is how `fastlane/metadata/es-ES/description.txt` vanished in `f543679c9` — and `deliver`'s `assign_defaults` then fills the gap from `default/`, so the locale ships the **English** copy in place of its translation. Silently reverting a locale to English is the failure this guards against.

Translations run longer than English, so each English source needs room to spare under its cap. Two have almost none: `keywords.txt` is 95 of 100 and translates to exactly 100 in `it`; `description.txt` was 3946 of 4000, translating to 3990 in fr-FR, before this branch trimmed it.

## Budgets are per file, not one factor

Measured expansion differs by five times — 1.1% for the description, 5.3% for the keywords, and 5.3% of a 100 character cap is the entire margin. So each file carries its own budget, and `subtitle.txt` and `release_notes.txt`, which have no committed translations to measure against, carry the App Store Connect maximum and no budget at all.

Going over a maximum is a hard error; going over a budget only warns. The description's 3400 is the one figure that is a judgement call rather than a measurement — the observed expansion alone would allow ~3950 — so happy to move it. #4941 trimmed the copy to 3278 to get under it, and keywords sit on 95, today's source length, so nothing warns as things stand while any growth does.

## Where the check lives

The length decision is in `fastlane/lib/helpers.rb` rather than the `Fastfile` so `fastlane/test/helpers_test.rb` can reach it. That suite runs under bare `ruby` with fastlane not loaded, so the helper cannot call `UI` — the reporting stays in the lane.

That placement is most of the value here. `RUN_FASTLANE_TESTS` enables the suite for any PR touching `fastlane/*`, which matches the metadata too, so over-long copy fails on the PR that writes it rather than mid-release on whoever is running the release.

## To test

```
bundle exec fastlane update_app_store_strings
```

Silent as things stand — every file in both metadata folders is inside its budget. Pad `fastlane/metadata/default/description.txt` past 3400 characters to see the warning, and past 4000 to see it stop with a hard error instead. Both paths were exercised locally; `git restore` the file afterwards.

```
ruby fastlane/test/helpers_test.rb
```

Padding any of the four files, in either metadata folder, past its budget turns `test_shipped_metadata_fits_its_limits` red — all eight were checked that way.

## Checklist

</details>


fastlane/Fastfile
```diff
@@ -18,6 +18,11 @@ APP_STORE_SCREENSHOTS_FOLDER = File.join(FASTLANE_FOLDER, 'screenshots')
 APP_STORE_TVOS_SCREENSHOTS_FOLDER = File.join(FASTLANE_FOLDER, 'screenshots-tvos')
 APP_STORE_STRINGS_PO_PATH = File.join(FASTLANE_FOLDER, 'AppStoreStrings.po')
 APP_STORE_TVOS_STRINGS_PO_PATH = File.join(FASTLANE_FOLDER, 'AppStoreStringsTVOS.po')
+# https://developer.apple.com/help/app-store-connect/reference/app-information/platform-version-information/
+APP_STORE_DESCRIPTION_MAX_SIZE = 4000
+# Budget for the English source. Translations run longer than English, and `gp_downloadmetadata` writes
+# nothing for a locale whose translation is over the maximum — leaving it with no description at all.
```
AliSoftware: Ideally we should also fix that in `gp_downloadmetadata` itself (in `release-toolkit`) so that instead of writing nothing then deleting the `.txt` and making the translation vanish, it would error.

Maybe with an option to escape the situation when that happens so that at least it doesn't _completely_ block a release but instead offer a judgement call to decide if (1) they want to keep the existing translation of the old description instead (which could be outdated but at least it's there) or (2) they prefer removing the description and letting it fallback to the English (which at least is up-to-date), all while suggesting that the real right solution is to fix the translation in GlotPress and retry the lane.
But at least definitively not ok to leave the `description.txt` file existing but empty (which would not make it fallback at all, as the PR description indicates)

fastlane/Fastfile
```diff
@@ -1434,6 +1441,26 @@ platform :ios do
...
   end
 
+  # Checks the English App Store description against the App Store Connect maximum and the source budget.
+  #
+  # @param metadata_folder [String] Path to the Fastlane deliver metadata folder
+  #
+  def lint_app_store_description_length(metadata_folder:)
+    path = File.join(metadata_folder, 'default', 'description.txt')
+    # Read as UTF-8 explicitly: under a non-UTF-8 `Encoding.default_external` this file's 3946
+    # characters would be counted as its 4006 bytes, failing the check on a legal description.
```
AliSoftware: This comment containing specific numbers like 3946 characters vs 4006 bytes will quickly drift and not reflect reality when `description.txt` will change again. I suggest getting rid of it completely.

```suggestion
```

fastlane/Fastfile
```diff
@@ -1434,6 +1441,26 @@ platform :ios do
...
+    length = File.read(path, mode: 'r:UTF-8').length
+
+    UI.user_error!("#{path} is #{length} characters, over the #{APP_STORE_DESCRIPTION_MAX_SIZE} App Store Connect accepts.") if length > APP_STORE_DESCRIPTION_MAX_SIZE
+
+    return if length <= APP_STORE_DESCRIPTION_SOURCE_MAX_SIZE
+
+    UI.important(
```
AliSoftware: I'm wondering if `UI.important` is enough. It's gonna be visible in the logs, but who looks at the CI logs when doing a release apart when there's a CI failure?

That's a tough one because it being a hard error is also a bit harsh (especially if it ends up being over budget by one or two characters and we feel it's acceptable since that's not a hard limit but just an estimate…).

I thus wonder if this specific part shouldn't be a Dangermattic check (maybe **in addition** to keeping that check in this lane with `UI.important` like you put there), so that when someone updates the `default/description.txt` in a PR they get a very visible Dangermattic comment on the PR directly.

AliSoftware: > I thus wonder if this specific part shouldn't be a Dangermattic check (maybe **in addition** to keeping that check in this lane with `UI.important` like you put there), so that when someone updates the `default/description.txt` in a PR they get a very visible Dangermattic comment on the PR directly.

After reading Claude's code reviews, I just realized this matches its point 4 from first review:

> #### 4. The check only runs in the release lane, so it fires at the worst moment
> `fastlane/Fastfile:1434` — the lint hangs off `update_app_store_strings_source`, which only `update_app_store_strings` and `update_app_store_strings_tvos` call. Nothing in `.buildkite/pipeline.yml` invokes those. So a PR that pads `description.txt` past 4000 passes CI cleanly and the new `UI.user_error!` lands mid-release, during code freeze, on whoever is running the release rather than on whoever wrote the copy.
>
> `.buildkite/pipeline.yml:51` already has a "Linters" group with a "Lint Localized Strings Format" step. Extracting the helper into its own lane and adding it there would move the failure to the PR that causes it. That's arguably a follow-up rather than this PR's job, but it's most of the value of the check.

AliSoftware: > 'm wondering if UI.important is enough. It's gonna be visible in the logs, but who looks at the CI logs when doing a release apart when there's a CI failure?

Note that Claude's first review also mentions something along the same line for that one too:

> #### 5. The warning fires on every run today with nothing anyone can do about it
> The source is 3946 against a 3400 budget, so update_app_store_strings now prints a warning every single time, indefinitely, until someone trims 546 characters of marketing copy. A warning that's always on trains people to scroll past it

I really think a Danger warning would be a good solution for this, to warn at the time it's modified.

AliSoftware (commented):
Left some improvement suggestions, and also agree with some feedback from Claude, especially that this should ideally apply to other other metadata (though the 85% threshold chosen for the description's budget might not be the best budget estimation for things like `keywords`?)

> 1. The same bug is live in keywords.txt and subtitle.txt, which the guard doesn't cover `fastlane/Fastfile:1487-1497` — the delete-on-overflow behaviour is a property of `gp_downloadmetadata` over every entry in `app_store_metadata_target_files`, not something specific to the description. The other three caps are much tighter, and two of them are already at the edge

mokagio:
Gosh how it frustrates me when Claude opens PRs as ready for review when that's not the case. I need better rules...

fastlane/Fastfile
mokagio: Good point tracked in https://linear.app/a8c/issue/AINFRA-2880.

fastlane/Fastfile
mokagio: > so that when someone updates the `default/description.txt` in a PR they get a very visible Dangermattic comment on the PR directly.

That sound like a nice addition to Dangermattic, regardless of this particular use case.

Tracked in https://linear.app/a8c/issue/AINFRA-2881

mokagio:
Copilot's suppressed comment on `fastlane/Fastfile:1456` — the over-max message reading as an incomplete clause — is addressed in ca5503f75d45492ce861cc1cb124ca12ce358be8. Both branches now name the unit (`-character maximum` / `-character budget`); verified by padding `default/description.txt` past 4000 and running `bundle exec fastlane update_app_store_strings`.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/135">8.19</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

mokagio:
## Review audit: every open suggestion, addressed or answered

Since the `.po` regeneration moved here from #4931, this walks back through every outstanding suggestion on this PR so none is lost in the move.

### @AliSoftware

| Suggestion | Status |
| --- | --- |
| `gp_downloadmetadata` should error rather than delete the file and write nothing | Answered — tracked in [AINFRA-2880](https://linear.app/a8c/issue/AINFRA-2880) |
| The comment's absolute character counts will drift | Addressed — the comment now carries only ratios, and the one absolute figure left is the budget itself |
| `UI.important` is too quiet; this wants a Dangermattic check at edit time | Answered — tracked in [AINFRA-2881](https://linear.app/a8c/issue/AINFRA-2881) |

The two tracked threads are still open on the PR; nothing further is pending in this branch.

### Claude review

Fixed in `57b5afa`, `c210f89`, `e015f66`, `2659030`:

- **`chomp` vs `rstrip`.** This one turned out to be substantive rather than stylistic. `PoFileGenerator#create_standard_entry` stores `content.rstrip` as the msgid, so `rstrip` is what GlotPress actually receives. No behaviour change today — all eight files measure identically — but `keywords.txt` has zero slack at 95 of 95.
- **The failure message never named Apple's hard limit.** It now reports budget and maximum separately, and files without a budget still show one number. Both shapes were checked by forcing real failures.
- **The keywords budget looked like an arbitrary reserve.** The reason it cannot move — `metadata/it/keywords.txt` sitting at 100 of 100 — is now next to the number.
- **`188e985` retired the `NEW:` heading but left "new" in the sentence beneath it.** Fixed here rather than as a follow-up, because editing the description resets every GlotPress translation of it and this branch already pays that cost once. Source is now 3274 of 3400.

Answered rather than changed:

- **The `v8.17` pin is not reproducible from the lane.** Correct, and worth recording how correct: the lane re-derived the key and undid the pin again on the regeneration after the copy fix, so this branch carries two manual re-pins. The pin still beats the alternative — re-keying orphans the 8.17 translations and puts nothing in their place — but the real fix is the stale `fastlane/metadata/default/release_notes.txt`, which no automation writes. Not this PR's job; see below.
- **`update_metadata_on_app_store_connect` defaults to `with_release_notes: true`.** Confirmed. `finalize_release` passes `false`, so the automated path is safe, but a human running the lane bare would push the 8.17 copy as the current release notes. Worth folding into the release-notes follow-up rather than flipping a default here.
- **Bullet style.** Lines 9-10 of `description.txt` use `-` where the rest of the file uses ` • `. Left alone deliberately: it changes how the listing renders, so it is a copy decision rather than a cleanup.
- **`APP_STORE_METADATA_LIMITS.freeze` is shallow.** Agreed and left alone; nothing mutates the nested hashes.

### One correction to the earlier review, and one confirmation

The last review reasoned about the untranslated-locale path without being able to read the gem. Reading it: `MetadataDownloader#delete_existing_metadata` does `FileUtils.rm_f` over **every** target file for the locale before writing back only the strings GlotPress reports as `current`. So the conclusion holds, and is slightly broader than stated — after a msgid change the locale's `description.txt` disappears rather than going stale.

Also worth pinning down, since it looks like a mismatch and is not one: the what's-new msgid is 459 characters against a 458-character source. `create_whats_new_entries` guarantees a trailing newline where `create_standard_entry` rstrips. The file is right; the two entry kinds simply follow different rules.

### Still not covered by anything

Nothing fails when the `.po` drifts from its sources. The suite added here checks lengths, not freshness — a test asserting the `.po` regenerates to itself would have caught the 8.18 link regression on the PR that caused it. That is the cheapest guard of the three discussed across these reviews, and it belongs in a follow-up rather than here.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

mokagio:
Both actionable points from the last review are in, plus two corrections to the review's own verification notes.

**The failure message was asserting something false** — that was my regression in `c210f89`, and the diagnosis is exactly right: `over` distributed across both nouns. Fixed in `b07c313` with the subordinating form. Checked by forcing real failures rather than reading it:

```
description.txt is 3474 characters, over its 3400-character budget (App Store Connect's maximum is 4000).
subtitle.txt is 35 characters, over its 30-character maximum.
```

**The translator comment now carries the consequence** (`f5b3687`, regenerated in `df0c581`), on all three entries rather than just the description — `subtitle.txt` has a 30-character ceiling that German and Italian will find long before the description does. Worth noting for anyone worried this resets translations: comments are `.po` metadata, not msgids. The regeneration diff is the timestamp, three `#.` lines and the key, and no msgid moved.

**Two things the review flagged as unverified, now verified.** The sandbox couldn't run the suite or read the gem, so:

- `ruby fastlane/test/helpers_test.rb` runs green — 12 runs, 29 assertions, 0 failures — after every commit here.
- The `rstrip` claim is from the implementation, not just the commit message: `PoFileGenerator#create_standard_entry` is literally `create_entry(msgctxt, content.rstrip, comment)`. Reading the same file also settles something that looks like a mismatch and isn't — the what's-new msgid is 459 against a 458-character source because `create_whats_new_entries` guarantees a trailing newline where `create_standard_entry` rstrips.
- `MetadataDownloader#delete_existing_metadata` reads as `FileUtils.rm_f` over **every** target file for the locale before anything is written back, which makes the earlier inference about untranslated locales correct and slightly broader than stated.

**On the pin**: `commit_changes: true` committing the revert silently is the sharpest version of this yet, and it happened twice more while addressing this round — `dc16fa0` is the third re-pin on this branch. Dropping `whats_new` from `app_store_metadata_source_files` is the right shape for the fix, and it's going in the follow-up ticket rather than here.

**Left alone deliberately**: the `- ` versus ` • ` bullets and the curly/straight apostrophe mix. Both change how the listing renders, which makes them a copy decision rather than a cleanup, and I'd rather not fold a rendering change into a PR about length budgets.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

fastlane/AppStoreStrings.po
```diff
@@ -1,13 +1,13 @@
 msgid ""
 msgstr ""
-"PO-Revision-Date: 2026-08-03 19:53+0100\n"
```
iangmaia, line 3: I guess we should re-generate the `AppStoreStringsTVOS.po` as well?

fastlane/Fastfile
```diff
@@ -1444,15 +1471,15 @@ platform :ios do
...
       },
       app_store_desc: {
         path: File.join(source_metadata_folder, 'description.txt'),
-        comment: 'translators: Multi-paragraph text used to display in the Apple App Store.'
+        comment: "translators: Multi-paragraph text used to display in the Apple App Store. Limit to #{app_store_metadata_max_size('description.txt')} characters including spaces! A translation over that length is dropped, and the locale then ships with no description at all."
       },
       app_store_keywords: {
         path: File.join(source_metadata_folder, 'keywords.txt'),
-        comment: 'translators: Comma separated Keywords used for Search Engine Optimization in the Apple App Store. Limit to 100 characters including spaces and commas!'
+        comment: "translators: Comma separated Keywords used for Search Engine Optimization in the Apple App Store. Limit to #{app_store_metadata_max_size('keywords.txt')} characters including spaces and commas! A translation over that length is dropped, and the locale then ships with no keywords at all."
```
iangmaia: When the translation is dropped, isn't the English source uploaded instead for that locale or is it really dropped?

fastlane/Fastfile
```diff
@@ -1434,6 +1436,31 @@ platform :ios do
...
+  # @param metadata_folder [String] Path to the Fastlane deliver metadata folder
+  #
+  def lint_app_store_metadata_lengths(metadata_folder:)
+    APP_STORE_METADATA_LIMITS.each do |file_name, limits|
+      path = File.join(metadata_folder, 'default', file_name)
+      UI.user_error!("Could not find the App Store metadata at #{path}.") unless File.exist?(path)
+
+      # Mirror `PoFileGenerator#create_standard_entry`, which stores `content.rstrip` as the msgid:
+      # trailing whitespace never reaches GlotPress, so counting it here would be stricter than reality.
+      length = File.read(path, mode: 'r:UTF-8').rstrip.length
```
iangmaia: Could the release notes be off by one here as, from my understanding, the check removes the final newline, but the PO generator adds it back? 🤔

iangmaia: Also, probably too much of a nitpick, but the UTF-8 made me wonder whether the keywords should be measured in bytes instead of characters?

fastlane/AppStoreStrings.po
mokagio: Yes — it was out of sync, and it had drifted further than the new comments: `AppStoreStringsTVOS.po` had not been regenerated since 2026-08-03, so it was still missing the length limits entirely.

Both catalogs are regenerated now, in 7951cd2b211c5d1715eb86139232ce0a1e9ffa4a (iOS) and bbc81aae721bd768b6bade382c67e0bcd7326ac8 (tvOS), with the what's-new key re-pinned at `v8.17` in cbfd85d32658b5f4609313de4fdd58d646a7a4e9.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

fastlane/Fastfile
mokagio: You're right, and the comment was wrong. The English source *is* what gets uploaded.

Two stages, and I had only traced the first:

1. `gp_downloadmetadata` deletes the locale's files up front (`MetadataDownloader#parse_data` → `delete_existing_metadata`) and `update_key` then rejects the over-length translation without writing anything. So the locale genuinely ends up with no `description.txt` on disk.
2. But `deliver` runs `UploadMetadata#assign_defaults` before uploading, which fills every language missing a value from the `default/` folder — which is the English source.

Verified rather than read: a fixture metadata folder with `default/` plus an `es-ES/` holding only `keywords.txt` comes back as

```
description: {"fr-FR" => "DESCRIPTION FRANCAISE\n", "es-ES" => "ENGLISH DESCRIPTION\n"}
```

So the failure is a locale silently reverting to English, which is if anything harder to spot than a blank listing. Corrected in the translator comments, the lane's warning, and the `helpers.rb` header in 7c871ed97f5e67da31802cf75feff3a8315fc7c4, regenerated into both `.po` files, and fixed in the PR description too.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

fastlane/Fastfile
mokagio: Good catch — off by one for `release_notes.txt`, and only that file.

`release_notes.txt` maps to the `whats_new` key, which `PoFileGenerator` routes through `create_whats_new_entries`:

```ruby
msgid = content.end_with?("\n") ? content : "#{content}\n"
```

No `rstrip` — the trailing newline stays in the msgid. The other three go through `create_standard_entry`, which does `content.rstrip`. Measured against the current sources:

```
release_notes.txt  file= 459  rstrip= 458  msgid= 459  delta=+1
subtitle.txt       file=  27  rstrip=  27  msgid=  27  delta=+0
description.txt    file=3274  rstrip=3274  msgid=3274  delta=+0
keywords.txt       file=  95  rstrip=  95  msgid=  95  delta=+0
```

Fixed in 0a46d40685e7aa4925b7f2d1f34e40bf7ed39301: `app_store_metadata_source_length` now branches on the file, and the test suite covers both shapes.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

mokagio: Not a nitpick, but characters is the right unit here — left as is.

The thing this budget guards is `MetadataDownloader#update_key`, which is what drops a translation:

```ruby
message_len = msg.length
if data.key?(:max_size) && (data[:max_size] != 0) && (message_len > data[:max_size])
```

`msg` comes out of `JSON.parse`, so it is a UTF-8 `String` and `#length` counts characters. Counting bytes here would reject copy that the downloader would have happily kept — for `keywords.txt` in particular, where the accented and CJK locales sit right on the cap, a byte count would be far stricter than the rule it is trying to anticipate.

App Store Connect's own 100-character keyword limit is likewise documented in characters, so the two agree.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

fastlane/Fastfile
```diff
@@ -1444,15 +1471,15 @@ platform :ios do
       whats_new: File.join(source_metadata_folder, 'release_notes.txt'),
       app_store_subtitle: {
         path: File.join(source_metadata_folder, 'subtitle.txt'),
-        comment: 'translators: Subtitle to be displayed below the application name in the Apple App Store. Limit to 30 characters including spaces and commas!'
+        comment: "translators: Subtitle to be displayed below the application name in the Apple App Store. Limit to #{app_store_metadata_max_size('subtitle.txt')} characters including spaces and commas! A translation over that length is dropped, and the locale then ships with no subtitle at all."
```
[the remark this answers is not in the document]
mokagio: Correct, and it had drifted further than this: `AppStoreStringsTVOS.po` had not been regenerated since 2026-08-03, so it was missing the length limits entirely, not just the new overflow sentence.

`update_app_store_strings_tvos` has been run and committed in bbc81aae721bd768b6bade382c67e0bcd7326ac8, alongside the iOS catalog in 7951cd2b211c5d1715eb86139232ce0a1e9ffa4a, with the what's-new key re-pinned at `v8.17` in cbfd85d32658b5f4609313de4fdd58d646a7a4e9.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

mokagio:
## Review feedback addressed

@iangmaia's third question found a real error in the premise, so the correction runs wider than the diff for it: the translator comments, the lane's warning, both `.po` files and the PR description all claimed a locale ships with *no* metadata when its translation overflows. It ships the **English** source — `deliver`'s `assign_defaults` fills every missing value from `default/` before upload. Verified against `Deliver::UploadMetadata` with a fixture folder rather than by reading.

| Comment | Status |
| --- | --- |
| `AppStoreStrings.po:3` (@iangmaia) — regenerate `AppStoreStringsTVOS.po` too | Done in 7951cd2b211c5d1715eb86139232ce0a1e9ffa4a, bbc81aae721bd768b6bade382c67e0bcd7326ac8, cbfd85d32658b5f4609313de4fdd58d646a7a4e9. It was stale since `2026-08-03`, so it was missing the limits entirely |
| `Fastfile:1482` (@iangmaia) — is the English source uploaded instead, or is it really dropped? | Uploaded. Corrected in 7c871ed97f5e67da31802cf75feff3a8315fc7c4 |
| `Fastfile:1450` (@iangmaia) — release notes off by one against the PO generator | Confirmed, `whats_new` keeps the trailing newline while the rest are `rstrip`ed. Fixed in 0a46d40685e7aa4925b7f2d1f34e40bf7ed39301, with tests for both shapes |
| `Fastfile:1450` (@iangmaia) — should keywords be measured in bytes? | Not changed. `MetadataDownloader#update_key` counts characters, and that is the check the budget anticipates |
| `Fastfile:1474` (Copilot) — tvOS catalog out of sync | Same as the first row |

Verified with `ruby fastlane/test/helpers_test.rb` (14 runs, 0 failures), `bundle exec rubocop`, and a run of both `update_app_store_strings` lanes.

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

iangmaia:
@mokagio FYI I've added 3d20ac7778cf6304cb2e86d3a3557ea197a6f3e1 re-generating the POs and updating a comment

iangmaia (approved):
Thank you for considering / addressing the issues we discussed!

PR #4940, opened 2026-08-11
[DO NOT MERGE] Resolve playback sources in one place instead of scattered HLS checks

kean:
| 📘 Part of: # |  
|:---:|

**Note**: is going to be used as a reference, not planning to merge this massive change as is.

HLS was handled by ~20 scattered `hasHLSStream(episode)` checks, each re-deriving what the player would actually open. Worst of them was `EpisodeManager.isVideo(_:)` — called per episode cell — which transitively hit `FileManager.fileExists` and read mutable `PlaybackManager` state, a back-edge from a static utility into the player singleton.

This adds `PlaybackSource` (a url plus a kind: `.localFile` / `.progressive` / `.hls`), resolved in exactly one place, and a capability table stating how the kinds differ — buffering, rate cap, cacheability, track visibility, advertised content type. Call sites now ask what a source *can do* rather than what it *is*, so resolution is pure and the differences are stated once.

Two bugs fall out of it:

- **tvOS speed cap.** The cap asked whether the episode *advertised* an HLS stream, not what it would play. A downloaded episode was capped to 2x in the UI while playing its local file at 3x.
- **Video detection.** `setAudioSessionVideoProperties()` and the watch now-playing media type asked `videoPodcast()`, missing video carried by a stream the feed doesn't describe as video.

Commits are self-contained and worth reviewing in order. No other behaviour changes.

## To test

Needs the `hls` feature flag enabled.

1. Play an episode that advertises an HLS stream. Video renders, and the speed control caps at 2x.
2. Download that same episode and play it. It plays the local file: speed now goes above 2x, and the cell shows no video indicator.
3. With a downloaded video episode playing, toggle video on in the player shelf. Playback switches to the stream and video appears.
4. Turn on Settings → Audio only. Video is hidden and the per-episode shelf toggle disappears.
5. Cast an HLS episode to a Chromecast. The receiver renders it as video, not audio-only.
6. On tvOS, play a downloaded episode that advertises HLS and confirm the speed control goes past 2x.
7. Regular progressive (mp3) episodes: streaming, stream-and-cache, and downloaded playback all behave as before.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

kean:
Local workflow-based code review:
```

⏺ Code review(xhigh · 13 findings)
  ⎿  podcasts/EpisodeManager.swift
       ● 487 [correctness]    `urlForEpisode` no longer consults `PlaybackManager.shouldStreamVideoDespiteDownload`, so the URL the player actually opens ignores
                              the user's "watch video" choice for a downloaded episode.
       ● 483 [correctness]    `isVideo(_:)` with default options no longer reflects the playback manager's streaming choice, so list UI stops marking the episode
                              the user is currently watching as video.
       ● 483 [efficiency]     `isVideo` now resolves the full playback source (two `fileExists` stats for downloaded episodes) where it previously
                              short-circuited on the cheap HLS-url check.
       ● 440 [simplification] Both local-file branches in `playbackSource` construct the identical `PlaybackSource`, duplicating a long line.
     podcasts/PlaybackManager.swift
       ● 277 [correctness]    `setAudioSessionVideoProperties()` now fires for any episode with an HLS stream, switching the audio session out of `.spokenAudio`
                              into `.moviePlayback` for audio-only podcasts.
       ● 855 [test-coverage]  `sourceOptions(for:)` — the only new logic that re-establishes the "user turned video on for a downloaded episode" preference — has
                              no test coverage; every new test passes `PlaybackSource.Options` explicitly to `EpisodeManager` or asserts the static capability
                              table, so nothing exercises the PlaybackManager seam where the diff's behavior actually lives.
       ● 912 [simplification] `VideoPresentation` enum and the `videoPresentation` property are only ever consumed by `shouldRenderVideo()`; `.hidden` and
                              `.unavailable` are never matched.
       ● 858 [simplification] `streamingSource(for:)?.kind.mayCarryUndeclaredVideo == true` is duplicated in `sourceOptions` and `canToggleVideoRendering` with
                              no shared name.
       ● 870 [naming]         The new `currentSource` property collides with the pre-existing, unrelated `currentSource` analytics concept used a dozen times in
                              the same file, so one identifier now means both "the media the player opened" and "the UI surface playback was started from".
     podcasts/WatchNowPlayingHelper.swift
       ●  41 [correctness]    A per-episode now-playing builder now reads global current-episode state, reporting `MPNowPlayingInfoMediaType.video` for
                              audio-only HLS episodes on watchOS.
     PocketCastsTests/Tests/Playback/PlaybackSourceTests.swift
       ●   9 [test-coverage]  `allKinds` is a hand-maintained duplicate of `PlaybackSource.Kind`'s case list; because `Kind` isn't `CaseIterable`, a newly added
                              kind silently escapes every "all other kinds" assertion in the file that is supposed to pin the capability table.
     PocketCastsTests/Tests/Utilities/EpisodeManagerTests.swift
       ● 253 [test-coverage]  `testStreamingSourceIsNeverHLSForUserEpisode` asserts inequality against a value that is `nil` in the test environment, so the
                              assertion is satisfied without ever resolving a UserEpisode source.
     podcasts/Google Cast/GoogleCastManager.swift
       ● 193 [simplification] The streaming source is resolved twice on the cast path — once directly and once inside `isVideo(_:options:)`.
```

pocketcasts:

PR #4941, opened 2026-08-12
Trim the App Store description under its budget

mokagio:
Stacked on #4932 — merge that one first.

#4932 gives `description.txt` a 3400 character budget so translations, which run longer than the English source, stay under the 4000 App Store Connect accepts. The source is 3946, so as it stands the lane would warn on every release run forever, which is the one open concern on that PR. This clears it: 3278, with room for the copy to grow before the warning comes back.

**This needs a marketing sign-off, not an infrastructure one.** The trim is redundancy rather than features — two paragraphs of filler, the Airplay/Chromecast/Sonos/CarPlay bullets that the `Everywhere` line four lines above already names, a duplicate Smart Playlists bullet, and the `Audible audiobooks podcasts.` fragment. Themes and Apple Watch say the same thing in fewer words. Nothing the app does went missing, but which 650 characters to give up is a copy call, so cut differently if you'd rather.

`app_store_desc` is a single msgid for the whole description, so any edit invalidates every locale all-or-nothing. That is why two aged references ride along rather than waiting for a follow-up: one retranslation cycle instead of three. Until translators catch up, non-English locales serve the pre-trim copy — worth not landing immediately before a release cut.

Last commit ratchets the metadata test from the maximum to the budget, so drift is caught on the PR that causes it rather than in a release log nobody reads. It leaves `keywords.txt`, already on exactly 95 of its 95 budget, no headroom — deliberate, given its translations reach 100 of 100, but say so if you'd rather 90.

## To test

```
ruby fastlane/test/helpers_test.rb
```

Append 130 characters to `fastlane/metadata/default/description.txt` and re-run: `test_shipped_metadata_fits_its_limits` goes red with `is 3408 of 3400 characters` — over budget, still far under the 4000 maximum the previous assertion allowed. `git restore` afterwards.

All eight tracked files across both metadata folders now return `:ok`, so `bundle exec fastlane update_app_store_strings` warns about nothing.

## Checklist

### DIFF
--- a/fastlane/metadata/default/description.txt
+++ b/fastlane/metadata/default/description.txt
@@ -9,11 +9,9 @@ Your listening, your way. Organize your episodes with our powerful new Playlists
 - Playlists: Curate your perfect mix. Combine episodes from different podcasts, build collections around specific themes or guests, and even add episodes from podcasts you don't follow.
 - Smart Playlists: We’ve upgraded our famous Filters. Build lists that automatically update based on rules you set, like release date, duration, or download status.
 
-With a unique set of simple yet powerful features and a limitless podcast database, you're sure to find something you love.
-
 WE SWEAT THE DETAILS
  • Design: Easily listen to, manage and find new podcasts.
- • Themes: Whether you're a dark or light theme person we've got you covered. We even have you OLED lovers covered with our Extra Dark theme.
+ • Themes: Dark, light, and an Extra Dark theme for those of you on OLED screens.
  • Universal: Custom but familiar iPad interface, supporting Slide Over, Split View and Picture in Picture.
  • Everywhere: CarPlay, AirPlay, Chromecast and Sonos. Listen to your podcasts in more places than ever before.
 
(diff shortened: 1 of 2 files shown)


AliSoftware:
> This needs a marketing sign-off, not an infrastructure one.

Any idea who we should ping / add as reviewer for that?

iangmaia (approved):

iangmaia:
Approved it to unblock as the code looks ok but as [mentioned](https://github.com/Automattic/pocket-casts-ios/pull/4941#issuecomment-5265999199) (only seen it afterwards) it'd be good to get an ✅ from someone in the team (cc @kean).

fastlane/metadata/default/description.txt
```diff
@@ -26,28 +24,21 @@ POWERFUL PLAYBACK
...
  • Sync: Subscriptions, Up Next, listening history, playback and filters are all securely stored in the cloud. You can pick up where you left off on another device and even the web.
  • Refresh: Let our servers check for new episodes, so you can get on with your day.
  • Notifications: We'll let you know when new episodes arrive, if you like.
  • Auto download: Automatically download episodes for offline playback.
- • Smart Playlists: Organize your episodes using custom rules.
  • Storage: All the tools you need to keep your podcasts tamed.
- • Audible audiobooks podcasts.
 
 ALL YOUR FAVORITES
- • Discover: Subscribe to any podcast in iTunes and more. Browse by charts, networks and categories.
```
kean, line 45: iTunes, wow

kean (changes requested):
I see the point of making the source description shorter, but it did remove some items like "Smart Playlist" outright, which doesn't seem ideal.

(I'd keep the "iTunes" replacement)

kean:
> Approved it to unblock as the code looks ok but as https://github.com/Automattic/pocket-casts-ios/pull/4941#issuecomment-5265999199 (only seen it afterwards) it'd be good to get an ✅ from someone in the team (cc @kean).

Let's ping someone from the product, @david-gonzalez-a8c would you mind giving a sign-off? Do you think it's worth trimming the source (English) version to increase the chances that other translations fit?

I can provide more context if needed. TL;DR the app infra team is restoring the scripts that pull the App Store listing localizations from GlotPress.

david-gonzalez-a8c:
Looks good, thanks for coming up with the suggestions.

mokagio:
I'm going to cash in @david-gonzalez-a8c 's approval to merge this into #4932 and move forward.

@kean please don't let this stop you from editing the description further. As much as I love Pocket Casts as a dev and a user, I'm looking at this with dry apps infra lenses and Claude's pass met my "reduce budget" criterion. You are in a much better position to update the description than I am.

PR #4948, opened 2026-08-12
PlaybackManager: Incremental @MainActor adoption from the UI side – Part 2

kean:
Stacked on #4947.

Part 1 left three per-method `@MainActor` annotations as placeholders where the whole type should be isolated. This promotes them to type-level and closes the ring of callers that pulled in.

- `SearchResultCellModel`, `EpisodeRowViewModel` (tvOS) and `EpisodeUpNextActions` are now `@MainActor`. The redundant `@MainActor` on the inner `Task`s and on `play()` goes away with it.
- Five call sites built `EpisodeRowViewModel` off-main inside `Task.detached`. Rather than weaken the annotation, each one now keeps the database work off-main and constructs the view models on main. `HomeViewModel` gets slightly faster out of this — it was calling `parentPodcast(dataManager:)` for the Up Next rows *inside* `MainActor.run`, so that query moves off-main.

One latent off-main call surfaced and is fixed: `DiscoveryEpisodeMenuButtons.load(_:)` ran its completion on the generic executor, so Play Next / Play Last in the Discover and Search context menus mutated the Up Next queue, wrote a `@Binding`, and showed a toast off the main thread.

The `Task { @MainActor in }` added to `PodcastDetailViewModel`'s archive observer is redundant at runtime — the publisher is already `.receive(on: DispatchQueue.main)`. It is there only because the class itself isn't isolated yet. `PodcastDetailViewModel`, `HomeViewModel` and `UpNextViewModel` should follow, which would delete the `MainActor.run` dance in all three, but that first needs `TVDataManager.loadPodcast` and `fetchEpisodes` to move off-main or the podcast-detail query lands on the main thread.

## To test

tvOS — the episode lists still populate and act correctly:

1. Home → Up Next and New Releases rows populate; play an episode from each.
2. Podcasts → open a podcast; the episode list loads. Toggle Show Archived and change the sort order; the list reloads in the new order.
3. Podcasts → archive an episode from the context menu; the list refreshes.
4. Up Next → the queue populates; Play Next, Play Last and Remove From Up Next work.
5. Starred and Listening History → both lists populate.
6. Discover → open the context menu on an episode; Play Next and Play Last add it to Up Next and show the toast.

iOS:

7. Search → play an episode result; the spinner appears for slow loads, then playback starts. Pause it again.

## Checklist

### DIFF
--- a/Pocket Casts TV App/UI/History/ListeningHistoryViewModel.swift
+++ b/Pocket Casts TV App/UI/History/ListeningHistoryViewModel.swift
@@ -28,18 +28,18 @@ class ListeningHistoryViewModel {
 
     private func fetchLocalData() {
         Task.detached { [dataManager] in
-            let episodes = self.loadEpisodeViewModels(using: dataManager)
+            let episodes = self.loadEpisodes(using: dataManager)
             await MainActor.run { [weak self] in
                 guard let self else { return }
-                self.episodes = episodes
+                self.episodes = episodes.map { EpisodeRowViewModel(episode: $0.episode, podcast: $0.podcast, source: .listeningHistory) }
                 self.state = episodes.isEmpty ? .empty : .ready
             }
         }
     }
 
-    nonisolated private func loadEpisodeViewModels(using dataManager: DataManager) -> [EpisodeRowViewModel] {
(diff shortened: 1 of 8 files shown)


kean:
Closing in favor of https://github.com/Automattic/pocket-casts-ios/pull/4953

PR #4956, opened 2026-08-14
Make AnalyticsCoordinator and AnalyticsSourceProvider MainActor-isolated

kean:
| 📘 Part of: # |
|:---:|

Marks `AnalyticsSourceProvider` and `AnalyticsCoordinator` (and therefore `AnalyticsEpisodeHelper` / `AnalyticsPlaybackHelper`) as `@MainActor`.

`track()` used to check `Thread.isMainThread` and re-dispatch itself with `DispatchQueue.main.async`. That check is gone — the isolation now guarantees it. It also fixes a real bug: only `track()` hopped, so the surrounding `currentSource` bookkeeping ran inline on the caller's thread, and `currentAnalyticsSource` would reach `SceneHelper.rootViewController()` and `UIApplication.shared.applicationState` off the main thread.

### ⚠️ The `Task { @MainActor in … }` hops are temporary

**Draft** because the hops added at non-isolated call sites are scaffolding, not the destination. They exist only so this can land ahead of the callers, and they should disappear one by one as `PlaybackManager`, `EpisodeManager`, `SiriShortcutsManager`, `DownloadManager` and friends become main-actor isolated themselves — continuing the incremental top-down adoption from #4953. Once a caller is isolated, its hop collapses into a plain synchronous call, which is what most of the analytics call sites in this diff already look like.

Reviewing the hops as permanent API would be reading them wrong; the thing to check here is the isolation of the two analytics types and that no call site changed meaning.

All 110 call sites were checked. The ones already on the main actor (view controllers, `MultiSelectHelper`, `PlaybackActionHelper`, `PlaybackManager.playBookmark`, the tvOS SwiftUI views) call through synchronously with no hop at all. The genuinely non-isolated ones got an explicit `Task { @MainActor in … }`: `PlaybackManager`, `EpisodeManager`, `SiriShortcutsManager`, `ServerSyncManager`, `DownloadManager`, `GoogleCastManager`, `WatchManager`, `UserEpisodeManager`, and four `DispatchQueue.global()` blocks in `PodcastViewController`. Ordering within each call chain is preserved because the hops are enqueued in source order — which matters for the set-`currentSource`-then-track and `ignoreNextSeek` patterns.

Other notes:

- On tvOS, `SearchAnalytics.episodeTapped`, `AppCoordinator.remotePlayPauseToggle`, `PlayerStatusObserver.updatePlayState` and the `SearchableViewModel.playEpisode` requirement were annotated `@MainActor` instead of hopping — their callers are all main-actor already.
- `OnboardingFlow`'s `AnalyticsSourceProvider` conformance moved to an extension so the struct itself doesn't inherit `@MainActor` (it's reached from the non-isolated `IAPHelper`).
- The pure HLS property builders on `AnalyticsPlaybackHelper` are `nonisolated` — they only read a feature flag, the episode and `PlaybackManager.shared`, so non-isolated callers and the tests don't need a hop.
- `AnalyticsEpisodeHelper`'s `ThreadSafeDictionary` is now a plain dictionary, and its two notification observer bodies were extracted into named methods.
- `PlaybackManager.playbackDidFail` saves/overrides/restores `currentSource` around `pause(userInitiated: false)`. I kept it faithful rather than move `pause()` into the hop, since `cleanupCurrentPlayer` below depends on the pause having run. Worth flagging that this block is inert today: it was written in 5064b7142 when the call was `pause()`, and `userInitiated: false` now suppresses the very event it was meant to attribute. Happy to remove it in a follow-up.

## To test

1. Build and run — no functional change is expected anywhere.
2. Start playing an episode, then pause, skip back and skip forward from the full player. Confirm `playback_play` / `playback_pause` / `playback_skip_back` / `playback_skip_forward` are logged with `source: player`.
3. Do the same from the mini player, lock screen and CarPlay, and confirm the `source` matches the surface you used rather than falling back to `unknown`.
4. Scrub the player and confirm a `playback_seek` fires, and that skipping back/forward does *not* additionally fire one.
5. Download an episode from the podcast screen, then let it finish. Confirm `episode_download_queued` and `episode_download_finished` both report `source: podcast_screen`.
6. Multi-select several episodes in a filter and archive/star/mark as played. Confirm the bulk events report `source: filters`.
7. On the podcast screen use Download All / Archive All / Unarchive All (these run on a background queue) and confirm the bulk events still report `source: podcast_screen`.
8. Ask Siri to pause/resume playback and confirm the events report `source: siri`.
9. Play/pause from the Apple Watch and confirm `source: watch`.
10. tvOS: play an episode from search and confirm `search_result_tapped` plus `playback_play` with `source: search`.

## Checklist

### DIFF
--- a/Pocket Casts TV App/Analytics/SearchAnalytics.swift
+++ b/Pocket Casts TV App/Analytics/SearchAnalytics.swift
@@ -7,6 +7,7 @@ enum SearchAnalytics {
 
     static let source = "search"
 
+    @MainActor
     static func episodeTapped(_ episode: EpisodeSearchResult) {
         AnalyticsPlaybackHelper.shared.currentSource = .search
         Analytics.track(.searchResultTapped, properties: [
--- a/Pocket Casts TV App/UI/AppCoordinator.swift
+++ b/Pocket Casts TV App/UI/AppCoordinator.swift
@@ -194,6 +194,7 @@ class AppCoordinator {
         UserDefaults.standard.synchronize()
     }
 
+    @MainActor
     func remotePlayPauseToggle() {
         AnalyticsPlaybackHelper.shared.timestampOfLastRemoteAction = .now
         PlaybackManager.shared.remotePlayPauseToggle()
--- a/Pocket Casts TV App/UI/Player/PlayerStatusObserver.swift
+++ b/Pocket Casts TV App/UI/Player/PlayerStatusObserver.swift
@@ -37,6 +37,7 @@ class PlayerStatusObserver {
         remainingEventsToSkip = amount
     }
 
+    @MainActor
     private func updatePlayState() {
         guard let player else {
             return
--- a/Pocket Casts TV App/UI/Search/SearchViewModel.swift
+++ b/Pocket Casts TV App/UI/Search/SearchViewModel.swift
@@ -73,7 +73,7 @@ protocol SearchableViewModel: AnyObject, Observation.Observable {
 
     func saveHistory(_ term: String)
 
-    func playEpisode(_ episode: EpisodeSearchResult) async -> Bool
+    @MainActor func playEpisode(_ episode: EpisodeSearchResult) async -> Bool
 
     var isInSearchMode: Bool { get }
 }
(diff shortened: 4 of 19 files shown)


pocketcasts:

PR #4957, opened 2026-08-14
Cover the fingerprint pipeline end to end

kean:
Ports the fingerprint test suite from #4920 onto trunk's API, without the Swift Concurrency refactor those tests were written against. 21 tests, no production changes.

The fixtures synthesize an episode's audio and build its reference fingerprint from that same audio, then run the real pipeline — decode, matcher, score and dominance gates, drift filter — so the expected mapping is exact and drift fails an assertion rather than looking plausible. Covers the dynamic-ad case the subsystem exists for (an ad spliced into the listener's copy shifts the mapping by exactly its length, and highlighting stops while it plays), both one-shot resolves, the mapping cache, teardown, and the preparation guards. Plus unit tests for reference decoding.

Preparation takes its episode from `PlaybackManager` and its files from `DownloadManager`, so a fixture is an episode with audio written where those lookups point, plus a stub Up Next that reports it as now playing (restored in teardown, no rows written). `FingerprintTestSupport` holds that plumbing and the polling helpers the tests wait on, since the manager reports no completion — a pass is finished once it has left `.preparing` and its committed mapping has stopped growing.

### Two tests differ from #4920

Both because of the bug that PR also fixes: `AVAudioFile.read` throws at EOF rather than returning no frames (reproduced on wav, caf and m4a), so every completed pass exits through the error path.

- **Mapping cache.** Trunk never persists one, so the round trip can't be tested. The cache is seeded through `FingerprintMappingCache.save` and the load path asserted instead — instant `.active`, cached entries, no decode.
- **Unmatched audio.** The pass lands in `.failed` rather than `.unavailable`, so the test asserts what it committed (nothing) instead of its terminal state.

#4920's `FingerprintPCMTests` isn't here: it targets `FingerprintPCM.interleave`, which that PR extracted. On trunk it's `private` inside `FingerprintTimingManager`, and `@testable` can't reach it.

## To test

Nothing user-facing changed — this is a test-only PR.

1. Run the new suites: `make test_staging ONLY_TESTING=PocketCastsTests/FingerprintPreparationTests`, then the same for `FingerprintResolveTests` and `ReferenceFingerprintTests`. 21 tests, all green, ~25s.
2. Run them alongside the existing `FingerprintTimingManagerTests` and `FingerprintMappingCacheTests`, plus a DB-backed class (`EpisodeManagerTests`, `PodcastManagerTests`), to confirm the stubbed `DataManager` and the files written into the podcasts directory are cleaned up after each test. 81 tests, all green.
3. `make lint` is clean.

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Fingerprint/FingerprintFixtures.swift
+++ b/PocketCastsTests/Tests/Fingerprint/FingerprintFixtures.swift
@@ -1,0 +1,258 @@
+import AVFoundation
+import Fingerprint
+import Foundation
+
+@testable import podcasts
+
+/// The two on-disk inputs a fingerprint run needs: an audio file, and a reference
+/// fingerprint built from that same audio the way the server builds one from the
+/// publisher's copy of the episode.
+enum FingerprintFixtures {
+
+    static let defaultSampleRate: Double = 44100
+
+    /// The reference grid `makeReference` fingerprints on, matching the shape of a
+    /// server-generated reference: a checkpoint every 2 s, each covering 8 s.
+    static let checkpointIntervalSeconds = 2
+    static let checkpointDurationSeconds = 8
+
+    enum FixtureError: Error {
+        case bufferAllocationFailed
+        case openForWritingFailed(String)
+        case openForReadingFailed(String)
+        case readFailed(String)
+    }
+
+    static let contentSeed: UInt64 = 0xF1_9E_2B_71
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #4958, opened 2026-08-14
Remove unreachable branches in PlaybackManager

kean:
| 📘 Part of: # |  
|:---:|

Removes two unreachable code paths in `PlaybackManager`. No behavior change.

`populateFrom(episodes:startingAtEpisode:)` took an optional `[BaseEpisode]?` and branched on `episodes == nil`, but its only caller — `play(playlist:)` — passes a non-optional `[Episode]` that is also guaranteed non-empty (it's `guard`ed on `playlistEpisodes.first`). The nil branch was never taken, and the `episodes!` force-unwrap in the else branch was only safe by accident of that single call site. The parameter is now non-optional and the branch is gone.

The nil branch held the only call to `switchTo(...)` with `moveExistingToUpNext: true` being meaningful as a choice; with it removed, the sole remaining caller always passes `true`, so the `!moveExistingToUpNext` branch that removed the previous episode from the queue was also unreachable. The parameter has been dropped.

With the branching gone, `populateFrom` was five unconditional lines behind a name that no longer described it, and its `startingAtEpisode` argument was always `episodes.first`. It's now inlined into `play(playlist:)`, which already `guard`s on that same first element. The `isEmpty` check is kept because `PlaybackQueue.bulkAdd` fires `bulkOperationDidComplete` even for an empty array.

Both branches have been dead since the method was introduced in 708d2d941 — it was copied from the older `populateFromEpisodes(_:startingAtEpisode:)`, which already had the same single non-nil caller. That predecessor has since been deleted along with its feature flag.

## To test

Playback behavior should be unchanged. To verify the affected path:

1. Add several episodes to Up Next.
2. Go to a filter/playlist and tap Play All.
3. Confirm the playlist replaces Up Next and starts playing from the first episode.
4. With a non-empty Up Next, tap play on a single episode elsewhere in the app.
5. Confirm it starts playing and the existing Up Next queue is preserved below it.

## Checklist

### DIFF
--- a/podcasts/PlaybackManager.swift
+++ b/podcasts/PlaybackManager.swift
@@ -194,7 +194,7 @@ class PlaybackManager: ServerPlaybackDelegate {
         // if the user has built an Up Next list, preserve that but make this the currently playing episode
         if !overrideUpNext && !switchingToDifferentUpNextEpisode && queue.upNextCount() > 0 {
             if let currEpisode = currentEpisode, currEpisode.uuid != episode.uuid {
-                switchTo(episodeToPlay: episode, moveExistingToUpNext: true, autoPlay: autoPlay, completion: completion)
+                switchTo(episodeToPlay: episode, autoPlay: autoPlay, completion: completion)
 
                 return
             }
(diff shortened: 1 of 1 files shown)


kean:
Good find by Claude, and as straightforward as it gets – merging.

PR #4961, opened 2026-08-14
PlaybackManager: Incremental @MainActor adoption from the UI side – Part 3

kean:
Part 3 of the incremental `@MainActor` adoption, continuing top-down from the UI side toward `PlaybackManager` (after #4947 and #4953).

UIKit and SwiftUI types already inherit isolation from their base classes and protocols, so this pass covers the plain helper types that call into `PlaybackManager` from the main thread without saying so: `SwipeActionsHelper`, `GeneratedChapterSeeker`, the `Constants` extension in `LiquidGlass.swift`, `SharingModal` (which pulls in `SharingHelper`), `CheckTranscriptAvailability`, and `UpNextHistoryModel`.

Two are more than annotations:

- **`UpNextHistoryModel`** moves from two per-method `@MainActor`s to type-level isolation. `reAddMissingItems` was the one method without an annotation, so its `Task { }` ran on the concurrent executor and mutated the Up Next queue off the main thread. Its DB reads move to main as a result, matching `loadEntries` and `loadEpisodes`.
- **`CheckTranscriptAvailability`**'s two notification observers now use `MainActor.assumeIsolated`. They're registered with `queue: .main`, but the closure is `@Sendable`, so without the explicit hop it would silently inherit isolation rather than assert it.

Types *below* `PlaybackManager` are left alone — annotating those pushes isolation upward into it instead of toward it. `PlayerColorHelper` is main-only too, but cascades through the SwiftUI theming layer into isolated-default-value errors in `Toast.show` and `EmptyStateView.init`; that needs its own PR.

The review also found call sites that genuinely run off the main thread today and will need fixing before `PlaybackManager` can be isolated: `WatchManager.handleWatchAction` (WCSession delegate thread — `skipBack` is hopped to main, `skipForward` two lines later isn't), `ShortcutManager.updateShortcuts` (`DispatchQueue.global()`), `ServerSyncManager`'s sync-delegate callbacks, and the archive/cleanup paths. Follow-ups, not part of this PR.

## To test

Compile-time change plus one behavioral fix; no visible difference except Up Next History.

1. Open the full player → Chapters tab, tap a chapter (including on an episode with generated chapters).
2. Swipe an episode row and use the Up Next / archive / delete actions.
3. Share a podcast and an episode, including "share from current position".
4. Profile → Up Next History: open an entry and restore it. The queue should restore correctly without hitching.
5. Switch between episodes with and without transcripts; the transcript shelf button still enables/disables.

## Checklist

### DIFF
--- a/podcasts/Episode Cells/SwipeActionsHelper.swift
+++ b/podcasts/Episode Cells/SwipeActionsHelper.swift
@@ -43,6 +43,7 @@ protocol SwipeHandler: AnyObject {
     func removeFromManualPlaylist(episode: Episode, at: IndexPath)
 }
 
+@MainActor
 enum SwipeActionsHelper {
     // Contrast themes prioritise readability; the green `support02` background
     // doesn't pass against the white "+" icon, so fall back to `support06` (the
@@ -227,6 +228,7 @@ enum SwipeActionsHelper {
     }
 }
 
+@MainActor
 fileprivate extension TableSwipeAction {
     static func removeAction(indexPath: IndexPath, tableView: UITableView, swipeHandler: SwipeHandler, episode: Episode) -> TableSwipeAction {
         return TableSwipeAction(indexPath: indexPath, title: L10n.delete, removesFromList: true, backgroundColor: ThemeColor.support05(), icon: UIImage(named: "delete"), tableView: tableView, handler: { _ -> Bool in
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #4981, opened 2026-08-19
Fix the episode details screen insets on watchOS 26

kean:
| 📘 Part of: # |
|:---:|

Noticed when testing https://github.com/Automattic/pocket-casts-ios/pull/4976.

On watchOS 26 the episode details content ran all the way to the display edge, so the title, date, description and the play button were clipped by the rounded corners. The action buttons were already the correct system width, but the leading-aligned `VStack` pinned them to the edge instead of centering them, which left them visibly offset.

Everything below the artwork now sits in a container with `scenePadding(.horizontal)`, putting it on the same content column the system button style already uses. The artwork keeps its full-bleed treatment.

The SwiftUI preview now feeds in a sample title and description so this layout is actually visible in the canvas — previously it used an empty `Episode()`.

Before / After
Screenshot 2026-08-19 at 12 38 51 PM Screenshot 2026-08-19 at 1 32 49 PM

## To test

1. Run the watch app on a watchOS 26 simulator or device.
2. Open Up Next (or any episode list) and tap an episode.
3. The title, date, duration and description should be inset from the display edge rather than clipped by the rounded corners, and should line up with the artwork's play button.
4. Scroll down: the divider and the action buttons (Download, Play Next, Archive, …) should be centered on the display, not pushed against the leading edge.
5. The artwork should still span the full width.

## Checklist

### DIFF
--- a/Pocket Casts Watch App/UI/Episode/EpisodeView.swift
+++ b/Pocket Casts Watch App/UI/Episode/EpisodeView.swift
@@ -12,11 +12,14 @@ struct EpisodeView: View {
             ScrollView {
                 VStack(alignment: .leading, spacing: 7) {
                     artwork
-                    episodeDetails
-                        .offset(y: geo.size.width * -0.6)
-                        .padding(.bottom, geo.size.width * -0.6)
-                    Divider()
-                    episodeActions
+                    VStack(alignment: .leading, spacing: 7) {
+                        episodeDetails
+                            .offset(y: geo.size.width * -0.6)
+                            .padding(.bottom, geo.size.width * -0.6)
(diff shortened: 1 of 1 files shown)


CookieyedCodes:
Looking forward to this, are the other watch os changes still going to happen?

kean:
> other watch os changes still going to happen?

@CookieyedCodes, anything specific you'd like to see? This one was just a one-off bug I noticed when testing something else, so I fixed it.

CookieyedCodes:
@kean ruan had posted some screenshots on the watch app in June so I'm mainly curious is that happening

PR #4982, opened 2026-08-19
Update smart playlists live when a download changes their contents

kean:
Fixes PCIOS-899

Since 8.16, downloading an episode from a smart playlist whose rules include Download Status = "Not Downloaded" left the row in the list until the playlist was re-entered.

[#4648](https://github.com/Automattic/pocket-casts-ios/pull/4648) removed the `episodeDownloaded` / `episodeDownloadStatusChanged` observers from the playlist detail screen to stop the list from jumping, on the grounds that `EpisodeCell` updates itself in place. That holds for cell state, but not for membership: when a playlist filters by download status, a finished download takes the episode out of the query's result set. The screen has no other live-update path, so the row stayed until `viewWillAppear` re-ran the query. Archiving kept working because only the download observers were removed.

The observers are back, but they now request a reload only when the playlist's rules actually key on the download status, so playlists that include every status (New Releases, In Progress, and anything else created with "All") stay as quiet as they are today and don't reintroduce PCIOS-804.

The gate is `EpisodeFilter.filtersByDownloadStatus`, which mirrors the download status clause in `PlaylistQueryBuilder` — the clause is only emitted when some, but not all, of the statuses are included. `PlaylistQueryBuilder` now uses the same property at both of its (duplicated) sites, so the gate can't drift from the SQL. `EpisodeFilter.isAffected(by: .downloadStatus)` uses it too, which also stops the playlist metadata cache from recounting playlists a download can't change.

## To test

1. In the Playlists tab, open a smart playlist and set Smart rules > Download Status = **Not Downloaded**.
2. Download an episode from the list and stay on the screen.
3. When the download completes, the row is removed. Delete the download from elsewhere and it comes back.
4. Set Download Status = **Downloaded** on the same playlist: the episode should appear as the download finishes and disappear when the file is deleted.
5. Set Download Status = **All** (or open New Releases), scroll down, and start a download on an off-screen-ish episode: the list must not jump or reload while the download runs and finishes, and the cell still updates its own download state (this is the PCIOS-804 case #4648 fixed).
6. Archiving from a playlist still removes the row immediately.

## Checklist

### DIFF
--- a/podcasts/New Detail/PlaylistDetailViewController+Observers.swift
+++ b/podcasts/New Detail/PlaylistDetailViewController+Observers.swift
@@ -19,9 +19,17 @@ extension PlaylistDetailViewController {
         addCustomObserver(Constants.Notifications.episodePlayStatusChanged, selector: #selector(refreshEpisodesFromNotification))
         addCustomObserver(Constants.Notifications.episodeArchiveStatusChanged, selector: #selector(refreshEpisodesFromNotification))
         addCustomObserver(Constants.Notifications.episodeStarredChanged, selector: #selector(refreshEpisodesFromNotification))
+        addCustomObserver(Constants.Notifications.episodeDownloaded, selector: #selector(refreshEpisodesIfFilteringByDownloadStatus))
+        addCustomObserver(Constants.Notifications.episodeDownloadStatusChanged, selector: #selector(refreshEpisodesIfFilteringByDownloadStatus))
         addCustomObserver(Constants.Notifications.manyEpisodesChanged, selector: #selector(refreshEpisodesFromNotification))
         addCustomObserver(UIResponder.keyboardWillShowNotification, selector: #selector(keyboardWillShow(_:)))
         addCustomObserver(UIResponder.keyboardWillHideNotification, selector: #selector(keyboardWillHide(_:)))
+    }
+
+    @objc func refreshEpisodesIfFilteringByDownloadStatus(notification: Notification) {
+        guard viewModel.playlist.isAffected(by: .downloadStatus) else { return }
+
+        reloader.request(.episodes)
     }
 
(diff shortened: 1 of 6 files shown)


podcasts/New Detail/PlaylistDetailViewController+Observers.swift
```diff
@@ -19,11 +19,19 @@ extension PlaylistDetailViewController {
...
         addCustomObserver(Constants.Notifications.episodeArchiveStatusChanged, selector: #selector(refreshEpisodesFromNotification))
         addCustomObserver(Constants.Notifications.episodeStarredChanged, selector: #selector(refreshEpisodesFromNotification))
+        addCustomObserver(Constants.Notifications.episodeDownloaded, selector: #selector(refreshEpisodesIfFilteringByDownloadStatus))
+        addCustomObserver(Constants.Notifications.episodeDownloadStatusChanged, selector: #selector(refreshEpisodesIfFilteringByDownloadStatus))
         addCustomObserver(Constants.Notifications.manyEpisodesChanged, selector: #selector(refreshEpisodesFromNotification))
         addCustomObserver(UIResponder.keyboardWillShowNotification, selector: #selector(keyboardWillShow(_:)))
         addCustomObserver(UIResponder.keyboardWillHideNotification, selector: #selector(keyboardWillHide(_:)))
     }
 
+    @objc func refreshEpisodesIfFilteringByDownloadStatus(notification: Notification) {
```
kean, line 29: The app should really just be observing the SQL queries using GRDB tooling, but that's for another day.

kean:
Tested, works as expected now.

kean:
Adressed the comments.

> Streaming buffers will make rows vanish from a "Not Downloaded" playlist 

I left a note to check this one. Not sure about the expected behavior.

PR #4990, opened 2026-08-21
Run the API sync tasks as async operations

kean:
| 📘 Part of: # |
|:---:|

Under the new `asyncApiTasks` feature flag (off by default), `ApiBaseTask` becomes an asynchronous `Operation`: `start()` returns immediately and the operation finishes when its async work completes, so no thread is held for the duration of a request. With the flag off, `start()` falls through to `Operation`'s default and blocks the queue thread until the work finishes, so queues, ordering and `waitUntilAllOperationsAreFinished()` behave exactly as they do today.

All the requests now go through `URLConnection.send(request:)` instead of `sendSynchronousRequest`, and all 40 task subclasses perform their work in an async closure. `TokenHelper` gained non-blocking `acquireToken()`/`acquirePasswordToken()` counterparts (the existing blocking ones are still used by their non-task callers).

A few things that came with the conversion:

- Nested task calls (`SyncTask` → retrieve podcasts/playlists/bookmarks/last sync) `await runTask()` instead of blocking a thread each.
- Unbounded blocking waits inside task bodies were converted so they can't clog the cooperative pool: `importQueue.waitUntilAllOperationsAreFinished()` → a barrier block, plus the bookmark import semaphore, the starred episode `DispatchGroup` and the year history groups. Waits that are bounded by a timeout and run on their own queue threads were left alone.
- `YearListeningHistory.sync` called `task.start()` and read `task.success` on the next line, which silently breaks with an asynchronous operation, so it now calls `runTaskSynchronously()`.
- `SyncTask.processSyncData` held an `objc_sync` lock across the new `await`. That lock is owned by the thread that took it, so it would be released from the wrong thread and eventually wedge every sync. It now uses `AsyncLock`, a non-reentrant mutex that can be held across suspension points.

## To test

1. Enable `asyncApiTasks` in Settings → Developer → Feature Flags.
2. Pull to refresh on the Podcasts screen and confirm the sync completes and the grid updates.
3. Play an episode, scrub, star it, and add a bookmark. Force a sync and confirm the changes reach another device.
4. Change the Up Next queue and confirm it syncs both ways.
5. Log out and log back in and confirm the full sync imports podcasts, filters and bookmarks.
6. Repeat steps 2-5 with the flag off and confirm nothing changed.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsServer/Private/API Tasks/ApiBaseTask.swift
+++ b/Modules/Sources/PocketCastsServer/Private/API Tasks/ApiBaseTask.swift
@@ -12,11 +12,53 @@ class ApiBaseTask: Operation, @unchecked Sendable {
     private let urlConnection: URLConnection
     private let tokenHelper: TokenHelper
 
+    /// Read once, so that flipping the flag while an operation is in flight can't leave it in a state
+    /// where it starts asynchronously but reports its progress synchronously (or vice versa).
+    private let runsAsynchronously = FeatureFlag.asyncApiTasks.enabled
+
+    private let stateLock = NSLock()
+    private var _isExecuting = false
+    private var _isFinished = false
+
     init(dataManager: DataManager = .sharedManager, urlConnection: URLConnection = URLConnection(handler: URLSession.shared)) {
         self.dataManager = dataManager
         self.urlConnection = urlConnection
         self.tokenHelper = TokenHelper(urlConnection: urlConnection)
         super.init()
+    }
+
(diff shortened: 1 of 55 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #4995, opened 2026-08-21
Remove nine feature flags that have shipped enabled

kean:
Follow-up to #4994. Removes nine feature flags that have defaulted to `true` since 2024 — the enabled branch is inlined and the dead fallback deleted.

| Flag | Added | What the removed fallback did |
|---|---|---|
| `useDefaultPlayerTapCookie` | 2024-11 | passed an unretained `self` as the tap's `clientInfo` |
| `defaultPlayerFilterCallbackFix` | 2024-05 | `unsafeBitCast` of the tap storage pointer |
| `onlyMarkPodcastsUnsyncedForNewUsers` | 2024-05 | marked all podcasts unsynced on every sign-in |
| `ignoreRouteDisconnectedInterruption` | 2024-08 | treated a route disconnect as a real interruption |
| `playerIsReadyToPlay` | 2024-09 | seeked without checking `isReadyToPlay()` |
| `useMimetypePackage` | 2024-09 | trusted the HTTP `Content-Type` header |
| `customPlaybackSettings` | 2024-10 | hid the global/local segmented control in Effects |
| `markAllSyncedInSingleStatement` | 2024-11 | one `UPDATE` per episode instead of a single `IN (…)` |
| `manageDownloadedEpisodes` | 2024-11 | hid the low-storage banner/modal |

Notes:

- The two `DefaultPlayer` flags were a chained `if/else if/else`, so they're removed together — `unretainedDefaultPlayer(for:)` collapses to the cookie lookup.
- `customPlaybackSettings` was read inside an `#if APPCLIP` guard; the property stays so the App Clip keeps returning `false`.
- The removed `markAllSynced(episodeIDs:)` fallback bound `"'\(episodeId)'"` — quotes included — so it never matched a row. Only the surviving single-statement path is correct.
- `processEpisode`'s `reportedContentType` parameter is now unused, since the mimetype lookup always overwrites it. Unwinding it also means removing `ExportStatus.reportedType` and changing the media exporter callback, so I left it for a separate PR rather than widening this one.

## To test

Regression testing around the code these flags gated; none of it should behave differently, since all nine were already on in production.

1. **Playback effects** — play an episode, open the Effects panel, and confirm the Global/Local segmented control appears. Repeat from a podcast's Settings → Playback Effects.
2. **Audio processing** — enable Trim Silence and Volume Boost, play for a minute, then skip between episodes several times. Audio stays clean and no crash on episode change (this exercises the tap cookie path).
3. **Interruptions** — play audio over Bluetooth headphones and power them off. Playback pauses and does not resume unexpectedly when they reconnect.
4. **Seeking** — start streaming an episode, pause it immediately, then scrub the scrubber. The seek takes effect rather than being ignored.
5. **Downloads** — download an episode and confirm it plays; a feed serving a wrong `Content-Type` should still download successfully.
6. **Sign-in** — sign in on a fresh install and confirm podcasts sync up. Sign out and back in on the same device; existing podcasts should not be re-marked unsynced.
7. **Low storage** — with the device near full and downloaded episodes present, confirm the Manage Downloads banner still appears.

Automated: `DownloadManagerTests` (2 tests) and `PocketCastsDataModelTests/PodcastDataManagerTests` (68 tests) pass.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
+++ b/Modules/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
@@ -916,14 +916,8 @@ class EpisodeDataManager {
         dbQueue.write { db in
             do {
                 db.beginTransaction()
-                if FeatureFlag.markAllSyncedInSingleStatement.enabled {
-                    let ids = episodes.map({"\($0.id)"})
-                    try db.executeUpdate("UPDATE \(DataManager.episodeTableName) SET playingStatusModified = 0, playedUpToModified = 0, durationModified = 0, keepEpisodeModified = 0, archivedModified = 0 WHERE id IN (\(ids.joined(separator: ",")))", values: nil)
-                } else {
-                    for episode in episodes {
-                        try db.executeUpdate("UPDATE \(DataManager.episodeTableName) SET playingStatusModified = 0, playedUpToModified = 0, durationModified = 0, keepEpisodeModified = 0, archivedModified = 0 WHERE id = ?", values: [episode.id])
-                    }
-                }
+                let ids = episodes.map({"\($0.id)"})
+                try db.executeUpdate("UPDATE \(DataManager.episodeTableName) SET playingStatusModified = 0, playedUpToModified = 0, durationModified = 0, keepEpisodeModified = 0, archivedModified = 0 WHERE id IN (\(ids.joined(separator: ",")))", values: nil)
                 db.commit()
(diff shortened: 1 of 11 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #4996, opened 2026-08-21
Update the app icon badge after a background refresh, not before it starts

kean:
Fixes PCIOS-896

Both background refresh paths called `badgeHelper.updateBadge()` right after *starting* the refresh, so the badge was written from pre-refresh counts. The correction relied on `BadgeHelper`'s 3s-debounced `podcastsRefreshed` subscription, which never fires in the background — iOS suspends the process as soon as the completion handler is signalled.

Both paths now badge after the refresh completes and wait for the `setBadgeCount` write before signalling `completionHandler` / `setTaskCompleted`. Behind `FeatureFlag.updateBadgeAfterBackgroundRefresh`.

`BadgeHelper.updateBadge` gains an optional completion; the count computation moves into `badgeCount() -> Int?` so there's one write site and one exit path instead of five early returns that could drop it. Behaviour-preserving, not behind the flag.

This doesn't fix the ticket on its own — the pushes carry no `badge` field and don't wake the app, so the badge still only moves when the app gets a background slot. Making it update on notification arrival needs the Notification Service Extension to set `content.badge`.

## To test

1. Profile → Settings → Notifications → App Badge → **Total Unplayed**.
2. Note the badge, background the app, and wait for a new episode to be available.
3. In Xcode, pause and run:
   `e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"au.com.shiftyjelly.podcasts.Refresh"]`
4. Resume — the badge should include episodes fetched by *that* refresh.
5. Turn `updateBadgeAfterBackgroundRefresh` off in the developer overrides and repeat: the badge shows the stale pre-refresh count.
6. Confirm the badge still updates in the foreground across each App Badge option.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Tapping a Discover collection's poster now opens the expanded collection, the same as tapping "Show All" [#5028](https://github.com/Automattic/pocket-casts-ios/pull/5028)
 - Add support for "Networks" row in "Discover" [#5026](https://github.com/Automattic/pocket-casts-ios/pull/5026)
 - Fix the Download button doing nothing in Search and Discover episode results [#4702](https://github.com/Automattic/pocket-casts-ios/pull/4702)
+- Fix the app icon badge showing a stale count after a background refresh: the badge is now updated once the refresh has finished instead of before it starts [#4996](https://github.com/Automattic/pocket-casts-ios/pull/4996)
 
 8.19
 -----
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

kean:
I'm still not 100% positive on this one – moving to draft and 8.21; will test more.

pocketcasts:

PR #4998, opened 2026-08-21
Add Bluesky to the About screen social links

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-893

Profile > Settings > About listed Website, Instagram, and X, but not Bluesky — even though we link it in the pocketcasts.com footer. This adds a Bluesky row following the existing pattern: `SocialsHelper.openBluesky()` tries the `bluesky://` app scheme and falls back to `https://bsky.app/profile/pocketcasts.com`, with a new `settingsAboutBlueskyTapped` analytics event.

Two notes for review:

- **`bluesky` added to `LSApplicationQueriesSchemes`.** `canOpenURL` returns `false` for undeclared schemes, so without this the app-scheme step silently never fires and every tap goes to the web. Unrelated but worth knowing: `instagram` isn't declared there today (only `instagram-stories`), so the existing Instagram row already always falls through to the web. Left alone as out of scope.
- **Secondary text is `@pocketcasts.com`.** The issue flagged this as a design call. Instagram and X share `socialHandle` ("@pocketcasts"), but the Bluesky handle is `pocketcasts.com` — which is also exactly what the Website row displays two lines up. I went with `@pocketcasts.com` so the row isn't the only one with a blank trailing slot, and the `@` (how Bluesky itself renders handles) keeps it visually distinct from the Website row. Happy to drop it to nothing if design prefers — it's a one-line change.

Reddit and Threads rows were both considered and deliberately excluded, per the issue.

## To test

1. Go to Profile > Settings > About.
2. Confirm a **Bluesky** row appears below **X**, showing `@pocketcasts.com`.
3. With the Bluesky app installed, tap the row — it should open the Pocket Casts profile in the app.
4. Without the Bluesky app installed, tap the row — it should open `bsky.app/profile/pocketcasts.com` in the browser.
5. Confirm the Website, Instagram, and X rows still behave as before.
6. Check a couple of themes (light/dark) to confirm the row is styled consistently with its neighbours.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #4999, opened 2026-08-22
PCIOS-902: iOS: "You've been signed out" alert repeatedly shown to a user who has never signed in (TOKEN_DEAUTH cleanup loop)

pocketcasts:
Resolves https://linear.app/a8c/issue/PCIOS-902/ios-youve-been-signed-out-alert-repeatedly-shown-to-a-user-who-has

## Summary
- When `tokenCleanUp()` runs after a `TOKEN_DEAUTH` error, it now distinguishes between users who have actual credentials (password or SSO refresh token) and users with only stale keychain data (email but no credentials).
- For the stale-credentials case, it silently clears the stale data without calling `SyncManager.signout()`, which prevents the "You've been signed out" alert from being shown to users who were never properly signed in.

## Changes
- `Modules/Sources/PocketCastsServer/Private/TokenHelper.swift` — Modified `tokenCleanUp()` to track whether the user has a password or refresh token. When neither exists, silently clear the stale keychain data (via `SyncManager.clearTokensFromKeyChain()`, `ServerSettings.setSyncingEmail(email: nil)`, and `ServerSettings.userId = nil`) instead of calling `SyncManager.signout()` which posts the `.serverUserWillBeSignedOut` notification and triggers the alert.
- `PocketCastsTests/Tests/Utilities/TokenHelperTests.swift` — Added `testTokenCleanUpSilentlyRemovesStaleCredentials` test that sets up stale keychain state (email only), calls `acquireToken()`, and verifies: (1) no token is returned, (2) the sign-out notification is NOT posted, and (3) the login state is cleared.

## Verification
- **Lint**: PASS
- **Tests**: FAIL — pre-existing GenerateCredentials build script failure prevents test execution locally (not related to this change)
- **Diff review**: PASS — only the 2 intended files modified, no unintended changes

## Confidence
**MEDIUM** — The fix correctly addresses the root cause identified in the issue (stale keychain data triggering the sign-out alert). The logic is straightforward and preserves existing behavior for users who have actual credentials. The uncertainty is that tests could not be run locally due to a pre-existing build configuration issue (missing credentials script), so the new test has not been executed.

---
Triage complexity: medium | [Linear issue](https://linear.app/a8c/issue/PCIOS-902/ios-youve-been-signed-out-alert-repeatedly-shown-to-a-user-who-has)

### DIFF
--- a/Modules/Sources/PocketCastsServer/Private/TokenHelper.swift
+++ b/Modules/Sources/PocketCastsServer/Private/TokenHelper.swift
@@ -236,13 +236,17 @@ class TokenHelper {
             logMessages.append("no email address")
         }
 
-        if ServerSettings.syncingPassword() == nil {
+        let hasPassword = ServerSettings.syncingPassword() != nil
+        if !hasPassword {
             logMessages.append("no password")
         }
 
+        var hasRefreshToken = false
         do {
             if try ServerSettings.refreshToken() == nil {
                 logMessages.append("no SSO token")
+            } else {
+                hasRefreshToken = true
             }
         } catch {
             if case let KeychainHelper.KeychainError.status(status) = error, status == errSecInteractionNotAllowed {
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5002, opened 2026-08-24
Add CHANGELOG entry for the watch background task crash fix

kean:
The code fix already landed in trunk — it rode in on the branch for #5001 and shipped in 8.20. This PR is now just the release note that was missing for it.

The watch app aborts when `pendingWatchBackgroundTask` is completed twice: `urlSessionDidFinishEvents(forBackgroundURLSession:)` completed the task but never cleared the ivar, so the next delivery for the same session id completed a stale task and WatchKit killed the process. The fix clears the ivar wherever the task is completed and guards the delegate callback by session identifier, so `wifiOnlyBackgroundSession` can't complete the cellular task early.

## To test

Check `CHANGELOG.md` renders the new 8.20 line correctly. The code change is already on trunk and unmodified here.

## Checklist


kean:
Heads up: the code change from this branch was already merged in the scope of another PR — it went in on the branch for #5001 as e1069a7ac (identical content) and shipped in 8.20.0.0.

Rebasing on trunk dropped the commit as already-applied, so all that's left here is the CHANGELOG entry, which was never added for this fix. Added under 8.20 since that's the release it actually shipped in.

PR #5005, opened 2026-08-24
Fix data races on the DataManager in-memory caches

kean:
| 📘 Part of: # |
|:---:|

Fixes a `_xzm_corruption_detected` crash reported inside `PodcastDataManager.cachePodcasts`. That stack is the victim, not the culprit: malloc only notices a corrupted freelist at the next allocation, which here happened to be the `String` GRDB allocates while decoding a podcast row during a refresh.

The corruption comes from unsynchronised access to `DataManager`'s in-memory caches. `FolderDataManager` and `UpNextDataManager` guard most cache access with a serial queue, but four reads bypassed it:

- `FolderDataManager.allFolders(includeDeleted:)`
- `FolderDataManager.save(folder:)` (legacy path)
- `UpNextDataManager.movePlaylistEpisode(from:to:)`
- `UpNextDataManager.saveOrdering(dbQueue:)`

These arrays are read on the main thread while background refresh/sync work replaces them in `cacheFolders`/`cacheEpisodes`. A racing reader can load the old buffer pointer and retain it after the writer has already released the last reference, so the buffer is used after free and eventually released twice — which is exactly what corrupts malloc metadata for every later allocation in the process, including the podcast cache reload.

Separately, all three cache queues (`FolderDataManager`, `UpNextDataManager`, `PodcastDataManager`) were `lazy var`. Swift instance `lazy var` is not thread-safe: two threads racing on first access both run the initialiser and both store to the same reference, over-releasing one queue and, worse, briefly handing the two threads different "locks" so the cache itself is left unguarded. The same applies to `DataManager`'s three lazy sub-managers.

`PodcastDataManager` itself only needed the queue change — every `cachedPodcasts` access there was already guarded.

## To test

This is a race, so there is no deterministic repro. To exercise the paths:

1. Run the new `DataManagerConcurrencyTests` with the Thread Sanitizer enabled — they hammer `allFolders` and `movePlaylistEpisode` from one thread while another writes.
2. Build and run the app, sign into an account with folders and a populated Up Next queue.
3. Pull to refresh on the Podcasts tab and, while the refresh runs, switch between the Podcasts tab (folder grid) and Up Next, and reorder a few Up Next episodes.
4. Confirm folders and Up Next stay correct and the app does not crash.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Private/Managers/FolderDataManager.swift
+++ b/Modules/Sources/PocketCastsDataModel/Private/Managers/FolderDataManager.swift
@@ -16,11 +16,7 @@ class FolderDataManager {
     ]
 
     private var cachedFolders = [Folder]()
-    private lazy var cachedFolderQueue: DispatchQueue = {
-        let queue = DispatchQueue(label: "au.com.pocketcasts.FolderDataQueue")
-
-        return queue
-    }()
+    private let cachedFolderQueue = DispatchQueue(label: "au.com.pocketcasts.FolderDataQueue")
 
     func setup(dbQueue: PCDBQueue) {
         cacheFolders(dbQueue: dbQueue)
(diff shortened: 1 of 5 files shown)


kean:
It seems potentially unsafe. Im going to look into converting it to use Swift Concurrency instead.

PR #5008, opened 2026-08-25
Fix the transcript layout on the bookmark details and edit screens

kean:
| 📘 Part of: # |
|:---:|

Fixes three transcript layout problems on the bookmark screens.

Both transcript text views drew a gradient fade over the top of their frame but kept `textContainerInset.top` at zero, so the first and last lines came to rest under the fade, unreachable. They now reserve the same room at their edges as the fade covers, kept in one place per screen so the two can't drift apart.

- **Edit screen** — the opening lines sat under "Select a transcript passage for this bookmark". The subtitle is now a sibling above the transcript rather than an overlay on top of it, which also drops the `GeometryReader` measurement and the inset re-sync it drove on every update.
- **Details screen** — the screen opened mis-laid-out and corrected itself a frame later: the content block was only width-constrained, so it was vertically centred until the transcript claimed the rest of the height, and `isLoadingTranscript` started `false`, so the first frame showed the bare passage. The layout is now top-anchored, and the view model starts out loading wherever a fetch is coming.

## To test

1. Open a bookmark with a captured passage from the Bookmarks list.
2. Confirm the header doesn't shift or re-centre after the screen appears.
3. Scroll the transcript to both ends — the first and last lines should come to rest clear of the fades, not under them.
4. Tap ⋯ → Edit and continue to the transcript step. Confirm it starts just below the subtitle with no large gap, and that its first line isn't trapped under it.
5. Repeat on a bookmark whose passage sits near the start of the episode, and on one with no transcript available.

## Checklist

### DIFF
--- a/podcasts/Bookmarks/Details/BookmarkDetailsView.swift
+++ b/podcasts/Bookmarks/Details/BookmarkDetailsView.swift
@@ -27,13 +27,14 @@ struct BookmarkDetailsView: View {
     @ViewBuilder
     private var layout: some View {
         if viewModel.passage != nil {
-            VStack(alignment: .leading, spacing: 24) {
-                Group {
+            VStack(alignment: .leading, spacing: 0) {
+                VStack(alignment: .leading, spacing: 24) {
                     header
                     details
                 }
                 .padding(.horizontal, 16)
 
+                // The room the transcript keeps clear at the top is the gap above it
                 transcriptSection
             }
             .padding(.top, 16)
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5014, opened 2026-08-26
Replace the _unsafeWait bridge in VideoExporter with structured concurrency (5 warnings)

kean:
| 📘 Part of: # |
|:---:|

Clears five concurrency warnings in `VideoExporter`, one of which is an error in Swift 6 language mode.

- `AVAssetWriter` / `AVAssetExportSession` captured by `@Sendable` closures (`onCancel`, the progress `Timer`) — those locals are now `nonisolated(unsafe)`.
- The `_unsafeWait` deprecation and an `AVAssetWriterInput` self-capture both came from `unsafeRequestMediaDataWhenReady`, which blocked a global-queue thread on a `DispatchSemaphore`. Replaced with a plain `for frame in 0...frameCount` loop plus a suspending `waitUntilReadyForMoreMediaData()`. The `Counter` actor is gone.

Frame count, timing, and the `markAsFinished()` error path are unchanged. Also drops an `adaptor` parameter `pixelBuffer` never used.

Worth a look on review: the readiness wait now polls every 5 ms instead of getting an immediate callback (every frame is a main-actor SwiftUI render, so it should never be hit), and `UnsafeWait.swift` is now dead code — left in place to keep the diff focused.

## To test

Needs a Plus/Patron account and a downloaded episode.

1. Play an episode, open the player, tap share → "Clip", set a start point and length, then "Clip and Share".
2. Pick a video destination — Instagram if installed, otherwise "More" → Save Video.
3. Confirm progress advances smoothly and the UI stays responsive during export.
4. Confirm the video plays back with the animated card, the correct audio segment, and the full clip length.
5. Repeat, cancelling mid-export, and confirm it stops cleanly with no hang or crash.

## Checklist

### DIFF
--- a/podcasts/Sharing/Clip/VideoExporter.swift
+++ b/podcasts/Sharing/Clip/VideoExporter.swift
@@ -68,7 +68,7 @@ enum VideoExporter {
     // Step 1 of video export
     private static func exportInitialVideo<Content: AnimatableContent>(of view: Content, to outputURL: URL, with parameters: Parameters, frameCount: Int, progress: Progress) async throws {
         let videoSize = CGSize(width: parameters.size.width * parameters.scale, height: parameters.size.height * parameters.scale)
-        let videoWriter = try AVAssetWriter(outputURL: outputURL, fileType: parameters.fileType)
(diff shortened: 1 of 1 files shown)


kean:
I'll split this into separate PRs and slightly one of the changes.

PR #5030, opened 2026-08-27
Inline the @GRDBRecord macro expansion for Folder

kean:
Replaces the `@GRDBRecord` macro on `Folder` with the code the macro was already generating, and drops the `GRDBMacros` import.

This is a pilot for removing the `GRDBMacros` target entirely. That target is used by exactly 6 model types and is the **only** reason the repo depends on `swift-syntax` — removing it would delete ~1,600 lines (a 507-line macro implementation plus a 1,031-line test suite that no longer matches what the macro emits and isn't in any test plan), take a macro plugin out of every clean build, and drop a 9.7 MB dependency checkout. `Folder` is the smallest of the 6, so it goes first to prove the approach before touching `Podcast`, `Episode`, `UserEpisode`, `EpisodeFilter`, and `NetworkDataUsageRecord`.

## Why this is safe

**The new code was generated by the macro, not written by hand.** I built `GRDBMacrosPlugin` and ran the real expansion:

```
swiftc -typecheck Folder.swift \
  -load-plugin-executable .build/debug/GRDBMacrosPlugin-tool#GRDBMacrosPlugin \
  -Xfrontend -dump-macro-expansions
```

The output is pasted in verbatim; the only edit is re-indenting the macro's ragged continuation lines. This matters because the expansion has details that are easy to get wrong by hand and fail silently — in particular `addedDate` encodes as `.timeIntervalSince1970` rather than GRDB's default `Date` serialization, and every property decodes via `decodeIfPresent(...) ?? default` rather than `decode(...)`, which is what keeps a NULL column from throwing.

**Nothing outside this file changes.** `databaseTableName`, `CodingKeys`, `Columns`, and the conformances keep identical names and visibility, so every query and call site is untouched. The macro stays in place for the other 5 types.

**`Folder` is the least hazardous of the 6.** 8 columns, no `@GRDBColumn` custom column-name mappings, no `@GRDBNullDateAsEpoch`, and a single optional `Date`. The `@GRDBIgnore` removed from `cachedUnreadCount` was already a no-op — the macro only reads `@objc` properties, and that one isn't `@objc`.

**Covered by existing CI tests.** `FolderColumnConsistencyTests` runs in `UnitTests.xctestplan` and does a real save/load round-trip against both the legacy SQL and GRDB paths (`FeatureFlag.grdbQueryInterface`, which defaults to on), plus asserts `cachedUnreadCount` is not persisted.

## Test results

- `FolderColumnConsistencyTests` — 3/3 pass.
- Full `PocketCastsDataModelTests` — 473 tests, 2 failures, both in `EpisodeColumnConsistencyTests.testSaveAndLoadPreservesAllFields` (`hasGeneratedTranscript`). **Pre-existing**: verified failing identically on clean `trunk` with this change stashed. Unrelated to `Folder`.
- SwiftLint clean on the changed file.

## To test

1. Create a folder, add a few podcasts to it, and give it a name and a colour.
2. Change the folder's sort order.
3. Force quit and relaunch — the folder, its name, colour, podcasts, and sort order should all persist.
4. Rename the folder, then delete it; confirm both stick after a relaunch.
5. Sync with another device or the web player and confirm folders round-trip.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:

PR #5033, opened 2026-08-28
Inline the @GRDBRecord macro expansion for the remaining models

kean:
> **Stacked on #5030** — review that one first; this PR targets `inline-grdb-record-folder`, so the diff here is only the remaining 5 models.

Applies the same technique as #5030 to the other five `@GRDBRecord` users: `Podcast`, `Episode`, `UserEpisode`, `EpisodeFilter`, and `NetworkDataUsageRecord`. After this, nothing in the app uses the macro.

**`GRDBMacros` itself is deliberately left in place** — the target, the plugin, the `swift-syntax` dependency, and the `Package.swift` wiring are all untouched. Deleting them is a separate follow-up, so that if anything here misbehaves the revert is a one-file change rather than a dependency-graph change.

## Why this is safe

**Every line was generated by the macro, not written by hand.** Same method as #5030, scripted across all five: for each model I took the real source file, swapped `import GRDBMacros` for local macro declarations, and ran the actual plugin:

```
swiftc -typecheck <model>.swift \
  -load-plugin-executable .build/debug/GRDBMacrosPlugin-tool#GRDBMacrosPlugin \
  -Xfrontend -dump-macro-expansions
```

The dumped `fMm_` (members) and `fMe_` (extension) sections are pasted in verbatim. The only transformation is re-indentation by brace depth — the macro emits ragged continuation lines. Crucially the expansion was dumped **before** the marker attributes were stripped, so `@GRDBColumn` / `@GRDBIgnore` / `@GRDBNullDateAsEpoch` all still informed the generated code.

That matters most for the details that fail silently if a human transcribes them:

- **`NetworkDataUsageRecord`'s 8 custom column names** (`episodeUuid` → `episode_uuid`, etc.) and **`Podcast.autoArchiveEpisodeLimit` → `episodeKeepSetting`** land in both the `CodingKeys` raw values and the `encode(to:)` keys.
- **14 `Date` columns** encode as `.timeIntervalSince1970`, not GRDB's default `Date` serialization.
- **3 `@GRDBNullDateAsEpoch` columns** (`Episode.lastDownloadAttemptDate`, `Episode.lastArchiveInteractionDate`, `UserEpisode.lastDownloadAttemptDate`) still encode `nil` as `Date(timeIntervalSince1970: 0)` — these back `NOT NULL DEFAULT 0` columns, so a plain `nil` would trip the constraint.
- Every property decodes via `decodeIfPresent(...) ?? default`, which is what keeps a NULL column from throwing.

**Nothing outside these five files changes.** `databaseTableName`, `CodingKeys`, `Columns`, and the conformances keep identical names and visibility, so all queries and call sites are untouched.

**Ignored properties stay ignored.** `Episode.hasOnlyUuid` and the other `@GRDBIgnore` properties are absent from `CodingKeys`, `encode(to:)`, and `Columns`, as before. Several `@GRDBIgnore` markers were already no-ops — the macro only ever read `@objc` properties, and those aren't `@objc`.

## Test results

Full `PocketCastsDataModelTests`: **473 tests, 2 failures** — the same count and the same two as on `trunk`.

- `PodcastColumnConsistencyTests`, `UserEpisodeColumnConsistencyTests`, `EpisodeFilterColumnConsistencyTests`, `FolderColumnConsistencyTests` — all pass, each doing a real save/load round-trip against both the legacy SQL and GRDB paths (`FeatureFlag.grdbQueryInterface`).
- `PodcastDataManagerTests`, `EpisodeDataManagerTests`, `UserEpisodeDataManagerTests`, `FolderDataManagerTests`, `NetworkDataUsageManagerTests`, `UserEpisodeEncodingTests` — all pass.
- `EpisodeColumnConsistencyTests.testSaveAndLoadPreservesAllFields` — 2 failures on `hasGeneratedTranscript`. **Pre-existing**, verified failing identically on clean `trunk`. See note below.
- SwiftLint clean on all five files.

### Pre-existing bug worth a separate issue

`Episode.hasGeneratedTranscript` is declared `public var hasGeneratedTranscript: Bool? = nil` **without** `@objc`. The macro only reads `@objc` properties, and the legacy SQL `columnNames` list doesn't include it either — so it is persisted by neither path, and the consistency test correctly catches it. Not introduced or changed here; flagging it since the red test is easy to misread as fallout from this PR.

## To test

`NetworkDataUsageRecord` has no consistency test, so it's worth manual attention:

1. Settings → check the data-usage figures render, then stream and download an episode on cellular and confirm the counters move.
2. Browse and subscribe to podcasts; check artwork colours, episode lists, and Up Next survive a force quit and relaunch.
3. Star, archive, and mark episodes as played; confirm each sticks after a relaunch.
4. Create a filter with several rules; confirm its episode list and settings persist.
5. Upload/inspect a Files (user) episode, including custom artwork.
6. Sync with another device or the web player and confirm podcasts, filters, folders, and playback position round-trip.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Public/Model/Episode.swift
+++ b/Modules/Sources/PocketCastsDataModel/Public/Model/Episode.swift
@@ -1,16 +1,13 @@
 import Foundation
 import GRDB
-import GRDBMacros
 import PocketCastsUtils
 
-@GRDBRecord(table: "SJEpisode")
 public class Episode: NSObject, BaseEpisode {
     private static let bonusType = "bonus"
     private static let trailerType = "trailer"
 
     @objc public var id = 0 as Int64
     @objc public var addedDate: Date?
-    @GRDBNullDateAsEpoch
     @objc public var lastDownloadAttemptDate: Date?
     @objc public var detailedDescription: String?
     @objc public var downloadErrorDetails: String?
@@ -46,10 +43,8 @@ public class Episode: NSObject, BaseEpisode {
     @objc public var episodeType: String?
     @objc public var archived = false
     @objc public var archivedModified = 0 as Int64
-    @GRDBNullDateAsEpoch
     @objc public var lastArchiveInteractionDate: Date?
     @objc public var excludeFromEpisodeLimit = false
-    @GRDBIgnore
     @objc public var hasOnlyUuid = false
     @objc public var deselectedChapters: String?
     @objc public var deselectedChaptersModified = 0 as Int64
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:

kean:
I will do these one by one, covering with more unit test and testing in the app.

PR #5039, opened 2026-08-28
Opt every podcast out of new episode pushes when the setting is off

kean:
Fixes PCIOS-803

Recreated from #4989 to clear out the earlier reviews. Same change, rebased on `trunk`.

New episode pushes are server-driven: the server decides who gets them from the `push_messages_on` mask sent on every refresh. Turning off Profile → Settings → Notifications → New Episodes only wrote a device-local setting, so refresh kept sending all `1`s and the pushes kept arriving.

Behind `newEpisodeNotificationsPushOptOut`, refresh now sends `push_messages_on` as all zeros while the group is off. `push_on` is untouched, so push registration is unaffected, and a failed request self-heals on the next refresh.

Supporting fixes:

- `checkDefaults()` seeds `notifications.newEpisodes` from the legacy `PushEnabled` flag when it was never written, so long-time users aren't silently opted out.
- `NotificationsCoordinator` now owns the ordering: the setting is written before the refresh is triggered, not after.
- Tapping the bell on a single podcast no longer routes through `setupNotifications`, which was turning push back on for every podcast.

## To test

1. Sign in, turn ON Profile → Settings → Notifications → New Episodes.
2. Turn it OFF, then pull to refresh on the Podcasts tab. In Proxyman/Charles, the `user/update` request sends `push_messages_on` as all `0`s (`push_on` follows the global push state as before: still `true` if another notification group is on).
3. Turn it back ON and refresh: `push_messages_on` is all `1`s.
4. With New Episodes ON, deselect some podcasts in "Choose podcasts" and refresh: only the selected podcasts are `1`.
5. Disable the feature flag in the debug menu and repeat step 2 — `push_messages_on` stays all `1`s (old behaviour).
6. Upgrade check: on a build from `trunk`, enable notifications for a single podcast from its settings screen without opening the Notifications screen, then install this branch. New Episodes reads ON, and a refresh sends `1` only for that podcast.

## Checklist


kean:
Tested, removed some unrelated changes – looks good, has a way to revert just in case.

PR #5040, opened 2026-08-28
Open a podcast's network from its header

kean:
| 📘 Part of: # |
|:---:|

Stacked on #5038, which persists `Podcast.networkListId`.

The author line in the podcast header is now tappable when the podcast belongs to a network, and opens that network's list of podcasts — the same screen Discover and search already push.

- The header's "Category · Author" line was already one `Text` over an `AttributedString`, with the category carrying a dummy link that `OpenURLAction` intercepted. Both parts now carry a `PodcastHeaderLink` URL instead, so the same mechanism routes each one to its own action. No extra views, no layout change.
- The author is a link only when the podcast has a `networkListId` and `FeatureFlag.networkDiscovery` is on; otherwise it stays plain text, exactly as today. It's drawn in `podcast.iconTintColor(for:)`, set as a run-level `foregroundColor` so the category keeps `primaryText01`.
- `SearchNetworkNavigator` becomes `NetworkNavigator` and moves out of `New Search/`. It takes `show(listId:title:)`, derives the source with `ServerHelper.listUrlString`, loads the collection and pushes `ExpandedCollectionViewController` as its own `DiscoverDelegate`. Callers hand over a list id and nothing else — the podcast page never sees a `DiscoverItem` or `DiscoverPodcast`. `title` is optional now, since the podcast page doesn't know the network's name until the list loads; the collection names itself in that case.
- `search_network_fail_to_load` → `network_fail_to_load`, now that it isn't search-only. It was English-only (added on this branch lineage), so no translations were lost.

Analytics: `podcast_screen_network_tapped` with `podcast_uuid` and `list_id`. The event is already registered in the schema but is marked `[web]` only — Automattic/EventHorizonSchemas#127 adds `ios`.

## To test

1. Build a debug configuration, where `networkDiscovery` defaults on (or enable it under Profile > Settings > Developer > Feature Flags).
2. Open a podcast that belongs to a network — Analog(ue) (Relay) is a good one — and pull to refresh so `networkListId` gets populated by #5038.
3. Tap the podcast title to expand the header. The "Category · Author" line appears above it.
4. The author should be drawn in the podcast's tint colour rather than the text colour. Tap it: the Relay network's list of podcasts opens, titled by the list.
5. Tap a podcast in that list, and follow/unfollow one — both should behave as they do when the same screen is opened from Discover.
6. Back on the podcast page, tap the category half of the line — it should still take you to Discover filtered by that category.
7. Open a podcast with no network (or turn `networkDiscovery` off) — the author should be plain text and not tappable.
8. Search for a network and open it from the results, to confirm the shared navigator still works there.

https://github.com/user-attachments/assets/992fe0bc-cd26-4b79-8198-f8f5e8c73a1d

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Discover/NetworkNavigatorTests.swift
+++ b/PocketCastsTests/Tests/Discover/NetworkNavigatorTests.swift
@@ -5,43 +5,46 @@ import XCTest
 @testable import podcasts
 @testable import PocketCastsServer
 
-/// Opening a network tapped in search: one screen per tap, and nothing for a tap the user moved on from.
-final class SearchNetworkNavigatorTests: XCTestCase {
+/// Opening a network: one screen per tap, and nothing for a tap the user moved on from.
+final class NetworkNavigatorTests: XCTestCase {
     private var window: UIWindow!
     private var navigationController: RecordingNavigationController!
-    private var searchHost: UIViewController!
+    private var host: UIViewController!
     private var presenter: UIViewController!
     private var serverHandler: MockDiscoverServerHandler!
-    private var navigator: SearchNetworkNavigator!
+    private var navigator: NetworkNavigator!
 
-    private let network = NetworkSearchResult(uuid: "c73d120f-c174-4324-b0a3-18f9b239a59d", title: "WNYC")
-    private let otherNetwork = NetworkSearchResult(uuid: "cdb75bc0-9f5a-4217-b1ca-f573821a7913", title: "Relay")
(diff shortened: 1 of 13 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5043, opened 2026-08-28
Refresh podcast metadata for existing podcasts

kean:
| 📘 Part of: # |
|:---:|

`isExplicit` and `networkListId` are only written when the cache server returns a full podcast body. Subscribed podcasts send `lastUpdatedAt` as `If-Modified-Since`, so a podcast whose data hasn't changed answers `304` and the new columns keep their defaults forever.

This clears `lastUpdatedAt` once for all subscribed podcasts, so the next refresh of each returns the full body. No extra network requests — those refreshes already happen on podcast open, playback, download and starred sync; they just return a body instead of a `304`. Same approach `ImageManager` uses to force artwork re-downloads.

Runs once per install, gated on `networkDiscovery`. The flag check sits outside `performUpdateIfRequired` because that helper writes its key as soon as it runs — checking inside would burn the key while the flag was off.

Also makes the explicit flag symmetric in `ServerPodcastManager+Update`; it could only latch on, never clear.

## To test

1. Follow a podcast with no recent episodes, so its metadata is stable.
2. On `trunk`, confirm its `isExplicit` / `networkListId` stay unset no matter how often you open its page.
3. Build this branch with `networkDiscovery` on, launch, and open that podcast's page.
4. Both fields are now populated — explicit podcasts show the badge, networked ones show the network link.
5. Relaunch and confirm it doesn't run again (`RefreshPodcastMetadataForExplicitAndNetworkList` is `true` in `UserDefaults`).
6. With `networkDiscovery` off, nothing runs and the key stays unset.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
+++ b/Modules/Sources/PocketCastsDataModel/Private/Managers/PodcastDataManager.swift
@@ -574,6 +574,10 @@ class PodcastDataManager {
         setOnAllPodcasts(value: version, propertyName: "colorVersion", subscribedOnly: true, dbQueue: dbQueue)
     }
 
+    func clearLastUpdatedAtForAllPodcasts(dbQueue: PCDBQueue) {
+        setOnAllPodcasts(value: NSNull(), propertyName: "lastUpdatedAt", subscribedOnly: true, dbQueue: dbQueue)
+    }
+
     private func saveSingleValue(name: String, value: Any?, podcastUuid: String, dbQueue: PCDBQueue) {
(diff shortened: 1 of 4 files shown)


kean:
I tested it by following the steps, and, without this fix the app doesn't reload and doesn't get a chance to re-parse the fields. It does refetch (once) when this patch is applied. The change seems very safe, so should be OK to proceed with it.

geekygecko (approved):
This makes sense. I added some break points in the code and it behaved like I expected.

pocketcasts:
Version `8.20` has now entered code-freeze, so the milestone of this PR has been updated to `8.21`.

PR #5044, opened 2026-08-28
Draw the podcast header's category in the podcast's tint

kean:
| 📘 Part of: # |
|:---:|

The category in the podcast header opens Discover filtered by it, but it was drawn in `primaryText01` with nothing to suggest it was tappable — you had to know. It now gets the same treatment the author got in #5040 when it leads to a network: the podcast's own tint colour.

- One line in `displayCategoryAndAuthor`, setting `foregroundColor` on the category run the same way the author already does. The `·` separator and the explicit badge keep the text colour, so only the two tappable parts are tinted.
- `networkTint` becomes `linkTint`, since it now marks both parts of the line rather than just the network.
- The author is still plain text when the podcast has no network, so a podcast outside a network shows a tinted category next to a plain author.

## To test

1. Open any podcast and tap its title to expand the header — the "Category · Author" line appears.
2. The category should now be drawn in the podcast's tint colour rather than the text colour. Tap it: Discover opens, filtered by that category, as before.
3. Open a podcast that belongs to a network (Analog(ue) on Relay, with `networkDiscovery` on) — category and author are both tinted, the `·` between them isn't.
4. Open a podcast with no network — the category is tinted, the author stays plain text and isn't tappable.
5. Switch themes, including the dark ones, and check the tint stays legible against the header background.

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Podcasts/PodcastHeaderViewModelTests.swift
+++ b/PocketCastsTests/Tests/Podcasts/PodcastHeaderViewModelTests.swift
@@ -20,14 +20,14 @@ final class PodcastHeaderViewModelTests: XCTestCase {
     }
 
     func testTheAuthorOpensTheNetworkThePodcastBelongsTo() {
-        let line = viewModel(networkListId: "cdb75bc0-9f5a-4217-b1ca-f573821a7913").displayCategoryAndAuthor(networkTint: .red)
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5055, opened 2026-09-02
Support text/plain podcast transcripts

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-910

Feeds hosted on Transistor advertise their transcripts as `text/plain` (for example every episode of *The 80,000 Hours Podcast*). `TranscriptFormat` only knew about `text/vtt`, `application/srt`, `application/json` and `text/html`, so `transcriptFormat` resolved to `nil` and the player showed a "this format is not supported" error instead of the transcript.

Adds `text/plain` as a supported format, rendered as untimed text like the existing HTML transcripts (no cues, so no karaoke highlighting — plain text carries no timings). It sits last in `supportedFormats`, so a timed transcript is still preferred whenever a feed offers one. Media type matching now also ignores parameters and case, so `text/plain; charset=utf-8` resolves the same way.

## To test

1. Search for and follow *The 80,000 Hours Podcast*.
2. Open any episode and tap the transcript button in the player.
3. The transcript is displayed. On trunk this shows "Sorry, we can't display this transcript" with a `text/plain` format error.
4. Open an episode of a podcast with a VTT transcript (most shows) and confirm transcripts still highlight along with playback.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 8.21
 -----
 - Fix New Episodes push notifications still arriving after turning off Profile → Settings → Notifications → New Episodes [#5039](https://github.com/Automattic/pocket-casts-ios/pull/5039)
+- Fix transcripts not showing for podcasts whose feeds serve them as plain text, such as those hosted on Transistor [#5055](https://github.com/Automattic/pocket-casts-ios/pull/5055)
 
 8.20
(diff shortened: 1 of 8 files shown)


kean:
It works.

Screenshot 2026-09-03 at 2 28 52 PM

PR #5057, opened 2026-09-02
Keep the download task id of an in-flight download

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-867

Episodes auto-added to Up Next sometimes stayed at "not downloaded" forever, with no error — even though the download really ran and received data.

`addToQueue` writes `episode.downloadTaskId` as soon as an episode is queued, but the real `URLSessionDownloadTask` is only created later, after `updatePodcastIfRequired` returns (seconds to minutes on a background-refresh launch). `clearStuckDownloads()` treats "has a `downloadTaskId` but no live task" as stale and nulls the column, so when it ran inside that window it wiped the mapping — and `resumeDownload()` never wrote it back, it only set `taskDescription` on the task. From then on `episodeForTask` looked the episode up by a `downloadTaskId` that no longer matched anything, and every callback bailed out, including `didFinishDownloadingTo`, whose temporary file iOS deletes as soon as it returns.

Two changes:

- `performDownload` re-asserts `downloadTaskId` right before the task is created, so the lookup used by the delegate callbacks is correct no matter what happened while the download was being prepared. This also stops `clearStuckDownloads()` from cancelling the live task for an episode it can no longer resolve.
- `clearStuckDownloads()` skips episodes whose download was attempted in the last minute, closing the remaining window between it reading the episode list and the task actually being created.

The new `DataManager.saveEpisode(downloadTaskId:episode:)` writes only that column, so re-asserting it doesn't disturb the episode's download status (which matters for episodes being buffered for streaming).

## To test

The race is timing-dependent, so the regression tests are the reliable check: `DownloadManagerStuckDownloadsTests` fails on trunk and passes here.

1. Enable Settings > Auto Download > Up Next.
2. Play an episode so new episodes get added to Up Next, and background the app so a background refresh triggers a launch.
3. Episodes added to Up Next download to completion rather than staying at "not downloaded".
4. Start a download, force-quit the app mid-download, relaunch: the download resumes or is re-queued as before.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
+++ b/Modules/Sources/PocketCastsDataModel/Private/Managers/EpisodeDataManager.swift
@@ -857,6 +857,12 @@ class EpisodeDataManager {
         save(fields: fields, values: values, dbQueue: dbQueue)
     }
 
+    func saveEpisode(downloadTaskId: String?, episode: Episode, dbQueue: PCDBQueue) {
+        episode.downloadTaskId = downloadTaskId
+
+        save(fieldName: "downloadTaskId", value: DBUtils.replaceNilWithNull(value: downloadTaskId), episodeId: episode.id, dbQueue: dbQueue)
+    }
+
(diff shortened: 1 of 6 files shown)


pocketcasts:

PR #5066, opened 2026-09-04
Run the network header collage under the navigation bar

kean:
| 📘 Part of: # |
|:---:|

Runs the collection header's collage up to the top of the screen so it sits behind the transparent iOS 26 navigation bar, instead of starting below it.

The collage strip's fixed 140pt height is replaced by a constraint pinning its *bottom* 140pt below the header's top, so only its top constant moves. `collageTopBleed` pulls that top up by the safe-area inset. The avatar, subtitle, title and description are all measured from the strip's bottom, so nothing below it shifts and the header's height is unchanged.

Gated on `LiquidGlass.isEnabled && podcastCollection != nil`: pre-26 bars are opaque so there'd be nothing to see behind them, and a header-less screen (descriptive list) has no collage.

Two knock-on fixes the dark strip forces:

- **Navigation title** would sit dark on dark. It now lives in a `titleView` label that fades in once the header's own title scrolls under the bar — the pattern `PodcastViewController` and `PlaylistDetailViewController` already use.
- **Status bar** takes `.lightContent` while the collage is behind it.

Both thresholds are measured off the live header — `headerBottom(of:)` converts one of its subviews into the controller's coordinates — rather than off hardcoded heights, so they flip at the right moment at any Dynamic Type size and revert once the header scrolls away.

Also pushes the controller onto a root controller in the previews instead of making it the root, so previews carry the back button the real screen has.

## To test

1. Discover → scroll to the networks row → tap a network.
2. The dark collage runs to the top of the screen, behind the navigation bar — no white band above it.
3. The avatar, "NETWORK", the name and the description sit exactly where they did before.
4. The status bar glyphs are light over the collage, and go back to the theme's style once you scroll past it.
5. The navigation bar has no title at rest; scroll until the header's own title passes under the bar and the network name fades in there.
6. Pull down past the top and let go — the collage returns to the top edge.
7. Repeat 1–5 in a dark theme, and at a large Dynamic Type size.
8. Discover → a curated collection ("Show all" on a collection row) gets the same treatment.
9. Discover → a `descriptive_list` item still has no header, and keeps its always-visible navigation title.
10. On iOS 18, every one of these screens looks the way it did before: opaque bar, collage starting below it.

## Checklist

https://claude.ai/code/session_01S86ieisS3zwUkTdF7GvXDv

### DIFF
--- a/podcasts/DiscoverCollectionHeader.swift
+++ b/podcasts/DiscoverCollectionHeader.swift
@@ -18,6 +18,17 @@ class DiscoverCollectionHeader: UICollectionReusableView {
         }
     }
 
+    @IBOutlet var collageTopConstraint: NSLayoutConstraint!
+
+    /// How far the collage runs above the header, so that it fills the space behind the
+    /// navigation bar. Its bottom is pinned to the header, so the avatar and everything
+    /// below it stay put however far it bleeds.
+    var collageTopBleed: CGFloat = 0 {
+        didSet {
+            collageTopConstraint.constant = -collageTopBleed
+        }
+    }
+
     @IBOutlet var avatarImageView: UIImageView! {
         didSet {
             avatarImageView.layer.cornerRadius = 40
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:

PR #5069, opened 2026-09-04
Fix Hashable violation in PodcastFolderSearchResult

kean:
| 📘 Part of: # |
|:---:|

Fixes a `EXC_BREAKPOINT` crash reported from `LazyLayoutViewCache.addItem`:

```
Fatal error: Duplicate keys of type 'Canonical' were found in a Dictionary.
This usually means either that the type violates Hashable's requirements, or
that members of such a dictionary were mutated after insertion.
```

`PodcastFolderSearchResult` declares `Hashable` and hand-writes `==` over `kind` and `uuid`, but lets the compiler synthesize `hash(into:)` over all seven stored properties. Two results for the same podcast with a different `title`, `author`, `isLocal`, `explicit` or `isVideo` are therefore equal but hash differently, which makes `Set`/`Dictionary` behaviour undefined.

That type is the identity of `SearchHistoryView`'s `ForEach(searchHistory.entries, id: \.self)` (via `SearchHistoryEntry`), inside the `LazyVStack` in `SearchListView`. `LazyLayoutViewCache` is a dictionary that outlives a single update, so when `SearchHistoryModel.updateFolders()` rewrites a folder entry in place — same uuid, new name — inserting the new key finds the stale-but-equal one during probing and traps. Intermittent, because it depends on the two hashes landing in the same probe run.

This PR adds a `hash(into:)` that matches `==`. It also de-duplicates the results returned by `PodcastSearchTask.search(term:)`: the API can repeat a podcast, and `OnboardingRecommendationsView` feeds that array straight into a `LazyVStack` `ForEach` keyed by uuid, which would crash the same way.

Unit tests cover the hash/equality consistency, the `Set` behaviour, podcast-vs-folder results sharing a uuid, and the de-duplication. Removing `hash(into:)` fails three of them.

## To test

There is no reliable manual repro — the crash depends on hash bucketing. The regression is covered by tests:

1. Run `PocketCastsServerTests/PodcastFolderSearchResultTests` (6 tests).
2. Run `PocketCastsTests/SearchHistoryModelTests` (8 tests).
3. Sanity-check search still behaves: open Search, search for a podcast, tap a result, reopen Search and confirm it shows in Recent.
4. With a Plus account, tap a folder from search results, rename the folder, and reopen Search — the Recent row should show the new name and not crash.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
+++ b/Modules/Sources/PocketCastsServer/Public/Search/PodcastSearchTask.swift
@@ -97,6 +97,11 @@ public struct PodcastFolderSearchResult: Codable, Hashable {
     public static func ==(lhs: PodcastFolderSearchResult, rhs: PodcastFolderSearchResult) -> Bool {
         lhs.kind == rhs.kind && lhs.uuid == rhs.uuid
     }
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(kind)
+        hasher.combine(uuid)
+    }
 }
 
 extension PodcastFolderSearchResult: Identifiable {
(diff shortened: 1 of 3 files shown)


pocketcasts:

PR #5078, opened 2026-09-08
Add the What's New message detail screen

kean:
|:---:|

Part of PCIOS-927, PCIOS-928

Opens a feed row onto its message: a horizontal pager over `content.pages` with page indicators (hidden for single-page messages), each page scrolling vertically under a translucent navigation bar. Renders every block the app models — `heading`, `paragraph`, `image` (sized from the published `width`/`height` so the page doesn't reflow on load), `video`, and `action`. Tapping a row pushes the detail and marks the message read.

Videos autoplay muted and looping over their poster, only while their page is on screen; Reduce Motion, Low Power Mode, or a load failure leaves the poster up with a play button. The audio is taken off the item rather than only muted, since a muted player still claims the shared audio session and would interrupt whatever is playing in another app. Sidecar WebVTT captions are parsed with SwiftSubtitles, already used by transcripts.

Decisions worth flagging:

- **The action allowlist is by scheme.** `WhatsNewLink` resolves a URL before the button is drawn: `pktc://` routes as a deep link, `https://` opens in an in-app Safari view, everything else is dropped and logged with no button rendered. A host allowlist would break research links to third-party surveys. This answers the plan's open "link/deep-link allowlist" question.
- **`pocketcasts://` is accepted as an alias for `pktc://`**, since the plan's examples use it and the app doesn't register it. The feed could publish `pktc://` instead and the alias would go.
- **Actions are pinned to the bottom of the page** rather than drawn where the catalog put them, so a button stays reachable however far the content scrolls.
- **Media is capped at ~52% of the page height**, so portrait screenshots narrow instead of pushing text off screen while landscape images still fill the width.
- **A message this build can't draw never reaches the feed.** Every block on a page can be dropped — an action the allowlist rejects is the app's own doing — and a message left with no pages would open onto an empty pager.
- **Read state stays in the feed view model**, the only path onto this screen; syncing is PCIOS-924, analytics PCIOS-933, and `poll` stays dropped by decoding until its contract exists.

The mock catalog gains a `known_issue` message built around a video and a long release notes message that has to scroll, and its media now points at assets that actually load, so the previews render something.

## To test

No entry point yet (PCIOS-925), so this is reviewed through the previews:

1. `WhatsNewFeedView.swift` → "Feed in a navigation controller". Tap a row — the message opens and its dot clears when you come back.
2. `WhatsNewMessageView.swift`: "Multiple pages" swipes with the dots tracking, "Try transcripts" switches to the Podcasts tab, "A single page" shows no dots, "A long page" scrolls under the bar with its button staying put, "A video demo" plays.
3. `WhatsNewMessageView_Previews` renders the secondary call to action once per theme.
4. `WhatsNewVideoView.swift` previews — check autoplay, tap to pause, and looping, and that a podcast playing in another app keeps going.
5. Turn on Reduce Motion or Low Power Mode and reopen a video: it waits on its poster until tapped.
6. At the largest accessibility text size, text grows and the page scrolls instead of clipping.

## Checklist

### DIFF
--- a/Modules/Sources/PocketCastsServer/Public/WhatsNew/WhatsNewCatalog+Mock.swift
+++ b/Modules/Sources/PocketCastsServer/Public/WhatsNew/WhatsNewCatalog+Mock.swift
@@ -5,7 +5,7 @@ public extension WhatsNewCatalog {
     ///
     /// The messages are published relative to now rather than on fixed dates, so the feed keeps
     /// rendering the same spread of relative dates however long after this was written it's read.
-    static let mock = mock(publishedDaysAgo: [0, 8, 27, 36, 62])
+    static let mock = mock(publishedDaysAgo: [0, 8, 27, 36, 62, 90])
 
     /// A catalog whose messages were published the given number of days ago, most recent first.
     ///
@@ -39,8 +39,9 @@ public extension WhatsNewCatalog {
 
     /// The messages the mock catalog is built from, each missing the `$publishedAt` the catalog fills in.
     ///
-    /// The user research message carries a `poll` block the app doesn't model, so anything rendering
-    /// the mock exercises a page that drops a block and still has something left to show.
+    /// Between them they cover every message type and every block the app renders. The user research
+    /// message carries a `poll` block the app doesn't model, so anything rendering the mock exercises
+    /// a page that drops a block and still has something left to show.
     private static let mockMessages = [
         """
         {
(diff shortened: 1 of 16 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5080, opened 2026-09-09
Auto download episodes added to a manual playlist

kean:
| 📘 Part of: # |
|:---:|

Fixes PCIOS-946

Turning on **Auto Download** in Playlist Options for a manual playlist had no effect until the next server sync: the user had to tap download on each episode manually.

`PlaylistManager.checkForAutoDownloads()` was only ever invoked from `ServerSyncManager.performActionsAfterSync()`. Adding an episode to a manual playlist goes through `DataManager.add(episodes:to:)`, which writes the join rows and nothing else, so nothing queued a download. The auto-download logic itself was already correct for manual playlists — `PlaylistQueryBuilder.query(clause:for:...)` branches on `playlist.manual`, and `EpisodeFilter.maxAutoDownloadEpisodes()` falls back to `Constants.Values.defaultPlaylistDownloadLimit`.

### What changed

`PlaylistManager` gains a per-playlist entry point:

- `queueAutoDownloads(for:downloadsAllowedNow:dataManager:downloadManager:)` holds the logic that used to be inline in the loop, with the dependencies injectable so it can be unit tested. It keeps the existing wifi/mobile-data gate (`NetworkUtils.shared.isConnectedToUnexpensiveConnection()` / `Settings.autoDownloadMobileDataAllowed()`) and the `queueForLaterDownload` vs `addToQueue` split, and still skips episodes that are already downloaded or queued.
- `checkForAutoDownloads(in:)` runs that off the main thread and posts `manyEpisodesChanged` when it queued anything, so the episode rows pick up the new download state.
- `checkForAutoDownloads()` (the sync path) is now a loop over the same function. Its behaviour is unchanged.

It is called from every path that adds episodes to a manual playlist:

- `ManualPlaylistsChooserViewController.doneTapped()` — single and multi-episode. This also covers `MultiSelectHelper.addToPlaylist` and the `SwipeActionsHelper` add-to-playlist swipe, which both present this chooser.
- `LocalSearchCoordinator.handleAddEpisode` — the search screen `PlaylistDetailViewController.addEpisodes()` presents.
- `NewPlaylistViewController` — both the `.addEpisode` and `.addEpisodes` creation branches. `EpisodeFilter.makeDefault()` leaves auto download off, so this is a no-op today; it's here so the invariant holds at every add path rather than depending on that default.

### Turning the toggle on downloads existing episodes

The reporter's screenshot implies they expect flipping the switch to download what's already in the playlist, not just what they add afterwards. **That is the behaviour shipped here.** `FilterEditOptionsViewController.viewWillDisappear` and the playlist picker in `DownloadSettingsViewController` both run the check after saving, and it scans the whole playlist up to the auto-download limit. It is idempotent — episodes already downloaded or queued are skipped — so dismissing the options screen repeatedly costs one query.

### Related

`PCIOS-867` (PR #5057, in progress) is a `downloadTaskId` race between `addToQueue` and `clearStuckDownloads()` that can silently drop an auto-download that did start. It compounds this bug: with this PR the downloads are now queued promptly, but until #5057 lands some of them can still be dropped on a background-refresh launch. Worth landing both before closing out the user report.

## To test

1. Create a manual playlist and add a couple of episodes to it.
2. Open **Playlist Options** and turn **Auto Download** on, then leave the screen.
3. The episodes already in the playlist start downloading.
4. Add another episode to the playlist — via the playlist's "Add Episodes" search, the episode detail "Add to Playlist", a swipe on an episode row, or multi-select — and confirm it starts downloading straight away, without waiting for a sync.
5. Repeat step 4 with **Settings → Downloads → Only on WiFi** and mobile data active: the episode should show *Waiting for WiFi* rather than downloading.
6. Turn Auto Download off and add another episode: nothing should download.

## Checklist

https://claude.ai/code/session_01ELNEVm8HCgqS7qJhhhvcmC

### DIFF
--- a/PocketCastsTests/Tests/Playlists/PlaylistManagerAutoDownloadTests.swift
+++ b/PocketCastsTests/Tests/Playlists/PlaylistManagerAutoDownloadTests.swift
@@ -1,0 +1,149 @@
+import XCTest
+import PocketCastsDataModel
+@testable import podcasts
+
+final class PlaylistManagerAutoDownloadTests: DBTestCase {
+    func testQueuesEpisodeAddedToManualPlaylist() throws {
+        let playlist = makeManualPlaylist(autoDownload: true)
+        let episode = makeEpisode()
+        dataManager.add(episodes: [episode], to: playlist)
+
+        let queued = PlaylistManager.queueAutoDownloads(for: playlist, downloadsAllowedNow: false, dataManager: dataManager, downloadManager: downloadManager)
+
+        XCTAssertEqual(queued, [episode.uuid])
+        let refreshed = try XCTUnwrap(dataManager.findEpisode(uuid: episode.uuid))
+        XCTAssertEqual(refreshed.episodeStatus, DownloadStatus.waitingForWifi.rawValue)
+        XCTAssertEqual(refreshed.autoDownloadStatus, AutoDownloadStatus.autoDownloaded.rawValue)
+    }
+
+    func testDownloadsImmediatelyWhenConnectionAllowsIt() throws {
+        let playlist = makeManualPlaylist(autoDownload: true)
+        let episode = makeEpisode()
+        dataManager.add(episodes: [episode], to: playlist)
+
+        let queued = PlaylistManager.queueAutoDownloads(for: playlist, downloadsAllowedNow: true, dataManager: dataManager, downloadManager: downloadManager)
+
+        XCTAssertEqual(queued, [episode.uuid])
+        let refreshed = try XCTUnwrap(dataManager.findEpisode(uuid: episode.uuid))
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pocketcasts:

PR #5081, opened 2026-09-10
Give the tab bar back when a multi-select screen goes off-screen

### LINKED ISSUE #5070: Tab bar vanishes - Quitting app is needed
### Description

When you add episodes from a podcast to a manual playlist and click the „view“ button in the banner that appears on the bottom of the player, the manual playlist opens but the tab bar is gone.

I would like to mention that this small banner at the bottom is pretty „annyoing“ as it blocks all interactions for some seconds. Would be better to have a small banner or a banner above the mini-player and tab bar.

1. Select Epsiodes from a Podcast
2. Add them to a (manual) playlist
3. click the „view“ button on the bottom Corner
4. the Tab bar and Mini Player are gone

To get the Tab bar back you need to force quit the app.

https://github.com/user-attachments/assets/94e9dac5-4bd2-4a24-98f2-0bcf192ff665

- [x] I have searched for existing bug reports.

iPhone 15 Pro Max
PocketCast Beta 8.20.0.0

kean:
Fixes PCIOS-945
Closes #5070

Multi-select hides the tab bar and mini player through `setEnclosingTabBarHidden`, which writes to the *shared* `UITabBarController` — but it's driven by a *per-screen* mode flag that only gets cleared when the user cancels or an action completes. Leaving a screen any other way left the tab bar and mini player hidden app-wide until a force quit.

The reported path: `MultiSelectHelper.addToPlaylist` never called `multiSelectActionCompleted()`, so tapping **View** in the "added to playlist" toast switched tabs with multi-select still on. It wasn't the only path — any `NavigationManager`-driven tab switch (deep link, push notification, widget, CarPlay) leaked the same way from Downloads, Starred, Listening History, Uploaded, Playlist detail, the podcast page, the profile Bookmarks list, and podcast-list edit mode. `BookmarksListView` looked covered by `.onDisappear`, but that never fires on a UIKit tab switch since the hosting view isn't removed from the hierarchy. Up Next and Folders were already safe — both reset their mode in `viewWillDisappear`.

Two changes:

- **`PCViewController` tracks the hidden tab bar** — screens call `setHidesEnclosingTabBar(_:animated:)` instead of `setEnclosingTabBarHidden(_:animated:)`. `PCViewController` remembers it, hands the tab bar and mini player back in `viewWillDisappear`, and takes them again in `viewWillAppear`, so individual screens don't need to do anything. Profile → Bookmarks is a hosting controller rather than a `PCViewController`, so it does the same in its own `viewWillAppear`/`viewWillDisappear`.
- **Add to Playlist now finishes** — the chooser gained an `onCompletion` closure, fired on Done (not on Close), wired to `multiSelectActionCompleted()`. Multi-select now exits after adding, like every other action.

## To test

1. Open a podcast, long-press an episode to enter multi-select, and select a few episodes.
2. Tap **Add to Playlist**, pick a manual playlist, tap **Done**.
3. Tap **View** in the toast — the Playlists tab opens with the tab bar and mini player intact, and the podcast screen is no longer in multi-select.
4. Re-enter multi-select on the podcast screen, then switch tabs directly — the tab bar comes back on the new tab, and returning to the podcast screen restores multi-select and hides it again.
5. Repeat step 4 on Downloads, Starred, Listening History, Uploaded, a playlist detail screen, and Profile → Bookmarks.
6. Enter podcast-list reorder mode, trigger a deep link that switches tabs (e.g. tap a push notification), and confirm the tab bar returns.

## Checklist


pocketcasts:

PR #5082, opened 2026-09-10
Answer a poll inside a What's New message

kean:
| 📘 Part of: # |
|:---:|

> [!WARNING]
> **The poll isn't fully specified yet.** The published contract only sketches one, and a response has nowhere to go, so this is a draft: the screen works end to end against mocked data while the shape of a poll and the transport for an answer are still being decided.

Adds the `poll` content block to What's New, so a User Research message can ask questions the reader answers in place instead of only linking out to a survey.

Two pieces are stubbed with `TODO:` comments until the contract lands:

- **Server parsing** — the plan publishes a single `question` with its `options` on the block itself and leaves the rest "to be defined when the poll use case is implemented". `WhatsNewPoll` decodes that shape *and* a `questions` array, so a poll can ask more than one thing without another release.
- **Sending the answers** — the destination, auth, validation, idempotency, retention and analytics for a response are all undecided, so rather than inventing an endpoint (or a task that wraps one), `WhatsNewPollViewModel.submit` records the answers in the log, waits a beat and reports success. The sending and failed states are already wired for whatever replaces it.

Behaviour worth calling out:

- A question with no options is dropped, and a poll left with nothing to ask is dropped with it, so a message never opens onto a button that can't be enabled.
- A second poll on the same page is dropped: one pinned button sends one poll's answers.
- Questions scroll with the rest of the page; only the button is pinned, so a long poll keeps it in reach.
- The whole feature stays behind the existing `whatsNewFeed` flag.

## To test

1. Enable the `whatsNewFeed` feature flag and open **Profile → What's New**.
2. Open **Exploring social features**.
3. First page: the question sits above four options. Tap one — it fills in with a checkmark and a blue border, and **Continue** goes from disabled to enabled. Tap another and the choice moves rather than piling up; tap the same one twice and it clears, disabling **Continue** again.
4. Swipe to the second page: two questions, and **Continue** stays disabled until both are answered. The second takes more than one answer at a time.
5. Tap **Continue**. The button shows a spinner, then is replaced by "Thanks for your feedback", and the options stop taking taps. Check the log for the `poll … answered with …` line — nothing leaves the device yet.
6. Turn text size up to an accessibility size and check the option rows grow with it rather than clipping.
7. With VoiceOver on, check each option reads its label and reports itself as selected once chosen.
8. Switch themes and check the chosen option's fill and border in each.

## Checklist

https://claude.ai/code/session_01DGuVU4nMDm3EhtkXqysmTS


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

kean:
Superseded by https://github.com/Automattic/pocket-casts-ios/pull/5091

PR #5084, opened 2026-09-10
Add the What's New entry point to Profile

kean:
|:---:|

Fixes PCIOS-925

Adds the What's New row to Profile, behind the `whatsNewFeed` flag, so the feed built in #5077 and #5078 finally has a way in. The row sits at the top of the Profile list and pushes `WhatsNewFeedViewController`.

The feed view model gains a second init that takes no messages and fetches the catalog itself: `load()` runs from the view's `.task`, asks `WhatsNewCatalogTask` for the published catalog (falling back to its cached copy when the network is out), and fills the list in. The `messages:` init stays for previews and tests and never touches the network.

Read state moves into a `readMessageIDs` set on the view model rather than living only in the item structs, so a reload doesn't resurrect dots the reader already cleared. It's still in-memory only — syncing is PCIOS-924.

Tests cover the three paths through the new init: the catalog fills an empty feed, a reload keeps what was read, and a message-backed feed is left alone. They run against a `URLProtocol` stub with an ephemeral cache directory.

## To test

1. Enable the `whatsNewFeed` flag in the developer/feature flag settings and restart.
2. Go to the Profile tab — **What's New** is the first row in the list.
3. Tap it. The feed loads from the published catalog and shows the messages with unread dots.
4. Open a message, go back — its dot is gone.
5. Leave the screen and come back. The feed reloads and the message you read stays read.
6. Turn on Airplane Mode and reopen the feed — the cached catalog still renders.
7. Turn the flag off — the row is gone from Profile.

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Whats New/WhatsNewFeedViewModelTests.swift
+++ b/PocketCastsTests/Tests/Whats New/WhatsNewFeedViewModelTests.swift
@@ -88,7 +88,42 @@ final class WhatsNewFeedViewModelTests: XCTestCase {
         XCTAssertFalse(viewModel.hasUnreadItems)
     }
 
+    /// The Profile row hands over an empty feed, so nothing shows until the catalog arrives.
+    func testLoadingFillsTheFeedFromTheCatalog() async {
+        let viewModel = WhatsNewFeedViewModel(catalogTask: catalogTask(publishing: Self.catalogJSON))
+        XCTAssertTrue(viewModel.items.isEmpty)
+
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5085, opened 2026-09-10
Filter out What's New messages the user isn't targeted by

kean:
|:---:|

Part of PCIOS-924 (the targeting half; read-state sync is still to come)

The CDN publishes one catalog per platform and locale, so the rest of the targeting is the client's to apply. `WhatsNewMessageFilter` is the one place that decides whether a message is meant for this user: its audiences against the account's tier, its `minimumAppVersion` against the running build, and its `publishedAt`/`expiresAt` against now. The dates are checked here rather than trusted to the catalog because a cached one can outlive a message.

`WhatsNewFeedViewModel` applies the filter before anything else, so it covers both the catalog and the messages-in-hand paths, and the unread count behind the Profile dot can never include a message the feed won't show.

`Version` (in `PocketCastsUtils`) parses a dotted version and compares it component by component — what comparing strings with `.numeric` doesn't give, since `8.1` and `8.1.0` are the same version but aren't equal as text. An unreadable version, an unparsable minimum, or an audience this build doesn't understand hides the message rather than guessing.

## To test

Targeting isn't visible without published test content, so this is mostly covered by the unit tests. To see it end to end:

1. Enable the `whatsNewFeed` flag and restart.
2. Point the app at a catalog with a message targeted at `["patron"]` and one with `"audiences": []`.
3. Open **Profile → What's New** on a free account — only the untargeted message is listed, and it's the only one the unread dot counts.
4. Add a message with `"minimumAppVersion"` above the running build — it doesn't appear. Set it to the running version, or that version with a `.0` on the end, and it does.
5. Add a message with an `expiresAt` in the past — it doesn't appear even once the catalog is cached and you're offline.

Unit tests: `PocketCastsUtilsTests/VersionTests`, `PocketCastsServerTests/WhatsNewMessageFilterTests`, `PocketCastsTests/WhatsNewFeedViewModelTests`.

## Checklist

https://claude.ai/code/session_01DGuVU4nMDm3EhtkXqysmTS

### DIFF
--- a/Modules/Sources/PocketCastsServer/Public/WhatsNew/WhatsNewMessageFilter.swift
+++ b/Modules/Sources/PocketCastsServer/Public/WhatsNew/WhatsNewMessageFilter.swift
@@ -1,0 +1,66 @@
+import Foundation
+import PocketCastsUtils
+
+/// Decides which of the catalog's messages a user is meant to see.
+///
+/// The CDN publishes one catalog per platform and locale, so every remaining rule — who a message
+/// is for, which builds can render it, and when it's live — travels on the message itself. Anything
+/// reading the feed has to apply the same rules, whether it's drawing the list or only deriving the
+/// unread dot, or a message could be counted unread on the Profile row and then be nowhere to be
+/// found in the feed it sends the user to.
+public struct WhatsNewMessageFilter {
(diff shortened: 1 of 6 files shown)


kean:
Merging. I left a note to look into how "audience" should be handled.

PR #5086, opened 2026-09-10
Add loading, empty and error states to the What's New feed

kean:
|:---:|

Part of PCIOS-926

Gives the What's New feed something to show while it has no rows, using the standard `ContentUnavailableView`:

- **Loading**: a spinner while the catalog is fetched.
- **Empty**: "No messages yet" when the catalog has nothing this build can show.
- **Error**: "Unable to load What's New" with **Try Again** when the catalog can't be reached and there's no cached copy to fall back to. The copy and button follow Discover's no-network state.

The view model gains a `state` (`loading` / `loaded` / `failed`) and `retry()`. A load cancelled by leaving the screen doesn't flip to the error, so the next visit just tries again. The states sit in an overlay shown only while there are no rows, so the reload that runs when coming back from a message never swaps the list for a spinner or an error.

## To test

1. Delete and reinstall the app so there's no cached catalog.
2. Enable the `whatsNewFeed` flag and restart.
3. Turn on Airplane Mode and go to Profile → **What's New**. A spinner shows, then "Unable to load What's New" with **Try Again**.
4. Turn off Airplane Mode and tap **Try Again**. The spinner shows again, then the feed loads.
5. Open a message and go back. The list stays put with no spinner.
6. For the empty state, map the catalog response to `{"schemaVersion": 1, "messages": []}` (e.g. with Proxyman) and open the feed, or check the **Empty** preview in `WhatsNewFeedView.swift`.

## Checklist

### DIFF
--- a/PocketCastsTests/Tests/Whats New/WhatsNewFeedViewModelTests.swift
+++ b/PocketCastsTests/Tests/Whats New/WhatsNewFeedViewModelTests.swift
@@ -92,9 +92,32 @@ final class WhatsNewFeedViewModelTests: XCTestCase {
     func testLoadingFillsTheFeedFromTheCatalog() async {
         let viewModel = WhatsNewFeedViewModel(catalogTask: catalogTask(publishing: Self.catalogJSON))
         XCTAssertTrue(viewModel.items.isEmpty)
+        XCTAssertEqual(viewModel.state, .loading)
 
         await viewModel.load()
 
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5096, opened 2026-09-11
Let users turn off the What's New unread dot

kean:
|:---:|

Fixes PCIOS-932

> [!NOTE]
> Stacked on #5109 (the `release/8.21` backmerge) for the protobuf types from #5103. Merge that first; this PR retargets to `trunk` when it lands.

Adds a **Show unread dot** setting under **General → What's New**, behind `whatsNewFeed`. It's on by default and syncs as the global `show_whats_new_dot` named setting, the same way as Generated chapters. The `show_whats_new_dot` proto fields come from the `api.pb.swift` regeneration in #5103.

Turning it off hides the dots on the Profile tab and the What's New row. Messages stay unread, the feed keeps its own dots, and **Read all** is unchanged. Toggling tracks `settings_general_whats_new_unread_dot_toggled` with `enabled` ([EventHorizonSchemas#129](https://github.com/Automattic/EventHorizonSchemas/pull/129)).

Notes:
- Web put this under Settings → Notifications (Automattic/pocket-casts-webplayer#4813). On iOS it's in General because the Notifications screen locks every row when push permission is denied, and the event is `settings_general_*`.
- iOS only fetches named settings when it has a local change to send, so a change made on another device shows up once iOS syncs a setting of its own. This is the same for every named setting.

## To test

1. Enable the `whatsNewFeed` flag, run the app and sign in.
2. Go to Profile → Settings → Developer → **Reset Read State**. The Profile tab shows a dot.
3. Go to Settings → General. At the bottom, under **What's New**, **Show unread dot** is on. Turn it off: the tab dot disappears right away, and the console logs `settings_general_whats_new_unread_dot_toggled` with `enabled: false`.
4. Go back to Profile. The What's New row has no dot. Open What's New: messages still show their unread dots, and **Read all** clears them.
5. Reset the read state again and turn **Show unread dot** back on. The Profile tab and What's New row dots come back.
6. Turn it off again, then open the web player's Settings → Notifications with the same account. **Show unread dot** is off there too.

## Checklist


pocketcasts:

kean:
Tested  that sync works both ways using `Api_NamedSettingsRequest`. Found an issue with initial sync re-setting the "needs sync" flag – fixed.

PR #5108, opened 2026-09-16
Use the async AVAsset load API in AVFileUtil (9 warnings)

kean:
Replaces the deprecated `loadValuesAsynchronously(forKeys:)`, `statusOfValue(forKey:error:)`, `asset.duration`, `asset.commonMetadata`, `stringValue` and `dataValue` calls in `AVFileUtil` with `try await load(...)`, clearing nine iOS 16 deprecation warnings. `AVFileUtil` reads the title, embedded artwork and duration of a file added in Profile → Files.

Behavior is unchanged:

- The old code loaded the keys `"title"` and `"artwork"`, which aren't `AVAsset` properties. AVFoundation logs "unrecognized key" for them and reports them as `.loaded` whenever the asset opens, so the artwork check always passed. The new code loads `commonMetadata`, which the old code was reading synchronously anyway, and processes the artwork whenever that load succeeds. On an unreadable file, the title handler still gets `nil` and the artwork handler still isn't called.
- Duration still loads in its own task, and a failed load still skips the handler.
- The handlers are still called off the main thread, and `AddCustomViewController` already dispatches its UI work to main.

I checked this with a macOS script that runs the old and new code on the same files. Both returned the same title, artwork bytes and duration for an m4a with an embedded title and artwork and for an aiff with no metadata. They also matched on a missing file.

## To test

1. Go to Profile → Files and tap **+** to add a file.
2. Pick an audio file with an embedded title and artwork (e.g. an MP3/M4A with ID3 tags).
3. Confirm the name field shows the embedded title, the artwork appears (with an active subscription), and the file size is shown.
4. Save and confirm the file appears in the Files list with the correct duration.
5. Repeat with a file that has no embedded metadata. Confirm the name falls back to the file name and the color picker is shown instead of artwork.

## Checklist


kean:
I tested that the embedded metadata still gets added:

Screenshot 2026-09-16 at 2 49 19 PM

PR #5118, opened 2026-09-16
Fix implicit strong capture warnings in nested closures

kean:
Fixes the 22 new `'weak' ownership of capture 'self' differs from implicitly-captured strong reference in outer scope` warnings ([ImplicitStrongCapture](https://docs.swift.org/compiler/documentation/diagnostics/implicit-strong-capture)).

The compiler warns when an inner closure captures `self` weakly but an outer escaping closure captures it strongly without saying so. The outer closure keeps `self` alive, so the inner `[weak self]` does nothing. Following the compiler's guidance, the outer closure now states which ownership it wants:

- **`[weak self]`** where the outer closure only passes `self` to the inner one. This is the same pattern `archiveAll` already uses. In some places the outer closure is stored by a view the owner holds, so the strong capture could create a retain cycle: the banner close handlers in Listening History and Playlists, the Discover no-network cell registration, the podcast page "Try Again" action, and `PodcastHeaderCell`'s SwiftUI content. The rest held `self` for the length of a network call or background task.
- **`[self]`** where the outer closure uses `self` itself: the `PlaybackManager` retry `Task`, `unarchiveAll` (it reads `episodeCount()`), and the watchOS `RefreshManager.refresh`, which needs a strong capture to avoid the watchOS 10 crash. Behavior doesn't change.
- **`PlusPaywallContainer`**: removed `[weak viewModel]`. The closure also captures the view struct (for `presentSubscriptionView`), and that struct holds the view model strongly, so the weak capture had no effect.

## To test

A clean build shows no `ImplicitStrongCapture` warnings. To smoke test the affected flows:

1. Pull to refresh on the Podcasts tab and in Profile > Stats. Refresh finishes, and stats load.
2. Open a podcast. The header lays out and resizes correctly. Use "Unarchive All" from the episode options, and the list updates.
3. Open a folder on the Podcasts tab. The preview artwork loads.
4. Close the informational banner in Listening History and in Playlists. The banner goes away.
5. Settings > Import & Export > Export. The export share sheet appears.
6. Open the Plus paywall (upgrade experiment) and tap the upgrade button. Prices load and the subscription view appears.
7. In CarPlay, open More > Listening History and More > Files.
8. With the network off, open Discover and tap retry on the no-network state. Discover reloads when the network is back.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5127, opened 2026-09-16
Load the audio track with the async AVAsset API in DefaultPlayer

kean:
| 📘 Part of: # |
|:---:|

Fixes the `'tracks' was deprecated in iOS 16.0 / watchOS 9.0: Use load(.tracks) instead` warnings in `DefaultPlayer`.

When the player item became ready, `playerStatusDidChange` read `asset.tracks` to find the audio track for the volume boost audio mix. That synchronous getter is deprecated, and it can block the main thread while tracks load.

Player items are now created with `automaticallyLoadedAssetKeys: [.tracks]`, so an item doesn't report `readyToPlay` until its tracks are loaded. `loadAudioTrack(for:)` then reads them without blocking through `asset.status(of: .tracks)` and attaches the audio mix in the same callback as before.

An earlier version of this PR loaded the tracks in a `Task` instead. That attached the mix only after playback could already have started, so an episode began without Volume Boost and then jumped in loudness.

Other changes:

- The ready handler now runs once per item, gated on `isWaitingForInitialPlayback` instead of `assetTrack == nil`. HLS streams never set `assetTrack`, so they used to rerun it on every status change.
- `loadEpisode` resets `assetTrack` and `audioMix`, so a reused `DefaultPlayer` sets up its new item instead of keeping the previous item's track and mix.
- A failed track load is now logged to FileLog.
- The share and clip players create their items through the same `DownloadManager.downloadParallelToStream(of:)`, so they preload tracks too.

**Why HLS gets no mix:** the volume boost mix uses an `MTAudioProcessingTap`, which needs a concrete audio asset track. HLS streams don't expose one, and attaching the mix anyway breaks playback at speeds other than 1x: the audio ignores the rate while the video honors it. The mix is only attached when an audio track is found.

## To test

Keep Trim Silence off so these episodes play through `DefaultPlayer` (the logs show `Using DefaultPlayer`).

1. Turn on Volume Boost, then play a downloaded episode from the start. It's louder from the first second, with no jump in volume.
2. Repeat with a streamed (non-HLS) episode.
3. Play an HLS episode at 2x. The audio plays at 2x.
4. Play a downloaded episode on the watch. It plays normally.
5. Share a clip from an episode. The clip preview plays.

## Checklist


kean:
I found a better way to implement it with `automaticallyLoadedAssetKeys: [.tracks]`, so no concurrency is introduced.

PR #5129, opened 2026-09-16
Fix non-Sendable export session captures in VideoExporter

kean:
| 📘 Part of: # |
|:---:|

Fixes the `capture of 'exportSession' with non-Sendable type 'AVAssetExportSession' in a '@Sendable' closure` warnings in `VideoExporter.exportFinalComposition`.

The progress `Timer` closure and the `onCancel` handler are both `@Sendable`. They now read the session through `UnsafeTransfer`, the same way `AudioClipExporter` handles its export session. Both reads are safe: `progress` can be read from any thread, and `cancelExport()` is documented as safe to call from any thread.

## To test

1. Open an episode's share sheet and pick a clip.
2. Share it as a video. The export progress updates and the video exports.
3. Start another video export and cancel it midway. The export stops.

## Checklist


kean:
Makes sense to ignore it. There is no possible race condition here.

PR #5131, opened 2026-09-16
Fix isolated conformance warnings in VideoExporter

kean:
| 📘 Part of: # |
|:---:|

Fixes the Xcode 27 `conformance of 'Content' to protocol ... may be isolated and cannot be passed to ... context` warnings (`IsolatedConformances`) in `VideoExporter`.

`VideoExporter` passes its generic `Content` view from the main actor into nonisolated code, and from there back to the main-actor `pixelBuffer`. The compiler can't rule out that `Content`'s conformance is isolated to an actor, and such a conformance can't cross those boundaries. Two changes fix this:

- `AnimatableContent` now refines `SendableMetatype`, so conforming types can't have isolated conformances. Its only conformer, `AnimatedShareImageView`, already qualifies.
- `pixelBuffer` takes `some AnimatableContent` instead of `some View`, so the requirement also applies there.

Both changes are needed. Without either one, one of the two warnings remains.

## To test

1. Share a clip as a video. The export finishes, and the video shows the animation.

## Checklist


podcasts/Sharing/Clip/VideoExporter.swift
```diff
@@ -3,7 +3,7 @@ import AVFoundation
 import UIKit
 import PocketCastsUtils
 
-protocol AnimatableContent: View {
+protocol AnimatableContent: View, SendableMetatype {
```
kean, line 6: TIL: `SendableMetatype`

PR #5132, opened 2026-09-16
Remove the addMissingPodcastAndEpisode deprecation and document that it blocks

kean:
| 📘 Part of: # |
|:---:|

Fixes the two `'addMissingPodcastAndEpisode(episodeUuid:podcastUuid:shouldUpdateEpisode:completion:)' is deprecated` warnings. #4721 deprecated the synchronous overload. This PR removes the deprecation and keeps the overload.

Moving the last two callers to the async API is risky. `SyncTask.importPlaylist` and `SyncYearListeningHistoryTask.updateEpisodes` both have to finish before their thread moves on. Calling the async API from them means blocking a thread on a semaphore or dispatch group while the work runs on the Swift concurrency pool. `EndOfYearStoriesBuilder.build()` already blocks one of those threads during the year sync. To do this properly, the callers need to run in an async context.

Until then, the overload's documentation warns that it performs synchronous networking and blocks the calling thread. Both call sites have a comment saying they block.

No behavior changes. The diff only touches documentation and removes the `@available` attribute.

## To test

No manual testing needed. Build the app and confirm the two deprecation warnings are gone.

## Checklist


kean:
I decided to remove the deprecation warnings I previously added and just leave a comment instead. There is no way to cleanly change it without refactoring the entire scope to be asynchronous. That would involve a major refactor.

PR #5143, opened 2026-09-17
Isolate clip sharing playback to the main actor

kean:
| 📘 Part of: # |
|:---:|

Marks `ClipPlaybackManager` and `ClipTime` as `@MainActor`. Both are `ObservableObject`s used only by SwiftUI, but `ClipPlaybackManager` was writing its `@Published` state from AVFoundation callbacks that aren't guaranteed to run on main. That's a real "Publishing changes from background threads" violation.

Fixes the annotation surfaced:

- The `seek(to:)` completion and the loop-back seek completion (which also pauses and resets `currentTime`) now hop to main. AVPlayer doesn't promise a queue for these.
- The `timeControlStatus` KVO sink now has `.receive(on: DispatchQueue.main)`. The compiler didn't flag this one, because the non-Sendable closure silently inherited main isolation while KVO delivers on whatever thread changed the status.
- The periodic time observer is registered with `queue: .main`, so its body runs in `MainActor.assumeIsolated`. The body moved into `handlePeriodicTime(_:)` unchanged.

The seek hops go through the main queue, so they stay in order with the periodic observer that shares it. Clearing `isSeeking` one turn later skips at most one 0.5s tick.

`SharingModal.Option.shareURL` and `ShareDestination.logClipShared` read `ClipTime` and are now `@MainActor`; every caller already was. `VideoExporter` only receives plain values and is untouched. `static var shared` → `static let`.

## To test

1. Play an episode, open Share → Clip.
2. Tap play in the trim bar: audio plays, the playhead moves, the button shows pause.
3. Drag the playhead while playing: it seeks and doesn't jump back.
4. Let playback reach the end handle: it jumps to the clip start, pauses, and the button returns to play.
5. Close the sheet: audio stops.
6. Tap Next → Copy Link (URL has `?t=start,end`), then More (system sheet shows the exported clip).

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5128, opened 2026-09-16
Replace deprecated NavigationLink(isActive:) in Suggested Folders

kean:
| 📘 Part of: # |
|:---:|

Fixes the `'init(destination:isActive:label:)' was deprecated in iOS 16.0` warning in `SuggestedFoldersView`.

The "Create custom folder" link is now a button that sets `createFolderActive`, and `.navigationDestination(isPresented:)` pushes `CreateFolderView`. The view is already inside a `NavigationStack` (`NavigationContainer`). The button now tracks `suggestedFoldersCreateCustomFolderTapped`, which replaces the `onChange` that did the same thing.

## To test

With a Plus account and enough podcasts to get suggestions:

1. On the Podcasts tab, open the menu and create a folder to show Suggested Folders.
2. Tap **Create custom folder**. The create folder screen opens in the same sheet, and `suggested_folders_create_custom_folder_tapped` is tracked once.
3. Create the folder. The sheet closes and the new folder opens.
4. Repeat, but go back from the create folder screen. Suggested Folders shows again.

## Checklist

- [ ] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [ ] I have considered adding unit tests for my changes.
- [ ] I have updated (or requested that someone edit) [the Event Horizon schema](https://github.com/Automattic/EventHorizonSchemas/blob/trunk/schema/pocket-casts.yml) to reflect any new or changed analytics.

PR #5145, opened 2026-09-17
Isolate the folder view models to the main actor

kean:
| 📘 Part of: # |
|:---:|

Marks the folder view models as `@MainActor`: `PodcastPickerModel`, `ChoosePodcastFolderModel`, `SuggestedFoldersModel`, `FolderModel` and `FolderHistoryModel`. They're only created and called from SwiftUI views and UIKit controllers, do no background work of their own, and post their notifications with `postOnMainThread`.

- `FoldersCoordinator` is `@MainActor` too. It creates `SuggestedFoldersModel` and only presents view controllers. Its sole owner, `PodcastListViewController`, first touches it from UI actions or inside `DispatchQueue.main.sync`, and its sinks already receive on main.
- `SuggestedFoldersModel.load()` was a nonisolated `async` function, so its `loadingState == .loading` guard read a `@Published` property on the global executor while the inner `Task { @MainActor in }` wrote it. The guard now runs on main. The network call still runs off-main and resumes on main.
- `FolderHistoryModel`'s per-method `@MainActor` moved to the class, which now also covers `restore()` (it shows a toast).
- In Swift 5 mode a default argument can't call a main-actor init, so `SuggestedFoldersView.init` no longer defaults `model`. The preview was the only caller relying on it. This touches two lines of the file #5128 also edits, well away from its changes.

No hops were needed; every consumer was already on main.

## To test

1. Podcasts → create folder → pick podcasts → name → color → Save. The folder appears and opens.
2. With more than 7 podcasts, Plus and `suggestedFolders` on, tap create folder. The suggestions sheet shows loading, then suggestions. Try "Use suggested folders" and "Create custom folder".
3. Open a folder → Edit folder. Rename it, change the color, then delete it. The grid updates each time.
4. Open a folder → Add/remove podcasts. Search, change the sort, select/deselect all, dismiss. The contents update.
5. From a podcast page, choose a folder, move the podcast, then set "No folder".
6. Settings → Folders history → open an entry → Restore. The toast shows and the folders come back.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5146, opened 2026-09-17
Isolate the playlist view models to the main actor

kean:
| 📘 Part of: # |
|:---:|

Marks the playlist view models as `@MainActor`: `PlaylistDetailViewModel`, `PlaylistPreviewViewModel`, `PlaylistCellViewModel`, `NewPlaylistCellViewModel` and `SmartRuleToggleViewModel`. Every consumer is a view controller, cell or SwiftUI view.

Off-main access fixed along the way:

- **`podcastsRefreshed` observer on the playlist detail screen.** The notification is raw-posted from `RefreshOperation` (operation queue) and from `RefreshManager`'s throttled path (`DispatchQueue.global()`). The selector ran on that thread and called `ReloadScheduler.request(.episodes)`, which is `@MainActor`, mutating its state and spawning its Task off-main. The compiler didn't catch it because the view controller is inferred main-actor. The observer now has its own handler that hops to main. The post precedes the refresh completion on the same thread, so the request still lands before `didFinishRefresh`. All other observers in that file are posted with `postOnMainThread`.
- **`PlaylistDetailViewModel.reloadPlaylistAndEpisodes`** wrote `playlist` and called `reloadEpisodeList` from a global queue. Now only `findPlaylist` runs in the background. The state update and reload happen together on main.
- **Download all / Queue all** read `viewModel.episodes` inside `DispatchQueue.global()`. The empty check now uses the `episodes` argument (which callers pass from `viewModel.episodes`) before dispatching.

Kept off main despite the new isolation:

- `saveUpNextAndPlay`'s `Task {}` would have silently inherited the main actor, moving the Up Next read and the batched `createManualPlaylists` writes onto main. It's now `Task.detached`, with the same order: read Up Next, play, save, toast.
- The image-URL task groups and `loadListEpisodes` copy `imageManager` / `episodesDataManager` into locals instead of reaching through `self` off-main.

`PlaylistPreviewViewModel.removeObserver()` is `nonisolated` because `deinit` calls it.

## To test

1. Filters → open a smart playlist → pull to refresh. It shows "Refresh complete", and the list reloads if new episodes arrived.
2. Pull to refresh again immediately (throttled path). The list still reloads about 1s later, with no Main Thread Checker warnings.
3. Smart playlist → More → Edit rules → change a rule → close. The title and episodes update.
4. More → Download all (and Queue for later on cellular) on a non-empty playlist. Episodes download or queue.
5. Manual playlist → Play all with a non-empty Up Next and "save Up Next" on → Replace and play. Playback starts, then the "Up Next saved" toast shows and a new manual playlist appears.
6. Search inside a playlist, clear, end search. The list restores.
7. Create a smart playlist, add rules (the preview fills in), Save. The playlist opens.
8. Edit the Podcasts and Starred rules and their toggle headers. The Filters list cells show counts and artwork.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5154, opened 2026-09-17
Isolate Theme and ThemeObserver to the main actor

kean:
| 📘 Part of: # |
|:---:|

Marks `Theme`, `ThemeObserver`, the generated `ThemeColor` and `AppTheme` as `@MainActor`.

`Theme.sharedTheme.activeTheme` was read from any thread before this change: the `ThemeColor` defaults, `AppTheme`, `Theme.isDarkTheme()`, analytics and the podcast grid loader. Most of those callers already ran on main and only needed an annotation. Three really run off the main thread, and each now gets the theme another way. There's no thread-safe copy of the theme and no lock.

**`Theme`**
- `isDarkTheme()` and the rest of the class are main-actor isolated.
- These stay `nonisolated`: the UserDefaults keys, `sharedTheme` and `init()` (`.sharedTheme` is used in ~15 default arguments, and Swift 5 mode doesn't allow main-actor reads there), and `preferredDarkTheme()` / `preferredLightTheme()`.
- New `nonisolated savedTheme()` reads the persisted theme from UserDefaults, like the two preferred-theme getters do. `activeTheme.didSet` writes UserDefaults synchronously, so the saved value never lags behind.

**Off-main readers**
- `AnalyticsAppThemeProvider` runs inside `Analytics._track` on whatever thread called it. `theme_selected` now uses `savedTheme()`, like the other three theme properties in that dictionary.
- `UserEpisode.urlForImage` is called by `UserEpisodeManager`, `WidgetHelper` and others. It uses `savedTheme().isDark`.
- `HomeGridDataHelper.gridListItems` runs on `PodcastListViewController`'s refresh queue. It now takes the theme as a parameter; `refreshGridItems()` reads it on main and passes it in. `HomeGridListItem.empty` is gone.
- `refreshGridItems()` is also called by about ten notifications, and some are posted from background threads (`podcastsRefreshed` from `RefreshOperation`, `syncCompleted`). It now hops to main first, following the `AnalyticsCoordinator` pattern.

**Generated colors.** `scripts/themes/generate_themes.rb` emits `@MainActor struct ThemeColor`. `ThemeColor.swift` is regenerated at build time.

**`AppTheme`** is `@MainActor`. `appTintColor()` stays `nonisolated` because it's a fixed color used by `EpisodeTableHelper`'s background loads. `color(for:theme:)` is split into `color(for:)` and `color(for:theme:)`.

**`ThemeObserver`** needs a `nonisolated init()` because its subclasses are created in default arguments (`Toast.show`, `EmptyStateView`, …). It now observes `themeChanged` with a selector instead of a Combine sink. `themeChanged` is only posted from `Theme.didSet` via `postOnMainThread`, so the handler runs on main. The one behavior difference is that `objectWillChange` fires during that post rather than one run-loop turn later. `ToastTheme`, `EmptyStateViewStyle` and `BookmarksStyle` are `@MainActor` protocols, and their `.defaultTheme` / `.playerTheme` / `.defaultStyle` / `.playerStyle` factories are `nonisolated` so they still work as default arguments.

**Call sites annotated `@MainActor`** (all already main-only in practice): `PlayerColorHelper`, `SwipeActionsHelper`, `ExplicitBadgeHelper`, the `Podcast` tint helpers, the `ImageManager` loaders that take a `UIImageView` plus `placeHolderImage`, `EpisodeFilter.playlistColor()`, `PodcastCollectionColors.activeThemeColor`, `OptionsPickerRootController.Colors.init(theme:)`, the carousel card colors, the player/transcript fade colors, the TV episode row tint, `SharingModal` / `SharingHelper`, the `BannerAdView.Colors` factories, three `IntroCarouselDataSource` color members and `PlusLandingViewModel.make`. `PlayerColorHelper`'s theme defaults changed from `Theme.sharedTheme.activeTheme` to `nil`, resolved inside. `FolderModel` and `ChoosePodcastFolderModel` get the same class-level `@MainActor` as #5145, so the two PRs merge cleanly.

No `assumeIsolated` or `nonisolated(unsafe)` was added. The only thread hop is the one in `refreshGridItems()`.

## To test

1. Settings → Appearance: switch through several themes. Tabs, lists, the player, bookmarks and toasts all re-theme.
2. Turn on "Use system theme" and flip iOS dark mode. The app follows your preferred light/dark themes.
3. Fresh install: Appearance shows "Use system theme" on by default.
4. Player → Bookmarks and Profile → Bookmarks (empty and non-empty): colors are right and update when the theme changes while the screen is open.
5. Trigger toasts (shuffle Up Next; add to a full manual playlist) in light and dark themes.
6. Share an episode from the player shelf and via a list swipe. The share modal keeps its fixed contrast-light styling.
7. Podcasts tab: check the empty state, pull to refresh, the banner ad colors if shown, and folder colors after switching between light and dark themes. Grid badges update after a background refresh or sync.
8. Files: a file with a color (no custom artwork) shows the right light/dark default artwork in the list and in the Up Next widget.
9. Fresh install: the intro carousel indicators and the Plus upgrade screen look right.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

scripts/themes/generate_themes.rb
```diff
@@ -138,7 +138,7 @@ def write_theme_value(hex_val, opacity, token_name, file_path, theme_name)
 File.truncate(file_path_styles, 0) if File.exist?(file_path_styles)
 
 File.write(file_path_colors,
-           "import PocketCastsUtils\nimport PocketCastsServer\nimport UIKit\n\n// ************ WARNING AUTO GENERATED, DO NOT EDIT ************\nstruct ThemeColor {\n", mode: 'a')
+           "import PocketCastsUtils\nimport PocketCastsServer\nimport UIKit\n\n// ************ WARNING AUTO GENERATED, DO NOT EDIT ************\n@MainActor\nstruct ThemeColor {\n", mode: 'a')
```
dangermattic, line 141: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/pocket-casts-ios/blob/e772b9845cc46f608b97d56d4f12b0ec65d71465/scripts/themes/generate_themes.rb#L141"></span>Layout/LineLength: Line is too long. [191/180]</td>
    </tr>
  </tbody>
</table>

PR #5158, opened 2026-09-17
Fix crash reading player state when syncing to the watch

kean:
Fixes Sentry issue 7735252146: `EXC_BAD_ACCESS` in `-[AVPlayer rate]` from `DefaultPlayer.playing()`.

`WatchManager.sendStateToWatch` runs on its background `sessionQueue` and calls `serializeNowPlaying()`, which reads `PlaybackManager.isPlaying`, `duration()`, and `currentTime()`. Those read the current player's `AVPlayer`. At the same moment, the main thread can switch episodes (`performLoad` → `cleanupCurrentPlayer` → `setupPlayer`), which sets the player to `nil` and releases it. The background thread was left holding a pointer to an `AVPlayer` that had already been freed. In the crash report, the main thread is inside `setupPlayer` after the user tapped an episode in Up Next.

The fix builds the now-playing snapshot on the main thread with `DispatchQueue.main.sync`, since the main thread owns the player's state. This can't deadlock: nothing waits synchronously on `sessionQueue` (every use is `async`), and `sendStateToWatch` already requires that it runs on `sessionQueue`, never on main. The rest of the payload (Up Next, playlists, settings) is still built in the background.

I didn't add a unit test because the crash is a timing-dependent race that a test can't reproduce reliably.

## To test

1. Pair an Apple Watch that has Pocket Casts installed. Start playing an episode on the phone.
2. Add a few episodes to Up Next.
3. Open Up Next and quickly tap different episodes to switch playback between them several times.
4. ✅ The app doesn't crash, and the watch's Now Playing screen shows the new episode with the correct play/pause state, position, and duration.
5. On the phone, play, pause, and skip.
6. ✅ The watch shows each change.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/138">8.21 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

PR #3410, opened 2025-08-07
GRDB: corruption recovery

leandroalonso:
| 📘 Part of: #2773 | 
|:---:|

We have a small set of users with corrupted databases. When the database is in this state we need to remove (or rebuild it). This PR adds code to:

1. Check if the database is corrupted
2. If it is, starts a "recovery"

The way this recover works is as it follows:

1. First it renames the file
2. Then it creates a new, healthy, empty database
3. It copies all the data that it can from the previous database (excluding `SJEpisode` which seems to be the culprit for all crashes)
4. It starts a login with the user credentials so all their data is correctly synced back to the device

In the end, the user has the same podcasts with the same settings. They lose per-episode details tho, as this table is not recoverable.

## To test

Best way to test that is a bit involved but it aims at reproducing the issue. You'll need a corrupted database to test though.

1. Login with an account that has the same podcasts as your corrupted database
2. Replace the database of your simulator with the corrupted database
3. Re-run the app
4. ✅ The app should detect the corrupted database and backup your data and login you again
5. ✅ Check that you didn't lose per-podcast settings (Profile > Settings > Auto Download and Auto Add to Up Next)
6. Re-run the app
7. It shouldn't do the same process again as your database is now healthy

## Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `CHANGELOG.md` if necessary.
- [x] I have considered adding unit tests for my changes.
- [x] I have updated (or requested that someone edit) [the spreadsheet](https://docs.google.com/spreadsheets/d/107jqrutZhU0fVZJ19SBqxxVKbV2NWSdQC9MFYdLiAxc/edit?usp=sharing) to reflect any new or changed analytics.

PR #5159, opened 2026-09-17
Show a blocking screen when the database can't be opened

kean:
Fixes a launch crash when the database file is corrupted (Sentry issue 7736804566):

```
DataManager.swift:65: Fatal error: 'try!' expression unexpectedly raised an error: SQLite error 11: malformed database schema (0) - while executing `SELECT * FROM sqlite_master LIMIT 1`
```

GRDB checks that the file is readable when it opens a connection. If the schema is corrupted, `DatabasePool.init` throws and the `try!` crashes the app. This happens on every launch, so the user can't open the app at all until they reinstall it, which throws away whatever hasn't synced.

Rather than crash, or quietly start over with an empty library, the app now stops and says what happened. The database stays on disk exactly as it was, so support can still ask for it and recover it.

**Changes**

- If the database at `pathToDb()` can't be opened, for any reason, `DataManager` keeps the file untouched, records the error in `databaseError`, and opens a throwaway database in the temporary directory. Nothing is moved, reset or deleted. The error is reported to Sentry as a handled error.
- `SceneDelegate` makes `DatabaseErrorViewController` the root view controller when `databaseError` is set. It's a full screen with no way out: it shows what went wrong (including the error and the path), tells the user not to reinstall the app, and offers **Contact Support** and **Export Database**, the existing export that zips up the database, preferences and logs.
- `AppDelegate` returns early from `didFinishLaunchingWithOptions` and skips its lifecycle work while the error screen is up, so nothing refreshes, syncs, downloads or writes against the throwaway database.
- `PCBundleDoc` no longer fails the whole export when there's no `-wal` file next to the database, which is exactly the case when the export matters most.

The throwaway database exists only so the app can get far enough to draw this screen: `DataManager.sharedManager` is reachable from the support form and from plenty of code that runs during launch, and all of it would crash against a manager with no database at all.

**Why only on open**

#3410 ran `PRAGMA integrity_check` on every launch and was reverted in #3415 after beta users couldn't open the app. This does nothing extra on a healthy launch. It only runs where the app would crash today.

## To test

1. Run the unit tests: `DataManagerDatabaseErrorTests`. They cover a malformed schema (the same `malformed database schema (0)` error as the crash), a file that isn't a database, the fallback database, and opening a healthy database, and they check the file on disk is left byte for byte as it was.
2. Manual test on a simulator:
   1. Sign in, then quit the app.
   2. Find `Library/Application Support/Pocket Casts/podcast_newDB.sqlite3` in the app container and overwrite it with junk: `head -c 8192 /dev/urandom > podcast_newDB.sqlite3`. Delete the `-wal` and `-shm` files next to it.
   3. Launch the app.
   4. ✅ The app doesn't crash. It shows the error screen with the SQLite error on it, and there's no way past it.
   5. ✅ **Export Database** produces a "Pocket Casts Export" zip in the share sheet.
   6. ✅ **Contact Support** opens the support form.
   7. ✅ The database file is still on disk, untouched, and no `podcast_newDB_backup.sqlite3` was created.
   8. Restore the database and relaunch. ✅ The app starts normally.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/pocket-casts-ios/milestone/138">8.21 ❄️</a>. The due date for this milestone has already passed.<br />
    </tr>
  </tbody>
</table>

</p>

PR #5163, opened 2026-09-17
Run SwiftLint once, in the podcasts target

kean:
| 📘 Part of: # |  
|:---:|

Fixes PCIOS- 

SwiftLint runs twice on every build of the app.

The `PreBuildActions` aggregate target — which `podcasts`, `WidgetExtension`, `Pocket Casts Watch App` and `Pocket Casts TV App` all depend on — runs `make lint`. The `podcasts` target then separately ran `make lint_lenient` in an unnamed shell script phase. Both phases are `alwaysOutOfDate = 1`, so neither is ever skipped, including on no-op builds.

This PR collapses the two into a single phase in the `podcasts` target, placed immediately after `Sources` and running `make lint_lenient`, skipped on CI.

**Position matters.** The lint has to run *after* the app target compiles, not before it. `PreBuildActions` is a dependency, so a lint phase there runs ahead of every `SwiftCompile` in `podcasts`; Xcode then recompiles those files and replaces their diagnostics, and the violations never reach the Issue Navigator or the editor gutter. The phase still runs and still prints `warning:` lines into the build log — they are just invisible where it counts. Keeping the phase where the old `podcasts` one was preserves the feedback developers actually see.

**Leniency matters.** `make lint` and `make lint_lenient` share the same `swiftlint lint --working-directory .. --config .swiftlint.yml --quiet` invocation over the same file set, but several rules are configured with error severity — `control_statement` (`.swiftlint.yml:150`), `string_interpolation_in_localized_string`, `swiftui_localization`. Under a strict run those fail the build rather than annotate it, which is not what a build-phase lint should do to someone mid-edit. `--lenient` downgrades them to warnings and exits 0.

**CI is unaffected.** `.buildkite/pipeline.yml` runs `swiftlint` as its own step, strict and independent of any build, and that remains the gate. Since the build-phase lint can no longer fail anything, `[ $CI ] && exit 0` keeps it from burning CI time — this is the same guard the original `podcasts` phase had.

The trade-off is that `WidgetExtension` and the Watch and TV apps no longer lint when built on their own; only the app scheme does.

Measured on an M5 Max (18 cores), clean DerivedData, packages pre-resolved, idle machine:

| | Before | After |
|---|---|---|
| No-op build (best of 3, steady state) | 20.9 s | 16.9 s |
| `PhaseScriptExecution` tasks per build | 6 | 5 |

**~4 s (−20%) off every no-op build.** Clean-build wall clock is unchanged within run-to-run noise (±3 s on this machine) — with a full compile to hide behind, the phase overlaps other work. The win is in the edit-build loop, where there is nothing to hide behind.

## To test

1. `make clean && make build_staging` — confirm it succeeds.
2. Add an error-severity violation to a file in the app target, e.g. `if (flag) { }` (`control_statement`).
3. Build from Xcode and confirm it **succeeds** with the violation shown as a *warning* in the Issue Navigator and in the editor gutter, attributed to the `SwiftLint` phase of the `podcasts` target — not as an error.
4. Revert the violation, build twice, and confirm the second (no-op) build succeeds and is faster than before.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5165, opened 2026-09-18
Invoke the pinned SwiftLint binary directly

kean:
| 📘 Part of: # |  
|:---:|

`make lint` ran SwiftLint through `swift package plugin`, which spends ~0.4s on SPM startup before any linting happens. The pinned binary is already on disk in `BuildTools/.build`, so invoke it directly. A Make rule resolves BuildTools when the binary isn't there yet, so a fresh checkout still works.

The version stays pinned exactly as before: the artifact bundle comes from `BuildTools/Package.resolved`, whose version is read out of `swiftlint_version` in `.swiftlint.yml`.

Measured on a clean tree with a warm SwiftLint cache: **0.75s → 0.35s** per run. The Xcode `SwiftLint` build phase still calls `make lint`, so it picks this up with no project change.

## To test

1. `make lint` — passes, and is faster than on trunk.
2. Add `if (true) { print(1) }` to any Swift file, then `make lint` — reports the `control_statement` error and exits non-zero.
3. `make format` — autocorrects that file.
4. `make lint_lenient` — reports the same violation as a warning and exits 0.
5. `make help` still lists `lint` and `format`.
6. Build the app in Xcode — the SwiftLint phase behaves as before.

## Checklist


kean:
It is a documented approach: https://github.com/realm/SwiftLint#xcode-run-script-build-phase.

PR #5172, opened 2026-09-18
Move L10n into a PocketCastsLocalization module

kean:
| 📘 Part of: # |
|:---:|

Moves `L10n` out of the app target into a new `PocketCastsLocalization` package module. The module holds the SwiftGen output, `Strings+L10n.swift` and `LocalizationHelpers.swift`.

- **Compiled once.** Before, the app, Watch, TV, App Clip and widget targets each compiled the 5,000-line generated file. Now they all link the one module.
- **Unblocks other modules.** About 450 app files use `L10n`, so no UI code could leave the app target without it.
- **No new imports.** The five targets compile `podcasts/Exports.swift` (`@_exported import PocketCastsLocalization`), which takes the place of `Strings+L10n.swift` in each of them. App code, and tests that use `@testable import podcasts`, see `L10n` exactly as before, even with `MemberImportVisibility` on.
- **The module generates its own code.** A local build tool plugin, `GenerateL10n`, runs SwiftGen 6.6.2, the same version as before. This PR removes the `L10n` phase in `PreBuildActions`, `swiftgen.yml`, `make generate_code` and the BuildTools SwiftGenPlugin dependency.
  - Why a plugin: SwiftPM fixes a package target's file list when it loads the package graph, and nothing makes a package target build after an Xcode aggregate target. The old script phase couldn't feed the module.
  - Why a prebuild command: Xcode builds the module for both iOS and watchOS into the same plugin directory, so a regular build command fails with "Multiple commands produce". On a no-op build, SwiftGen takes about 0.3 s and doesn't rewrite the file, so nothing recompiles.
- **Translations stay in the app bundle.** Each target still bundles `Localizable.strings`, because App Intents (`LocalizedStringResource`) and What's New look keys up in the main bundle. `L10n` now reads `Bundle.main` directly and no longer force-unwraps the `en.lproj` lookup. In module tests it returns the English text, which the new `PocketCastsLocalizationTests` checks.
- `L10n.bulkDownloadMax` depends on an app constant, so it moved to the app. It now sits next to the other constant-based strings in `Settings.swift`.

The new test target runs in the `Modules-Package` scheme. `UnitTests.xctestplan` doesn't include module test targets yet.

## To test

1. Build and run the app. Strings look the same as before. Switch the app language (Settings › Pocket Casts › Language) and check that a few screens are translated.
2. Build the Watch app, widgets, App Clip and TV app. Check that the widget gallery and a Watch screen show their strings.
3. Add a key to `podcasts/en.lproj/Localizable.strings`, use it in app code as `L10n.<newKey>`, and build once. It compiles on the first build.
4. On a podcast page, select more episodes than the bulk download limit and choose Download. The limit warning still shows.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>

.</td>
    </tr>
  </tbody>
</table>

</p>

PR #5177, opened 2026-09-18
Forward Discover search bar appearance via appearance transitions

kean:
Every time the Discover tab appeared or disappeared, UIKit logged "Calling -viewDidAppear: directly on a view controller is not supported, and may result in out-of-order callbacks" (and the same for -viewDidDisappear:) for `PCSearchBarController`. `DiscoverCollectionViewController` returns `false` from `shouldAutomaticallyForwardAppearanceMethods`, so it has to forward appearance to its child search bar itself, and it did that by calling `viewDidAppear`/`viewDidDisappear` directly. Those calls can't just be removed: without them the search bar never sets `isVisible`, so picking a search history item or suggestion no longer fills in the Discover search field (the behavior added in 64b5fe975). This switches them to `beginAppearanceTransition`/`endAppearanceTransition`, which is the supported way to forward manually and matches how the same controller already drives its cells and search results.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

kean:
I no longer see the warnings in Console.

PR #5178, opened 2026-09-18
Give image-only buttons VoiceOver labels instead of asset names

kean:
VoiceOver read several image-only buttons by their asset name ("discover_tick", "nav cast off", "chapter skipforward", "effects off", "cross-little") because none of them set an accessibility label, so UIKit and SwiftUI fell back to the image name. Each control now has a localized label built from existing strings, and the stateful ones keep it correct as their state changes.

- Podcast page Follow/Unfollow button: "Follow" or "Unfollow", depending on whether you follow the podcast
- Nav bar Cast button: "Cast to", plus a "Connected" value while casting
- Full player chapter buttons: "Previous chapter" / "Next chapter"
- Full player Effects button: "Playback Effects", with an "On"/"Off" value
- Profile banner close button: "Dismiss"
- Notifications permission checkbox: the checkmark image is decorative and the nested button is hidden, so the row reads as one element and uses the Selected trait for its state


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5175, opened 2026-09-18
Move vendored C, Objective-C and AEXML code into SwiftPM targets

kean:
| 📘 Part of: # |  
|:---:|

Moves the vendored C, Objective-C and Swift code out of the app targets into four targets in `Modules/Package.swift`:

| Target | Files | Linked into |
|---|---|---|
| `MNAVChapters` | `MNAVChapterReader.{h,m}` | app, App Clip, Watch, TV |
| `VoiceBoostN` | `VoiceBoostN.{c,h}`, `VoiceBoostN_Internal.h` | app, App Clip, TV |
| `SJUtils` | `SJCommonUtils.{h,m}`, `SJMediaMetadataHelper.{h,m}` | app, App Clip, Watch, TV |
| `AEXML` | `AEXML.swift` | app |

**Why:** each of these files was compiled separately into 2–4 targets, and each of those targets compiled its own bridging-header PCH just to reach them. Now each file compiles once per platform, and Swift files `import` the module. The Swift-visible API is unchanged.

**Bridging headers:** all four (`podcasts-Bridging-Header.h`, `app-clip-Bridging-Header.h`, `watch-bridging-header.h`, `tv-bridging-header.h`) contained only these imports. They're deleted, along with `SWIFT_OBJC_BRIDGING_HEADER` in every configuration. The `Third Party Libraries` synchronized folder is gone too.

**Imports:** the bridging headers also pulled UIKit, Foundation and AVFoundation into every Swift file in those targets. About 115 files relied on that, so they now import the framework directly. That accounts for most of the diff. Nothing else changed in those files.

**Target membership changes:** the TV app used to compile `AEXML.swift` but never used it, so TV doesn't link `AEXML`. The Watch app now also compiles `SJMediaMetadataHelper.m`, since it shares the `SJUtils` target with `SJCommonUtils.m`. `VoiceBoostN` links Accelerate itself instead of relying on `DefaultPlayer.swift`'s `import Accelerate`.

**Left behind:** nothing from the list. `NSNull+Length.m` is now the only Objective-C file in the app target. It's a runtime category that isn't exposed to Swift, so it didn't need a bridging header and is out of scope.

AEXML's upstream ([tadija/AEXML](https://github.com/tadija/AEXML), MIT) is a Swift package. I kept the vendored copy to keep this a pure move. We could switch to the remote dependency in a follow-up.

## To test

[...]

PR #5179, opened 2026-09-18
Move the analytics core into a PocketCastsAnalytics module

kean:
| 📘 Part of: # |
|:---:|

Moves the analytics core out of the app targets into a new `PocketCastsAnalytics` package module. Follows #5175.

Analytics is used by about 260 app files, second only to `L10n`. With it in a module, the app, App Clip, Watch and TV targets compile it once, and other modules can send events.

- **Moved:** `Analytics`, `AnalyticsEvent`, `String+Analytics`, the logging, os_log and Tracks adapters, and `ABTest`/`ABTestProvider`, which `TracksAdapter` needs.
- **Call sites don't change.** The four targets compile `podcasts/Analytics/Exports+Analytics.swift` (`@_exported import PocketCastsAnalytics`). It's a separate file from `Exports.swift` so the widget extension doesn't pull in Tracks, GRDB and the server module.
- **Stays in the app** because it depends on app code:
  - The opt-in/opt-out flow, now in `Analytics+OptOut.swift`. It depends on `Settings` and `AppDelegate`, and keeps its `#if !APPCLIP` branches. The module makes `setAdaptersRegisteredStatus` and `logCurrentAdapters` public for it.
  - `LiveAnalyticsStreamer` and the crash logging adapter. They depend on `Settings`, `ApiCredentials` and the `STAGING` flag, which a package target can't see.
  - `AnalyticsAppThemeProvider`, which depends on `Theme`. Its protocol moved to the module, and the `#if` around it is gone: only the app registers a provider.
  - `AnalyticsPlaybackHelper` and the other helpers.
- **Seams:**
  - `TracksAdapter` takes the app group `UserDefaults` from its three callers instead of defaulting to `SharedConstants`.
  - `userLoginDidChange` moved to `PocketCastsServer`, next to `serverUserWillBeSignedOut`. Every file that uses it already imports the server module.
- **`AnalyticsDescribable` conformances of module types now live in the module.** That's 14 conformances (`ThemeType`, `PlayerAction`, `LibrarySort`, the search results, …) in `AnalyticsDescribable+Modules.swift`, next to the 12 that were already there. Left in the app, each would have needed `@retroactive` and a `public` `analyticsDescription`. The app keeps the rest of those extensions.
- **Tests:** `AnalyticsTrackPropertiesTests` and `AnalyticsAppThemeProviderTests` moved to the new `PocketCastsAnalyticsTests` target, which is added to `UnitTests.xctestplan`.

## To test

1. Build and run the iOS app with Settings › Developer › Tracks Logging enabled. Browse a few screens and check that events still show up in the logs with the same names and properties, such as `settings_appearance_theme_changed` with `theme_selected`.
2. In Settings › Privacy, turn analytics off, then back on. Events stop and then resume.
3. Sign out and sign back in. Tracks switches between the anonymous and signed-in user (`TracksAdapter` observes `userLoginDidChange`).
4. Build and run the Watch app, App Clip and TV app. Events still show up in the console under the `Analytics` category.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>

.</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PR #5183, opened 2026-09-18
Show SwiftLint violations to Claude after each edit

kean:
| 📘 Part of: # |  
|:---:|

The Claude Code `PostToolUse` hook ran `swiftlint --fix` through the SwiftPM plugin, printed "SwiftLint completed" and exited 0, so the agent never saw a violation that `--fix` can't correct. With a file containing `.textAlignment = .left` and an interpolated `NSLocalizedString` (an error-level rule), the hook said nothing while `swiftlint lint` reported three violations.

The hook now lives in `.claude/hooks/swiftlint.sh`:

- After fixing, it runs `lint --quiet` on the same file and exits 2 with the violations on stderr, which Claude Code feeds back to the agent. Files the config excludes are skipped (`--force-exclude`).
- It calls the pinned binary in `BuildTools/.build` directly and checks that its version matches `swiftlint_version`. If it's missing or stale, it resolves `BuildTools` once, the same way `make lint` does.
- Paths come from `$CLAUDE_PROJECT_DIR` and the edited file's checkout instead of the shell's working directory, so it keeps working after a `cd` and in worktrees.

| | Before | After |
|---|---|---|
| Per edit, warm | 0.8–2.7 s | 0.25–0.35 s |
| First edit in a new worktree | 8.3 s | 0.35 s (reuses the main checkout's binary) |

## To test

1. Save this as `podcasts/LintProbe.swift`:
   ```swift
   import UIKit

   final class LintProbe {
       func configure(label: UILabel, count: Int) {
           label.textAlignment = .left
           label.text = NSLocalizedString("probe \(count)", comment: "")
       }
   }
   ```
2. Run `echo "{\"tool_input\":{\"file_path\":\"$PWD/podcasts/LintProbe.swift\"}}" | CLAUDE_PROJECT_DIR=$PWD .claude/hooks/swiftlint.sh; echo $?`. It prints the `localization_comment`, `natural_text_alignment` and `string_interpolation_in_localized_string` violations and exits 2.
3. Replace the file with `let a = [1,2]` and run the same command. It exits 0 with no output, and the file now reads `let a = [1, 2]`.
4. In a Claude Code session, ask it to add a `.left` text alignment somewhere. The agent is told about the `natural_text_alignment` violation right after the edit.
5. Delete the probe file.

## Checklist


kean:
Works as expected.

PR #5184, opened 2026-09-18
Stop union-merging project.pbxproj

kean:
| 📘 Part of: # |  
|:---:|

`.gitattributes` has merged `*.pbxproj` with the `union` driver since cb379f56d (April 2023). A union merge keeps both sides' lines for a conflicting hunk and never reports a conflict, so a rebase can finish cleanly and leave `project.pbxproj` unreadable. That's how an earlier rebase produced a broken project "without conflict", and agents trust a clean rebase.

Reproduced on the current project: two branches that each add a `PBXFileReference` at the top of the section. With `merge=union`, `git rebase` exits 0 and drops the `};` between the two objects, and `plutil -lint` rejects the file. Without the rule, Git reports a conflict in `project.pbxproj`.

This PR:

- Removes `*.pbxproj merge=union`, so these merges surface as conflicts. `CHANGELOG.md` keeps its union rule.
- Adds a `Lint project.pbxproj` step to the Linters group that runs `plutil -lint` on the project (about 30 ms). It runs on the `mac` queue because Danger's `linter` queue doesn't have `plutil`.

`plutil` catches syntax damage like the dropped brace. It doesn't catch a duplicated key, which is still a valid property list.

## To test

1. Check that the "Lint project.pbxproj" step passes on this PR.
2. Locally, delete one `};` line from `podcasts.xcodeproj/project.pbxproj` and run `plutil -lint podcasts.xcodeproj/project.pbxproj`. It fails. Revert the change.
3. Optional: on two scratch branches off this one, add a different file to the same group in Xcode and commit. Rebasing one onto the other now stops with a conflict in `project.pbxproj`.

## Checklist


.buildkite/pipeline.yml
```diff
@@ -75,6 +75,18 @@ steps:
         agents:
           queue: mac
 
+      # A merge or rebase can leave the project unreadable without reporting a
+      # conflict, e.g. when it drops a closing brace between two added objects.
+      - label: ":xcode: Lint project.pbxproj"
+        command: |
+          plutil -lint podcasts.xcodeproj/project.pbxproj || {
```
AliSoftware, line 82: On one hand I like the idea of lining the project file, but on the other hand:

1. Having a dedicated step/job like here will likely be a bit of a waste because it'll take more time to wait for an agent, have that agent checkout the repo, boot the VM inside that Mac agent, set everything up etc… than the <30ms it takes to run `plutil -lint`. If we really want to run `plutil -lint` it would probably be more efficient to run it as part of one more command in an existing job's script, so we don't pay again for all the checkout + VM boot overhead… all for running such a small/ultrafast command
2. The other jobs that exist in our pipeline, like the ones that do "Build and Test" for example, are probably already going to surface an error very early in their run if the Xcode project is invalid (as the `xcodebuild` command that they run would fail to read the project file anyway).
3. And if the point is to have a more explicit error message when such a case happen, see point (1), i.e. we could call `plutil -lint || { echo "You might have a badly-resolved merge conflict on your xcodeproj file"; exit 1 }` very early in the `.buildkite/commands/build.sh` script instead of a dedicated job requiring its own agent and bootstrapping time for a <30ms command 😛

So maybe remove this new Buildkite step completely… then either don't bother calling `plutil -lint` (and rely on "Build and Test"'s `xcodebuild` failing if a bad merge were to mess up the project, or still call `plutil -lint` (to get a nicer error message) but at the start of `build.sh`?

.gitattributes
```diff
@@ -1,4 +1,3 @@
 CHANGELOG.md merge=union
-*.pbxproj merge=union
```
AliSoftware, line 2: 👍

AliSoftware:
Whoops I started my review when the PR was still open… but now I see you closed it while I was writing so it was already closed by the time I submitted my comments 🤣 nvm then!

kean:
Hey, @AliSoftware. Sorry about that! Claude suggested it in the scope of "how can I make this repo better optimized for agents". I didn't have the capacity to review and understand it, so I decided to skip it. If you think it's an improvement, let's reopen it.

kean:
I'm also considering trying the new JSON-based pbxproj.

AliSoftware:
> I'm also considering trying the new JSON-based pbxproj.

As much as I'm excited for this new JSON format, in the fastlane team slack (I'm a contributor there) we're currently discussing the potential impacts of this on fastlane. In particular, some tools of fastlane depends on the `xcodeproj` gem and on reading the project file.

I haven't yet checked in details, but depending on which parts of fastlane are relying on that (and if our own a8c release-toolkit tooling use that as well?) it might mean that we can't migrate to this new JSON format until the `xcodeproj` gem is updated to support it?

PR #5185, opened 2026-09-18
Remove the grdbQueryInterface feature flag

kean:
Removes `FeatureFlag.grdbQueryInterface`, which has defaulted to on, and the code that only existed to support turning it off:

- **Save paths**: `save`/`bulkSave` for Episode, UserEpisode, Podcast, Folder and EpisodeFilter now always use GRDB's `PersistableRecord`. The raw SQL fallback and its `createValuesFrom` value builders are deleted.
- **`columnNames`**: the managers' column lists were only used by the legacy SQL path. They now live in the `*ColumnConsistencyTests`, which were their only other users. They still guard the table schema, and #5030/#5180 use them to check the encoder output.
- **Test harness**: `DataManagerTestCase.runWithBothImplementations` ran every test twice, once for "SQL" and once for "GRDB". Because the flag defaulted to on, both runs actually used GRDB. It's now `runWithDataManager { dataManager in … }` and runs once, and the `\(impl):` message prefixes are gone. Most of the diff is this mechanical change.

⚠️ Heads-up for #5030 and #5180: both add assertions that use `implementationName`, so after this lands they need a rebase that drops the `"\(implementationName): "` prefix from those messages.

## To test

1. CI is green.
2. Optional: run the DataModel tests locally with `cd Modules && xcodebuild test -scheme Modules-Package -destination 'platform=iOS Simulator,id=<sim>' -only-testing:PocketCastsDataModelTests`. All 473 pass.
3. Smoke test: subscribe to a podcast, create a folder and a playlist, play and archive an episode, and add a file. Everything persists after relaunching the app.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

kean:
going to split it – too many changes

PR #5188, opened 2026-09-18
Remove the placeholder Modules target

kean:
| 📘 Part of: # |  
|:---:|

Removes the `Modules` library and its `ModulesTests` target from `Modules/Package.swift`. They only held `Modules.sayHello()`, a stub from the package scaffolding that nothing imports, and `UnitTests.xctestplan` never listed `ModulesTests`, so it didn't run locally or on CI. 325b67456 removed both, and the rename in c9dbbea2a brought them back.

`GRDBMacrosTests` is also missing from the plan. It's being removed separately.

## To test

1. Build the app and run the unit tests. Nothing referenced the `Modules` product, so both still pass.
2. Check that CI's Build and Test step passes.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>

.</td>
    </tr>
  </tbody>
</table>

</p>

PR #5189, opened 2026-09-18
Skip SwiftGen when the English strings haven't changed

kean:
| 📘 Part of: # |
|:---:|

Follow-up to #5172. Two build-time fixes for `PocketCastsLocalization`:

- **SwiftGen ran on every build.** A `prebuildCommand` runs every build, and SwiftGen took 0.5–1.5 s inside Xcode's plugin sandbox even when nothing had changed. The command now runs through `/bin/sh` and skips SwiftGen when a stamp shows it already ran with the same arguments on the current `Localizable.strings` (~0.04 s). A regular `buildCommand` with input/output tracking isn't an option: the target builds for iOS and watchOS in one build, both platforms share one plugin work directory, and Xcode fails with "Multiple commands produce `Strings+Generated.swift`".
- **Editing a string rebuilt the whole app.** Adding a key changes the module's interface, and because `Exports.swift` re-exports the module, every file in the app targets recompiled (1,107 `podcasts` + 129 Watch/App Clip/widget files, 54 s). With `-enable-incremental-imports` (Debug only), the same edit recompiles only the module (~24 s against a ~19 s no-op build on my machine).

## To test

1. Build the app twice. The second build's "Generate L10n from Localizable.strings" step doesn't run SwiftGen, and nothing in `PocketCastsLocalization` recompiles.
2. Add a key to `podcasts/en.lproj/Localizable.strings` and build. `L10n` has the new member, and the build recompiles `PocketCastsLocalization` but not the app's Swift files.
3. Use the new `L10n` member in one app file and build. Only that file recompiles.
4. Remove the key and build. The member is gone.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>

.</td>
    </tr>
  </tbody>
</table>

</p>

kean:
Closing in favor of reverting the module: #5190.

PR #5190, opened 2026-09-18
Revert "Move L10n into a PocketCastsLocalization module (#5172)"

kean:
| 📘 Part of: # |  
|:---:|

Reverts #5172. `L10n` goes back into the app targets, generated by the `L10n` build phase in `PreBuildActions` (`make generate_code` with the BuildTools SwiftGen plugin).

The module version ran SwiftGen as a prebuild command on every build (0.5–1.5 s), and any change to the English strings changed the module interface, which recompiled every app file through the `@_exported import` (about 54 s for adding one key). The fixes in #5189 rely on a shell-script stamp and an unsafe compiler flag, which isn't a trade worth making.

The only conflict was in `Modules/Package.swift`, where #5179 added `PocketCastsAnalytics` next to `PocketCastsLocalization` in the Xcode support targets. The `PocketCastsAnalytics` entries stay. I also dropped a dangling `Debounce.swift` build file that the revert would have brought back into the App Clip; #5174 moved that file into PocketCastsUtils.

## To test

1. Delete `podcasts/Strings+Generated.swift` if you have one locally.
2. Build the `pocketcasts` scheme. It should succeed, and the `L10n` phase in `PreBuildActions` should regenerate `podcasts/Strings+Generated.swift`.
3. Build the `Pocket Casts TV App` scheme. It should succeed.
4. Build again without changes. The `L10n` phase should be skipped.
5. Launch the app and check that strings are localized, e.g. by switching the simulator language.

## Checklist


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>

.</td>
    </tr>
  </tbody>
</table>

</p>

